PR #3, opened 2022-09-27
fix: Change the Dependabot ecosystem from pip-compile to pip

MthwRobinson:
### Summary

Fixes [this build failure](https://github.com/Unstructured-IO/unstructured/runs/8577828844). Corrects the ecosystem name by updating it to `pip` from `pip-compile`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "pip-compile"
+  - package-ecosystem: "pip"
     directory: "/requirements"
     schedule:
       interval: "weekly"


.github/dependabot.yml
```diff
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "pip-compile"
```
qued, line 3: Weird, I just saw an example of using this [here](https://github.com/mozilla-releng/tooltool/blob/fb7d180f5c21286f443a709e57538643fd8b6707/.github/dependabot.yml).

cragwolfe (approved):

PR #21, opened 2022-10-03
chore: Integration test to show LabelStudio brick working with SDK

yuming-long:
### Summary
Shows that LabelStudio brick `stage_for_label_studio` work with `label_studio_sdk` to upload data.

The HTTP request to `label-studio` is recorded by the [vcr](https://vcrpy.readthedocs.io/en/latest/usage.html) decorator for the first time and stored in `label_studio.yaml` file. The cassette is used later to replay the  HTTP response.

### Test
Instructions on how to rebuild the fixture are included in the test comment.


cragwolfe (approved):
LGTM!

test_unstructured/staging/test_label_studio.py
```diff
@@ -3,12 +3,57 @@
...
     return [Title(text="Title 1"), NarrativeText(text="Narrative 1")]
 
 
+@vcr.use_cassette("fixtures/cassettes/label_studio.yaml")
+def test_upload_label_studio_data_with_sdk_on_real_instance(caplog, elements):
+    # The HTTP request to label studio is recorded by vcr decorator in the first time
+    # and replayed later.
+    log = logging.getLogger("urllib3")
+    log.setLevel(logging.DEBUG)
+    # Define the URL where Label Studio is accessible and the API key for your user account
```
qued, line 24: I think this comment should be clarified. This sounds like it needs to match credentials of the current user for testing purposes. I think it should be clear that...
1. this is a fake key that needs to match what's in the vcr cassette and
2. this would need to by updated if the vcr cassette is updated (by someone with a different access token).

PR #26, opened 2022-10-09
feat: Implement LabelBox staging brick

asymness:
Resolves https://github.com/Unstructured-IO/community-tasks/issues/7

**Summary of changes:**
- [x] Implement `stage_for_label_box` function according to specifications in `staging/label_box.py` file.
- [x] Add unit tests for `stage_for_label_box` and ensure 100% coverage.
   image

- [x] Update docs with example and description of `stage_for_label_box`.
   image

- [x] Update changelog and version


unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def stage_for_label_box(
+    elements: List[Text],
+    output_directory: str,
+    url_prefix: str,
+    external_ids: Optional[List[str]] = None,
+    attachments: Optional[List[List[Dict[str, str]]]] = None,
+    create_directory: bool = False,
+) -> List[Dict[str, Any]]:
```
qued, line 39: Need docstring.

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
+import os
+
+from typing import Any, Dict, List, Optional
+from unstructured.documents.elements import Text
+
+
+VALID_ATTACHMENT_TYPES = ["IMAGE", "VIDEO", "RAW_TEXT", "TEXT_URL", "HTML"]
+
+
+def _validate_attachments(attachments, element_index):
```
qued, line 10: Need type hints + docstring.

docs/source/bricks.rst
```diff
@@ -562,3 +562,62 @@ Examples:
...
+
+``stage_for_label_box``
+--------------------------
+
+Formats outputs for use with `LabelBox <https://docs.labelbox.com/docs/overview>`_. LabelBox accepts cloud-hosted data 
+and does not support importing text directly. The ``stage_for_label_box`` does the following:
+
+* Stages the data files in the ``output_directory`` specified in function arguments to be uploaded to a cloud storage service.
+* Returns a config of type ``List[Dict[str, Any]]`` that can be written to a ``json`` file and imported into LabelBox.
+
```
MthwRobinson, line 576: Additional note:

>  Note, `stage_for_label_box` does not upload the data to remote storage such as S3. Users can upload the data to S3 using `aws s3 sync ${output_directory} ${url_prefix}` after running the `stage_for_label_box` staging brick.

MthwRobinson: The example snippet covers the S3 upload part, just wanted to make explicit that `stage_for_label_box` only stages and doesn't do the upload step.

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+            "The length of external_ids parameter must be the same as the length of elements parameter."
+        )
+    elif not external_ids:
+        ids = [element.id for element in elements]
+    else:
+        ids = external_ids
+
+    if attachments and len(attachments) != len(elements):
+        raise ValueError(
+            "The length of attachments parameter must be the same as the length of attachments parameter."
```
qued, line 52:
```suggestion
            "The length of attachments parameter must be the same as the length of elements parameter."
```

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+from typing import Any, Dict, List, Optional
+from unstructured.documents.elements import Text
+
+
+VALID_ATTACHMENT_TYPES = ["IMAGE", "VIDEO", "RAW_TEXT", "TEXT_URL", "HTML"]
+
+
+def _validate_attachments(attachments, element_index):
+    for attachment_index, attachment in enumerate(attachments):
+        error_message_prefix = f"Error at index {attachment_index} of attachments parameter for element at index {element_index}."
```
MthwRobinson, line 12: This line is too long and triggering the line length error in the lint job. Max width is 100. I've been formatting long strings like this:

```python
error_message_prefix = f"Error at index {attachment_index} of attachments "
                                          f"parameter for element at index {element_index}."
```

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+    output_directory: str,
+    url_prefix: str,
+    external_ids: Optional[List[str]] = None,
+    attachments: Optional[List[List[Dict[str, str]]]] = None,
+    create_directory: bool = False,
+) -> List[Dict[str, Any]]:
+    ids: List[str]
+    if external_ids and len(external_ids) != len(elements):
+        raise ValueError(
+            "The length of external_ids parameter must be the same as the length of elements parameter."
```
MthwRobinson, line 43: Same comment as above on string length

MthwRobinson (changes requested):
Implementation looks great! Just a few comments on linting and doc updates.

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+    elif not external_ids:
+        ids = [element.id for element in elements]
+    else:
+        ids = external_ids
+
+    if attachments and len(attachments) != len(elements):
+        raise ValueError(
+            "The length of attachments parameter must be the same as the length of attachments parameter."
+        )
+    elif not attachments:
```
qued, line 54: If the changes I suggested on line 50 are applied, then we should've already dealt with the case where `attachments` is an empty list but `elements` has 1 or more elements, but I still think it'd be better to use:
```suggestion
    elif attachments is None:
```
... here since I think the intent of this block is to deal with the `None` case, not the empty list case.

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+        attachments: List[List[Dict[str, str]]] = [{} for _ in range(len(elements))]
```
qued, line 55: This seems to make `attachments` have type `List[Dict[str, str]]` instead of `List[List[Dict[str, str]]]`, am I missing something?

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+def stage_for_label_box(
+    elements: List[Text],
+    output_directory: str,
+    url_prefix: str,
+    external_ids: Optional[List[str]] = None,
+    attachments: Optional[List[List[Dict[str, str]]]] = None,
+    create_directory: bool = False,
+) -> List[Dict[str, Any]]:
+    ids: List[str]
+    if external_ids and len(external_ids) != len(elements):
```
qued, line 41: Similar to my comment on line 50, I think this should be:
```suggestion
    if (external_ids is not None) and len(external_ids) != len(elements):
```

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+        raise ValueError(
+            "The length of external_ids parameter must be the same as the length of elements parameter."
+        )
+    elif not external_ids:
```
qued, line 45: Similar to my comment on line 54, I think this should be:
```suggestion
    elif external_ids is None:
```

unstructured/staging/label_box.py
qued: Also, this is a nit, but why not iterate over `elements` instead of `range(len(elements))` since it's cleaner and we don't really care about the iteration value anyway?

unstructured/staging/label_box.py
asymness: Added 👍

unstructured/staging/label_box.py
asymness: Added 👍

docs/source/bricks.rst
asymness: Added the note before example 👍

docs/source/bricks.rst
```diff
@@ -562,3 +562,62 @@ Examples:
...
+.. code:: python
+
+  import os
+  import json
+
+  from unstructured.documents.elements import Title, NarrativeText
+  from unstructured.staging.label_box import stage_for_label_box
+
+
+  S3_BUCKET_NAME = "labelbox-staging-bucket"  # The S3 Bucket name where data files should be uploaded.
```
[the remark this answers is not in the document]
asymness: Done 👍

docs/source/bricks.rst
```diff
@@ -562,3 +562,62 @@ Examples:
...
+  S3_BUCKET_KEY_PREFIX = "data/"  # The S3 key prefix (I.e. directory) where data files should be stored.
+  S3_URL_PREFIX = f"https://{S3_BUCKET_NAME}.s3.amazonaws.com/{S3_BUCKET_KEY_PREFIX}"  # The URL prefix where the data files will be accessed. In case of S3, the bucket should have public access enabled.
```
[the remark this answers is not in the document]
asymness: Fixed

unstructured/staging/label_box.py
```diff
@@ -0,0 +1,84 @@
...
+        ids = [element.id for element in elements]
+    else:
+        ids = external_ids
+
+    if attachments and len(attachments) != len(elements):
```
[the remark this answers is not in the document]
asymness: Good catch. I have updated this.

asymness:
Fixed all the issues. Sorry for all the lint issues, I must have forgotten to run `make check`. 🤐

MthwRobinson (approved):
LGTM - thanks!

PR #35, opened 2022-10-13
chore: Update PDFDocument to use from_file method

MthwRobinson:
### Summary

This PR updates the `PDFDocument` object to use the `from_file` method. This makes it consistent with `HTMLDocument` and also makes it consistent with the example that's in the documentation [here](https://unstructured-io.github.io/unstructured/examples.html#pdf-parsing).

### Testing

```python
from unstructured.documents.pdf import PDFDocument

 doc = PDFDocument.from_file("OER_Filled.pdf")
print(doc)
```


qued (approved):
LGTM!

PR #58, opened 2022-11-08
docs: Added long description to PyPi

benjats07:
This PR add long description to the package.

The description is taked from README.md

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,22 +17,29 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-import os
-
 from setuptools import setup, find_packages
 
 from unstructured.__version__ import __version__
 
-cwd = os.path.dirname(os.path.abspath(__file__))
-
-# Read in README.md for our long_description
-with open(os.path.join(cwd, "README.md"), encoding="utf-8") as f:
-    long_description = f.read()
-
 setup(
     name="unstructured",
     description="A library that prepares raw documents for downstream ML tasks.",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,14 +17,22 @@
...
+cwd = os.path.dirname(os.path.abspath(__file__))
+
+# Read in README.md for our long_description
+with open(os.path.join(cwd, "README.md"), encoding="utf-8") as f:
+    long_description = f.read()
+
 setup(
     name="unstructured",
     description="A library that prepares raw documents for downstream ML tasks.",
+    long_description=long_description,
```
MthwRobinson, line 35: We should be able to do this without importing `os`, as seen in `transformers` [here](https://github.com/huggingface/transformers/blob/main/setup.py#L416-L417). While we're in here, let's also add `keywords`, `url`, `python_requires`, and `classifiers`.  For `classifiers` we can borrow what the classifiers from `transformers`, except drop `Programming Language :: Python :: 3.7` and change development status to `Development Status :: 4 - Beta`

MthwRobinson (changes requested):
Check out the inline comment.

setup.py
benjats07: @MthwRobinson  What keywords you suggest? I think maybe: 
"NLP PDF HTML parsing preprocessing"

MthwRobinson: I'd add `CV` and `XML`, the others look good to me!

MthwRobinson (approved):
LGTM, nice work! Can you do the same in the `setup.py` for `unstructured-api-tools`?

benjats07:
Yes, already done.

PR #63, opened 2022-11-10
feat: Cleaning bricks to extract text before/after a pattern

MthwRobinson:
### Summary

- Adds a cleaning brick to extract text before a pattern
- Adds a cleaning brick to extract text after a pattern
- Bumps to `0.2.3` to support using these cleaning bricks in the OER pipeline

#### `extract_text_before`

```python

  text = "Here I am! STOP Look at me! STOP I'm flying! STOP"

  # Returns "Here I am!"
  extract_text_before(text, r"STOP")
```

#### `extract_text_after`

```python
  from unstructured.cleaners.extract import extract_text_after

  text = "SPEAKER 1: Look at me, I'm flying!"

  # Returns "Look at me, I'm flying!"
  extract_text_after(text, r"SPEAKER \d{1}:")
```

### DIFF
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -391,7 +391,7 @@ Extracts text that occurs after the specified pattern.
 Options:
 
 * If ``index`` is set, extract after the ``(index + 1)``th occurence of the pattern. The default is ``0``.
-* Strips trailing whitespace is ``strip`` is set to ``True``. The default is ``True``.
+* Strips trailing whitespace if ``strip`` is set to ``True``. The default is ``True``.
 
 
 Examples:
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -368,7 +368,7 @@ Extracts text that occurs before the specified pattern.
 Options:
 
 * If ``index`` is set, extract before the ``(index + 1)``th occurence of the pattern. The default is ``0``.
-* Strips leading whitespace is ``strip`` is set to ``True``. The default is ``True``.
+* Strips leading whitespace if ``strip`` is set to ``True``. The default is ``True``.
 
 
 Examples:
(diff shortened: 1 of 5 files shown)


docs/source/bricks.rst
```diff
@@ -360,6 +360,52 @@ Examples:
...
 
+``extract_text_before``
+-----------------------
+
+Extracts text that occurs before the specified pattern.
+
+Options:
+
+* If ``index`` is set, extract before the ``(index + 1)``th occurence of the pattern. The default is ``0``.
+* Strips leading whitespace is ``strip`` is set to ``True``. The default is ``True``.
```
qued, line 371:
```suggestion
* Strips leading whitespace if ``strip`` is set to ``True``. The default is ``True``.
```

docs/source/bricks.rst
```diff
@@ -360,6 +360,52 @@ Examples:
...
+
+``extract_text_after``
+----------------------
+
+Extracts text that occurs after the specified pattern.
+
+Options:
+
+* If ``index`` is set, extract after the ``(index + 1)``th occurence of the pattern. The default is ``0``.
+* Strips trailing whitespace is ``strip`` is set to ``True``. The default is ``True``.
```
qued, line 394:
```suggestion
* Strips trailing whitespace is ``strip`` is set to ``True``. The default is ``True``.
```

unstructured/cleaners/extract.py
```diff
@@ -0,0 +1,46 @@
+import re
+
+
+def _get_indexed_match(text: str, pattern: str, index: int = 0) -> re.Match:
+    if not isinstance(index, int) or index < 0:
+        raise ValueError(f"The index is {index}. Index must be a positive integer.")
```
qued, line 6: It looks like `index` is allowed to be zero, so shouldn't this say "Index must be a non-negative integer"?

qued (approved):
Other than a few minor doc nits, LGTM!

unstructured/cleaners/extract.py
MthwRobinson: Indeed it should!

MthwRobinson: Updated

Issue #56, opened 2022-11-07
Add Colab Notebook


nateraw:
Would be nice to have Colab notebook with examples shown in the readme here. Happy to contribute this, since I already did it to play with the repo the other day.

PR #66, opened 2022-11-11
docs: add Colab notebook line in README

yuming-long:
Addresses https://github.com/Unstructured-IO/unstructured/issues/56 and add Colab notebook with examples from README.

You can add a comment for a block to suggest changes for the Colab notebook.


MthwRobinson (changes requested):
I looks like that colab notebook has an `IndexError` in one of the cells right now

MthwRobinson (approved):
LGTM!

PR #74, opened 2022-11-14
fix: fix abstract class

qued:
Changed where `_read` sits in the inheritance structure since `PDFDocument` doesn't really need lazy document processing and it was causing a typing issue with the new version of `mypy`.

#### Testing

- In `requirements/test.txt`, change the `mypy` version to 0.990
- In your `unstructured` environment, run `pip install -r requirements/test.txt`
- run `make check`
- There should be no errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.2.6-dev0
 
-* Typing corrections
+* Small change to how _read is placed within the inheritance structure since it doesn't really apply to pdf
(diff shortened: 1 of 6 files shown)


MthwRobinson (commented):
Want to pull in the updates from main and then I'll do a final review? May have to juggle some conflicts because of the `mypy` bump from earlier.

MthwRobinson (approved):
LGTM!

PR #75, opened 2022-11-18
chore: Remove PDF parsing code and dependencies

mallorih:
Local PDF parsing code is removed in favor of the remote API call. Also any tests for that code is removed.


MthwRobinson (changes requested):
Looks good! Just one addition, could you update the PDF parsing section in the README to reference the new `partition_pdf` function instead of the `PDFDocument` class?

MthwRobinson (approved):
Code looks good! LGTM once the linting job is passing

PR #77, opened 2022-11-22
feat: Add entities kwarg to datasaur bricks

MthwRobinson:
### Summary

Adds and `entities` kwarg that allows users to pass entities to the `stage_for_datasaur` brick. If passed, the entities are included in the staged output.

### Testing

Entities should be included in the output of the follow snippet.

```python
  from unstructured.staging.datasaur import stage_for_datasaur

  elements  = [Text("Hi my name is Matt.")]
  entities = [[{"text": "Matt", "type": "PER", "start_idx": 11, "end_idx": 15}]]
  datasaur_data = stage_for_datasaur(elements, entities)
```


unstructured/staging/]
```diff
@@ -0,0 +1,33 @@
...
+    return result
+
+
+def _validate_datasaur_entity(entity: Dict[str, Any]):
+    """Raises an error if the Datasaur entity is invalid."""
+    keys_and_types = {"text": str, "type": str, "start_idx": int, "end_idx": int}
+
+    for key, _type in keys_and_types.items():
+        if key not in entity:
+            raise ValueError(f"Key '{key}' was expected but not present in the Datasaur entity.")
```
qued:31]: Pretty sure this file should be deleted.

qued:31]: The one named `]`

qued (approved):
LGTM other than accidentally duplicated file.

unstructured/staging/]
MthwRobinson:31]: But it's the best file!

MthwRobinson:31]: Even better than `:q!` 🤣

PR #84, opened 2022-11-29
feat: Add html escape quotes to cleaning brick

MthwRobinson:
### Summary

Adds the `&apos;` HTML quote escape to the `replace_unicode_quotes` cleaning brick. 

### Testing

```python
from unstructured.cleaners.core import replace_unicode_quotes

# The output should be "My dog's bowl"
replace_unicode_quotes("My dog&apos;s bowl")
```


qued (approved):
LGTM!

PR #86, opened 2022-11-30
fix: partition pdf bad responses

LaverdeS:
- raise ValueError for bad responses in `partition_pdf`
- remove unused dependency Text from imports


MthwRobinson (approved):
LGTM!

PR #88, opened 2022-11-30
test: mock model api calls; full coverage for partition_pdf

MthwRobinson:
### Summary

- Updates the `partition_pdf` tests to use mock API calls instead of calling the live API.
- Expands `partition_pdf` tests to achieve 100% test coverage

### Testing

`make test` should work with the updated tests and show 100% coverage


qued (approved):
LGTM!

PR #89, opened 2022-12-01
chore: add type hint to requests.post arg

qued:
In cases where the `types-requests` package was installed, `mypy` would throw an error on the `files` argument of `requests.post` in `unstructured/partition/pdf.py`. This PR fixes that without any functional change to the code by providing a type hint to the argument passed to `requests.post`.

#### Testing:
Verify the error from the `main` branch by installing `types-requests` in your environment and running `make check`.

Verify the fix from this branch in the same manner. With `types-requests` installed, `make check` should find no errors.

### DIFF
--- a/unstructured/partition/pdf.py
+++ b/unstructured/partition/pdf.py
@@ -3,10 +3,16 @@ import requests  # type: ignore
 import sys
 
 if sys.version_info < (3, 8):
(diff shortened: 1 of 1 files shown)


cragwolfe (approved):
Works for me as described in PR description. Also validated `main` currently fails.

PR #90, opened 2022-12-06
docs: update readme with gif

LaverdeS:
relative path to gif might have to be changed once the branch is merged. For now the gif was uploaded directly to the readme, path is automatically generated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@
   
 </div>
 
+<div align="center">
+<img src="https://user-images.githubusercontent.com/38184042/205945013-99670127-0bf3-4851-b4ac-0bc23e357476.gif" title="unstructured in action!">
+</div>
+	
 <h3 align="center">
   <p>Open-Source Pre-Processing Tools for Unstructured Data</p>
 </h3>


MthwRobinson (changes requested):
The terminal embedded in the README looks great! Two requests:

- Since we're focusing more on PDF -> JSON at the moment, can we change the example to one that uses `partition_pdf`? That'll also show an import from the brick yard.
- The resolution on the gif is a little low (not sure if that's just me or my minor). I'm good with the currently resolution though if it's tough to make it sharper.

LaverdeS:
> The terminal embedded in the README looks great! Two requests:
> 
> * Since we're focusing more on PDF -> JSON at the moment, can we change the example to one that uses `partition_pdf`? That'll also show an import from the brick yard.
> * The resolution on the gif is a little low (not sure if that's just me or my minor). I'm good with the currently resolution though if it's tough to make it sharper.

considering the last 1:1 ´, shall we count this as done? I can always make more or add more unstructured calls in the future. for now is HTML and PDF calling partition_pdf. Once partition_html is done I can update it

MthwRobinson:
Yep, just approved. ✅

PR #91, opened 2022-12-09
feat: Add partition_html brick

MthwRobinson:
### Summary

Adds a `partition_html` brick with an API that is similar to the `partition_pdf` bricks. No new functionality is added, this simply brings makes our HTML parsing capabilities consistent with the bricks concept.

### Testing

```python

  elements = partition_html(filename="example-docs/example-10k.html")

  with open("example-docs/example-10k.html", "r") as f:
      elements = partition_html(file=f)

  with open("example-docs/example-10k.html", "r") as f:
      text = f.read()
  elements = partition_html(text=text)
```


CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
-## 0.3.0-dev7
+## 0.3.0-dev8
 
 * Implement staging brick for Argilla. Converts lists of `Text` elements to `argilla` dataset classes.
 * Removing the local PDF parsing code and any dependencies and tests.
 * Reorganizes the staging bricks in the unstructured.partition module
 * Allow entities to be passed into the Datasaur staging brick
 * Added HTML escapes to the `replace_unicode_quotes` brick
 * Fix bad responses in partition_pdf to raise ValueError
+* Adds the `partition_html` for partitioning HTML documents.
```
qued, line 9:
```suggestion
* Adds `partition_html` for partitioning HTML documents.
```

README.md
```diff
@@ -68,35 +68,17 @@ of the features in the library.
...
-
-Actual results could differ materially from those expressed or implied in our forward-looking statements. Our future financial condition and results of operations, as well as any forward-looking statements, are subject to change and to inherent known and unknown risks and uncertainties. You should not assume at any point in the future that the forward-looking statements in this report are still valid. We do not intend, and undertake no obligation, to update our forward-looking statements to reflect future events or circumstances.
-```
-
-If you then run:
-
-```python
-doc.pages[2].elements
+elements = partition_html("example-docs/example-10k.html")
```
qued, line 73: I slightly prefer showing the actual text contents of the elements here since someone approaching this for the first time will see the beginning and end points in terms they understand without knowing anything about the `Title` and `NarrativeText` objects. Not a blocker though.

unstructured/partition/html.py
```diff
@@ -0,0 +1,38 @@
+from typing import IO, List, Optional
+
+from unstructured.documents.elements import Element
+from unstructured.documents.html import HTMLDocument
+
+
+def partition_html(filename: str = "", file: Optional[IO] = None, text: str = "") -> List[Element]:
```
qued, line 7: Nit: Could we use `None` as the default for arguments that weren't specified? If this sits in an automated pipeline somewhere I can imagine an empty string being passed for `filename` (in which case some kind of "invalid filename" from the file handler response would be more helpful than "One of filename, file, or text must be specified.") or for `text` (in which case we'd want to return an empty `HTMLDocument`).

qued (approved):
LGTM other than nits.

unstructured/partition/html.py
MthwRobinson: Yep! Just updated the defaults

README.md
MthwRobinson: Added the text output back into the readme

PR #92, opened 2022-12-09
build(deps): Bump certifi to address security scans

MthwRobinson:
### Summary

Bumps and pins the `certifi` version to address vulnerabilities from the dependabot scans.


qued (approved):
LGTM!

PR #100, opened 2022-12-14
fix: add __init__.py to the partition module

MthwRobinson:
### Summary

- Add `__init__.py` to the `partition` module so it can be imported.


qued (approved):
LGTM, I can import `partition`

PR #101, opened 2022-12-15
feat: `translate_text` cleaning brick

MthwRobinson:
### Summary

Adds a `translate_text` cleaning brick for translating content between languages. Uses the [Helsinki NLP](https://huggingface.co/Helsinki-NLP) models from `transformers` for translation. If the source language is not specified, the language is detected with `langdetect`.

### Testing

```python
  from unstructured.cleaners.translate import translate_text

  # Output is "I'm a Berliner!"
  translate_text("Ich bin ein Berliner!")

  # Output is "I can also translate Russian!"
  translate_text("Я тоже можно переводать русский язык!", "ru", "en")
```

### DIFF
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -450,7 +450,7 @@ Examples:
 ``translate_text``
 ------------------
 
-The ``translate_text`` cleaning bricks translates text between language. ``translate_text``
+The ``translate_text`` cleaning bricks translates text between languages. ``translate_text``
 uses the `Helsinki NLP MT models <https://huggingface.co/Helsinki-NLP>`_ from
 ``transformers`` for machine translation. Works for Russian, Chinese, Arabic, and many
 other languages.
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -452,7 +452,8 @@ Examples:
 
 The ``translate_text`` cleaning bricks translates text between language. ``translate_text``
 uses the `Helsinki NLP MT models <https://huggingface.co/Helsinki-NLP>`_ from
-``transformers`` for machine translation. 
+``transformers`` for machine translation. Works for Russian, Chinese, Arabic, and many
+other languages.
 
 Parameters:
 
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -447,6 +447,36 @@ Examples:
   extract_text_after(text, r"SPEAKER \d{1}:")
 
 
+``translate_text``
+------------------
+
+The ``translate_text`` cleaning bricks translates text between language. ``translate_text``
+uses the `Helsinki NLP MT models <https://huggingface.co/Helsinki-NLP>`_ from
+``transformers`` for machine translation. 
+
+Parameters:
+
+* ``text``: the input string to translate.
+* ``source_lang``: the two letter language code for the source language of the text. 
+  If ``source_lang`` is not specified,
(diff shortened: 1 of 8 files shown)


docs/source/bricks.rst
```diff
@@ -447,6 +447,37 @@ Examples:
   extract_text_after(text, r"SPEAKER \d{1}:")
 
 
+``translate_text``
+------------------
+
+The ``translate_text`` cleaning bricks translates text between language. ``translate_text``
```
qued, line 453:
```suggestion
The ``translate_text`` cleaning bricks translates text between languages. ``translate_text``
```

unstructured/cleaners/translate.py
```diff
@@ -0,0 +1,82 @@
...
+def _get_opus_mt_model_name(source_lang: str, target_lang: str):
+    """Constructs the name of the MarianMT machine translation model based on the
+    source and target language."""
+    return f"Helsinki-NLP/opus-mt-{source_lang}-{target_lang}"
+
+
+def _validate_language_code(language_code: str):
+    if not isinstance(language_code, str) or len(language_code) != 2:
+        raise ValueError(
+            f"Invalid language code: {language_code}. Language codes must be " "two letter strings."
```
qued, line 20: Is there a reason this is broken up into two strings? Did `black` join two lines or something?

unstructured/cleaners/translate.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+
+def translate_text(text, source_lang: Optional[str] = None, target_lang: str = "en"):
+    """Translates the foreign language text. Uses the stage
```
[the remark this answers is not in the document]
MthwRobinson: Fixed that up

unstructured/cleaners/translate.py
MthwRobinson: Yeah it was a long string and `black` must have combined them. I'll fix that

unstructured/cleaners/translate.py
```diff
@@ -0,0 +1,82 @@
...
+    text: str
+        The text to translate
+    target_lang: str
+        The two letter language code for the target langague. Defaults to "en".
+    source_lang: Optional[str]
+        The two letter language code for the language of the input text. If source_lang is
+        not provided, the function will try to detect it.
+    """
+    if text.strip() == "":
+        raise ValueError("An empty string was passed to translate_text.")
```
qued, line 39: What are the concerns around trying to translate an empty string? Is it just that we think this was a mistake, or does this cause problems if we allow it to go into the models?

docs/source/bricks.rst
MthwRobinson: Fixed

qued:
Is it intentional that `install-huggingface` is not included in the make target `install`? If so that's fine, but the help text on the make target says "installs all test, dev, and experimental requirements" so it seems like it should with that text.

unstructured/cleaners/translate.py
MthwRobinson: Empty strings were causing a problem with `langdetect` because there string didn't have any features it could use to detect the language. Realized we don't really need to run the translation though if it's an empty string. Now just returning the original text if it's empty.

MthwRobinson:
Added `install-huggingface` to `make install`. If it's already doing the `dev` requirements, might as well just install everything.

qued (approved):
LGTM!

PR #103, opened 2022-12-16
feat: Added `EmailElement` for email documents

mallorih:
### Summary

Adds an `EmailElement` class to `unstructured.documents` to make it easier to parse email documents into elements. Also added test that are basically the same as the tests for `Text` class but modified for the new `Name` class that is apart of the  `email_elements`.


MthwRobinson (changes requested):
Looking good, just a few small suggestions. Also, started to pull together some code that parses out the `text/html` content from the email file (link on the branch below). We can think through how we might want to work that in with your new data structure.

https://github.com/Unstructured-IO/unstructured/blob/robinson/partition-eml/unstructured/partition/email.py

unstructured/documents/email_elements.py
```diff
@@ -0,0 +1,149 @@
...
+        
+        {self.meta_data}
+
+        Body: 
+        
+        {self.body}
+
+        Attachment:
+
+        {self.attachment}
```
MthwRobinson, line 148: Could the attachments wind up being files? If so would it make more sense to print the filenames of the attachments instead of printing the attachment itself? Just so we don't wind up printing out the bytes for the attachment.

MthwRobinson (commented):
One questions on attachments, otherwise this is looking good.

unstructured/partition/text.py
```diff
@@ -0,0 +1,64 @@
...
+from unstructured.partition.text_type import (is_possible_narrative_text
+                                              ,is_possible_title
+                                              ,is_bulleted_text
+                                             )
+
+
+def partition_text(
+    filename: Optional[str] = None, file: Optional[IO] = None, text: Optional[str] = None
+) -> List[Element]:
+    """Partitions an .eml documents into its constituent elements.
```
MthwRobinson, line 16: `.eml` -> `.txt`

MthwRobinson: Or actually probably something "partitions a plain text document ..."

unstructured/partition/text.py
```diff
@@ -0,0 +1,64 @@
...
+    Parameters
+    ----------
+     filename
+        A string defining the target filename path.
+    file
+        A file-like object using "r" mode --> open(filename, "r").
+    text
+        The string representation of the .eml document.
```
MthwRobinson, line 24: `of the .eml document` -> `of a plain text document`

unstructured/partition/text.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    if not any([filename, file, text]):
+        raise ValueError("One of filename, file, or text must be specified.")
+
+    if filename is not None and not file and not text:
+        with open(filename, "r") as f:
+            msg = email.message_from_file(f)
```
MthwRobinson, line 31: These should all deal with plain text `.txt` documents rather than emails. And we'll use it (1) when we want to partition a plain text document and (2) if your partitioning an email and choose to process the `text/plain` content.

unstructured/partition/text.py
```diff
@@ -0,0 +1,64 @@
...
+        _text: str = str(text)
+        msg = email.message_from_string(_text)
+
+    else:
+        raise ValueError("Only one of filename, file, or text can be specified.")
+
+    content_map: Dict[str, str] = {
+        part.get_content_type(): part.get_payload() for part in msg.walk()
+    }
+    content = content_map.get("text/plain", "")
```
MthwRobinson, line 47: This should go in `partition_email` and we'll call `partition_text` from there if the user decides to process the `text/plain` content.

unstructured/partition/text.py
```diff
@@ -0,0 +1,64 @@
...
+    if not content:
+        raise ValueError("text/plain content not found in email")
+
+    content = re.split(r"\n\n\n|\n\n|\n", content)
```
MthwRobinson, line 51: I'd make `split_by_paragraph` a separate helper function and also `\r` and other line-ending variants a well. The links below provide some background on that.

- https://stackoverflow.com/questions/20056306/match-linebreaks-n-or-r-n
- https://stackoverflow.com/questions/1761051/difference-between-n-and-r

MthwRobinson (changes requested):
Added some initial comments. I know this one is still in flight, just a reminder to add the following to the final PR:

- Sphinx docs on any new bricks
- Unit tests
- A section in the `README` showing how to use `partition_text`.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.4-dev2
+
+* Add
```
MthwRobinson, line 3: You can add your bullet under `0.3.3-dev1` and then change the version to `0.3.3-dev2`. We'll move to `0.3.4-devx` once we do the `0.3.3` release.

CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
-## 0.3.3-dev1
+## 0.3.3-dev2
```
MthwRobinson, line 1: This should be `0.3.5-dev0` now since we just did a `0.3.4` release

unstructured/documents/email_elements.py
```diff
@@ -0,0 +1,151 @@
...
+
+        {self.meta_data}
+
+        Body:
+
+        {self.body}
+
+        Attachment:
+
+        {self.attachment_name}
```
MthwRobinson, line 150: Should this be a list of attachments in case there are multiple attachments?

unstructured/documents/email_elements.py
```diff
@@ -0,0 +1,151 @@
...
+    """An email class with it's attributes"""
+
+    def __init__(self, recipient: Recipient, sender: Sender, subject: Subject, body: BodyText):
+        self.recipient = recipient
+        self.sender = sender
+        self.subject = subject
+        self.body = body
+        self.received_info: ReceivedInfo
+        self.meta_data: MetaData
+        self.attachment: Attachment
```
MthwRobinson, line 127: Similar here, should this be `self.attachments: List[Attachment]` in case there are multiple?

unstructured/documents/email_elements.py
```diff
@@ -0,0 +1,151 @@
...
+        self.attachment_name: Attachment
```
MthwRobinson, line 128: Can we have a `Attachment.name` attribute instead of creating an extra attribute on `Email` for the attachment name?

MthwRobinson: Actually `name` is already inherited so I think we can probably just drop this attribute

unstructured/documents/email_elements.py
```diff
@@ -0,0 +1,151 @@
...
+class MetaData(Name):
+    """A text element for capturing header meta data of an email (e.g. Subject,
+    To, From, etc)."""
+
+    category = "MetaData"
+
+    pass
+
+
+class Attachment(Name):
```
MthwRobinson, line 108: Should attachment also have a `bytes` or file-like attribute that contains the actual attachment? If we don't have code to deal with attachments yet, we can also spin attachments off and deal with it in a separate PR.

MthwRobinson (changes requested):
Looking good, just a couple of questions related to attachments. Can approve if we want to spin attachments off and deal with it in a separate PR.

unstructured/documents/html.py
```diff
@@ -9,7 +9,7 @@
 
 from lxml import etree
 
-from unstructured.logger import get_logger
+from unstructured.logger import logger
```
MthwRobinson, line 12: Do you know why these files are showing up in your diffs? They look the same as what's on `main`

MthwRobinson (approved):
LGTM!

PR #105, opened 2022-12-19
build(deps): bump `certify` to incorporate security patches

MthwRobinson:
### Summary

Bumps and pins the `certify` version to incorporate security patches in version `2022.12.07`.


qued (approved):
LGTM!

PR #106, opened 2022-12-19
chore: logger should not be setting up a BasicConfig

yuming-long:
### Summary
Clean up loggers for `unstructured` so it doesn't have BasicConfig, which will be done in `unstructured-infra-tools`.
Pipeline repos could use a logger by:
```
# pipeline-api
from unstructured.logger import logger
```


cragwolfe:
LGTM.

More context: generally, packages should leave log formatting and at-what-level logs are emitted, i.e. the log handler, to the application. Otherwise an application has to undo / remove logging handlers if it wants a different behavior (yuck!).

reference: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

To configure this in the context of fastapi, a logging config file would include something like:
```
loggers:
  unstructured:
    level: INFO
    handlers:
      - default
    propagate: no
```

cragwolfe (approved):

PR #109, opened 2022-12-21
chore: add new pattern for dash bullet

mallorih:
This PR adds a new unicode character for dash `-` bullets, so the `is_bulleted_text` method can recognize plain text dashes.


MthwRobinson (changes requested):
Could you add a test? Otherwise looks good once the CHANGELOG entry and version bump are in.

MthwRobinson (approved):
LGTM, pending a CHANGELOG tweak

PR #110, opened 2022-12-21
fix: `partition_html` should process container divs that include text

MthwRobinson:
### Summary

Fixes an issue with `partition_html` that caused `div` tags that were both containers and included to text to be skipped. For example:

```html
<div>My name is Matt
    <div>This is my message</div>
</div>
```

would previously include `This is my message` but not `My name is Matt`. With the fix, all of the text is included.

### Testing

Using the sample doc from the `pipeline-email` repo:

```python

filename = "sample-docs/family_day.eml"
```

The output should look like:

```python
Hi All,

Get excited for our first annual family day!

They'll be face painting, a petting zoo, funnel cake and more.

Make sure to RSVP!

Best.

Mallori Harrell

Unstructured Technologies

Data Scientist
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
-## 0.3.5-dev0
+## 0.3.5-dev1
 
 * Add new pattern to recognize plain text dash bullets
 * Add test for bullet patterns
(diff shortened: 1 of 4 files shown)


mallorih (approved):
LGTM

PR #127, opened 2023-01-04
feat: add `convert_to_dataframe` staging brick

MthwRobinson:
### Summary

Adds the `convert_to_dataframe` staging brick for converting a list of document elements to a `pandas` DataFrame. The dataframe has a `text` column with the element text and a `type` column indicate the element type.

Re-runs `pip-compile` to add `pandas` as an explicit dependency and bump the `wheel` version. The `wheel` version pin is to resolve a security scan item.

### Testing

```python
  from unstructured.documents.elements import Title, NarrativeText
  from unstructured.staging.base import convert_to_dataframe

  elements = [Title(text="Title"), NarrativeText(text="Narrative")]
  df = convert_to_dataframe(elements)
```


qued (approved):
LGTM other than suggestions. Testing code worked for me!

PR #132, opened 2023-01-09
feat: generic `partition` brick with filetype detection

MthwRobinson:
### Summary

Adds a generic `partition` brick that detects the file type and then invokes the appropriate partitioning brick. In support of this functionality, this PR also adds the following:

- A module for detecting file type
- Includes `magic` in the dependencies and instructions for installing `libmagic` 
- Added support for processing file-like objects in `"rb"` mode to `partition_html` and `partition_eml`

### Testing

```python
  import docx

  document = docx.Document()
  document.add_paragraph("Important Analysis", style="Heading 1")
  document.add_paragraph("Here is my first thought.", style="Body Text")
  document.add_paragraph("Here is my second thought.", style="Normal")
  document.save("mydoc.docx")
  elements = partition(filename="mydoc.docx")

  with open("mydoc.docx", "rb") as f:
      elements = partition(file=f)
```

```python
  elements = partition(filename="example-docs/layout-parser-paper-fast.pdf")
```


MthwRobinson:
Test failure is due to `libmagic` detecting different MIME types for `.docx` and `.xlsx` files on Mac vs. Linux (see [this issue](https://github.com/python-openxml/python-docx/issues/545) for details). Currently working a fix.

cragwolfe (changes requested):
Also need to add support in the Dockerfile? I believe the package is `file-devel`

MthwRobinson:
Per offline convo, I'm removing the dockerfile since we're not using that for anything

cragwolfe (approved):

Makefile
```diff
@@ -20,7 +20,7 @@ install-base: install-base-pip-packages install-nltk-models
 install: install-base-pip-packages install-dev install-nltk-models install-test install-huggingface install-unstructured-inference
 
 .PHONY: install-ci
-install-ci: install-base-pip-packages install-test install-nltk-models install-huggingface install-unstructured-inference
+install-ci: install-base-pip-packages install-test install-nltk-models install-huggingface install-unstructured-inference local-inference
```
qued, line 23:
```suggestion
install-ci: install-base-pip-packages install-test install-nltk-models install-huggingface install-unstructured-inference install-local-inference
```

PR #143, opened 2023-01-11
chore: Fix parse received data

mallorih:
The `_parse_received_data` function current returns `None` when the `datetimestamp` variable is set. This PR fixes this bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.1-dev0
+
+* Fix bug in `_parse_received_data`.
+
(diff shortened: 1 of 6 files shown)


mallorih:
Weird, I accidentally reverted the change.

MthwRobinson (changes requested):
Changes look could, could you add a test too?

mallorih:
@MthwRobinson I changed my fix in the Name class in `email_elements` a little and added the test.

MthwRobinson (approved):
LGTM!

PR #144, opened 2023-01-12
feat: partition image

qued:
Adds `partition_image` to partition image file types, which is integrated into the `partition` brick. This relies on the 0.2.2 version of `unstructured-inference`.

#### Testing:

From a Python terminal:
```python

res1 = partition_image(filename="example-docs/example.jpg", url=None)
res2 = partition(filename="example-docs/example.jpg")
res1 == res2

```
Should run and output should be `True`.


unstructured/partition/image.py
```diff
@@ -0,0 +1,32 @@
+from typing import List, Optional
+
+from unstructured.documents.elements import Element
+from unstructured.partition.pdf import partition_pdf_or_image
+
+
+def partition_image(
+    filename: str = "",
+    file: Optional[bytes] = None,
+    url: Optional[str] = "https://ml.unstructured.io/",
```
MthwRobinson, line 10: Should the url here default to `/layout/image`?

unstructured/partition/image.py
```diff
@@ -0,0 +1,32 @@
...
+) -> List[Element]:
+    """Parses an image into a list of interpreted elements.
+    Parameters
+    ----------
+    filename
+        A string defining the target filename path.
+    file
+        A file-like object as bytes --> open(filename, "rb").
+    template
+        A string defining the model to be used. Default None uses default model ("layout/pdf" url
```
MthwRobinson, line 22: Should `layout/pdf` -> `layout/image` in the doc string?

MthwRobinson (commented):
Couple questions on the URL for the partitioning via API for images. Also, do we need to update the deployed inference API to include the image endpoint or is that done already?

unstructured/partition/image.py
MthwRobinson: Per offline convo, we're going to do write up a design doc for the inference API and then circle back to this.

MthwRobinson (approved):
LGTM pending the one doc-string update!

MthwRobinson (approved):
LGTM pending the additional test for partitioning via API

PR #164, opened 2023-01-18
chore: return `Element` objects in `partition_pdf` and `partition_image`

### LINKED ISSUE #147: `partition_pdf` should return `unstructured` document `Element` objects
Currently, if you run `elements = partition_pdf("example-docs/layout-parser-paper.pdf", url=None)` and look at `elements[0]`, the type is `LayoutElement`, which is a type from `unstructured-inference`. Instead, we should return standard `unstructured` document elements for consistency with the other partition functions. As part of this, we may consider whether to include extract information such as coordinates as optional attributes in `Element`.

MthwRobinson:
### Summary

Closes #147. Updates `partition_pdf` and `partition_image` to return standard `unstructured` `Element` objects instead of the `DocumentLayout` object from `unstructured_inferenence`.

Fixes the healthcheck url when partitioning via API. Previously it constructed the URL as `http://127.0.0.1:8000/layout/pdf/healthcheck` instead of `http://127.0.0.1:8000/healthcheck`, resulting in a `404`.

###

- Run `unstructured-inference` locally with `make run-app-dev`
- Run the following to verify partitioning via API. The outputs should be `unstructured` `Element` objects.

```python

filename = "example-docs/layout-parser-paper.pdf"
elements = partition_pdf_or_image(filename=filename, url="http://127.0.0.1:8000")
```

- Run the following to verify partitioning locally

```python

filename = "example-docs/layout-parser-paper.pdf"
elements = partition_pdf_or_image(filename=filename, url=None)
```


qued (approved):
LGTM!

PR #165, opened 2023-01-19
feat: add ability to parse `LayoutElement` lists

### LINKED ISSUE #163: Lists from LayoutParser should be split into `ListItem`s
### Summary

Currently if you run code such as:

```python

filename = "example-docs/layout-parser-paper.pdf"
elements = partition_pdf(filename=filename)
```

List elements are a single `Text` blog that looks like:

```
'1. An oﬀ-the-shelf toolkit for applying DL models for layout detection, character recognition, and other DIA tasks (Section 3) 2. A rich repository of pre-trained neural network models (Model Zoo) that underlies the oﬀ-the-shelf usage 3. Comprehensive tools for eﬃcient document image data annotation and model tuning to support diﬀerent levels of customization 4. A DL model hub and community platform for the easy sharing, distribu- tion, and discussion of DIA models and pipelines, to promote reusability, reproducibility, and extensibility (Section 4)'
```

We should split up the list so each element in the list is its own `ListItem` element.

MthwRobinson:
### Summary

Closes #163. Adds the ability to split lists that are detected in LayoutParse with the List type instead of treating them as a single block of text.

### Testing

The following should now show `ListItem` objects in `elements` after the document is parsed.

```python

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
-## 0.4.4-dev0
+## 0.4.4-dev1
 
 * Updated `partition_pdf` and `partition_image` to return `unstructured` `Element` objects
 * Fixed the healthcheck url path when partitioning images and PDFs via API
 * Adds an optional `coordinates` attribute to document objects
 * Adds `FigureCaption` and `CheckBox` document elements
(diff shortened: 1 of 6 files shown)


qued (approved):
LGTM!

PR #173, opened 2023-01-24
fix: handle xml filetype detection on amazon linux

MthwRobinson:
### Summary

Updates `detect_filetype` to properly detect `.xml` and `.html` files on Amazon Linux systems. On Amazon linux, the MIME type is `application/xml` instead of `text/xml`. The logic now treats any MIME type that ends in `.xml` as an XML file. Also updates the tests to account for different operating systems potentially returning different MIME types from buffer.

### Testing

Tests pass on an Amazon Linux instance.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.4
+## 0.4.4-dev3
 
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM, verified XML tests fail on Amazon linux before this change but pass after.

PR #177, opened 2023-01-25
fix: cleanup from live `.docx` tests

MthwRobinson:
### Summary

This PR captures a number of small changes made while testing the `partition` code on real world `.docx` example documents. These changes include:

- Loosen the default caps ratio for `exceeds_cap_threshold` and add the ability to control the threshold with the `NARRATIVE_TEXT_CAP_THRESHOLD` env var. 
- Text that is all capitalized does not go through the the caps threshold text.
- Include unknown text types as generic `Text` elements. This change was also made for`.txt` and `.html`.
- Tighter equality conditions for `Text` objects. Now `category` is checked in addition to `text` and `coordinates`.
- The `Body Text` style no longer defaults to `NarrativeText` for `.docx`. It now defaults to `Text`. This is because the style is insufficient to be confident that text is narrative.
- All caps text is lower cased before checking for verbs. The upper case text seems to confuse the POS tagger in some cases.
- Add patterns for matching US city/state/zip address. These were previously incorrectly identified variously as narrative text or titles

#### Adjust cap threshold with env var

```python
import os
from unstructured.partition.text_type import is_possible_narrative_text

is_possible_narrative_text("This All Starts With Caps") # Should be False
os.environ["NARRATIVE_TEXT_CAP_THRESHOLD"] = "1.0"
is_possible_narrative_text("This All Starts With Caps") # Should be True
```

#### Caps ratio does not apply to all upper text

```python
import os
from unstructured.partition.text_type import exceeds_cap_ratio

exceeds_cap_ratio("THIS IS A SENTENCE IN ALL CAPS.") # Should be False
```

#### Include generic `Text` elements for HTML and text files

```python

html = "<p>Hi everyone,</p><p>This is a very important email!</p>"
# First item should be HTMLText, second should be HTMLNarrativeText
partition_html(text=html)
```

```python

text = "Hi everyone,\n\nThis is a very important email!"
# First item should be Text, second should be NarrativeText
partition_text(text=text)
```

#### Equality checks for Text elements

```python

assert (NarrativeText(text="hi") == Title(text="hi")) is False
assert (Title(text="hi") == Title(text="hi")) is True
```

#### All caps text is lowered before checking for verbs

```python
from unstructured.partition.text_type import contains_verb

# Should be True, was previously False
contains_verb("HE RUNS AROUND")
```

#### Capturing addresses

```python

text = """
VERY IMPORTANT MEMO
DOYLESTOWN, PA 18901
"""
# First element should be title and second element should be Address
partition_text(text=text)
```


qued:
First PR test fails, all others succeed: 
image

docs/source/bricks.rst
```diff
@@ -277,8 +279,8 @@ for consideration as a title. The function performs the following checks:
 
 * Empty text cannot be a title
 * Text that is all numeric cannot be a title
-* If a title contains more than one sentence that exceeds a certain length, it cannot be a title.
-  Sentence length threshold is controlled by the ``sentence_min_length`` kwarg and defaults to 5.
+* If a title contains more than one sentence that exceeds a certain length, it cannot be a title. Sentence length threshold is controlled by the ``sentence_min_length`` kwarg and defaults to 5.
+* If a segment of text ends in a command, it is not considered a potential title. This is to avoid salutations like "To My Dearest Friends," getting flagged as titles.
```
qued, line 283:
```suggestion
* If a segment of text ends in a comma, it is not considered a potential title. This is to avoid salutations like "To My Dearest Friends," getting flagged as titles.
```

unstructured/partition/text_type.py
```diff
@@ -27,6 +39,9 @@ def is_possible_narrative_text(text: str, cap_threshold: float = 0.3) -> bool:
         logger.debug(f"Not narrative. Text is all numeric:\n\n{text}")
         return False
 
+    # NOTE(robinson): it gets read in from the environment as a string so we need to
+    # case it to a float
```
qued, line 43:
```suggestion
    # cast it to a float
```

MthwRobinson:
@qued - Update the first PR test. I used the wrong function in the original version.

qued (approved):
LGTM, PR tests all work now!

PR #179, opened 2023-01-26
fix: no `UserWarning` when `partition_pdf` is called

### LINKED ISSUE #174: Get rid of `UserWarning` on the first call to `partition`
Currently, we get a `UserWarning` the first time you call `partition_pdf` or `partition_image`. The goal of this issue is add handling so that this `UserWarning` no longer appears.

```python

partition("example-docs/layout-parser-paper-fast.pdf")
```

```python
/home/ec2-user/anaconda3/envs/unstructured/lib/python3.8/site-packages/torch/functional.py:504: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at ../aten/src/ATen/native/TensorShape.cpp:3190.)
  return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
```

MthwRobinson:
### Summary

Closes #174. Suppresses the `UserWarning` when `partition_pdf` is called. Also fixes up a few of the PDF tests in `test_auto.py`.

### Testing

The following should no longer output a warning

```python

partition(filename=filename)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.5-dev0
+## 0.4.5-dev1
 
 * Loosen the default cap threshold to `0.5`.
 * Add a `NARRATIVE_TEXT_CAP_THRESHOLD` environment variable for controlling the cap ratio threshold.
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM, no warning when I run the test code on this branch, warning occurs on main branch.

PR #188, opened 2023-01-31
fix: titles and narrative text need at least one english word

MthwRobinson:
### Summary

Adds a check that requires narrative text and titles to have at least one English words. Prevents office symbols like `ASDF-JQL` and similar items from being flagged as titles.

### Testing

```python
from unstructured.partition.text_type import is_possible_narrative_text, is_possible_title

is_possible_title("PARROT BEAK") # should be True
is_possible_title("ASDF-JQL") # should be False

is_possible_narrative_text("The parrot is squawking!") # should be True
is_possible_narrative_text("ASDJ JQL; QWER. TUSP AHSAD.") # should be False
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.5-dev1
+## 0.4.5-dev2
 
 * Loosen the default cap threshold to `0.5`.
 * Add a `NARRATIVE_TEXT_CAP_THRESHOLD` environment variable for controlling the cap ratio threshold.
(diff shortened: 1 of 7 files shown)


qued (approved):
My only concern is that something like terse titles with acronyms (Like "3. SBIRS") may get misclassified, but that's sort of an edge case. LGTM!

MthwRobinson:
Yeah that definitely makes sense. We can always adjust the logic if we see that kind of thing popping up a lot.

PR #189, opened 2023-02-01
feat: set a user controlled max word length for titles

MthwRobinson:
### Summary

Adds a max word length for titles to avoid long text with title-like characteristics from getting categorized as text. Users can use the `title_max_word_length` kwarg in `is_possible_title` or the `UNSTRUCTURED_TITLE_MAX_WORD_LENGTH` env var control this setting. Also added an `UNSTRUCTURED_` prefix to the narrative text caps threshold environment variable.

### Testing

The following should return `False` for `is_possible_title`. Previously it returned `True`.

```python

is_possible_title("The box contains: 5 GOLDEN RINGS, 4 CALLING BIRDS, 3 FRENCH HENS, 2 TURTLE DOVES AND A PARTRIDGE IN A PEAR TREE")
```

### DIFF
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -382,9 +382,7 @@ Examples:
 Determines if the section of text exceeds the specified caps ratio. Used in
 ``is_possible_narrative_text`` and ``is_possible_title``, but can be used independently
 as well. You can set the caps threshold using the ``threshold`` kwarg. The threshold
-defaults to ``0.3``. Only runs on sections of text that are a single sentence.
-The environment variable takes precedence over the kwarg. The caps ratio
-check does not apply to text that is all capitalized.
+defaults to ``0.3``. Only runs on sections of text that are a single sentence. The caps ratio check does not apply to text that is all capitalized.
 
 Examples:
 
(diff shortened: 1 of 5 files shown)


docs/source/bricks.rst
```diff
@@ -279,7 +279,10 @@ The ``is_possible_title`` function determines if a section of text is a candidat
 for consideration as a title. The function performs the following checks:
 
 * Empty text cannot be a title
-* Text that is all numeric cannot be a title
+* Text that is all numeric cannot be a title.
+* If a title contains to many words it is not a title. The default max length is ``15``. You can change the max length with
```
qued, line 283:
```suggestion
* If a title contains too many words it is not a title. The default max length is ``15``. You can change the max length with
```

docs/source/bricks.rst
```diff
@@ -380,8 +383,7 @@ Determines if the section of text exceeds the specified caps ratio. Used in
 ``is_possible_narrative_text`` and ``is_possible_title``, but can be used independently
 as well. You can set the caps threshold using the ``threshold`` kwarg. The threshold
 defaults to ``0.3``. Only runs on sections of text that are a single sentence.
-You can also set the threshold using the ``NARRATIVE_TEXT_CAP_THRESHOLD`` environment
-variable. The environment variable takes precedence over the kwarg. The caps ratio
+The environment variable takes precedence over the kwarg. The caps ratio
```
qued, line 386: Don't we want to give the name of the env variable here (it now seems to be `UNSTRUCTURED_NARRATIVE_TEXT_CAP_THRESHOLD`)?

qued (approved):
LGTM, just a nit and a question.

docs/source/bricks.rst
MthwRobinson: Ooops actually meant to get rid of that section since the env var applies to `is_possible_narrative_text` and not `exceeds_caps_ratio`. Fixed

PR #191, opened 2023-02-02
fix: more english words; split on punctuation

MthwRobinson:
### Summary

Adds a broader list of English words and also splits on punctuation during the English word check to capture titles that a "jammed together" like `1.A.RISKS`. Also updated the default non-alpha threshold since it was a little too aggressive.

### Testing

The following should return `True`. It currently returns `False` on `main`.

```python

is_possible_title("1.A.RISKS")
```

Also run `make install &&  python setup.py install` and confirm that `english-words.txt` is present at `~/.pyenv/versions/3.8.13/envs/english_words/lib/python3.8/site-packages/unstructured-0.4.5.dev5-py3.8.egg/unstructured/nlp`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.5-dev6
+## 0.4.5-dev5
 
 * Loosen the default cap threshold to `0.5`.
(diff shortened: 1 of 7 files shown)


qued (approved):
One minor typo, otherwise LGTM!

PR #193, opened 2023-02-03
fix: preserve the order of shapes in `partition_pptx` output

MthwRobinson:
### Summary

Orders the elements extracted from `.pptx` files based on their position on the page. Proceeds from top to bottom and left to right. Previously the elements were not in the correct order.

### Testing

Run the following using the filename Matt provides you. The outputs should go top to bottom and left to right based on the position of the shapes.

```python

partition(filename=filename)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.5-dev5
+## 0.4.6
 
 * Loosen the default cap threshold to `0.5`.
 * Add a `UNSTRUCTURED_NARRATIVE_TEXT_CAP_THRESHOLD` environment variable for controlling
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Checks that titles and narrative text are at least 50% alpha characters.
 * Restricts titles to a maximum word length. Adds a `UNSTRUCTURED_TITLE_MAX_WORD_LENGTH`
   environment variable for controlling the max number of words in a title.
+* Updated `partition_pptx` to order the elementso n the page
```
qued, line 16:
```suggestion
* Updated `partition_pptx` to order the elements on the page
```

qued (approved):
LGTM, just 1 typo

PR #205, opened 2023-02-07
feat: optional page breaks for `.pptx`, `.pdf`, `.html` and images

MthwRobinson:
### Summary

Adds a `include_page_breaks` kwarg to optionally add page breaks for filetypes that support it (`.pptx`, `.pdf`, `.html` and images). This is especially helpful for documents like `.pptx` where content doesn't naturally flow from slide to slide. `include_page_breaks` defaults to `False` for backward compatibility. Note, `.docx` does not support paging due to the behavior described in [this SO post](https://stackoverflow.com/questions/36193159/page-number-python-docx). Page breaks are created by the renderer as opposed to the `.docx` source files.

### Testing

Run the following code. You should see `PageBreak` in `elements`.

```python

elements = partition(filename=filename, include_page_breaks=True)
```

### DIFF
--- a/unstructured/partition/__init__.py
+++ b/unstructured/partition/__init__.py
@@ -1,8 +1,6 @@
 import requests  # type: ignore
-from typing import BinaryIO, List, Optional, Union, Tuple, Mapping
+from typing import BinaryIO, Optional, Union, Tuple, Mapping
 from urllib.parse import urlsplit
-
-from unstructured.documents.elements import Element
 
 
 def _partition_via_api(
@@ -11,7 +9,8 @@ def _partition_via_api(
     url: str = "https://ml.unstructured.io/layout/pdf",
     token: Optional[str] = None,
     data: Optional[dict] = None,  # NOTE(alan): Remove after different models are handled by routing
-) -> List[Element]:
+    include_page_breaks: bool = False,
+):
     """Use API for partitioning."""
     if not filename and not file:
         raise FileNotFoundError("No filename nor file were specified")
@@ -40,6 +39,15 @@ def _partition_via_api(
 
     if response.status_code == 200:
         pages = response.json()["pages"]
-        return [element for page in pages for element in page["elements"]]
+        num_pages = len(pages)
+        elements = list()
+        for i, page in enumerate(pages):
+            for element in page["elements"]:
+                elements.append(element)
+            if include_page_breaks and i < num_pages - 1:
+                elements.append({"type": "PageBreak"})
+
+        return elements
+
     else:
         raise ValueError(f"response status code = {response.status_code}")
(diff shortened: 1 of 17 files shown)


unstructured/partition/__init__.py
```diff
@@ -1,17 +1,16 @@
...
-from unstructured.documents.elements import Element
-
 
 def _partition_via_api(
     filename: str = "",
     file: Optional[bytes] = None,
     url: str = "https://ml.unstructured.io/layout/pdf",
     token: Optional[str] = None,
     data: Optional[dict] = None,  # NOTE(alan): Remove after different models are handled by routing
-) -> List[Element]:
```
qued, line 14: Shouldn't we keep the return type hint?

unstructured/partition/__init__.py
```diff
@@ -40,6 +39,15 @@ def _partition_via_api(
...
     if response.status_code == 200:
         pages = response.json()["pages"]
-        return [element for page in pages for element in page["elements"]]
+        num_pages = len(pages)
+        elements = list()
+        for i, page in enumerate(pages):
+            for element in page["elements"]:
+                elements.append(element)
+            if include_page_breaks and i < num_pages - 1:
+                elements.append({"type": "PageBreak"})
```
qued, line 48: Just curious, why'd you decide to handle this here rather than at the endpoint?

unstructured/partition/html.py
```diff
@@ -57,9 +57,16 @@ def partition_html(
...
         raise ValueError("Only one of filename, file, or text can be specified.")
 
+    elements: List[Element] = []
+    num_pages = len(document.pages)
+    for i, page in enumerate(document.pages):
+        for element in page.elements:
+            elements.append(element)
+        if include_page_breaks and i < num_pages - 1:
+            elements.append(PageBreak())
+
```
qued, line 71: This looks pretty similar to [the block in pdf.py](https://github.com/Unstructured-IO/unstructured/blob/feat/page-breaks/unstructured/partition/pdf.py#L138-L144), maybe worth factoring out to a function.

qued (approved):
LGTM, just a typo, some suggestions, and a question.

unstructured/partition/__init__.py
MthwRobinson: The type hint for this one was actually wrong (it returns a list of dicts), but when I updated it `mypy` had a meltdown. I'll take another look and see if I can fix it.

unstructured/partition/html.py
MthwRobinson: Good call! I'll add a function to `common.py`

unstructured/partition/__init__.py
MthwRobinson: Main reason is that the endpoint returns representation of the document layout, so you already have `document["pages"]`. You don't really need to do anything additional to track pages there since it's already doing it.

unstructured/partition/html.py
MthwRobinson: Refactored and added to `common.py`

PR #206, opened 2023-02-08
Update bricks.rst

djacobs7:
Changed whitespace to extra_whitespace in documentation, to match options text.


MthwRobinson (approved):
LGTM! Thanks for the fix.

PR #207, opened 2023-02-08
docs: core concepts training notebook

MthwRobinson:
### Summary

Adds a training notebook cover core concepts in the `unstructured` library. Also adds a `to_dict` method to element objects.

### Testing

Run and check the `examples/training/0-Core Concepts.ipynb` notebook.


examples/training/0-Core Concepts.ipynb
```diff
@@ -0,0 +1,354 @@
...
+   "source": [
+    "elements"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e8009e7b",
+   "metadata": {},
+   "source": [
+    "You can also partitioning a document from a file-like object instead of a filename as follows:"
```
qued, line 102:
```suggestion
    "You can also partition a document from a file-like object instead of a filename as follows:"
```

examples/training/0-Core Concepts.ipynb
```diff
@@ -0,0 +1,354 @@
...
+    "        - `FigureCaption`\n",
+    "        - `NarrativeText`\n",
+    "        - `ListItem`\n",
+    "        - `Title`\n",
+    "        -  `Address`\n",
+    "    - `CheckBox`\n",
+    "    - `Image`\n",
+    "    - `PageBreak`\n",
+    "    \n",
+    "Other element types that we will add in the future include tables and figures. Different partioning functions use different methods for determining the element type and extracting the associated content. Document elements has a `str` representation. You can print them using the snippet below."
```
qued, line 190:
```suggestion
    "Other element types that we will add in the future include tables and figures. Different partioning functions use different methods for determining the element type and extracting the associated content. Document elements have a `str` representation. You can print them using the snippet below."
```

examples/training/0-Core Concepts.ipynb
```diff
@@ -0,0 +1,354 @@
...
+    "    print(element)\n",
+    "    print(\"\\n\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9d8ab7d6",
+   "metadata": {},
+   "source": [
+    "One helpful aspect of document elements is that they allow you to cut a document down to the elements that you need for your particular use case. For example, if you're training a summarization model you may only want to include narrative text for model training. You'll notice in the output above that the oput includes a lot of titles and other content that may not be suitable for a summarization model. The following code shows how you can limit your output to only narrative text with at least two sentences. As you can see, the output now only contains "
```
qued, line 235:
```suggestion
    "One helpful aspect of document elements is that they allow you to cut a document down to the elements that you need for your particular use case. For example, if you're training a summarization model you may only want to include narrative text for model training. You'll notice that the output above includes a lot of titles and other content that may not be suitable for a summarization model. The following code shows how you can limit your output to only narrative text with at least two sentences. As you can see, the output now only contains narrative text."
```

examples/training/0-Core Concepts.ipynb
```diff
@@ -0,0 +1,354 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "id": "5d13fc38",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "output = [el.to_dict() for el in elements]"
```
qued, line 287: This seems like what `convert_to_isd` does. Should we just show them how to use that?

qued (approved):
LGTM with 3 nits and 1 suggestion.

examples/training/0-Core Concepts.ipynb
MthwRobinson: Yeah I'll switch this to `convert_to_isd`

PR #216, opened 2023-02-14
docs: add info about os dependencies

qued:
Added information about OS dependencies `poppler` and `tesseract`.


cragwolfe (approved):
Thanks for adding this!

PR #217, opened 2023-02-14
docs: update make targets

qued:
Update suggested make targets for installation in README, and also add suggestion about using `make install-local-inference`.


cragwolfe (approved):
Thanks again :)

PR #218, opened 2023-02-14
feat: Sample ingest project with S3 connector

cragwolfe:
See the README updates (i.e. https://github.com/Unstructured-IO/unstructured/blob/crag/ingest/Ingest.md ).

End-to-end test added to CI to verify ingest runs and outputs match expected.

Also updates some requirements since `make pip-compile` needed to be run for the new `requirements/ingest-s3.txt`


.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+omit =
+    unstructured/ingest/*
```
MthwRobinson, line 3: Why don't we want coverage on the ingest files?

Ingest.md
```diff
@@ -0,0 +1,36 @@
+# Batch Processing Documents
+
+Several classes are provided in the Unstructured library
+to enable effecient batch processing of documents.
+
+## The Abstractions
+
+```mermaid
```
MthwRobinson, line 8: 🧜‍♀️

Makefile
```diff
@@ -50,6 +50,11 @@ install-dev:
 install-build:
 	pip install -r requirements/build.txt
 
+## install-ingest-s3:       install requirements for the s3 connector
+.PHONY: install-ingest-s3
```
MthwRobinson, line 54: Should the the s3 ingest requirements be optional dependencies in `setup.py`?

examples/ingest/s3-small-batch/main.py
```diff
@@ -0,0 +1,53 @@
...
+    def run(self):
+        self.initialize()
+
+        self.doc_connector.fetch_docs()
+
+        # fetch the list of lazy downloading IngestDoc obj's
+        docs = self.doc_connector.fetch_docs()
+
+        # Debugging tip: use this instead of the mp.Pool loop
+        #self.doc_processor_fn(docs[0])
```
MthwRobinson, line 31: Reminder to remove this comment

MthwRobinson (commented):
Couple of inline questions/comments

examples/ingest/s3-small-batch/main.py
cragwolfe: this was intentionally left as a hint to future users

.coveragerc
cragwolfe: ingest is not the primary point of the repo yet, it's provided more as a way for users to quickly roll their own. and it is hard to test multiprocessing well with codecov (i think).

there is an an end-to-end test which IMO is more valuable than a bunch of unittests IMO. can prioritize unittests later?

examples/ingest/s3-small-batch/main.py
cragwolfe: updated comment to make that more clear

Makefile
cragwolfe: also added to setup.py (so following the local-inference pattern here)

.coveragerc
MthwRobinson: Yeah that works for me. Just with the cautionary note that I always find it slower to add unit tests after the fact

Makefile
MthwRobinson: 🎉

MthwRobinson (approved):
Comments addressed - LGTM!

PR #222, opened 2023-02-14
chore: default to local inference for `partition_pdf` and `partition_image`

MthwRobinson:
### Summary

Updates `partition_pdf` and `partition_image` to use local inference by default since the hosted model inference endpoint is no longer running and most of our users are running inference locally.

### Testing

The following snippets should run in local inference mode (if they run at all at the moment, that means the changes work).

```python
 elements = partition_pdf("example-docs/layout-parser-paper-fast.pdf")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.4.9-dev0
+## 0.4.9-dev1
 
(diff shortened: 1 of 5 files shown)


qued (approved):
LGTM!

PR #223, opened 2023-02-14
Fix/adding optional kwarg encoding

sparkbrains:


test_unstructured/partition/test_email.py
```diff
@@ -126,7 +126,7 @@ def test_partition_email_from_text_file_with_headers():
 
 def test_partition_email_from_text():
     filename = os.path.join(DIRECTORY, "..", "..", "example-docs", "fake-email.eml")
-    with open(filename, "r") as f:
+    with open(filename, "r", encoding="utf-8") as f:
```
MthwRobinson, line 129: These tests should still pass without modification in this PR

MthwRobinson (changes requested):
Please add a summary section and testing instructions to your PR. Added one inline comment. Can you also add a test using a document that fails with the default encoding, but works if you pass in the correct encoding?

PR #224, opened 2023-02-14
feat: ability to skip English language specific checks with env var

MthwRobinson:
### Summary

Adds the ability to skip English language specific text type checks by setting the `UNSTRUCTURED_LANGUAGE` environment variable to `""` or by setting the `language` kwarg to `""`.

### Testing

```python
from unstructured.partition.text_type import is_possible_narrative_text, is_possible_title

text = "Я говорю по-русски. Вы тоже?"

is_possible_narrative_text(text, language="en") # should be False
is_possible_narrative_text(text, language="") # should be True

text = "Риски"

is_possible_narrative_text(text, language="en") # should be False
is_possible_narrative_text(text, language="") # should be False
is_possible_title(text, language="en") # should be False
is_possible_title(text, language="") # should be True
```

### DIFF
--- a/unstructured/partition/text_type.py
+++ b/unstructured/partition/text_type.py
@@ -49,6 +49,7 @@ def is_possible_narrative_text(
         logger.debug(f"Not narrative. Text is all numeric:\n\n{text}")
         return False
 
+    language = os.environ.get("UNSTRUCTURED_LANGUAGE", language)
     if language == "en" and not contains_english_word(text):
         return False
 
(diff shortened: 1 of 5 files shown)


unstructured/partition/text_type.py
```diff
@@ -49,6 +49,7 @@ def is_possible_narrative_text(
         logger.debug(f"Not narrative. Text is all numeric:\n\n{text}")
         return False
 
+    language = os.environ.get("UNSTRUCTURED_LANGUAGE", language)
```
cragwolfe, line 52: seems fine for now, but an env var doesn't seem great if processing several documents with different languages. is there a low overhead lang detection lib we could use that is the default?

cragwolfe (approved):

PR #227, opened 2023-02-16
feat: Ingest CLI flags and test fixture updates

cragwolfe:
* Many command line options added. The sample ingest project is now an easy to use CLI (no code editing necessary), capable of processing large numbers of files from S3 in a re-entrant manner. See [Ingest.md](https://github.com/Unstructured-IO/unstructured/blob/ingest/tweaks/Ingest.md).
* Fixes issue where text fixtures had been truncated
  * Adds a check to make sure this doesn't happen again
* Moves fixture outputs for the existing connector one subdir lower, to make room for future connector outputs.


unstructured/ingest/connector/s3_connector.py
```diff
@@ -170,11 +188,16 @@ def fetch_docs(self):
...
         return [
             S3IngestDoc(
-                self.config.s3_bucket,
+                self.config,
                 s3_key,
-                self.config.output_dir,
-                self._tmp_download_dir,
-                self.config.anonymous,
             )
+            # S3IngestDoc(
```
MthwRobinson, line 194: Should this commented out section be here?

MthwRobinson (approved):
One comment on some commented out code, otherwise LGTM

unstructured/ingest/connector/s3_connector.py
cragwolfe: no, removed. thanks for flagging.

PR #233, opened 2023-02-16
fix: `ElementMetadata` serializes when the filename is a `Path` object

### LINKED ISSUE #232: ISD dictionaries are not JSON serializable if the filename has a POSIX path
Currently if a filename has a path like "../../my-file.txt", the ISD dictionary is not JSON serializable. The goal of this issue is to only include the filename and not the full path in the metadata so that the ISD dictionary is JSON serializable.

MthwRobinson:
### Summary

Closes #232. Ensures ISD dictionaries are JSON serializable when the filename is a `Path` object.

### Testing

The following code should work. Previously you'd get a `JSONDecode` error.

```python
    import pathlib
    import json

    from unstructured.documents.elements import ElementMetadata, Title, NarrativeText
    from unstructured.staging.base import convert_to_isd
    

    metadata = ElementMetadata(filename=pathlib.PosixPath("../../fake-file.txt"))
    elements = [
        Title(text="Title 1", metadata=metadata),
        NarrativeText(text="Narrative 1", metadata=metadata),
    ]
    output = convert_to_isd(elements)
    json.dumps(output)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.10
+
+* Fixes `ElementMetadata` so that it's JSON serializable when the filename is a `Path` object.
+
 ## 0.4.9
 
 * Added ingest modules and s3 connector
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM!

PR #236, opened 2023-02-16
feat: add `partition_doc` for `.doc` files

MthwRobinson:
### Summary

Adds a `partition_doc` function for partition Word documents in the older `.doc` format. Relies on `libreoffice` to convert the documents to `.docx` and then calls `partition_docx`. As noted in [this SO discussion](https://stackoverflow.com/questions/38468442/multiple-doc-to-docx-file-conversion-using-python), for windows users you can also do this directly with the `win32com` client. Will address that in a follow on issue.

### Testing

The following code should work on this branch.

```python

  elements = partition_doc(filename="example-docs/fake.doc")
```


qued (approved):
LGTM!

PR #246, opened 2023-02-21
build(deps): automatically download `nltk` models when required

### LINKED ISSUE #239: Download `nltk` models if they're not available so people don't have to run `make install-nltk-models`
Currently you need to run `make install-nltk-models` to download the NLTK pre-reqs in `unstructured`. We can make this easier by downloading those models if they're not available when they're called in the relevant modules.

MthwRobinson:
### Summary

Closes #239. Simplifies the installation processes by checking for `nltk` models and downloading them if they are not already present. Users no longer need to run `make install-nltk-models` (or equivalent) before using `unstructured`.

### Testing`

`make install-nltk-models` was removed from CI, so if the build is passing this should be good to go. You can also remove `~/nltk_data` and run the tests locally.


benjats07 (approved):
LGTM

PR #251, opened 2023-02-22
feat: add allow custom parsers in partition_html

### LINKED ISSUE #250: partition_html incorrect encoding
**Describe the bug**
`partition_html` function incorrectly handles cyrillic (and probably other non-latin) text.
This happens when an html do not have correct encoding specified in a meta tag.

**To Reproduce**
```
text = '<html><body><p>Привет</p></body></html>'
parts = partition_html(text=text)
print(parts[0].text)
```
Prints `Ð\x9fÑ\x80Ð¸Ð²ÐµÑ\x82`

**Expected behavior**
```
text = '<html><body><p>Привет</p></body></html>'
parts = partition_html(text=text)
print(parts[0].text)
```
Prints `Привет`

 - OS: mac
 - Python version: 3.9.9

**Additional context**
As a possible solution, I would suggest to add an ability to provide your own parser to `partition_html` function.
This way we could use `lxml.etree.HTMLParser` with initialized with desired `encoding` argument.

asai95:
Close #250.

This will allow `partition_html` to use a custom XMLParser or HTMLParser.
It can be useful if one needs to specify additional arguments to these parsers (not only built-in `remove_comments=True`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.12-dev2
+
+* Add `parser` parameter to `partition_html`.
+
 ## 0.4.12-dev1
 
 * Adds console_entrypoint for unstructured-ingest, other structure/doc updates related to ingest.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,7 @@
-## 0.4.12-dev2
-
-* Add `parser` parameter to `partition_html`.
-
 ## 0.4.12-dev1
 
 * Adds console_entrypoint for unstructured-ingest, other structure/doc updates related to ingest.
+* Add `parser` parameter to `partition_html`.
 
 ## 0.4.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.12-dev1
+## 0.4.12-dev2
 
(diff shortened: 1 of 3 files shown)


MthwRobinson (approved):
LGTM, thanks for the contribution!

unstructured/partition/html.py
```diff
@@ -4,6 +4,7 @@
 
 from unstructured.documents.elements import Element
 from unstructured.documents.html import HTMLDocument
+from unstructured.partition.xml import VALID_PARSERS
```
MthwRobinson, line 7: I think this may need to be `from unstructured.documents.xml import VALID_PARSERS`

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.12-dev2
+
+* Add `parser` parameter to `partition_html`.
+
 ## 0.4.12-dev1
```
MthwRobinson, line 5: Could you change this line to `## 0.4.12-dev2` like you have above and add your bullet below?

unstructured/partition/html.py
asai95: Done, thank you!
I also noticed that I forgot to add the description in the docstring, fixed that too.

CHANGELOG.md
asai95: Done

PR #259, opened 2023-02-23
fix: Adds missing __init__.py

cragwolfe:
Fixes an issue where the packaged CLI `unstructured-ingest` was failing due to:

    ModuleNotFoundError: No module named 'unstructured.ingest'

Also updates the Ingest documentation to make clear that detectron and other OS dependencies are needed when processing PDF's.

Tested locally in a fresh virtual env where the `unstructured` package was installed from the local checkout with `pip install -e .`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.13
+
+* Fixes unstructured-ingest cli.
+
 ## 0.4.12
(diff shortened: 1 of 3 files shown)


ajjimeno (approved):
Tested in on a fresh environment and works. I checked the changed files and LGTM.

PR #268, opened 2023-02-23
feat: detect filetype with extension if libmagic is unavailable

### LINKED ISSUE #228: Add filetype check based on file extension if `libmagic` isn't available
Currently we use libmagic to determine the filetype for input files. However, `magic` requires the `libmagic` library, which not all users have available. The goal of this issue is to add a fallback that determines the filetype based on the file extension if `libmagic` is not available. If `libmagic` is available, `detect_filetype` should continue to use `magic`, as it does currently.

firedragonironfist:
This change resolves #228 I added the changes from the previous pull request and also rectified the black usage error. Please let me know if you have any further feedback or concerns.

Signed-off-by: Yousuf <yousufmohd223@gmail.com>


unstructured/file_utils/filetype.py
```diff
@@ -106,100 +112,217 @@ def __lt__(self, other):
...
+        """Use libmagic to determine a file's type. Helps determine which partition brick
+        to use for a given file. A return value of None indicates a non-supported file type."""
+        if filename and file:
+            raise ValueError("Only one of filename or file should be specified.")
+
+        if filename:
+            _, extension = os.path.splitext(filename)
+            extension = extension.lower()
+            mime_type = magic.from_file(filename, mime=True)
+            if not LIBMAGIC_AVAILABLE:
```
MthwRobinson, line 155: Move this check above `mime_type = magic.from_file(filename, mime=True)`, otherwise you'll get an error if `libmagic` isn't installed

MthwRobinson (changes requested):
Added inline comments. `filetype.py` is looking pretty good! This PR shouldn't change
`examples/ingest/s3-small-batch/main.py`, you can revert that with `git checkout main -- examples/ingest/s3-small-batch/main.py`. Looks like there are some merge conflicts to resolve as well.

firedragonironfist:
Ok i will resolve them

firedragonironfist:
Please let me know if you have any further feedback or concerns.

MthwRobinson (approved):

PR #279, opened 2023-02-24
feat: add ubuntu setup script

qued:
Added Ubuntu setup script.

#### Testing:

In a fresh environment, run the script with
```shell
./setup_ubuntu.sh <account name>
```
...where `<account name>` is the name of the user account you are setting up for.

After installation and restarting the shell, you should have the `pyenv` command available. `pyenv versions` should show that version 3.8.15 is installed.

Check that poppler is installed with `pdftoppm -h`.

Check that tesseract is installed with `tesseract`.

### DIFF
--- a/unstructured/__version__.py
+++ b/unstructured/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.4.16-dev16-dev0"  # pragma: no cover
+__version__ = "0.4.16-dev1"  # pragma: no cover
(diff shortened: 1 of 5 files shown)


unstructured/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.4.16-dev0"  # pragma: no cover
+__version__ = "0.4.16-dev16-dev0"  # pragma: no cover
```
MthwRobinson, line 1: This should be `dev1` right?

MthwRobinson (approved):
LGTM pending shellcheck and the `__version__` fix.

unstructured/__version__.py
qued: Yeah. I think `version-sync` is breaking 😕

PR #281, opened 2023-02-25
fix: add libmagic dep to ubuntu script

cragwolfe:


qued (dismissed):
LGTM!

scripts/setup_ubuntu.sh
```diff
@@ -80,6 +80,9 @@ $sudo $pac install -y poppler-utils
 # Install tesseract as well as Russian language
 $sudo $pac install -y tesseract-ocr libtesseract-dev tesseract-ocr-rus
 
+#### libmagic
+$sudo $pac install libmagic-dev
```
qued, line 84:
```suggestion
$sudo $pac install -y libmagic-dev
```

qued (changes requested):
Actually just one minor thing so it can be run non-interactively.

qued (approved):

Issue #215, opened 2023-02-13
feat/Add support for epub format


nestordemeure:
**Is your feature request related to a problem? Please describe.**
One of the leading open format for books is `epub`. Thus, when running algorithms on books I often have to find a way to process them.

**Describe the solution you'd like**
An `epub` is a compressed `html`. As `unstructured` already supports `html`, it would be very straightforward to add support for epub.

**Describe alternatives you've considered**
I am currently using [pandoc](https://pandoc.org/) to convert my `epub` data to `html`. This is straightforward but adds a special case to my pipelines.

MthwRobinson:
@nestordemeure - Thanks for adding this issue! We're tracking it and will prioritize it as soon as we can.

PR #284, opened 2023-02-25
feat: Add GitHub data connector; add Markdown partitioner

### LINKED ISSUE #256: Create a data conector for Github repos Part 1 (all supported filetypes but no source code)
Create a data connector that pulls data from GIthub repo.

See [Adding Data Connectors](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#adding-data-connectors) for details on how to get started. Make sure to include a link to this issue when submitting a PR.

**Definition of Done**

- The [checklist](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist) has been completed.
- The connector is able to process a single repository and recursively the entire documents in a repo.
- Should ignore code.
- Should process only doc types supported in [partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py), including markdown files.

    
Bonus points: the ability to filter by document type.

tomaarsen:
Closes #256

Hello!

## Pull Request overview
* Add GitHub data connector for ingest.
  * I have ticked off the checklist.
  * The connector can process a single repository and recursively load all documents in the repo.
  * Files with extensions that are not supported by `partition.auto` are skipped.
  * Supports downloading from a specific git branch.
  * Supports the ability to filter documents through globs.
* Add `partition_md` to support partitioning Markdown files.
  * Implemented using [`python-markdown`](https://pypi.org/project/Markdown/), which internally converts the markdown into HTML. This is perhaps not ideal, but it works well enough for now.
  * Add `FileType.MD` and support for `partition_md` in `auto.partition`.

## The data connector checklist
- [x] Create a new module under [unstructured/ingest/connector/](unstructured/ingest/connector/) implementing the 3 abstract base classes, similar to [unstructured/ingest/connector/s3_connector.py](unstructured/ingest/connector/s3_connector.py).
  - [x] The subclass of `BaseIngestDoc` overrides `process_file()` if extra processing logic is needed other than what is provided by [auto.partition()](unstructured/partition/auto.py).
- [x] Update [unstructured/ingest/main.py](unstructured/ingest/main.py) with support for the new connector.
- [x] Create a folder under [examples/ingest](examples/ingest) that includes at least one well documented script.
- [x] Add a script test_unstructured_ingest/test-ingest-\<the-new-data-source\>.sh. It's json output files should have a total of no more than 100K.
- [x] Git add the expected outputs under test_unstructured_ingest/expected-structured-output/\<folder-name-relevant-to-your-dataset\> so the above test passes in CI.
- [x] Add a line to [test_unstructured_ingest/test-ingest.sh](test_unstructured_ingest/test-ingest.sh) invoking the new test script.
- [x] If additional python dependencies are needed for the new connector:
  - [x] Add them as an extra to [setup.py](unstructured/setup.py).
  - [x] Update the Makefile, adding a target for `install-ingest-<name>` and adding another `pip-compile` line to the `pip-compile` make target. See [this commit](https://github.com/Unstructured-IO/unstructured/commit/ab542ca3c6274f96b431142262d47d727f309e37) for a reference.
  - [x] The added dependencies should be imported at runtime when the new connector is invoked, rather than as top-level imports.
- [x] Honors the conventions of `BaseConnectorConfig` defined in [unstructured/ingest/interfaces.py](unstructured/ingest/interfaces.py) which is passed through [the CLI](unstructured/ingest/main.py):
  - [x] If running with an `.output_dir` where structured outputs already exists for a given file, the file content is not re-downloaded from the data source nor is it reprocessed. This is made possible by implementing the call to `MyIngestDoc.has_output()` which is invoked in [MainProcess._filter_docs_with_outputs](ingest-prep-for-many/unstructured/ingest/main.py).
  - [x] Unless `.reprocess` is `True`, then documents are always reprocessed.
  - [x] If `.preserve_download` is `True`, documents downloaded to `.download_dir` are not removed after processing.
  - [x] Else if `.preserve_download` is `False`, documents downloaded to `.download_dir` are removed after they are **successfully** processed during the invocation of `MyIngestDoc.cleanup_file()` in [process_document](unstructured/ingest/doc_processor/generalized.py)
  - [x] Does not re-download documents to `.download_dir` if `.re_download` is False, enforced in `MyIngestDoc.get_file()`
  - [x] Prints more details if `.verbose` similar to [unstructured/ingest/connector/s3_connector.py](unstructured/ingest/connector/s3_connector.py).

### Installation
I've added `pygithub==1.57.0` as a dependency, installable via `pip install unstructured["github"]`. Note that the newer `1.58.0` does not work due to this issue: https://github.com/PyGithub/PyGithub/issues/2436, it prevents us from using a `Repository` instance in a multiprocessing loop.

I've also added `markdown` as a core dependency, required for `partition_md`, and `types-Markdown` as a test requirement to satisfy `mypy`. I re-ran `pip-compile` to update the requirements files afterwards.

### CLI Help
The usage of the `unstructured-ingest` CLI is now like so:
```
Usage: unstructured-ingest [OPTIONS]

Options:
  --s3-url TEXT                   Prefix of s3 objects (files) to download.
                                  E.g. s3://bucket1/path/. This value may also
                                  be a single file.
  --s3-anonymous                  Connect to s3 without local AWS credentials.
  --github-url TEXT               URL to GitHub repository, e.g.
                                  "https://github.com/Unstructured-
                                  IO/unstructured", or a repository owner/name
                                  pair, e.g. "Unstructured-IO/unstructured"
  --github-access-token TEXT      A GitHub access token, see https://docs.gith
                                  ub.com/en/authentication/keeping-your-
                                  account-and-data-secure/creating-a-personal-
                                  access-token
  --github-branch TEXT            The branch for which to fetch files from. If
                                  not given, the default repository branch is
                                  used.
  --github-file-glob TEXT         A comma-separated list of file globs to
                                  limit which types of files are accepted,
                                  e.g. '*.html,*.txt'
  --re-download / --no-re-download
                                  Re-download files from s3 even if they are
                                  already present in --download-dir.
  --download-dir TEXT             Where s3 files are downloaded to, defaults
                                  to tmp-ingest-<6 random chars>.
  --preserve-downloads            Preserve downloaded s3 files. Otherwise each
                                  file is removed after being processed
                                  successfully.
  --structured-output-dir TEXT    Where to place structured output .json
                                  files.
  --reprocess                     Reprocess a downloaded file from s3 even if
                                  the relevant structured output .json file in
                                  --structured-output-dir already exists.
  --num-processes INTEGER         Number of parallel processes to process docs
                                  in.  [default: 2]
  -v, --verbose
  --help                          Show this message and exit.
```
Note the four new `github`-related arguments. In particular, I want to point out that for `--github-url`, both `https://github.com/Unstructured-IO/unstructured` and `Unstructured-IO/unstructured` are valid. Beyond that, `--github-file-glob` can be used to select specific data, e.g. with `"*.html,*.txt,*.md"`.

### Sample usage
An example usage would be the following:
```
unstructured-ingest \
    --github-url Unstructured-IO/unstructured \
    --github-branch main \
    --structured-output-dir github-ingest-output \
    --num-processes 2 \
    --verbose
```

## Markdown partitioner
This PR does not only include a GitHub connector, but also a partitioner for markdown. I rely on [`python-markdown`](https://pypi.org/project/Markdown/), which internally converts the markdown into HTML text, after which I call `partition_html`. The new `partition_md` has this signature:
```python
def partition_md(
    file: Optional[IO] = None,
    text: Optional[str] = None,
    url: Optional[str] = None,
    parser: VALID_PARSERS = None,
) -> List[Element]:
```

I want to point out one caveat of this approach. Normally, when calling a `partition` method with `filename` and `include_metadata=True`, then the filename will be included in the output. However, my implementation always calls `partition_html` via `text=...`, even if `partition_md` is called with a `filename`. In short, the filename is not included in outputs by `partition_md`.
For the time being I think this is a minor issue - it is already miles better than not having any Markdown support, after all. I also wanted to keep the focus of this PR on the GitHub connector.

I also added some very simple tests to the markdown partitioner based on the HTML partitioner.

### Markdown sample output
See here a snippet from my output when running:
```
unstructured-ingest --github-url Unstructured-IO/unstructured --github-file-glob 'Readme.md' --structured-output-dir github-unstructured-readme-output --verbose
```
With other words, I downloaded from this repository, using the glob `Readme.md`, which means I only fetch the README. The output folder contained only `README.md.json` containing this output. As you can see, the titles are correctly discovered etc.

```json
...
  {
    "element_id": "6fc70513a3165acfe9986eda09da3e36",
    "text": "Open-Source Pre-Processing Tools for Unstructured Data",
    "type": "Title",
    "metadata": {
      "page_number": 1
    }
  },
  {
    "element_id": "dec0569ba095e8dfad354bbdd6a624c1",
    "text": "The unstructured library provides open-source components for pre-processing text documents\nsuch as PDFs, HTML and Word Documents. These components are packaged as bricks ð§±, which provide\nusers the building blocks they need to build pipelines targeted at the documents they care\nabout. Bricks in the library fall into three categories:",
    "type": "UncategorizedText",
    "metadata": {
      "page_number": 1
    }
  },
  {
    "element_id": "de32231a1dac79f3e9baea4e970ad305",
    "text": ":jigsaw: Partitioning bricks that break raw documents down into standard, structured\n  elements.",
    "type": "UncategorizedText",
    "metadata": {
      "page_number": 1
    }
  },
  {
    "element_id": "3c3ad9bc65fd3100568ecad1dbd768e5",
    "text": ":broom: Cleaning bricks that remove unwanted text from documents, such as boilerplate and\n  sentence\n  fragments.",
    "type": "UncategorizedText",
    "metadata": {
      "page_number": 1
    }
  },
  {
    "element_id": "6388255e5f0ea0819eed2c8b3b3bd02f",
    "text": ":performing_arts: Staging bricks that format data for downstream tasks, such as ML inference\n  and data labeling.",
    "type": "UncategorizedText",
    "metadata": {
      "page_number": 1
    }
  },
  {
    "element_id": "8329a04e799e36105f29583b4f10814c",
    "text": ":eight_pointed_black_star: Quick Start",
    "type": "Title",
    "metadata": {
      "page_number": 1
    }
  },
  {
    "element_id": "115d9086aac6be868679eb07e418e65d",
    "text": "Use the following instructions to get up and running with unstructured and test your\ninstallation.",
    "type": "UncategorizedText",
    "metadata": {
      "page_number": 1
    }
  },
...
```

---

Let me know if you need anything of me at this point! I hope this will be helpful to you all!

- Tom Aarsen


tomaarsen:
Some additional, unrelated comments:
* `reprocess` is an argument on the `BaseConnectorConfig`, but it doesn't have to be there: It is only used on `MainProcess` instead:
  https://github.com/Unstructured-IO/unstructured/blob/e61ce2cc0065b4afb2af89df648058909ffd4f74/unstructured/ingest/interfaces.py#L50
* The `initialize` method on the `BaseConnector` never gets called:
  https://github.com/Unstructured-IO/unstructured/blob/e61ce2cc0065b4afb2af89df648058909ffd4f74/unstructured/ingest/interfaces.py#L25-L29
  That also goes with the initialize method of the `S3Connector`, which actually has code in it. I didn't test thoroughly, but I don't think these lines are ever called:
  https://github.com/Unstructured-IO/unstructured/blob/e61ce2cc0065b4afb2af89df648058909ffd4f74/unstructured/ingest/connector/s3_connector.py#L166-L173

Perhaps I can make a PR to address these changes soon.

- Tom Aarsen

cragwolfe:
@tomaarsen , First off, lots of great stuff in this PR!

I did some testing on the markdown/detection partitioning, and hit an issue, so the rest of this comment is focused on that.

At least in my mac env, the `filetype` for .md files is resolving as `FileType.TXT` in `unstructured/partition/auto.py`, so it never gets into the clause:

        elif filetype == FileType.MD:

I was initially looking for diff's with this script between `main` and this branch, and noticed the outputs were the same.

```
$ cat ~/tmp/dump-elems.py
import sys
elems = partition(sys.argv[1])
output = "\n".join([f"{e.to_dict()['type']:20}  {e.text:<80}" for e in elems])
print(output)
```

I think the issue itself may have been misleading with markdown comment in that the text parser is actually ok-ish -- sorry about that!

For the sake of this PR, it is OK to punt on better markdown processing but also OK to try to improve it :)

FWIW, source code files also come through as `FileType.TXT`, which ends up with the text in multiple `Element`s even though each individual element is particularly meaningful.

cragwolfe:
Regarding:

> Some additional, unrelated comments:
> * reprocess is an argument on the BaseConnectorConfig, but it doesn't have to be there: It is only used on MainProcess instead:

Yes, you're right, should be removed.

> * The initialize method on the BaseConnector never gets called

Good catch! I pushed the quick fix here:
https://github.com/Unstructured-IO/unstructured/pull/285

unstructured/ingest/connector/github.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @property
+    def filename(self):
+        return (Path(self.config.download_dir) / self.path).resolve()
+
+    def _output_filename(self):
+        return Path(self.config.output_dir) / f"{self.path}.json"
+
+    def _create_full_tmp_dir_path(self):
+        """includes "directories" in s3 object path"""
```
cragwolfe, line 74: "s3 object path"

unstructured/ingest/connector/github.py
```diff
@@ -0,0 +1,188 @@
...
+        os.chdir("..")
+        if len(os.listdir(cur_dir)) == 0:
+            os.rmdir(cur_dir)
+
+    def initialize(self):
+        pass
+
+    def is_file_type_supported(self, path: str) -> bool:
+        # Workaround to ensure that auto.partition isn't fed with .yaml, .py, etc. files
+        # TODO: What to do with no filenames? e.g. LICENSE, Makefile, etc.
```
cragwolfe, line 142: I assume you mean "no filename extensions." Actually, libmagic does a reasonably good job of guessing them:

```
(unst) bash-5.1$ cp example-docs/fake-html.html example-docs/fake-h
(unst) bash-5.1$ cp example-docs/fake-power-point.pptx example-docs/fake-p
(unst) bash-5.1$ PYTHONPATH=. python -c "import sys; from unstructured.file_utils.filetype import detect_filetype; print(detect_filetype(sys.argv[1]))" example-docs/fake-h
FileType.HTML
(unst) bash-5.1$ PYTHONPATH=. python -c "import sys; from unstructured.file_utils.filetype import detect_filetype; print(detect_filetype(sys.argv[1]))" example-docs/fake-p
FileType.PPTX
```

which means `unstructured.partition.auto::partition()` would do the right thing with them.

Perhaps the check should be for [FileType.UNK](https://github.com/Unstructured-IO/unstructured/blob/9062d25/unstructured/file_utils/filetype.py#L221) ?

examples/ingest/github/ingest.sh
```diff
@@ -0,0 +1,19 @@
...
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --github-url Unstructured-IO/unstructured \
+    --github-branch main \
+    --structured-output-dir github-ingest-output \
+    --num-processes 2 \
+    --verbose
+
+# Alternatively, you can call it using:
+# unstructured-ingest --github-url ...
```
cragwolfe, line 19: I ran into the following error on mac. (Maybe mac only not linux?)

```
<snip>
The file 'unstructured/staging/label_box.py' is discarded as it does not contain a supported filetype.
The file 'unstructured/staging/label_studio.py' is discarded as it does not contain a supported filetype.
The file 'unstructured/staging/prodigy.py' is discarded as it does not contain a supported filetype.
The file 'unstructured/utils.py' is discarded as it does not contain a supported filetype.
  File "./unstructured/ingest/main.py", line 204, in <module>
    main()
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "./unstructured/ingest/main.py", line 195, in main
    MainProcess(
  File "./unstructured/ingest/main.py", line 63, in run
    results = pool.map(self.doc_processor_fn, docs)  # noqa: F841
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/pool.py", line 537, in _handle_tasks
    put(task)
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/synchronize.py", line 101, in __getstate__
    context.assert_spawning(self)
  File "/Users/cragwolfe/.pyenv/versions/3.8.15/lib/python3.8/multiprocessing/context.py", line 359, in assert_spawning
    raise RuntimeError(
RuntimeError: RLock objects should only be shared between processes through inheritance
```

I have a strong suspicion it is due to `<class 'github.Repository.Repository'>` which is a member of each `GitHubIngestDoc` instance not playing well with multiprocessing.

tomaarsen: Please retry with `pygithub==1.57.0`. 1.58.0 indeed has a problem with multiprocessing.

tomaarsen:
> @tomaarsen , First off, lots of great stuff in this PR!
> 
> I did some testing on the markdown/detection partitioning, and hit an issue, so the rest of this comment is focused on that.
> 
> At least in my mac env (with libmagic installed, which is preferred), the `filetype` for .md files is resolving as `FileType.TXT` in `unstructured/partition/auto.py`, so it never gets into the clause:

Quick comment - I'll reply in more depth later. I think this is because I had accidentally installed both `python-magic` and `python-magic-bin`. The first of which breaks the magic import on Windows and sets a conditional to false such that it uses a backup system for determining the filetype: a mapping from ".md" to FileType.MD.

examples/ingest/github/ingest.sh
cragwolfe: That fixes it! And I see you have it pinned in setup.py.

Can you add a `NOTE` in setup.py to change to that dep to `pygithub>1.58.0` if the future version resolves the issue?

cragwolfe: I appreciate your time on this as well! 😊

unstructured/ingest/connector/github.py
tomaarsen: Whoops, I did mean for no extension. I could rely on libmagic here, but this filtering is applied solely on the file paths: the actual documents have yet to be downloaded: that is done in the `GitHubIngestDoc.get_file` method. Through this filtering, less `GitHubIngestDoc` instances need to be created, and thus less API requests to GitHub are made.

What do you suggest? I can only check for `FileType.UNK` in `GitHubIngestDoc`.

examples/ingest/github/ingest.sh
tomaarsen: Done!

unstructured/ingest/connector/github.py
tomaarsen: Fixed!

tomaarsen:
@cragwolfe I'd love to add proper support for .md files: it seems like a valuable addition to the product beyond the GitHub data connector itself. However, the current implementation of `detect_filetype` is unable to detect `MD` files correctly. I'll try to work around this!

tomaarsen:
I've made the markdown detection more robust. Could you perhaps try again to see if it works as intended on your device now?
I appreciate it.

cragwolfe:
I think it is important to establish the pattern of parsing markdown files explicitly, as this PR does -- thanks!

I think there is room for improvement in the markdown parser -- we should be able to use markdown native "types" rather than convert to html. But not a blocker for this PR and can be tracked in a separate issue.

 [exploration
](https://gist.github.com/cragwolfe/5bc0d283f811cb377e8cdec3899e9992)

tomaarsen:
> I think there is room for improvement in the markdown parser -- we should be able to use markdown native "types" rather than convert to html. But not a blocker for this PR and can be tracked in a separate issue.

I agree, for the time being this "Markdown-via-HTML" support will be preferable to no markdown support or using the `partition_text` for it.

Alternatively, another PR can take it in the opposite direction and move to using `pandoc` as suggested in #215 for converting various different markup formats to e.g. html. This would theoretically add support for all filetypes that `pandoc` supports, although then the filetype detector would also have to be able to detect those other markup files correctly.
Perhaps an interesting discussion for later.

cragwolfe:
I'm seeing some errors that seem specific to github.py when processing the Unstructured repo:

```
bash -x examples/ingest/github/ingest.sh


fetching GitHubIngestDoc(repo=Repository(full_name="Unstructured-IO/unstructured"), path='examples/argilla-summarization/img/isw.png') - PID: 99063
ERROR:root:Failed to process GitHubIngestDoc(repo=Repository(full_name="Unstructured-IO/unstructured"), path='examples/argilla-summarization/img/isw.png')
  File "/Users/cragwolfe/r/unstructured/unstructured/ingest/doc_processor/generalized.py", line 23, in process_document
    doc.get_file()
  File "/Users/cragwolfe/r/unstructured/unstructured/ingest/connector/github.py", line 94, in get_file
    f.write(content_file.decoded_content)  # type: ignore
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/github/ContentFile.py", line 59, in decoded_content
    assert self.encoding == "base64", f"unsupported encoding: {self.encoding}"
AssertionError: unsupported encoding: none
```

But this file may currently be processed with:

    $ PYTHONPATH=. python -c "from unstructured.partition.auto import partition;
    print('len of elems is '+str(len(partition('examples/argilla-summarization/img/isw.png'))))
    len of elems is 6

Other docs this is an issue for: `example-docs/fake-email-header.eml`, `example-docs/factbook.xml`, `example-docs/example-10k.html`

Once these are clean, this PR should be good to go!

tomaarsen:
I've had a look into this, and it seems to be the result of poor documentation on the PyGitHub side. On the official GitHub documentation I could find some more information about this: https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#size-limits
As it turns out, files larger than 1MB are returned to have no contents and an encoding of "none". This caused an AssertionError on the `decoded_content` property, which expects to always see a `"base64"` encoding. I would have expected this to be documented on the PyGitHub side, or for a more useful error to be thrown.

Regardless, I can use the `raw.githubusercontent` link, e.g.:
https://raw.githubusercontent.com/Unstructured-IO/unstructured/main/unstructured/nlp/english-words.txt

As for the `.eml` file etc., it seems to consider that one a `message/rfc822`. I'm not sure why it does that - perhaps something with the file encoding?

cragwolfe (approved):
The examples/ingest/github/ingest.sh script works great!

There are two failures processing files, but that occurs outside the connector as well:

```
ValueError: Invalid file example-docs/fake-email-header.eml. File type not support in partition.
ValueError: Invalid file example-docs/factbook.xml. File type not support in partition.
```

PR #285, opened 2023-02-25
fix: Ingest main: actually initialize the connector

cragwolfe:
This brings back the connector initialization as was intended and got lost in a refactor.

Thanks @tomaarsen for catching this.

### DIFF
--- a/unstructured/ingest/main.py
+++ b/unstructured/ingest/main.py
@@ -20,6 +20,7 @@ class MainProcess:
 
     def initialize(self):
         """Slower initialization things: check connections, load things into memory, etc."""
+        self.doc_connector.initialize()
         initialize()
 
     def cleanup(self):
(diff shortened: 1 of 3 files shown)


unstructured/ingest/main.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, doc_connector, doc_processor_fn, num_processes, reprocess):
 
     def initialize(self):
         """Slower initialization things: check connections, load things into memory, etc."""
+        self.doc_connector.initialize()
```
james-see, line 23: Just adding this method checks the connections an everything?

james-see: nevermind I see it was just pass'ing before based on the github connector PR

qued (approved):
LGTM!

tomaarsen:
Looks good!

PR #287, opened 2023-02-25
fix: update script for docker

qued:
Updates so that script will run inside a docker container to set it up.

Tested versions:
- [x] 22.04
- [x] 20.04
- [x] 18.04

#### Testing:
Create a directory with only `scripts/setup_ubuntu.sh` in it.

Create the following `Dockerfile` in that directory:
```docker
FROM ubuntu
COPY setup_ubuntu.sh /
```
From within the directory, run `docker build . -t ub-test`.

Run with `docker run -it ub-test /bin/bash`.

Now within the container, run the script with:
```shell
./setup_ubuntu.sh root
```

After installation, run `source $HOME/.bashrc` and you should have the `pyenv` command available. `pyenv versions` should show that version 3.8.15 is installed.

Check that poppler is installed with `pdftoppm -h`.

Check that tesseract is installed with `tesseract`.

### DIFF
--- a/scripts/setup_ubuntu.sh
+++ b/scripts/setup_ubuntu.sh
@@ -11,13 +11,17 @@ set -ux
 
 # Set user account for which we're configuring the tools
 USER_ACCOUNT=$1
+USER_ACCOUNT_HOME=$(bash -c "cd ~$(printf %q "$USER_ACCOUNT") && pwd")
 
 # Set package manager command for this distribution
 pac="apt"
 
 # If we're not running as root, we want to prefix certain commands with sudo
 if [[ $(whoami) == 'root' ]]; then
+    $pac update -y
+    $pac install -y sudo
     sudo=''; else
+    type -p sudo >/dev/null || (echo "Please have an administrator install sudo and add you to the sudo group before continuing." && exit 1)
     sudo='sudo'
 fi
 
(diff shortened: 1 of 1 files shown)


cragwolfe:
When running this in a fresh docker container, I get this annoying interactive prompt:

```
debconf: unable to initialize frontend: Dialog
debconf: (No usable dialog-like program is installed, so the dialog based frontend cannot be used. at /usr/share/perl5/Debconf/FrontEnd/Dialog.pm line 78.)
debconf: falling back to frontend: Readline
Configuring tzdata
------------------

Please select the geographic area in which you live. Subsequent configuration questions will narrow this down by presenting a list of cities, representing the time zones in which they are located.

  1. Africa  2. America  3. Antarctica  4. Australia  5. Arctic  6. Asia  7. Atlantic  8. Europe  9. Indian  10. Pacific  11. US  12. Etc
Geographic area: 
```
not a blocker since it was there before, just flagging it. This is how I got around it: https://gist.github.com/cragwolfe/e230fea5491ce2704356ee17857ee73c#file-setup-deps-and-test-user-sh

cragwolfe (approved):

cragwolfe:
Confirmed this still works with: https://gist.github.com/cragwolfe/e230fea5491ce2704356ee17857ee73c

cragwolfe:
Also confirmed running this [script](https://gist.githubusercontent.com/cragwolfe/e230fea5491ce2704356ee17857ee73c/raw/6178a5641a14b907ffc5d02b7fadf01b9be4fb43/unstructured-pdf-smoke-test.sh) as root works, though need to be sure to `su - root` to pickup pyenv changes in a fresh shell.

PR #290, opened 2023-02-26
fix: get rid of tzdata halt when running as non-root

qued:
With a small change I think we can get rid of the interactive prompt when tzdata gets installed.

#### Testing:

Run per @cragwolfe 's [gist](https://gist.github.com/cragwolfe/e230fea5491ce2704356ee17857ee73c) but _without_ lines 9-17 of [setup-deps-and-test-user.sh](https://gist.github.com/cragwolfe/e230fea5491ce2704356ee17857ee73c#file-setup-deps-and-test-user-sh). 

If all goes well the script should complete without any interactive prompts.


MthwRobinson (approved):
LGTM!

cragwolfe:
Tested with the "[gist](https://gist.github.com/cragwolfe/e230fea5491ce2704356ee17857ee73c) but without lines 9-17 of [setup-deps-and-test-user.sh](https://gist.github.com/cragwolfe/e230fea5491ce2704356ee17857ee73c#file-setup-deps-and-test-user-sh)." as @qued suggested --  works well!

And learned a bit about sudo along the way 🙏 .
Neat that that one change fixes it.

cragwolfe (approved):

PR #298, opened 2023-02-27
fix: `process_document` behavior when exception is raised

alvarobartt:
Hi to whoever reviews this PR :hugs:

## What's in this PR?

I was running the example for the `S3Connector` at https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-unstructured-ingest-cli, but found out that it was not working as expected if some required dependencies are not installed, because an exception is thrown in the middle of the process and some documents may have been properly processed while some may have failed due to those missing dependencies.

So on, that leads to the `try-except` to jump to the `except` part and then to the `finally`, but the `else` is skipped, which means that the `tmp-...` dir created to store those documents is kept in the source directory. So on, if an exception happens, it needs to be removed (unlinked), as `doc.cleanup_file()` will never be reached, unless placing it too in the `finally` block.

Also as another minor nit, I saw that the `unstructured` logger is defined with the default configuration at `unstructured.logger` so I used that instead of default's `logging`.

Is the current `unstructured.logger` something you're willing to improve? e.g. defining it as `def get_logger` and adding to it a custom configuration to make the exceptions more readable? If so LMK and I'll have a look at it 👍🏻 

## What's missing in this PR?

Both keeping `doc.clean_up()` in the `finally` or just the `if filename.exists()` during the `except` block works, but I think that having both combined instead of using the `try-except-else` is more suitable. LMKWYT 👍🏻 

Thanks!


unstructured/ingest/doc_processor/generalized.py
```diff
@@ -30,9 +32,11 @@ def process_document(doc):
...
+        filename = doc.filename if isinstance(doc.filename, Path) else Path(doc.filename)
+        if filename.exists():
+            os.unlink(filename.as_posix())
         # TODO(crag) save the exception instead of print?
-        logging.error(f"Failed to process {doc}", exc_info=True)
-    else:
-        doc.cleanup_file()
+        logger.error(f"Failed to process {doc}", exc_info=True)
     finally:
+        doc.cleanup_file()
```
alvarobartt, line 41: Note that removing the file-deletion in the `except Exception` above is the same thing as keeping this line, since the code right after the `finally` will always be executed, so please LMK what option you prefer:
* The implemented one
* Or, the one that just moves the `doc.cleanup_file` from `else` to `finally` and doesn't perform the `filename.exists()` and `os.unlink()` under the `except`

tomaarsen:
Hello!

It is my understanding that the temporary files *should* be kept in case of failure, i.e. that the temporary files are only removed if the processing was successful. This behaviour has the desired effect that the downloading does not need to be repeated.

Please see the [Ingest documentation](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist)
> - [ ] Else if `.preserve_download` is `False`, documents downloaded to `.download_dir` are removed after they are **successfully** processed during the invocation of `MyIngestDoc.cleanup_file()` in [process_document](unstructured/ingest/doc_processor/generalized.py)

In short, the current behaviour is not a bug.

The logger changes that you suggest seem wise, though.

- Tom Aarsen

alvarobartt:
> Hello!
> 
> It is my understanding that the temporary files _should_ be kept in case of failure, i.e. that the temporary files are only removed if the processing was successful. This behaviour has the desired effect that the downloading does not need to be repeated.
> 
> Please see the [Ingest documentation](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist)
> 
> > * [ ]  Else if `.preserve_download` is `False`, documents downloaded to `.download_dir` are removed after they are **successfully** processed during the invocation of `MyIngestDoc.cleanup_file()` in [process_document](unstructured/ingest/doc_processor/generalized.py)
> 
> In short, the current behaviour is not a bug.
> 
> The logger changes that you suggest seem wise, though.
> 
>     * Tom Aarsen

Thanks for your input here!

So what I found out is that those `tmp-...` files were not used even when kept there, instead, the same file was downloaded again in another dir, but let me check that just in case.

But true, if the documentation says to keep the `download_dir` then I need to rever those changes, nice catch :)

tomaarsen:
> So what I found out is that those `tmp-...` files were not used even when kept there, instead, the same file was downloaded again in another dir, but let me check that just in case.

Perhaps the decision to keep the `tmp-...` file was to allow further testing on that file just using e.g. `partition` from `unstructured.partition.auto`? However, I do agree that it does not seem ideal.

alvarobartt:
> > So what I found out is that those `tmp-...` files were not used even when kept there, instead, the same file was downloaded again in another dir, but let me check that just in case.
> 
> Perhaps the decision to keep the `tmp-...` file was to allow further testing on that file just using e.g. `partition` from `unstructured.partition.auto`? However, I do agree that it does not seem ideal.

I'll check the `S3Connector`, because it seems that even when `self.config.verbose`, the print-statement at https://github.com/Unstructured-IO/unstructured/blob/5db94fdee61556cc25cf66166580e8f779195647/unstructured/ingest/connector/s3_connector.py#LL96C6-L96C6 is not actually printing anything even if the file is there... So that makes me think that files are not being reused by default or that there's a bug to fix somewhere...

I'll explore this further and let you know later @tomaarsen :hugs:

alvarobartt:
So... by default, a random one is chosen https://github.com/Unstructured-IO/unstructured/blob/5db94fdee61556cc25cf66166580e8f779195647/unstructured/ingest/main.py#L168-L171 so files are preserved but never re-used... So I assume that the patch in this PR makes sense?

Maybe it's worth adding a check in case the `preserve_files` is passed along with the `download_dir`?

cragwolfe:
Hi @alvarobartt , thanks for the code cleanup and digging into the logic! Also, thank you @tomaarsen for correctly interpreting the intent for the various flags and how they operate together.

@alvarobartt , nitpicky request: please update the PR description to a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) format. Also, is this fixing something, or mainly a cleanup (refactor?).

After following along the rest of the discussion, I think the main awkwardness is that:

If `--download-dir` is not specified (where the implication is directory with a random is suffix is chosen) AND --preserve-downloads is specified,
then the files are indeed preserved in that randomly-generated directory. So, if the ingest is run again with the same options, the previous downloads don't make a difference since of course they will not be found in the new random directory.

Couple of options:

1/ Complain and exit from main if the user does `--preserve-downloads` without `--download-dir`, so users can avoid this gotcha.
2/ Stop generating random suffixes and just store downloads in "tmp-ingest" by default. I added the random prefix to make clear differences between different runs (simultaneous or not), but I'm not sure it is worth it.

cragwolfe (approved):
LGTM

alvarobartt:
> Hi @alvarobartt , thanks for the code cleanup and digging into the logic! Also, thank you @tomaarsen for correctly interpreting the intent for the various flags and how they operate together.
> 
> @alvarobartt , nitpicky request: please update the PR description to a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) format. Also, is this fixing something, or mainly a cleanup (refactor?).
> 
> After following along the rest of the discussion, I think the main awkwardness is that:
> 
> If `--download-dir` is not specified (where the implication is directory with a random is suffix is chosen) AND --preserve-downloads is specified, then the files are indeed preserved in that randomly-generated directory. So, if the ingest is run again with the same options, the previous downloads don't make a difference since of course they will not be found in the new random directory.
> 
> Couple of options:
> 
> 1/ Complain and exit from main if the user does `--preserve-downloads` without `--download-dir`, so users can avoid this gotcha. 2/ Stop generating random suffixes and just store downloads in "tmp-ingest" by default. I added the random prefix to make clear differences between different runs (simultaneous or not), but I'm not sure it is worth it.

Thanks for the feedback @cragwolfe! I assume that the best option is to generate a hash (MD5) out of the provided path and store everything in `.cache/unstructured/unstructured_ingest/<HASH>/...`, if that's an option I can submit another PR.

cragwolfe (approved):

cragwolfe:
@alvarobartt , 

>  generate a hash (MD5) out of the provided path

You mean a hash of `--structured-output-dir`?
I think that makes sense. Probably the absolute resolved path of `--structured-output-dir`?

If `--preserve-downloads` is specified without a `--download-dir`, it's probably worth printing a message to the user where  that ".cache" download dir is.

>  if that's an option I can submit another PR.

That'd be great! 😊

PR #302, opened 2023-02-27
feat: add `requires_dependencies` decorator

alvarobartt:
## What's in this PR?

After playing around with `unstructured` for some time I've realized that having a Python decorator to check whether the required dependencies are installed before starting the execution and failing on the import e.g. the import is not at the top of the file.

So on, I've developed `@requires_dependencies()` which can be really useful now that there's an ongoing competition and many people may collaborate through PRs mainly in the connectors. Most likely each one of them will have specific dependencies.

## What's missing in this PR?

I've also thought about handling all the optional dependencies with `try-except ImportError` and providing a message in case those are not installed and how to do so, similar to the Python decorator but in a manual way. If this is something you'd also like please LMK and I'll address that too as part of this PR 👍🏻


MthwRobinson (changes requested):
@alvarobartt - Thanks for this! Could you add a test showing this decorator both working and not working? And can you also add a line to the `CHANGELOG` and bump `unstructured/__version__.py`?

alvarobartt:
> @alvarobartt - Thanks for this! Could you add a test showing this decorator both working and not working? And can you also add a line to the `CHANGELOG` and bump `unstructured/__version__.py`?

Sure @MthwRobinson I'll do that! Do you want me to bump it to a minor or just to the next `dev` release? Maybe minor is more suitable but LMK :) In the meantime I'll work on the integration tests!

MthwRobinson:
@alvarobartt - You can bump that to the next `-dev` release. We'll likely cut a new release sometime in the next few days.

MthwRobinson (approved):
LGTM with the updates! Seeing a test failure in CI but don't think that has to do with your PR. Going to run the job again and see if it passes.

MthwRobinson (commented):
Actually, any chance the decorator could be return a `None`? This is the error in CI

```
AttributeError: 'NoneType' object has no attribute 'initialize'
```

unstructured/utils.py
```diff
@@ -10,3 +11,29 @@ def save_as_jsonl(data: List[Dict], filename: str) -> None:
...
+                except ImportError:
+                    missing_deps.append(dep)
+            if len(missing_deps) > 0:
+                raise ImportError(
+                    f"Following dependencies are missing: {', '.join(missing_deps)}."
+                    + f"Please install them using `pip install unstructured[{extras}]`."
+                    if extras
+                    else f"Please install them using `pip install {' '.join(missing_deps)}`.",
+                )
+            func(*args, **kwargs)
```
tomaarsen, line 35:
```suggestion
            return func(*args, **kwargs)
```
This causes the test failures I reckon.

tomaarsen:
Note also that a simple decorator implementation like this can be detrimental to the usability of the methods/functions on which it is applied. Consider the following scripts:

#### Baseline
```python
import inspect

def foo(bar: int, var: bool) -> str:
    """My Foo Docstring"""
    return f"{bar} {var}"

print(f"foo signature: {inspect.signature(foo)}")
print(f"foo docstring: {foo.__doc__}")
```
```
foo signature: (bar: int, var: bool) -> str
foo docstring: My Foo Docstring
```

#### With naive decorator
```python
import inspect

@requires_dependencies(["numpy"])
def foo(bar: int, var: bool) -> str:
    """My Foo Docstring"""
    return f"{bar} {var}"

print(f"foo signature: {inspect.signature(foo)}")
print(f"foo docstring: {foo.__doc__}")
```
```
foo signature: (*args, **kwargs)
foo docstring: None
```

See for example https://github.com/argilla-io/argilla/pull/2380 for an example of how a decorator can be implemented while preserving this important information.

- Tom Aarsen

alvarobartt:
> Note also that a simple decorator implementation like this can be detrimental to the usability of the methods/functions on which it is applied. Consider the following scripts:
> 
> 
> 
> #### Baseline
> 
> ```python
> 
> import inspect
> 
> 
> 
> def foo(bar: int, var: bool) -> str:
> 
>     """My Foo Docstring"""
> 
>     return f"{bar} {var}"
> 
> 
> 
> print(f"foo signature: {inspect.signature(foo)}")
> 
> print(f"foo docstring: {foo.__doc__}")
> 
> ```
> 
> ```
> 
> foo signature: (bar: int, var: bool) -> str
> 
> foo docstring: My Foo Docstring
> 
> ```
> 
> 
> 
> #### With naive decorator
> 
> ```python
> 
> import inspect
> 
> from unstructured.utils import requires_dependencies
> 
> 
> 
> @requires_dependencies(["numpy"])
> 
> def foo(bar: int, var: bool) -> str:
> 
>     """My Foo Docstring"""
> 
>     return f"{bar} {var}"
> 
> 
> 
> print(f"foo signature: {inspect.signature(foo)}")
> 
> print(f"foo docstring: {foo.__doc__}")
> 
> ```
> 
> ```
> 
> foo signature: (*args, **kwargs)
> 
> foo docstring: None
> 
> ```
> 
> 
> 
> See for example https://github.com/argilla-io/argilla/pull/2380 for an example of how a decorator can be implemented while preserving this important information.
> 
> 
> 
> - Tom Aarsen

Thanks for the suggestion @tomaarsen I'll check this tomorrow morning! But it makes complete sense 🤗

MthwRobinson (approved):
LGTM, thanks for the contribution!

PR #307, opened 2023-02-28
build(deps): make pip-compile

cragwolfe:
Also skip the github ingest test 90% of the time in CI for now until creds are added.


ajjimeno (approved):
LGTM

PR #309, opened 2023-02-28
fix: track narrative text and figure captions in HTML documents

MthwRobinson:
### Summary

Fixes an issue introduced in [this commit](https://github.com/Unstructured-IO/unstructured/commit/74e6b84b41ff5456ba8f87958a2c7cde73d3c8d1#diff-9eb4b1248793302fa3b6bc9db25756503ea28f53ca556417c27f0be85d7a709b) that caused `NarrativeText` and `FigureCaption` elements to get represented as `Text` in HTML documents.

### Testing

The following previously showed no `NarrativeText` elements, but should now display `NarrativeText`.

```python
from unstructured.documents.elements import NarrativeText

url = "https://www.understandingwar.org/backgrounder/russian-offensive-campaign-assessment-february-8-2023"

narrative_text = [el for el in elements if isinstance(el, NarrativeText)][2:]
for element in narrative_text:
    print(element)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.17-dev1
+## 0.5.0
 
 ### Enhancements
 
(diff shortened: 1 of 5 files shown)


qued (approved):
LGTM!

PR #312, opened 2023-02-28
refactor: For S3 Ingest, write to file directly using `json.dump`

tomaarsen:
Hello!

## Pull Request overview
* Use `json.dump(..., file)` rather than `file.write(json.dumps(...))`.

## Details
Rather than writing to string in memory first, and then writing to file, this PR proposes to write to file directly using `json.dump`.

### References:
* [`json.dump`](https://docs.python.org/3/library/json.html#json.dump)
* [`json.dumps`](https://docs.python.org/3/library/json.html#json.dumps)

Personally, I always remember the difference as "dumps => dump string" (and idem for "loads => load string").

- Tom Aarsen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.4.17-dev1
+## 0.5.0
 
 ### Enhancements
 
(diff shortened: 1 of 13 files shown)


cragwolfe (approved):
LGTM!

tomaarsen:
I think all this needs is for the changelog (and __version__) version to be bumped. I can tackle this tomorrow if it hasn't been done by then.

cragwolfe:
@tomaarsen , I would have added the CHANGELOG / version change but don't think I can push to your branch.

tomaarsen:
You can't? Perhaps some of my settings are messed up.
No worries, though. I'll tackle it.

PR #314, opened 2023-02-28
feat: improve caching mechanism for `download_dir` on ingest

alvarobartt:
## What's in this PR?

Currently, the `download_dir` is not behaving as a cache, because it's randomly generated if not specified, which means that the files are downloaded over and over if `download_dir` is not specified, which is not ideal. So on, this PR introduces a caching mechanism to take into consideration the relevant fields that produce the data (in this case the remote path and/or params).


unstructured/ingest/main.py
```diff
@@ -221,6 +223,13 @@ def main(
             ),
         )
     elif github_url:
+        if not download_dir:
+            download_dir = (
+                cache_path
+                / hashlib.md5(
+                    f"{github_url}_{github_branch}_{github_file_glob}".encode("utf-8"),
```
cragwolfe, line 230: actually the glob shouldn't matter. if downloading "*html,*txt" in one run, in "*html" in another, i'd expect it to not redownload the html for the same github url and branch.

alvarobartt: Yeah probably that makes more sense, I just went fast over that connector because I was not super aware of it, but I'll remove that part now :)

P.S. Can you send me your no-reply email to add you as a co-author? or are you willing to add a suggestion so that I commit it directly through GitHub?

cragwolfe (commented):
@alvarobartt , thanks for taking on improving the experience of the defualt download_dir in the ingest.

A few quick comments:
* i cringe saying this, but please use a sha256 instead of a md5 hash. otherwise it is guaranteed a person wearing a security hat will raise eyebrows (even if they shouldn't).
* please update the PR title to be a bit more descriptive (like "caching mechanism for ingest downloads")
* please remove no longer relevant .gitignore: https://github.com/Unstructured-IO/unstructured/blob/3c1b089/.gitignore#L134-L136
* a bit nit picky, but for the sake of keeping each connector case (from line 210 downwards) as clean as possible, can we push the logic for `cache_path` to line 207?

E.g.:

```
if github_url:
    hashed_dir_name = hashlib.sha256(f"f"{github_url}_{github_branch}".encode("utf-8")).hexdigest()[0:10]
elif s3_url:
    ....
download_dir = cache_path / hashed_dir_name
```
Also another nitpick hidden in  ^^ -- probably don't need the full hexdigst string, just the first say 10 digits?

Thanks!

alvarobartt:
> @alvarobartt , thanks for taking on improving the experience of the defualt download_dir in the ingest.
> 
> A few quick comments:
> 
> * i cringe saying this, but please use a sha256 instead of a md5 hash. otherwise it is guaranteed a person wearing a security hat will raise eyebrows (even if they shouldn't).
> * please update the PR title to be a bit more descriptive (like "caching mechanism for ingest downloads")
> * please remove no longer relevant .gitignore: https://github.com/Unstructured-IO/unstructured/blob/3c1b089/.gitignore#L134-L136
> * a bit nit picky, but for the sake of keeping each connector case (from line 210 downwards) as clean as possible, can we push the logic for `cache_path` to line 207?
> 
> E.g.:
> 
> ```
> if github_url:
>     hashed_dir_name = hashlib.sha256(f"f"{github_url}_{github_branch}".encode("utf-8")).hexdigest()[0:10]
> elif s3_url:
>     ....
> download_dir = cache_path / hashed_dir_name
> ```
> 
> Also another nitpick hidden in ^^ -- probably don't need the full hexdigst string, just the first say 10 digits?
> 
> Thanks!

Sure @cragwolfe thanks for the feedback I'll address all that now 🤗

unstructured/ingest/main.py
cragwolfe: > P.S. Can you send me your no-reply email to add you as a co-author? or are you willing to add a suggestion so that I commit it directly through GitHub?

Thank you for asking -- I'm ok just being a reviewer on this one 😊

cragwolfe:
Oh, I forgot one one other suggestion: If the user has `--preserve-downloads` and `--download-dir` is not specified, `main` could print a line indication the name of the cache dir where the files will be saved. wdyt?

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 0.5.1-dev0
+
+### Enhancements
+
+* Default `download_dir` now lives under `$HOME/.cache/unstructured/ingest` instead
```
cragwolfe, line 5: Nit: let's call out `unstructured-ingest`, so "Default `download_dir`" -> "Default `unstructured-ingest` `download_dir`."

Also maybe instead of "instead of in the root directory under the name of `tmp-ingest-...`" something like "instead of a "tmp-ingest-" dir in the the working directory"

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+of in the root directory under the name of `tmp-ingest-...`.
+
+### Features
+
+### Fixes
+
+* As of the caching update, now files are not downloaded over and over when 
+`download_dir` is not specified since its path is not random anymore.
```
cragwolfe, line 13: Nits: again let's specified for `unstructured-ingest` but also "if invoked with --preserve-downloads". Also "are not downloaded over and over" -> "are not re-downloaded"

unstructured/ingest/connector/s3_connector.py
```diff
@@ -175,7 +175,8 @@ def initialize(self):
             raise ValueError(
                 f"No objects found in {self.config.s3_url} -- response list object is {response}",
             )
-        os.mkdir(self.config.download_dir)
+        if not Path(self.config.download_dir).exists():
+            os.mkdir(self.config.download_dir)
```
cragwolfe, line 179: This can be abbreviated to one line: `os.mkdir(self.config.download_dir, exist_ok=True)`.

It feels kind of odd that this would only need to be added to this connector, though.

cragwolfe: OK, looking more closely at s3_connector, this dir does not need to be created here in the first place because any time a file is downloaded, all parent dirs are created if needed:
https://github.com/Unstructured-IO/unstructured/blob/e52dd5c/unstructured/ingest/connector/s3_connector.py#L86 .
Tested to confirm, but can go ahead and remove this block.

alvarobartt: Ok I'll update both that suggestion and the changelog too, thanks 👍🏻

cragwolfe (approved):

Issue #306, opened 2023-02-28
Add github creds to CI for the github ingest test


cragwolfe:
The [github ingest test](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-github.sh) may fail due to rate limiting, which is easy to hit with unauthenticated requests. Authenticated requests are limited at a higher rate. Reference: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting .

Let's update the test to take advantage of CI creds for the higher rate limit.

PR #317, opened 2023-02-28
bump: release commit

cragwolfe:
Updates the github-ingest test's expected output, which is a result of https://github.com/Unstructured-IO/unstructured/pull/309 .
Includes tweaks so that the github ingest test script now runs 20% of the time. The above would have been caught sooner if this ran 100% of the time. Related issue: https://github.com/Unstructured-IO/unstructured/issues/306 .


ajjimeno (approved):
LGTM

Issue #257, opened 2023-02-23
Create a data conector for Azure Blob Storage


benjats07:
Create a data connector that pulls data from [Azure Blob Storage ](https://learn.microsoft.com/en-us/azure/storage/blobs/)

See [Adding Data Connectors](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#adding-data-connectors) for details on how to get started. Make sure to include a link to this issue when submitting a PR.

**Definition of Done**

- The [checklist](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist) has been completed.
- The connector is able to process a single object or a container.
- The connector is able to process the entire container.
- For private files, it requires the credentials of access.
- For now, it is OK to process only doc types that [partition.auto()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py) is capable of processing. 
    
Bonus points: the ability to filter by document type.

alvarobartt:
Hi @benjats07 I'd like to actually tackle this issue, can I? Thanks

benjats07:
@alvarobartt sure, we are receiving multiple solutions to all issues in the repo, so go for it!

benjats07:
Currently, we don't have any container for that, feel free to create an account for that, the free tier should be enough for that.

alvarobartt:
As mentioned in https://github.com/Unstructured-IO/unstructured/pull/318#issuecomment-1450619852, there are still some things to improve, such as the naming and also to add integration tests. For the latter, I'd appreciate it if someone from @unstructured-io could "clone" somehow the files from S3 to Azure Blob Storage (which can be done under the free-tier AFAIU) so that I can develop the integration tests using that URL.

alvarobartt:
Did you get the chance to think about the naming for this connector @benjats07? I kept `ABSConnector` but ABS acronym is not as well known/common as S3 for AWS S3 or GCS for Google Cloud Storage, so I'm happy to open the naming to discussion 👍🏻

benjats07:
Sorry about the delay @alvarobartt.
Could you check with this [link](https://datalakebenja22.blob.core.windows.net/demo-files?sp=r&st=2023-03-03T18:11:01Z&se=2023-04-04T01:11:01Z&sv=2021-06-08&sr=c&sig=%2FF3fRTWUJE1JWpdYhnMUjVBP7LZW6g8nSrVbotltL7k%3D) ? 

[...]

PR #318, opened 2023-03-01
feat: add `FsspecConnector` to easily integrate new connectors with a `fsspec` implementation available

alvarobartt:
## What's in this PR?

So as you may see this is a pretty big PR, that basically adds an "adapter" to easily plug in any connector with an available `fsspec` implementation. This is a way to standardize how the remote filesystems are used within `unstructured`.

I've additionally renamed `s3_connector.py` to `s3.py` for readability and consistency and tested that the current approach works as expected and is aligned with the expectations.


tomaarsen:
This isn't a cache issue. The CI performs the following installs prior to starting the ingest tests:
https://github.com/Unstructured-IO/unstructured/blob/95109db6b0441e2709c10a8303b0a5e6b3264a85/.github/workflows/ci.yml#L110-L112
We would expect for `make install-ingest-s3` to install the (new) required dependencies (`s3fs` and `fsspec`), but `make install-ingest-s3` is simply this:
https://github.com/Unstructured-IO/unstructured/blob/95109db6b0441e2709c10a8303b0a5e6b3264a85/Makefile#L52-L55

And this PR does not update the [`requirements/ingest-s3.txt` file](https://github.com/Unstructured-IO/unstructured/blob/main/requirements/ingest-s3.txt), which still happily installs the old dependency of `boto3`:
https://github.com/Unstructured-IO/unstructured/blob/95109db6b0441e2709c10a8303b0a5e6b3264a85/requirements/ingest-s3.txt#L19-L20
And does not install e.g. `s3fs` or `fsspec`. Resolving this issue is as simple as running
```
pip-compile --extra=s3 --output-file=requirements/ingest-s3.txt requirements/base.txt setup.py
```
and committing the changes.

- Tom Aarsen

alvarobartt:
I've also another doubt @cragwolfe for `gcs` and `s3` seems straightforward to define the protocol name and the name of the connector classes etc. but for Azure Blob Storage (ABS) is weird, because:

1. It's not usually shortened as ABS and that's not a well-known acronym
2. ABS is the acronym but the protocol is `adfs`
3. Since `adlfs` supports both Azure Blob Storage and Azure DataLake Gen1 & 2, should we rename the `ABSConnector` to `AzureConnector` and handle both filesystems? Or is that too wide?

I think that we need to discuss the Azure Blob Storage integration further, but that can be done separately in #257

.github/workflows/ci.yml
```diff
@@ -78,50 +77,53 @@ jobs:
...
+        id: changes
+        with:
+          filters: |
+            src:
+              - 'unstructured/**'
 
-    - if: steps.changes.outputs.src == 'true' && github.ref != 'refs/heads/main'
-      uses: dangoslen/changelog-enforcer@v3
+      - if: steps.changes.outputs.src == 'true' && github.ref != 'refs/heads/main'
+        uses: dangoslen/changelog-enforcer@v3
```
cragwolfe, line 129: why all the formatting differences? it makes it hard to find the real diff if there is one (is there?).
same with other .yml file.

Makefile
```diff
@@ -49,11 +49,18 @@ install-dev:
 install-build:
 	pip install -r requirements/build.txt
 
-## install-ingest-s3:       install requirements for the s3 connector
 .PHONY: install-ingest-s3
```
cragwolfe, line 52: The double `##` means that when the user types `make` those are the list of advertised targets. I think it is worth advertising all of the `install-ingest-` targets.

requirements/ingest-s3.txt
```diff
@@ -12,16 +22,16 @@ argilla==1.3.1
...
     # via
     #   -r requirements/base.txt
     #   argilla
-boto3==1.26.80
-    # via unstructured (setup.py)
-botocore==1.29.80
-    # via
-    #   boto3
-    #   s3transfer
+botocore==1.27.59
```
cragwolfe, line 33: interesting that s3fs wants a significantly older version of botocore, published Aug 24, 2022 🤔

cragwolfe (commented):
https://github.com/aio-libs/aiobotocore not being highly maintained gives me pause on making this the default for s3. In the context of ingest multiprocessing fetching on file per process, aiobotocore probably isn't helping much over boto3 either.

Besides that (and additional considerations below), Fsspec seems like a really neat library to support and quickly get a broad range of sources to connect to.

What about making `fsspec` the connector where `filesystem` could be the first param? Not entirely sure how the filesystem-specific extra params would be passed through but there are ways to solve that. Maybe easier if `ffspec` is a subcommand.  I'm also not entirely sure how to handle the dependencies as they may be different than a "native" implementation, e.g. s3. I mean, there could be a lot of `make install-fsspec-` targets and setup.py extras, but that's starting to get excessive. Maybe draw the line there and leave it to the user to the the deps right.

@tomaarsen 's prior comments on  Azure Blob Storage (ABS) also kind of align here. If there are quirks for other filesystems in fsspec that make them best handled natively, then that option is preserved by just having fsspec remain its own subcommand.

.github/workflows/ci.yml
alvarobartt: It's just formatting, as those were not compliant with the standard YAML formatting, mainly inconsistent use of tabs + spaces, as well as some indents missing (nothing changed in terms of functionality). Anyway, I committed those style changes separately so that the difference can be easily seen at https://github.com/Unstructured-IO/unstructured/pull/318/commits/ee58b522fefb047528bfdfc1c036998b15e46e7c

Makefile
alvarobartt: Ok I'll put that back in!

requirements/ingest-s3.txt
alvarobartt: So AFAIU `s3fs` is actively maintained (see https://github.com/fsspec/s3fs) and widely use by some popular libraries such as `transformers` or `datasets`, since it's easy to plug in new remote data sources based on `fsspec`.

.github/workflows/ci.yml
cragwolfe: OK, I appreciate that! I'd request there be a separate PR for formatting-only yaml changes for the sake of unrelated PR (and main branch commit) diff readability.

alvarobartt:
> https://github.com/aio-libs/aiobotocore not being highly maintained gives me pause on making this the default for s3. In the context of ingest multiprocessing fetching on file per process, aiobotocore probably isn't helping much over boto3 either.

So AFAIU s3fs is actively maintained (see https://github.com/fsspec/s3fs) and widely used by some popular libraries such as `transformers` or `datasets`, since it's easy to plug in new remote data sources based on `fsspec`.

> Besides that (and additional considerations below), Fsspec seems like a really neat library to support and quickly get a broad range of sources to connect to.
> 
> What about making `fsspec` the connector where `filesystem` could be the first param? Not entirely sure how the filesystem-specific extra params would be passed through but there are ways to solve that. Maybe easier if `ffspec` is a subcommand. I'm also not entirely sure how to handle the dependencies as they may be different than a "native" implementation, e.g. s3. I mean, there could be a lot of `make install-fsspec-` targets and setup.py extras, but that's starting to get excessive. Maybe draw the line there and leave it to the user to the the deps right.

I think the line should be drawn at this point, because support is offered for `s3`, `gcs`, and `abs`, which are the 3 main cloud storage providers. For the rest of them, we just offer the abstract connector so that they can either add their own based on `fsspec` or just improve the CLI so that if the protocol is an available `fsspec` implementation (registered in `fsspec`) then the user can still use the `unstructured-ingest` over that filesystem but being in charge and control of the specific dependencies that the filesystem may need.

> @tomaarsen 's prior comments on Azure Blob Storage (ABS) also kind of align here. If there are quirks for other filesystems in fsspec that make them best handled natively, then that option is preserved by just having fsspec remain its own subcommand.

Also, the Azyure Blob Storage comments above were mine not @tomaarsen's 😆, but sure I agree, maybe we can merge this time (just active S3 support) and I'll work in the CLI improvement in a separate PR to both keep the current functionality but also support a fast plug-and-play for any other filesystem with a `fsspec` interface, if that's OK with you?

cragwolfe:
> Also, the Azyure Blob Storage comments above were mine not @tomaarsen's 😆, 

my bad!

> maybe we can merge this time (just active S3 support) 

I guess i don't see the benefit of immediately replacing the existing connector vs keeping them side by side (hesitation being really old version of boto3, unneeded async layer from aiobotocore, weirder stack traces when something goes wrong through extra layers of abstraction) but if the [test](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-s3.sh) passes, sure. we could always revert.

alvarobartt:
> I guess i don't see the benefit of immediately replacing the existing connector vs keeping them side by side (hesitation being really old version of boto3, unneeded async layer from aiobotocore, weirder stack traces when something goes wrong through extra layers of abstraction) but if the [test](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-s3.sh) passes, sure. we could always revert.

Ok then I'll solve the merge conflicts and then we can evaluate whether the replacement is worth it or not 😄

alvarobartt:
Hi @cragwolfe so the CI is failing when running `apt-get` can you please have a look at it? I made sure everything's working fine for the `S3Connector` with different args both through the CLI and through the existing unit tests.

cragwolfe:
@alvarobartt , that was fixed in https://github.com/Unstructured-IO/unstructured/commit/a5da3de43b2d569d0ccd69dcf2874dcfa50e737b ,so should be fixed when you merge your branch with main.

alvarobartt:
Hi again @cragwolfe hope you're having a great weekend! So after the discussion with @benjats07 in #257, I realized that it might be better to keep this PR just for the replacement of `S3Connector` to be `fsspec`-based as well as to add the `FsspecConnector` to easily allow the ingestion of files from any available `fsspec` registered implementation.

So on, if you agree I'll just keep `FsspecConnector` and `S3Connector` in this PR, and submit separate ones for both `GCSConnector` and `AzureBlobStorageConnector` as well as unit tests for both of those, and if applicable another PR to easily allow the ingestion via CLI of files from any `fsspec` implementation.

unstructured/ingest/main.py
```diff
@@ -84,6 +88,22 @@ def run(self):
...
+)
+@click.option(
+    "--azure-account-name",
+    default=None,
+    help="Azure Blob Storage or DataLake account name.",
+)
+@click.option(
+    "--azure-connection-string",
+    default=None,
+    help="Azure Blob Storage or DataLake connection strin.",
```
cragwolfe, line 105:
```suggestion
    help="Azure Blob Storage or DataLake connection string.",
```

cragwolfe (commented):
This is looking good and have confirmed the s3 connector is working, but have a few requests:

1 / Please create a separate tidy-only PR for the bulk of the yaml changes in .github/workflows -- without that, it makes it hard to see what's actually changing in this PR.

2/ The content under Ingest.md und "Adding Data Connectors" should be updated in a few places since `s3_connector.py` no longer exists. Also, for the very first sentence in that section, perhaps refer to the `github` connector as the more straightforward connector for a user to refer to when writing one?

3/ aiobotocore was just [bumped](https://github.com/aio-libs/aiobotocore/pull/999) to not have an ancient version of botocore! 🎉  Please `make pip-compile` again to include that.

alvarobartt:
> This is looking good and have confirmed the s3 connector is working, but have a few requests:

Thanks for the feedback @cragwolfe :hugs:

> 1 / Please create a separate tidy-only PR for the bulk of the yaml changes in .github/workflows -- without that, it makes it hard to see what's actually changing in this PR.

Yes, I'll remove that from this PR and create a tiny one for cosmetic issues later today! Also, do you want me to also remove `AzureBlobStorageConnector` from this PR and create a separate PR for it as well as for `GCSConnector`? I think that may be cleaner?

> 2/ The content under Ingest.md und "Adding Data Connectors" should be updated in a few places since `s3_connector.py` no longer exists. Also, for the very first sentence in that section, perhaps refer to the `github` connector as the more straightforward connector for a user to refer to when writing one?

Sure, I'll update `Ingest.md` to explain both how to add custom connectors, and also how to use any `fsspec` implementation.

> 3/ aiobotocore was just [bumped](https://github.com/aio-libs/aiobotocore/pull/999) to not have an ancient version of botocore! 🎉 Please `make pip-compile` again to include that.

Cool, that's great news! I'll re-run `make pip-compile` :)

alvarobartt:
Hi @cragwolfe so everything should be ready :) I'll create a separate PR to fix YAML formatting, one for `AzureBlobStorageConnector` (ready to be published too), and `GCSConnector`. I may also submit another PR for some spelling/typos I've seen while reviewing the docs 👍🏻

unstructured/ingest/main.py
```diff
@@ -290,13 +290,13 @@ def main(
...
+                path=s3_url,
+                access_kwargs={"anon": s3_anonymous},
                 download_dir=download_dir,
-                s3_url=s3_url,
                 output_dir=structured_output_dir,
-                # set to False to use your AWS creds (not needed for this public s3 url)
-                anonymous=s3_anonymous,
                 re_download=re_download,
                 preserve_downloads=preserve_downloads,
+                verbose=verbose,
```
tomaarsen, line 299:
```suggestion
                verbose=verbose,
```
This should no longer be necessary, the verbosity should all be handled in `main.py` since the logger changes.

alvarobartt: Nice catch, probably there from a merge conflict resolved badly? I'll check the rest in case there's any other `verbose` still used :)

cragwolfe (approved):
LGTM! 🎉 

> So on, if you agree I'll just keep FsspecConnector and S3Connector in this PR, and submit separate ones for both GCSConnector and AzureBlobStorageConnector as well as unit tests for both of those, and if applicable another PR to easily allow the ingestion via CLI of files from any fsspec implementation.

That sounds good, thank you!

unstructured/ingest/main.py
cragwolfe: @alvarobartt , I just removed that `verbose` line before seeing your comment. The s3 connector at least runs fine after that.

alvarobartt:
Hi @cragwolfe is the CI supposed to take that long? It seems that it's been stuck there for about 20 minutes already 😕

tomaarsen:
I downloaded the logs, and it seems that it was previously stuck on
```
2023-03-08T08:33:38.8939803Z + ./test_unstructured_ingest/test-ingest-s3.sh
2023-03-08T09:14:56.7876303Z ##[error]The operation was canceled.
```
until the job was restarted. It seems it's stuck again, as it is now on attempt #2.
Does `./test_unstructured_ingest/test-ingest-s3.sh` work for you locally @alvarobartt?

alvarobartt:
> I downloaded the logs, and it seems that it was previously stuck on
> 
> ```
> 2023-03-08T08:33:38.8939803Z + ./test_unstructured_ingest/test-ingest-s3.sh
> 2023-03-08T09:14:56.7876303Z ##[error]The operation was canceled.
> ```
> 
> until the job was restarted. It seems it's stuck again, as it is now on attempt # 2. Does `./test_unstructured_ingest/test-ingest-s3.sh` work for you locally @alvarobartt?

It does work, but just if I comment https://github.com/Unstructured-IO/unstructured/blob/1580c1bf8eed5378797ab04d0ae80037a59d5e0b/test_unstructured_ingest/test-ingest-s3.sh#L6-L10, otherwise it fails, but that shouldn't be related to this PR...

Screenshots are attached below:

Screenshot 2023-03-08 at 11 25 10

Screenshot 2023-03-08 at 11 27 08

Screenshot 2023-03-08 at 11 29 17

tomaarsen:
That looks normal to me. I'm mostly surprised that GitHub is cool with jobs running for this long.

alvarobartt:
> That looks normal to me. I'm mostly surprised that GitHub is cool with jobs running for this long.

Assuming that the mean time for those is around 10 minutes or less, I think we could limit the running time of the job and the number of attempts too because otherwise, this is wasting resources for no reason, but again, not sure why it is stuck there :/

alvarobartt:
Hi, @cragwolfe so the `test-ingest-s3.sh` is failing as pointed out earlier by @tomaarsen but I cannot find the root cause of the issue because I reproduced everything locally and it works fine... I did some modifications to the script just before to see whether that could solve it because locally I couldn't run the `test-ingest-s3.sh`, but it seems to be stuck also there too... Any hint?

alvarobartt:
After exploring this a little bit the issue seems to be related to GitHub Actions and `s3fs` because I created a GitHub Actions from `main`, and works fine, but when checking out this branch instead it gets stuck... I tried to reproduce this locally but I'm unable to, so the issue just happens on GitHub Actions... What can do about this?

cragwolfe:
> After exploring this a little bit the issue seems to be related to GitHub Actions and s3fs because I created a GitHub Actions from main, and works fine, but when checking out this branch instead it gets stuck... I tried to reproduce this locally but I'm unable to, so the issue just happens on GitHub Actions... What can do about this?

@alvarobartt , I don't think there is a great answer. The s3 test script should remain a part of CI for a couple of good reasons.

The best idea I have at the moment is to bring back the old connector which is the default for --s3 args, but allow using the fsspec connector with an additional flag. I'm open to other ideas, but also have pretty high conviction the s3 ingest test script should remain.

alvarobartt:
> > After exploring this a little bit the issue seems to be related to GitHub Actions and s3fs because I created a GitHub Actions from main, and works fine, but when checking out this branch instead it gets stuck... I tried to reproduce this locally but I'm unable to, so the issue just happens on GitHub Actions... What can do about this?
> 
> @alvarobartt , I don't think there is a great answer. The s3 test script should remain a part of CI for a couple of good reasons.
> 
> The best idea I have at the moment is to bring back the old connector which is the default for --s3 args, but allow using the fsspec connector with an additional flag. I'm open to other ideas, but also have pretty high conviction the s3 ingest test script should remain.

Ok makes sense @cragwolfe I'll add the former `S3Connector` back, do you agree on still keeping this S3 connector in `_s3.py`? I'll continue exploring this issue further to find the root cause of the issue...

cragwolfe:
> Ok makes sense @cragwolfe I'll add the former S3Connector back, do you agree on still keeping this S3 connector in _s3.py? I'll continue exploring this issue further to find the root cause of the issue...

Yes, fine with me to keep the fsspec around in _s3.py and the old connector moved to s3.py , if that is what you mean.

alvarobartt:
So the issue seems to be related to `multiprocessing` and `fsspec` in Linux, as in macOS I'm not able to reproduce it, so I'll try to explore this further and hopefully create another PR on top to solve that since I think I have a stable solution but requires more testing.

But for now I'd say we're good to merge this PR 👍🏻

test_unstructured_ingest/test-ingest-s3.sh
```diff
@@ -1,24 +1,24 @@
...
-   echo
-   echo "There are differences from the previously checked-in structured outputs."
-   echo 
-   echo "If these differences are acceptable, copy the outputs from"
-   echo "s3-small-batch-output/ to test_unstructured_ingest/expected-structured-output/s3-small-batch/ after running"
-   echo 
-   echo "  PYTHONPATH=. python examples/ingest/s3-small-batch/main.py --structured-output-dir s3-small-batch-output"
-   echo
-   exit 1
+if [[ "$(diff -r test_unstructured_ingest/expected-structured-output/s3-small-batch/ s3-small-batch-output/ | wc -l)" -ne 0 ]]; then
```
cragwolfe, line 14: i had intentionally not piped the diff output so one could quickly see what did change without having to run anything

cragwolfe: that could be a lot of output though, maybe do `| head -100 ` instead?

alvarobartt: Hm but here we're printing just the newline counts with `wc -l`, right?

alvarobartt: I can revert that if needed

cragwolfe: Right, which means the user has no idea what the actual diff is.

> I can revert that if needed

Yes, please. 🙏 

There is a similar diff in the github ingest script, we can update both in a separate PR for whatever the best user experience is going to be.

tomaarsen:
> So the issue seems to be related to `multiprocessing` and `fsspec` in Linux, as in macOS I'm not able to reproduce it, so I'll try to explore this further and hopefully create another PR on top to solve that since I think I have a stable solution but requires more testing.

Would this interfere with Linux users' ability to use the functionality from this PR? 
Even if yes, I suppose it beats not having the functionality for anyone, haha

alvarobartt:
> > So the issue seems to be related to `multiprocessing` and `fsspec` in Linux, as in macOS I'm not able to reproduce it, so I'll try to explore this further and hopefully create another PR on top to solve that since I think I have a stable solution but requires more testing.
> 
> Would this interfere with Linux users' ability to use the functionality from this PR? Even if yes, I suppose it beats not having the functionality for anyone, haha

I'll explore this further today and come back with either the docstrings update or a solution on that :)

But yes it seems just a `s3fs` issue as far as I could check, anyway I'll explore this further today before we merge!

alvarobartt:
Ok I've already solved it and made sure it works on macOS, Windows, and Linux, so I'll push the commits now to use `_s3.py` instead, and if the unit tests pass, then I'll remove `s3.py` and rename `_s3.py` to `s3.py`

I'll give you more details once I make sure the CI is also passing @cragwolfe @tomaarsen

alvarobartt:
So the issue seems to be just occurring in Linux since the default `multiprocessing` start method is `fork` (which is not safe on macOS), while on both Windows and macOS the default start method is `spawn`. So on, for consistency, if we set the default start method to `spawn`, the `s3fs` will also work in Linux, as the issue happens due to `multiprocessing` using `fork` as the start method.

To check that Windows, Linux, and macOS support `spawn` one can run `multiprocessing.get_all_start_methods()` and `spawn` will always be listed there, so the introduced `multiprocessing.set_start_method("spawn")` won't break anything in any OS.

Also, I took the decision to include that on the top of the file surrounded with a `try-except`, and as far as I could check that's the safest way, even though it could also be placed inside the `if __name__ == "__main__"`.

I would have posted more references, but I checked a lot of resources and took ideas from here and there while trying to test this on Windows, Linux, and macOS to ensure it works consistently across OS.

* https://docs.python.org/3/library/multiprocessing.html
* https://superfastpython.com/multiprocessing-start-method/
* https://github.com/fsspec/s3fs/issues/464

alvarobartt:
Hi @cragwolfe, this has already been fixed! See comment above 👍🏻

alvarobartt:
Another finding is that when using `fork` the logs were duplicated as anyone can check in previous runs of the CI when running `test_unstructured_ingest/test-ingest-s3.sh`, plus this implementation being a little bit faster than the previous one (but this may have something to do with the latency), as the former run of `test_unstructured_ingest/test-ingest-s3.sh` was taking around 3 minutes in the CI job, while the one implemented here takes around 2:30 minutes.

unstructured/ingest/main.py
```diff
@@ -14,14 +15,17 @@
...
+from unstructured.ingest.connector.s3 import S3Connector, SimpleS3Config
 from unstructured.ingest.connector.wikipedia import (
     SimpleWikipediaConfig,
     WikipediaConnector,
 )
 from unstructured.ingest.doc_processor.generalized import initialize, process_document
 from unstructured.ingest.logger import ingest_log_streaming_init, logger
 
+with suppress(RuntimeError):
+    mp.set_start_method("spawn")
```
cragwolfe, line 27: Big change here and I agree it seems to be a good one.

To confirm, I tested all the other example scripts to confirm on both linux and mac. 😅

cragwolfe:
All very nice detailed comments, and thanks for the references @alvarobartt !

...the prior double logging on Linux 🤦

tomaarsen:
Glad to see this merged by the way. The use of `fsspec` is quite smart.

PR #321, opened 2023-03-01
refactor: Fully move towards logging; remove `if config.verbose` conditionals

tomaarsen:
Hello!

## Pull Request overview
* Move away from printing, use logging exclusively.
  * Rely on the recommended `logger = logging.getLogger(__name__)` rather than importing a logger from `unstructured/logger.py`.
  * ~~Initialize the logger from `unstructured/__init__.py`.~~
  * ~~When [`rich`](https://github.com/Textualize/rich) is installed, use its logging Handler.~~
* ~~When `rich` is installed, use its Traceback formatter.~~

## Details
Rather than partially using logging and partially using printing, this PR modernizes the project by moving fully to logging. Inspiration for this change was the unnecessary `if self.config.verbose:` conditionals throughout the ingest data connectors: these can easily be replaced with one `if verbose: logger.setLevel(logging.DEBUG)`, while all `print` statements are replaced with `logger.debug` calls.
As a result, the ingest data connector files are much cleaner. Note also that I was able to remove the `verbose` argument to the ingest `BaseConnectorConfig` classes, as the verbosity is now handled via one conditional in `unstructured/ingest/main.py`.

### ~~Optional `rich` support~~
~~Inspired by Argilla's recent move to `rich` (https://github.com/argilla-io/argilla/pull/2350), I believed that the support for it for `unstructured` would be equally beneficial. In particular, a logging handler and a traceback formatter are used if `rich` is installed. These are some screenshots of what these may look like:~~

<details><summary>With rich installed</summary>

image

image

image

</details>

<details><summary>Without rich installed</summary>

image

image

image

</details>

### ~~Logging support~~
~~As we now properly use logging, third party tools can enable or disable our logs as they wish, e.g. by modifying the log levels of our logger named "unstructured". The remaining loggers that are created using `logger = logging.getLogger(__name__)` are named something like "unstructured.partition.auto", and they will propagate their logs down to "unstructured", which actually handles the logs (i.e. via the logger).~~

## ~~Question~~
~~1. With this optional `rich` support in place, the question remains of how the dependency should be included. Perhaps in `dev` if we do not want to enforce people to use it? Or perhaps we can propose an `unstructured[rich]`? Or perhaps we just place it as a required dependency, and give people the option to uninstall it later?~~

~~The `rich` functionality is quite "loud" and subjective. With other words, they may not be liked by everyone. I am open to changing this PR to exclude the `rich` functionality, but I think the best solution would be to not include `rich` in any of the dependencies if users or maintainers are not all a fan.~~
I am of the belief that the logger changes are quite important, however.

- Tom Aarsen


alvarobartt:
I think we thought about that same here 🤣 But I created a draft PR using `structlog` instead of `rich` as a handler as you did! Nicely done BTW 💪🏻 

Anyway, I'm open to discussion, because `structlog` supports context vars which are really cool, and work smoothly in async environment.

Draft PR at #322

alvarobartt:
I can provide you a screenshot of `process_document` logs if you think that's useful for the sake of comparing both approaches, but I'm OK to drop my PR in behalf of yours, since as you're already using it at Argilla, it may make more sense to be aligned.

alvarobartt:
> I can provide you a screenshot of `process_document` logs if you think that's useful for the sake of comparing both approaches, but I'm OK to drop my PR on behalf of yours, since as you're already using it at Argilla, it may make more sense to be aligned.

image

unstructured/__init__.py
```diff
@@ -0,0 +1,5 @@
+from unstructured.logger import initialize_logger
+from unstructured.utils import initialize_rich_tracebacks
+
+initialize_logger(__name__)
+initialize_rich_tracebacks()
```
cragwolfe, line 5: So, the library itself should not be doing any logging initialization as that is in the domain of end-applications which are going to have different opinions on how to do that.

ref: https://docs.python.org/3/howto/logging-cookbook.html#adding-handlers-other-than-nullhandler-to-a-logger-in-a-library

Everywhere else in the library should just import the `logger` from `unstructured.logger` and use that to log message.

https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/main.py when invoked as a command is fair game i think, though, because it is an actual application. Though, dumb terminals should still work OK.

tomaarsen: > So, the library itself should not be doing any logging initialization as that is in the domain of end-applications which are going to have different opinions on how to do that.

> ref: https://docs.python.org/3/howto/logging-cookbook.html#adding-handlers-other-than-nullhandler-to-a-logger-in-a-library

Understood, I'll revert those changes.

> Everywhere else in the library should just import the logger from unstructured.logger and use that to log message.

I have always understood `logger = logging.getLogger(__name__)` to be the correct approach for logging, e.g. as suggested on the [Advanced Logging Tutorial](https://docs.python.org/3/howto/logging.html#logging-advanced-tutorial). The logger at e.g. `"unstructured.partition.auto"` will propagate the logs to the core logger at `"unstructured"`.
However, the docs on [logging for a library](https://docs.python.org/3/howto/logging.html#library-config) are not very clear on this matter. 

I can revert to using a single logger from `unstructured.logger` and import that throughout the project.

tomaarsen: Should be all set now. I appreciate the valuable review! It's always good to learn something new :)

cragwolfe: For sure, I appreciate PR's intending to better the code base!

cragwolfe:
OK, I started to test this out thinking it was a quick one, and ... issues 😆 (not the fault of this PR).

[1] Here's one issue: https://github.com/Unstructured-IO/unstructured/issues/331 : with --verbose , there is a ton out of output that is way too verbose.

[2] Another one: logger.info lines added in this PR do not emit in subprocesses if logging is configured under `__name__ == __main__`:

    logger = logging.getLogger("unstructured")
    handler = logging.StreamHandler()
    handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s %(message)s"))
    logger.addHandler(handler)
    logger.setLevel(logging.INFO)

To keep it simple for the time being, i think the thing to do for now is add an [initializer](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool) when the [Pool is instantiated](https://github.com/Unstructured-IO/unstructured/blob/707f92f/unstructured/ingest/main.py#L68) to set up logging. This could optionally be done given a keyword arg in `main`.

[3] I'm thinking it makes sense to have an `"unstructured.ingest"` logger as `--verbose` does not necessarily imply the user wants unstructured debug messages, even after [1] is addressed.

tomaarsen:
1 and 3 are quite sensible, but I am wondering how users would be able to specify that they *do* want debug logs for the 'regular' `"unstructured"` logger, if they call the program via the CLI?
As for 2, logging does not currently output in the terminal due to the handler that I removed in my last commit. This is not strictly a multiprocessing issue. Note that neither the `logger` nor its parent (the root logger) have handlers. They are also both set to the WARNING level by default, resulting in no logs. 
However, I can't manage to get `logger.info` to output any logs, even if I lower all of the levels and add handlers...

cragwolfe:
> 1 and 3 are quite sensible, but I am wondering how users would be able to specify that they do want debug logs for the 'regular' "unstructured" logger, if they call the program via the CLI?

Maybe. I was thinking of it from the perspective that if I'm processing hundreds or thousands of files, --verbose means give me all the details as files are downloaded and processed. If that is multiplied by 10 to 100 with unstructured-the-lib debug messages, eh. There should be a way to turn unstructured-the-lib debugging, though.

> However, I can't manage to get logger.info to output any logs, even if I lower all of the levels and add handlers...

Right, that is what I was saying with [2]. If you take out `mp.Pool` and do the processing in the same process, you see the logs.  I think it needs to be configured in each subprocess (or venture into true multiprocessing logging -- not super excited with that idea). 

> Oh, completely unrelated, but if you're on Python 3.7 onwards, you can use breakpoint() as a shortcut for import pdb; pdb.set_trace(). I'd recommend it!

🙏 😊

tomaarsen:
> > However, I can't manage to get logger.info to output any logs, even if I lower all of the levels and add handlers...
> 
> Right, that is what I was saying with [2]. If you take out `mp.Pool` and do the processing in the same process, you see the logs. I think it needs to be configured in each subprocess (or venture into true multiprocessing logging -- not super excited with that idea).

I actually don't get any logs, regardless of `mp.Pool`. Both with, and without, your 
```python
    logger = logging.getLogger("unstructured")
    handler = logging.StreamHandler()
    handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s %(message)s"))
    logger.addHandler(handler)
    logger.setLevel(logging.INFO)
```
in `__name__ == "__main__"`.
(And also with `logging.basicConfig(...)` in like in https://github.com/Unstructured-IO/unstructured/commit/d51ea3a8c744e5acbd65a89d48949792bdd1f21c#diff-c7866d4747abb72555855965dc619655bb451c16e30240489eeec69eec57cef0R310)

cragwolfe:
Hmm, in my branch unchanged the tail of the output (after a lot of unstructured debugging output) is:

```
2023-03-02 00:34:08,330 unstructured INFO Wrote wikipedia-ingest-output/Open-source software-1137981045-txt.json
Wrote wikipedia-ingest-output/Open-source software-1137981045-txt.json
2023-03-02 00:34:08,361 unstructured DEBUG Cleaning up WikipediaIngestTextDoc(page=<WikipediaPage 'Open-source software'>)
2023-03-02 00:34:08,361 unstructured DEBUG Cleaning up WikipediaIngestTextDoc(page=<WikipediaPage 'Open-source software'>)
Wrote wikipedia-ingest-output/Open-source software-1137981045-txt.json
Wrote wikipedia-ingest-output/Open-source software-1137981045-summary.json
Wrote wikipedia-ingest-output/Open-source software-1137981045-html.json
```
so that[ first doc](https://github.com/Unstructured-IO/unstructured/blob/d51ea3a/unstructured/ingest/main.py#L58) has logs

tomaarsen:
I was testing with `--github-url`, perhaps the difference originates from that? I'll look into it later!

tomaarsen:
I've got a bit of time to look into this now.

@cragwolfe How do you envision the direction of this PR from this point? I assume that the move from `print` to logging is quite desirable, but would you like to help me set up the remaining items for a definition of done for this PR?

cragwolfe:
Hi @tomaarsen , I agree logging over printing is desirable.

Thanks for asking. IMO, the Definition of Done for this PR is:

* The end user experience is preserved when run from the command line in that the same `unstructured-ingest`-related lines are emitted with or without `--verbose`.
* With `--verbose`, some `unstructured` logging output is ok, but not the thousands of lines of output dumped per doc currently at the DEBUG level.

I'd probably approach this by creating an `"unstructured.ingest"` logger and getting that to work.

tomaarsen:
I'm now encountering the multiprocessing issue - It's quite bothersome. The [tutorials/documentation](https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes) on the topic are all much more complex than seems advisable for `unstructured-ingest`. It's not ideal if the logger is more complex than whatever it's logging.

cragwolfe:
> It's not ideal if the logger is more complex than whatever it's logging.

Completely agree!

So, going back to my prior comment:

> To keep it simple for the time being, i think the thing to do for now is add an [initializer](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool) when the [Pool is instantiated](https://github.com/Unstructured-IO/unstructured/blob/707f92f/unstructured/ingest/main.py#L68) to set up logging. This could optionally be done given a keyword arg in main.

I think this would be pretty simple for having logs streamed to stderr. But, I think an `"unstructured.ingest"` logger would need to be added which also seems reasonable to me (assuming no significant changes to the default "unstructured" logger).

@tomaarsen , I should have time tomorrow or by Monday to give that a quick shot if you'd prefer not to / don't have the cycles.

tomaarsen:
I'm afraid I won't have access to my desktop this weekend, starting from Friday after work. With other words, I think the next time I can look at this will be Monday.

tomaarsen:
I've tackled it. Here are some sample outputs, using our favourite example:
```
> unstructured-ingest --wikipedia-page-title "Mango" --reprocess --wikipedia-auto-suggest=False   
2023-03-05 23:33:42,660 SpawnPoolWorker-1 INFO     Processing C:\Users\tom\.cache\unstructured\ingest\6455dbcb95\Mango-1136638036.txt
2023-03-05 23:33:42,870 SpawnPoolWorker-1 INFO     Wrote structured-output\Mango-1136638036-txt.json
2023-03-05 23:33:43,260 SpawnPoolWorker-1 INFO     Processing C:\Users\tom\.cache\unstructured\ingest\6455dbcb95\Mango-1136638036-summary.txt
2023-03-05 23:33:43,268 SpawnPoolWorker-1 INFO     Wrote structured-output\Mango-1136638036-summary.json
2023-03-05 23:33:44,777 SpawnPoolWorker-2 INFO     Processing C:\Users\tom\.cache\unstructured\ingest\6455dbcb95\Mango-1136638036.html
2023-03-05 23:33:46,175 SpawnPoolWorker-2 INFO     Wrote structured-output\Mango-1136638036-html.json
```
And using `--verbose`:
```
> unstructured-ingest --wikipedia-page-title "Mango" --reprocess --wikipedia-auto-suggest=False --verbose
2023-03-05 23:33:59,390 SpawnPoolWorker-2 DEBUG    Fetching WikipediaIngestHTMLDoc(page=<WikipediaPage 'Mango'>) - PID: 43924
2023-03-05 23:33:59,391 SpawnPoolWorker-1 DEBUG    Fetching WikipediaIngestTextDoc(page=<WikipediaPage 'Mango'>) - PID: 24476
2023-03-05 23:33:59,392 SpawnPoolWorker-1 INFO     Processing C:\Users\tom\.cache\unstructured\ingest\6455dbcb95\Mango-1136638036.txt
2023-03-05 23:33:59,611 SpawnPoolWorker-1 INFO     Wrote structured-output\Mango-1136638036-txt.json
2023-03-05 23:33:59,611 SpawnPoolWorker-1 DEBUG    Cleaning up WikipediaIngestTextDoc(page=<WikipediaPage 'Mango'>)
2023-03-05 23:33:59,924 SpawnPoolWorker-1 DEBUG    Fetching WikipediaIngestSummaryDoc(page=<WikipediaPage 'Mango'>) - PID: 24476
2023-03-05 23:34:00,111 SpawnPoolWorker-1 INFO     Processing C:\Users\tom\.cache\unstructured\ingest\6455dbcb95\Mango-1136638036-summary.txt
2023-03-05 23:34:00,116 SpawnPoolWorker-1 INFO     Wrote structured-output\Mango-1136638036-summary.json
2023-03-05 23:34:00,116 SpawnPoolWorker-1 DEBUG    Cleaning up WikipediaIngestSummaryDoc(page=<WikipediaPage 'Mango'>)
2023-03-05 23:34:01,450 SpawnPoolWorker-2 INFO     Processing C:\Users\tom\.cache\unstructured\ingest\6455dbcb95\Mango-1136638036.html
2023-03-05 23:34:02,812 SpawnPoolWorker-2 INFO     Wrote structured-output\Mango-1136638036-html.json
2023-03-05 23:34:02,812 SpawnPoolWorker-2 DEBUG    Cleaning up WikipediaIngestHTMLDoc(page=<WikipediaPage 'Mango'>)
```

I've indeed used the `initializer` keyword in the multiprocessing Pool, which caused the logger to work quite easily from that point onwards. Through this PR, the `verbose` CLI keyword only needs to live in `ingest/main.py`, rather than each of the data connectors being responsible for using this verbose keyword correctly.

I'm open to suggestions regarding the logging format, among others @cragwolfe.

unstructured/ingest/connector/github.py
```diff
@@ -78,31 +78,29 @@ def _create_full_tmp_dir_path(self):
     def cleanup_file(self):
         """Removes the local copy the file (or anything else) after successful processing."""
         if not self.config.preserve_downloads:
-            if self.config.verbose:
-                print(f"cleaning up {self}")
+            logger.debug(f"Cleaning up {self}")
+            breakpoint()
```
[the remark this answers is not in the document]
tomaarsen: My apologies! I should not have let that slip by me.

unstructured/ingest/logger.py
```diff
@@ -0,0 +1,11 @@
...
+
+logger = logging.getLogger("unstructured.ingest")
+handler = logging.StreamHandler()
+formatter = logging.Formatter("%(asctime)s %(processName)-10s %(levelname)-8s %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+
+def set_ingest_logging_level(level: int) -> None:
+    logger.setLevel(level)
```
cragwolfe, line 11:
```suggestion
def ingest_log_streaming_init(level: int) -> None:
    handler = logging.StreamHandler()
    formatter = logging.Formatter("%(asctime)s %(processName)-10s %(levelname)-8s %(message)s")
    handler.setFormatter(formatter)
    logger.addHandler(handler)
    logger.setLevel(level)
```

As is written, any outside application that imports anything under `unstructured.ingest.connector` would have this stream handler and formatter added which may not be desirable (if they wanted different log handling, they'd have to remove the ones created by virtue of this `logger` module being imported).

But, I think everything is fine if the handler / formatter (i.e. everything except the `logger` object)  is just moved into this function. I also suggested a tweak on the function name to reflect the update.

tomaarsen: I agree that the additional flexibility is desirable, albeit niche. I've made the change as suggested.

cragwolfe:
Just need
```
ingest_log_streaming_init(logging.DEBUG if self.verbose else logging.INFO)
```
below 
```
    def run(self):
```
i think, for the main process.

tomaarsen:
Ah, to allow logs before the multiprocessing pool is made. Well spotted. I was already curious why I didn't get a message like
```
All docs have structured outputs, nothing to do. Use --reprocess to process all.
```
When I was testing without `--reprocess`. I thought I just hallucinated the existence of that output haha, but it was actually a bug.

Thank you for the detailed reviews.

cragwolfe (approved):
🎉

PR #322, opened 2023-03-01
feat: add `structlog` to replace `print`-statements and default `logging`

alvarobartt:
## What's in this PR?

Basically the idea is to introduce `structlog` to replace the default `logging` used as well as the `print`-statements used along the codebase, to rely on a more standardized logging interface such as `structlog`. More information on `structlog` at https://www.structlog.org/en/stable/

## What's missing in this PR?

To identify which `contextvars` can be useful, make sure that the `structlog` logging messages are not annoying, make sure that the information is representative and accurate, ideantify more possible logging messages, etc.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ setup(
         "python-magic",
         "markdown",
         "requests",
+        "structlog",
         # NOTE(robinson) - The following dependencies are pinned
         # to address security scans
         "certifi>=2022.12.07",
--- a/unstructured/documents/html.py
+++ b/unstructured/documents/html.py
@@ -8,6 +8,7 @@ if sys.version_info < (3, 8):
 else:
     from typing import Final
 
+import structlog
 from lxml import etree
 
 from unstructured.cleaners.core import clean_bullets, replace_unicode_quotes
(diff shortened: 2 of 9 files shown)


cragwolfe:
@alvarobartt , please see this comment for the distinction between `unstructured` the library and `unstructured-ingest` the application:
https://github.com/Unstructured-IO/unstructured/pull/321#discussion_r1122167243

Also see further discussion in that PR on getting logging in *ingest-the-application* properly configured.

Structlog looks very cool (and rich is also cool! I have used that briefly in the past). However, I don't think it belongs as a dependency at the library level.

Also, to address the `logger.get_logger(__name__)` vs. `logger.get_logger("something-a-human-carefully-chose")`, i'm a bit biased towards the latter in applications. (There may be multiple something-a-human-carefully-chose nested strings within a library, doesn't have to be one)

The reasoning is: 1/ it's generally pretty easy to trace down where a log message came from. 2/ The application can configure log messages to include %(module) . 3/ more compact log messages. 4/ Intentionality matters, you might want to log against different intentions from the same module -- different logger objects, i.e. one for "coolstuff.network" and another for "coolstuff.security".

CC @tomaarsen that also suggested this pattern.

I can be convinced otherwise, but I think there are valid alternatives outside the suggestion in the Advanced Logging Tutorial, as mentioned by @tomaarsen here: https://github.com/Unstructured-IO/unstructured/pull/321#discussion_r1122193848 .

alvarobartt:
Due to the recent merge of #321 from @tomaarsen I'll just close this PR :) Feel free to re-open or get back to `structlog` anytime and just ping me!

PR #335, opened 2023-03-02
fix: ensure all text is maintained in html output

### LINKED ISSUE #332: bug/issue parsing html file
**Describe the bug**
With 0.4.13 I'm able to parse http://paulgraham.com/getideas.html and get "January 2023(Someone fed my ..." but with 0.5.1 I only get the text "January 2023Someone". Note the paren is missing and it may be confusing the system.

**To Reproduce**
import requests

filename = "paul.html"
res = requests.get("http://paulgraham.com/getideas.html")
with open(filename, "wt") as f:
    f.write(res.text)

<outputs> January 2023(Someone

**Expected behavior**
I expect the whole content of the html file.  Version 0.4.13 generates a 789 character string of the whole document.

**Screenshots**
N/A

 - OS: Mac
 - Browser Na
 - Python version 3.11 and 3.10

**Additional context**
None, other than I did not try the other intermediate versions.
[EDIT] Actually I just tried it with 0.5.0 and it worked so the issue may be with 0.5.1

Thanks.
Julio

MthwRobinson:
### Summary

Fixes a bug introduced in #313 and closes #332 . Strip `script` elements from the XML document rather than skipping them in the iteration step. The switch from `tag_elem.itertext()` to `tag_elem.iter()` had caused some text to be missed.

### Testing

The following should show a full paragraph of text output now:

```python

url = "http://paulgraham.com/getideas.html"
```

### DIFF
--- a/example-docs/ideas-page.html
+++ b/example-docs/ideas-page.html
@@ -1,0 +1,44 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
+<html><script type="text/javascript"> 
+ <!-- 
+ (new Image).src="https://store.yahoo.net/cgi-bin/refsd?e=http://paulgraham.com/getideas.html&h=paulgraham.com&v=1.0&dr=" + escape(document.referrer); 
+ --> 
+ </script>
+<head><title>How to Get New Ideas</title><!-- <META NAME="ROBOTS" CONTENT="NOODP"> -->
+<link rel="shortcut icon" href="http://ycombinator.com/arc/arc.png">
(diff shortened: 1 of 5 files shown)


unstructured/documents/html.py
```diff
@@ -97,6 +97,7 @@ def _read(self) -> List[Page]:
             return self._pages
         logger.info("Reading document ...")
         pages: List[Page] = []
+        etree.strip_elements(self.document_tree, ["script"])
```
tomaarsen, line 100: Good call 👍 This is a much conciser solution.

example-docs/ideas-page.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
```
LaverdeS, line 1: Nice addition. Maybe better to have also one of these in the main repo?

unstructured/documents/html.py
LaverdeS: LGTM!

LaverdeS (approved):

PR #344, opened 2023-03-06
chore: Remove file accidentally created/committed

tomaarsen:
Hello!

## Pull Request overview
* Remove the [`doc` file](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/doc) that seems to have been accidentally created.

## Details
#236 introduces this file when introducing `doc.py` as well. It was likely a reasonable mistake.

- Tom Aarsen


MthwRobinson (approved):
LGTM, thanks!

tomaarsen:
The label is a good idea. I'll update the changelog for the time being though, until the label works.

Issue #254, opened 2023-02-23
Create a data connector for processing the biomedical literature


ajjimeno:
Create a data connector that:
* fetches PDF files from [PMC Open Access Subset](https://www.ncbi.nlm.nih.gov/pmc/tools/openftlist).
* stores the files locally (at least temporarily for processing), and runs them through [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py).

See [Adding Data Connectors](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#adding-data-connectors) for details on how to get started. Make sure to include a link to this issue when submitting a PR.

**Definition of Done**

* The [checklist](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist) has been completed.
* The connector is able to process all files from the [PMC Open Access Subset FTP directory](https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_pdf).
* The connector is able to process a document using the [Individual article download](https://www.ncbi.nlm.nih.gov/pmc/tools/ftp/#indart).
* The connector is able to process a document using the [PDF download](https://www.ncbi.nlm.nih.gov/pmc/tools/ftp/#indart:~:text=a%20single%20article-,PDF%20download,-Available%20for%3A).
* The connector can accept any credentials, if necessary.
* The connector should be able of processing the PDF documents through [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py).

cragwolfe:
LGTM!

PR #345, opened 2023-03-07
Connector for Biomedical Literature

HAKSOAT:
Creates a connector for #254 

### Implementation details
The implementation involves the introduction of `SimpleBiomedConfig`, `BiomedIngestDoc` and `BiomedConnector` which ingests documents from the [PDF Download](https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_pdf/).

At the moment, it downloads from the [Individual Article Download](https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_package/)  as tar.gz files but yet to work on the ingestion of such files.


HAKSOAT:
Hi @cragwolfe I have this PR open which works for PDFs and can get a first review.

However, for the tar.gz files, I'm not quite sure what is desired. I have it in mind to extract contents of the tar.gz files and process them, is this the desired behaviour?

I can see there is a [Filetype.ZIP](https://github.com/Unstructured-IO/unstructured/blob/e43e9178ae3f1a5fda2b86ac8e264621b13dc208/unstructured/file_utils/filetype.py#L94) in the code but I'm not quite sure what the objective is for compressed files.

HAKSOAT:
> However, for the tar.gz files, I'm not quite sure what is desired. I have it in mind to extract contents of the tar.gz files and process them, is this the desired behaviour?
> 
> I can see there is a [Filetype.ZIP](https://github.com/Unstructured-IO/unstructured/blob/e43e9178ae3f1a5fda2b86ac8e264621b13dc208/unstructured/file_utils/filetype.py#L94) in the code but I'm not quite sure what the objective is for compressed files.

Hi @ajjimeno I'd appreciate your input on this too

HAKSOAT:
I gave this some thought and I have an idea on how to handle compressed files.

We have a Connector (or something similar for such files), instead of them downloading the files, the decompress.

Hence, at this point of the code:

https://github.com/Unstructured-IO/unstructured/blob/7c619f045b2c7ffdd2f79158bbd986db1c096452/unstructured/ingest/main.py#L72-L89

After the processing of files (where compressed files will fail), we try to extract the files that were not processed (compressed files only). 

In a pattern similar to:

```
        docs = self.COMPRESSED_CONNECTOR.get_decompressed_and_ingest_docs()

        # remove docs that have already been processed
        if not self.reprocess:
            docs = self._filter_docs_with_outputs(docs)
            if not docs:
                return
```

The call pass the decompressed files into the multiprocessing call as usual.

Please let me know if this makes sense and matches the intention for the project.

ajjimeno:
@HAKSOAT while Crag provides an answer. There are different ways in which files can be access from PMC. One is through the compressed files as you have seen already. An alternative way is to identify the PMIC ids of the PDFs and download each individual PDF file. I do not have a preferred way to do it, i.e. downloading the compressed files versus individual PDF documents.

HAKSOAT:
> @HAKSOAT while Crag provides an answer. There are different ways in which files can be access from PMC. One is through the compressed files as you have seen already. An alternative way is to identify the PMIC ids of the PDFs and download each individual PDF file. I do not have a preferred way to do it, i.e. downloading the compressed files versus individual PDF documents.

Thanks for the response @ajjimeno The download of individual PDF documents already works in this PR. I only worked on the compressed files because the issue document #254 asked that "The connector is able to process a document using the [Individual article download](https://www.ncbi.nlm.nih.gov/pmc/tools/ftp/#indart)." where articles refer to `tar.gz` files. If PDF suffices and this isn't necessary, I can take it out.

ajjimeno:
@HAKSOAT I see. In addition to the PDF, there might be additional data
files such as supplementary material linked to the document and available
in the tar.gz file, which is why there is a compressed version of the
article. It would be great to be able to get all the individual
documents linked to an article. I think processing the PDF documents is
already a good way forward. Being able to process the tar.gz files would be
really an interesting bonus. Thanks!

test_unstructured_ingest/test-ingest-biomed.sh
```diff
@@ -0,0 +1,24 @@
...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "$SCRIPT_DIR"/.. || exit 1
+
+if [[ "$(find test_unstructured_ingest/expected-structured-output/biomed-ingest-output/ -type f -size +10k | wc -l)" != 2 ]]; then
+    echo "The test fixtures in test_unstructured_ingest/expected-structured-output/biomed-ingest-output/ look suspicious. At least one of the files is too small."
+    echo "Did you overwrite test fixtures with bad outputs?"
+    exit 1
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py    --biomed-api-from "2019-01-02" --biomed-api-until "2019-01-02+00:03:10" --structured-output-dir biomed-ingest-output  --num-processes 2 --verbose --download-dir biomed-download --preserve-downloads
```
cragwolfe, line 12: formatting nit: please break into multiple lines, i.e.
```
    --biomed-api-from "2019-01-02" \
...
```

test_unstructured_ingest/test-ingest-biomed.sh
```diff
@@ -0,0 +1,24 @@
...
+   echo
+   echo "There are differences from the previously checked-in structured outputs."
+   echo 
+   echo "If these differences are acceptable, copy the outputs from"
+   echo "biomed-ingest-output/ to test_unstructured_ingest/expected-structured-output/biomed-ingest-output/ after running"
+   echo 
+   echo "PYTHONPATH=. ./unstructured/ingest/main.py    --biomed-api-from "2019-01-02" --biomed-api-until "2019-01-02+00:03:10" --structured-output-dir biomed-ingest-output  --num-processes 2 --verbose --download-dir biomed-download --preserve-downloads"
+   echo
+   exit 1
+fi
```
cragwolfe, line 24: Please add this script to:
https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest.sh

examples/ingest/biomed/ingest.sh
```diff
@@ -0,0 +1,46 @@
...
+# Through the ftp directories:
+# https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_pdf
+# https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_package
+
+# By providing the path, the documents existing therein are downloaded.
+# For example, to download the documents in the path: https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_pdf/07/
+# The path needed is oa_pdf/07/
+
+# To download the documents in the path: https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_package/0a/01/
+# The path needed is oa_package/0a/01/
```
cragwolfe, line 21: The First Approach is with `--biomed-path`, right? 

Maybe add a script call `ingest-with-path.sh` that just has an example ingest with that approach?

unstructured/ingest/connector/biomed.py
```diff
@@ -0,0 +1,345 @@
...
+                elif "command successful" in response:
+                    self.is_dir = True
+                else:
+                    raise ValueError("Something went wrong when validating the path: {path}.")
+
+
+@dataclass
+class BiomedIngestDoc(BaseIngestDoc):
+    config: SimpleBiomedConfig
+    file_meta: Dict
```
cragwolfe, line 136: might as well make this a `dataclass`?

test_unstructured_ingest/test-ingest-biomed.sh
```diff
@@ -0,0 +1,24 @@
...
+PYTHONPATH=. ./unstructured/ingest/main.py    --biomed-api-from "2019-01-02" --biomed-api-until "2019-01-02+00:03:10" --structured-output-dir biomed-ingest-output  --num-processes 2 --verbose --download-dir biomed-download --preserve-downloads
+
+if ! diff -ru biomed-ingest-output test_unstructured_ingest/expected-structured-output/biomed-ingest-output ; then
+   echo
+   echo "There are differences from the previously checked-in structured outputs."
+   echo 
+   echo "If these differences are acceptable, copy the outputs from"
+   echo "biomed-ingest-output/ to test_unstructured_ingest/expected-structured-output/biomed-ingest-output/ after running"
+   echo 
+   echo "PYTHONPATH=. ./unstructured/ingest/main.py    --biomed-api-from "2019-01-02" --biomed-api-until "2019-01-02+00:03:10" --structured-output-dir biomed-ingest-output  --num-processes 2 --verbose --download-dir biomed-download --preserve-downloads"
```
tomaarsen, line 21:
```suggestion
   echo "PYTHONPATH=. ./unstructured/ingest/main.py --biomed-api-from '2019-01-02' --biomed-api-until '2019-01-02+00:03:10' --structured-output-dir biomed-ingest-output  --num-processes 2 --verbose --download-dir biomed-download --preserve-downloads"
```
The double quotes do not work here due to those larger quotes also being used for the larger string being echoed.

test_unstructured_ingest/test-ingest-biomed.sh
cragwolfe, line 24: Please `chmod ugo+x` this script

unstructured/ingest/main.py
```diff
@@ -384,6 +415,21 @@ def main(
                 re_download=re_download,
             ),
         )
+    elif biomed_path or biomed_api_id or biomed_api_from or biomed_api_until or biomed_api_format:
```
cragwolfe, line 418: up in the block around line 309 under

        if not download_dir:
            cache_path = Path.home() / ".cache" / "unstructured" / "ingest"
           ...

A case needs to be added for biomed so that downloads may properly be stored under ~/.cache (which saves files from being re-download with the user having to think about that too much).

While your at it, you could change this very confusing [ValueError](https://github.com/Unstructured-IO/unstructured/blob/c51adb2/unstructured/ingest/main.py#L299) to something like:

`This connector does not support saving downloads to ~/.cache/  , --download-dir must be provided`

HAKSOAT:
Thanks for the review. I will attend to them shortly.

unstructured/ingest/connector/biomed.py
```diff
@@ -0,0 +1,345 @@
...
+
+            if dir_:
+                dir_.mkdir(parents=True, exist_ok=True)
+
+        urllib.request.urlretrieve(
+            self.file_meta.get("ftp_path"),  # type: ignore
+            self.file_meta.get("download_filepath"),
+        )
+
+        logger.debug(f"File downloaded: {self.file_meta.get('download_filepath')}.")
```
cragwolfe, line 170:
```suggestion
        logger.debug(f"File downloaded: {self.file_meta.get('download_filepath')}")
```

cragwolfe:
On the .tar.gz vs. PDF question, IMO the PDF solution in this PR is very much sufficient!

test_unstructured_ingest/test-ingest-biomed.sh
HAKSOAT: Thanks for spotting this. I simply copied the command in.

test_unstructured_ingest/test-ingest-biomed.sh
HAKSOAT: Hi @cragwolfe I'm not quite sure what you mean here. 

If I'm to chmod, my current understanding is that it has to be on the `test-ingest-biomed.sh` file just before it is called. At what point do I get to do that in our workflow here? Do I chmod in the `test-ingest.sh` file? I looked for other examples but can't seem to find other cases where such is being done in the codebase.

tomaarsen: Perhaps he means to run
```
git update-index --chmod=+x test_unstructured_ingest/test-ingest-biomed.sh
```
Which updates the files permissions to 755 on git, making it executable for everyone.

HAKSOAT:
> On the .tar.gz vs. PDF question, IMO the PDF solution in this PR is very much sufficient!

Noted. I'll stick to that.

test_unstructured_ingest/test-ingest-biomed.sh
HAKSOAT: Ah makes sense. I didn't even know that was possible. Thank you.

cragwolfe (approved):

PR #348, opened 2023-03-07
Added string formatter

mu4farooqi:
Absence of string formatter will make it print "{segment}" instead of actual segment.


cragwolfe (approved):
Thanks!

PR #351, opened 2023-03-09
feat: `UNSTRUCTURED_LANGUAGE_CHECK` env var to control 

MthwRobinson:
### Summary

Adds an `UNSTRUCTURED_LANGUAGE_CHECK` environment variable that controls whether language specific checks such as vocab and part of speech checks are applied. Users can set this to `"true"` for more accurate partitioning or `"false"` for faster processing.

### Testing

The following shows how to apply this language checks with both the kwarg and the environment variable.

```python

is_possible_title("BLI BLA BLARG") # should be True
is_possible_title("BLI BLA BLARG", language_checks=True) # should be False

os.environ["UNSTRUCTURED_LANGUAGE_CHECKS"] = "true"
is_possible_title("BLI BLA BLARG") # should be False
```

We should also observe a speed up when `partition_html` is applied on the feature branch relative to `main`. My tests locally show a ~3x speedup.

On the feature branch
```python
In [5]: %timeit partition_html(filename="example-docs/example-10k.html")
1.62 s ± 17.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each
```

On `main`:
```python
In [6]: %timeit partition_html(filename="example-docs/example-10k.html")
3.49 s ± 13.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/test_unstructured/partition/test_text_type.py
+++ b/test_unstructured/partition/test_text_type.py
@@ -50,11 +50,13 @@ def test_is_possible_narrative_text(text, expected, monkeypatch):
     monkeypatch.setattr(text_type, "word_tokenize", mock_word_tokenize)
     monkeypatch.setattr(text_type, "pos_tag", mock_pos_tag)
     monkeypatch.setattr(text_type, "sent_tokenize", mock_sent_tokenize)
+    monkeypatch.setenv("UNSTRUCTURED_LANGUAGE_CHECKS", "true")
     is_possible_narrative = text_type.is_possible_narrative_text(text, cap_threshold=0.3)
     assert is_possible_narrative is expected
 
 
-def test_text_type_handles_non_english_examples():
+def test_text_type_handles_non_english_examples(monkeypatch):
+    monkeypatch.setenv("UNSTRUCTURED_LANGUAGE_CHECKS", "true")
     narrative_text = "Я говорю по-русски. Вы тоже?"
     title = "Риски"
 
@@ -102,6 +104,7 @@ def test_text_type_handles_non_english_examples_with_env_var(monkeypatch):
 def test_is_possible_title(text, expected, monkeypatch):
     monkeypatch.setattr(text_type, "sent_tokenize", mock_sent_tokenize)
     monkeypatch.setattr(text_type, "word_tokenize", mock_word_tokenize)
+    monkeypatch.setenv("UNSTRUCTURED_LANGUAGE_CHECKS", "true")
     assert text_type.is_possible_title(text) is expected
 
 
--- a/test_unstructured/partition/test_text_type.py
+++ b/test_unstructured/partition/test_text_type.py
@@ -55,6 +55,13 @@ def test_is_possible_narrative_text(text, expected, monkeypatch):
     assert is_possible_narrative is expected
 
 
+def test_narrative_text_language_checks():
+    # NOTE(robinson) - This is true because we don't check english vocab if language checks
+    # are set to False
+    text = "Dal;kdjfal adawels adfjwalsdf. Addad jaja fjawlek"
+    text_type.is_possible_narrative_text(text, language_checks=False) is True
+
+
 def test_text_type_handles_non_english_examples(monkeypatch):
     monkeypatch.setenv("UNSTRUCTURED_LANGUAGE_CHECKS", "true")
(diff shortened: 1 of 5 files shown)


test_unstructured/partition/test_text_type.py
```diff
@@ -50,11 +50,20 @@ def test_is_possible_narrative_text(text, expected, monkeypatch):
...
     is_possible_narrative = text_type.is_possible_narrative_text(text, cap_threshold=0.3)
     assert is_possible_narrative is expected
 
 
-def test_text_type_handles_non_english_examples():
+def test_narrative_text_language_checks():
+    # NOTE(robinson) - This is true because we don't check english vocab if language checks
+    # are set to False
+    text = "Dal;kdjfal adawels adfjwalsdf. Addad jaja fjawlek"
+    text_type.is_possible_narrative_text(text, language_checks=False) is True
```
qued, line 62:
```suggestion
    assert text_type.is_possible_narrative_text(text, language_checks=False) is True
```

test_unstructured/partition/test_text_type.py
```diff
@@ -102,9 +111,17 @@ def test_text_type_handles_non_english_examples_with_env_var(monkeypatch):
...
     monkeypatch.setattr(text_type, "word_tokenize", mock_word_tokenize)
+    monkeypatch.setenv("UNSTRUCTURED_LANGUAGE_CHECKS", "true")
     assert text_type.is_possible_title(text) is expected
 
 
+def test_title_language_checks():
+    # NOTE(robinson) - This is true because we don't check english vocab if language checks
+    # are set to False
+    text = "BTAR ADFJA L"
+    text_type.is_possible_narrative_text(text, language_checks=False) is True
```
qued, line 122:
```suggestion
    assert text_type.is_possible_narrative_text(text, language_checks=False) is True
```

unstructured/partition/text_type.py
```diff
@@ -45,7 +46,13 @@ def is_possible_narrative_text(
         narrative text
     language
         The two letter language code for the text. defaults to "en" for English
+    language_checks
+        If True, conducts checks that are language to the specified language.
```
qued, line 50:
```suggestion
        If True, conducts checks that are specific to the chosen language.
```
Also maybe add a note that selecting `True` could negatively impact performance?

unstructured/partition/text_type.py
```diff
@@ -101,7 +109,13 @@ def is_possible_title(
         The minimum number of alpha characters the text needs to be considered a title
     language
         The two letter language code for the text. defaults to "en" for English
+    language_checks
+        If True, conducts checks that are language to the specified language.
```
qued, line 113:
```suggestion
        If True, conducts checks that are specific to the chosen language.
```
Same as above, maybe indicate that this could be a performance hit?

qued (approved):
LGTM other than a few minor issues! Nice speed up!

cragwolfe:
Note that with https://github.com/Unstructured-IO/unstructured/pull/380 , this workaround to speed up results is no longer necessary. For this particular example, performance is actually better with ` os.environ["UNSTRUCTURED_LANGUAGE_CHECKS"] = "true"`

```
In [18]: os.environ["UNSTRUCTURED_LANGUAGE_CHECKS"] = "false"
os.environ["UNSTRUCTURED_LANGUAGE_CHECKS"] = "false""

In [19]: is_possible_title("BLI BLA BLARG") 
is_possible_title("BLI BLA BLARG")
Out[19]: True

In [20]: timeit partition_html(filename="example-docs/example-10k.html")
timeit partition_html(filename="example-docs/example-10k.html"))
646 ms ± 2.94 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)



In [21]: os.environ["UNSTRUCTURED_LANGUAGE_CHECKS"] = "true"
os.environ["UNSTRUCTURED_LANGUAGE_CHECKS"] = "true""

In [22]: is_possible_title("BLI BLA BLARG") 
is_possible_title("BLI BLA BLARG")
Out[22]: False

In [23]: timeit partition_html(filename="example-docs/example-10k.html")
timeit partition_html(filename="example-docs/example-10k.html"))
588 ms ± 2.42 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #353, opened 2023-03-10
Add `AzureBlobStorageConnector`

### LINKED ISSUE #257: Create a data conector for Azure Blob Storage

alvarobartt:
## What's in this PR?

Upon #318 merge into `main`, this PR introduces the `AzureBlobStorageConnector` explicit `fsspec` connector so as to close #257. This PR also introduces a feature in the `ingest-cli` so as to use the generic `FsspecConnector` in case the protocol is not among the specifically implemented ones. Finally, a deprecation lifecycle starts for the `--s3-url` option in favor of `--remote-url`.

## What's missing in this PR?

Adding the `test_unstructured_ingest/test-ingest-azure.sh` script for testing, but it's still pending since the Azure Blob Storage container provided by @benjats07 in #257 contains some files that `unstructured` cannot parse/partition, in this case both XLSX and XML files.


alvarobartt:
Hi @cragwolfe, this should also be done already 👍🏻 Feel free to install it as `pip install -e ".[azure,local-inference]" and run it as `./examples/azure/ingest.sh` or using the `unstructured-ingest` CLI as `unstructured-ingest --remote-url abfs://demo-files/ --azure-account-name datalakebenja22 --structured-output-dir azure-demo-files-output --num-processes 2`

unstructured/ingest/main.py
```diff
@@ -95,18 +102,40 @@ def run(self):
 
 
 @click.command()
+@click.option(
+    "--remote-url",
+    default=None,
+    help="Remote fsspec URL formatted as `protocol://dir/path`, it can contain both "
+    "a directory or a single file.",
```
cragwolfe, line 109: Might as well explicitly mention the supported protocols here, to save the user any confusion.

unstructured/ingest/main.py
```diff
@@ -303,17 +343,52 @@ def main(
...
+                config=SimpleAzureBlobStorageConfig(
+                    path=remote_url,
+                    access_kwargs={
+                        "account_name": azure_account_name,
+                        "account_key": azure_account_key,
+                    }
+                    if azure_account_name
+                    else {"connection_string": azure_connection_string}
+                    if azure_connection_string
+                    else {},
```
cragwolfe, line 370: readability nit: top 8 lines are a bit awkward. break into an old fashioned if/else on line 360?

```
    if azure_connection_string:
       access_kwargs = {"connection_string": azure_connection_string}
    elif azure_account_name:
       ...
```

cragwolfe:
> Adding the test_unstructured_ingest/test-ingest-azure.sh script for testing, but it's still pending since the Azure Blob Storage container provided by @benjats07 in https://github.com/Unstructured-IO/unstructured/issues/257 contains some files that unstructured cannot parse/partition, in this case both XLSX and XML files.

You can now use:
```
        --remote-url abfs://container1/ \
        --azure-account-name azureunstructured1 \
```

cragwolfe:
Besides the minor comments and once the tests are in CI, should be good to go!

alvarobartt:
> > Adding the test_unstructured_ingest/test-ingest-azure.sh script for testing, but it's still pending since the Azure Blob Storage container provided by @benjats07 in #257 contains some files that unstructured cannot parse/partition, in this case both XLSX and XML files.
> 
> You can now use:
> 
> ```
>         --remote-url abfs://container1/ \
>         --azure-account-name azureunstructured1 \
> ```

Sure I'll update it now and retry to pass the CI :+1:

unstructured/ingest/main.py
```diff
@@ -303,17 +343,52 @@ def main(
...
+                    path=s3_url,
+                    access_kwargs={"anon": s3_anonymous},
+                    download_dir=download_dir,
+                    output_dir=structured_output_dir,
+                    re_download=re_download,
+                    preserve_downloads=preserve_downloads,
+                ),
+            )
+        elif protocol in ("abfs", "az"):
+            doc_connector = AzureBlobStorageConnector(  # type: ignore
```
tomaarsen, line 370: This `AzureBlobStorageConnector` will, as it is simply an empty subclass of `FsspecConnector`, make instances of `FsspecIngestDoc`, rather than of `AzureBlobStorageIngestDoc`. As a result, the latter class is never used, and its `@requires_dependencies(["adlfs", "fsspec"], extras="azure")` is never called.
This last bit is not really a problem, as there is also a `requires_dependencies` decorator around `AzureBlobStorageConnector`.

alvarobartt: Hmm true, maybe we can parametrize that to provide the `...IngestDoc` class to the `FsspecConnector` so that it's used instead of the default `FsspecIngestDoc`, or maybe we can just skip it and remove both `S3IngestDoc` and `AzureBlobStorageIngestDoc`?

unstructured/ingest/main.py
```diff
@@ -303,17 +343,56 @@ def main(
...
+            access_kwargs = (
+                {
+                    "account_name": azure_account_name,
+                    "account_key": azure_account_key,
+                }
+                if azure_account_name
+                else {"connection_string": azure_connection_string}
+                if azure_connection_string
+                else {}
+            )
```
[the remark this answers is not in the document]
alvarobartt: You're right! Ok I'll commit this manually and add you as co-author 👍🏻

unstructured/ingest/main.py
tomaarsen: I think both solutions are reasonable. The first one would be as simple as:
```python
# azure.py
@requires_dependencies(["adlfs", "fsspec"], extras="azure")
class AzureBlobStorageConnector(FsspecConnector):
    ingest_doc_class = AzureBlobStorageIngestDoc
```

```python
# ffspec.py
class FsspecConnector(BaseConnector):
    """Objects of this class support fetching document(s) from"""

    ingest_doc_class = FsspecIngestDoc

...

    def get_ingest_docs(self):
        return [
            self.ingest_doc_class(
                self.config,
                file,
            )
            for file in self._list_files()
        ]
```

```python
# s3.py
@requires_dependencies(["s3fs", "fsspec"], extras="s3")
class S3Connector(FsspecConnector):
    ingest_doc_class = S3IngestDoc
```

And the second one is as simple as just removing the ingest classes from existence. Perhaps the first approach allows for more flexibility in case there ever comes an `ffspec`-supported ingest connector that wants to do some additional processing? If that never happens though, then the second approach is simpler. 
I think it's best to defer to the maintainers here.

tomaarsen:
The logs are showing:
```
./test_unstructured_ingest/test-ingest.sh: line 9: ./test_unstructured_ingest/test-ingest-azure.sh: Permission denied
```

You likely have to set the permissions of this file such that they can be executed:
```
git update-index --chmod=+x ./test_unstructured_ingest/test-ingest-azure.sh
```
I ran into this issue myself with several of my ingest data connectors.

unstructured/ingest/main.py
alvarobartt: So what I did is basically the same as you mentioned above but using `ingest_doc_cls` as an attribute 😆 But maybe makes more sense to just remove it and rely on `FsspecIngestDoc`?

alvarobartt:
> The logs are showing:
> 
> ```
> ./test_unstructured_ingest/test-ingest.sh: line 9: ./test_unstructured_ingest/test-ingest-azure.sh: Permission denied
> ```
> 
> You likely have to set the permissions of this file such that they can be executed:
> 
> ```
> git update-index --chmod=+x ./test_unstructured_ingest/test-ingest-azure.sh
> ```
> 
> I ran into this issue myself with several of my ingest data connectors.

Hmm looks like this didn't work @tomaarsen I'll look for a solution! 👍🏻

alvarobartt:
So it worked already... but it seems that we're getting limitations from GitHub, which is not related to this PR 😖 Can we just re-run those failing tests? @cragwolfe

tomaarsen:
I suspect that all tests would pass if it wasn't for the rate limit on the GitHub tests, considering that the tests pass for Python 3.10 and the recent changes since the last passing tests should not have introduced any version-specific code. However, safest is just to wait a bit and rerun the tests.

cragwolfe (approved):
Nice! 🚀

PR #357, opened 2023-03-10
feat: add `"fast"` strategy for PDF parsing; fallback to `"fast"` if `detectron2` is not available

### LINKED ISSUE #311: PDF parsing fallback if `tesseract` and/or `detectron2` is not available
`partition_pdf` currently uses `detectron2` to detect document elements in a PDF and OCRs the document using `tesseract` when the text is not present. The goal of this issue is to have a fallback that extracts the raw text if `detectron2` and/or `tesseract` is not available and partitions using `partition_text`. The [`pdf`](https://github.com/hwchase17/langchain/blob/master/langchain/document_loaders/pdf.py) and [`paged_pdf`](https://github.com/hwchase17/langchain/blob/master/langchain/document_loaders/paged_pdf.py) loaders in `langchain` provide some options for this that use `pdfminer` and `pypdf`. This issue will require modifications to [`unstructured-inference`](https://github.com/Unstructured-IO/unstructured-inference)

MthwRobinson:
### Summary

Closes #311. Adds a `"fast"` strategy for partitioning PDFs that uses `pdfminer`. The default strategy is `"hi_res"` and is the original partitioning logic that uses `detectron2`. If `detectron2` is not available and the `"hi_res"` strategy is selected, `partition_pdf` fallsback to using the `"fast"` strategy. The implementation uses `pdfminer` because that's already installed as a dependency with the `local-inference` extra. There are other options for accomplishing this as well, but they would entail adding a new dependency. The `"fast"` strategy substantially speeds up processing.

### Testing

The following code should now run:

```python

elements = partition(filename=filename, strategy="fast")

    elements = partition(file=f, strategy="fast")
```

Additionally, we should see a speedup when we use the `"fast"` strategy.

```python
In [7]: %timeit partition(filename=filename, strategy="fast")
244 ms ± 646 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit partition(filename=filename, strategy="hi_res")
2.43 s ± 24.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


cragwolfe (approved):
Must faster with "fast" 🚀 

Did a quick test to see what the data looked like using:
https://gist.github.com/cragwolfe/104e98ad496d3c93235532db651f1a13

Text element length distrubution:
```
Counts for text length:
Length 0: 6
Length 1: 98
Length 2: 14
Length 2 to 10: 32
Length 11 to 99: 1260
Length 100 to 999: 12
Length 1000 or more: 0
````

PR #358, opened 2023-03-10
fix: parsing for files with `message/rfc822` MIME type; dir for unsupported files

### LINKED ISSUE #338: Unable to process two docs under example-docs/
**Describe the bug**

Attempting to parse `example-docs/factbook.xml` or `example-docs/fake-email-header.eml` results in an error, "File type not support in partition." All other docs in `example-docs/` are able to be partitioned without an error.

**To Reproduce**

```
>>> partition("example-docs/factbook.xml")
  File "<stdin>", line 1, in <module>
  File "/Users/cragwolfe/r/unstructured/unstructured/partition/auto.py", line 72, in partition
    raise ValueError(f"{msg}. File type not support in partition.")
ValueError: Invalid file example-docs/factbook.xml. File type not support in partition.
```

Discovered by running the script: https://github.com/Unstructured-IO/unstructured/blob/main/examples/ingest/github/ingest.sh

**Expected behavior**
The call to `.partition()` succeeds

 - OS: mac m1
 - Python version 3.8.15
 - on `main` `HEAD` which is currently 
 ```
 f5af87a Tom Aarsen      2023-03-02      feat: Expose Wikipedia `auto_suggest` argument to the ingest CLI (#336)
 ```

MthwRobinson:
### Summary

Closes #338. Adds the ability to process files with  a `message/rfc822` MIME type, which previously caused failures for `example-docs/fake-email-header.eml`. More information on RFC 822 messages is available [here](https://www.rfc-editor.org/rfc/rfc822). Also creates a separate directory under `example-docs` for filetypes that are not yet supported.

### Testing

```python

filename = "example-docs/fake-email-header.eml"
```


tomaarsen:
Should solve the problem nicely 👍

cragwolfe (approved):
LGTM! (also tested locally)

PR #362, opened 2023-03-13
fix: open xml files in read only mode

ryannikolaidis:
Currently when we create an XMLDocument from file we read the document in [r+ mode](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files). However, it doesn't look like we actually ever need to write to the file, in which case we should just use the read mode (default). This will also mean that a user doesn't need write-permission in order to parse the document.


cragwolfe (approved):
LGTM!

PR #363, opened 2023-03-13
chore: add libreoffice to ubuntu install script

MthwRobinson:
### Summary

Adds `libreoffice` to the dependencies that get installed in the `setup_ubuntu.sh` script. We previously omitted this dependency in the setup scripts we were unsure how many users would have the older `.doc`/`.ppt` files as a opposed to the newer `.docx`/`.pptx`.

### DIFF
--- a/scripts/setup_ubuntu.sh
+++ b/scripts/setup_ubuntu.sh
@@ -25,7 +25,7 @@ fi
 USER_ACCOUNT=$1
 
 # Update existing packages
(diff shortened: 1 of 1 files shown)


qued (approved):
LGTM. I confirmed this package gets the respective tests to pass on Ubuntu. Aesthetically I'd like the installation in the script to line up with where it is in the AL2 script, but I can push that as a formatting PR later.

PR #366, opened 2023-03-14
feat: add staging brick to clean non-ascii characters from unicode

natygyoon:
Motivation: Given a unicode string, one heavy-handed but often acceptable approach to extract the “clean text” from it is to simply remove non-ascii characters.


docs/source/bricks.rst
```diff
@@ -578,6 +578,23 @@ Examples:
   clean_postfix(text, r"(END|STOP)", ignore_case=True)
 
 
+``clean_non_ascii_chars``
+-----------------
```
cragwolfe, line 582: Please add `-` so length matches above

cragwolfe (approved):
LGTM!

PR #368, opened 2023-03-15
feat: add --metadata-include and --metadata-exclude parameters to unstructured-ingest

natygyoon:
Currently, the default behavior of [unstructured-ingest](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/main.py) is to [drop the metadata field](https://github.com/Unstructured-IO/unstructured/blob/70420b5/unstructured/ingest/interfaces.py#L97) “filename” from the json output.

However, it would be better if the user could specify what metadata fields to drop (if they exist) or what fields to include (drop all other fields).

### DIFF
--- a/unstructured/ingest/interfaces.py
+++ b/unstructured/ingest/interfaces.py
@@ -1,6 +1,7 @@
 """Defines Abstract Base Classes (ABC's) core to batch processing documents
 through Unstructured."""
 
+import json
 from abc import ABC, abstractmethod
 
 from unstructured.ingest.logger import logger
@@ -98,16 +99,18 @@ class BaseIngestDoc(ABC):
         self.isd_elems_no_filename = []
         for elem in isd_elems:
             # type: ignore
+            metadata_dict = json.loads(elem["metadata"])
             if self.config.metadata_exclude:
                 ex_list = self.config.metadata_exclude.split(",")
                 for ex in ex_list:
-                    elem["metadata"].pop(ex, None)
+                    metadata_dict.pop(ex, None)
             elif self.config.metadata_include:
                 in_list = self.config.metadata_include.split(",")
                 for k in elem["metadata"]:
                     if k not in in_list:
-                        elem["metadata"].pop(k, None)
-
+                        metadata_dict.pop(k, None)
+            elem["metadata"] = str(metadata_dict)
+            
             elem.pop("coordinates")  # type: ignore[attr-defined]
             self.isd_elems_no_filename.append(elem)
 
--- a/unstructured/ingest/interfaces.py
+++ b/unstructured/ingest/interfaces.py
@@ -59,7 +59,9 @@ class BaseIngestDoc(ABC):
 
     Crucially, it is not responsible for the actual processing of the raw document.
     """
+
     config: BaseConnectorConfig
+
     @property
     @abstractmethod
     def filename(self):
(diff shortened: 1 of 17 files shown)


unstructured/ingest/main.py
```diff
@@ -103,6 +133,20 @@ def run(self):
 
 
 @click.command()
+@click.option(
+    "--metadata-include",
+    cls=MutuallyExclusiveOption,
+    default="",
+    help="If set, include the specified metadata fields if they exist and drop all other fields.",
```
ryannikolaidis, line 140: How does one assign this list when calling? Is it a single string with comma separated values? or do you make multiple `--metadata-include X --metadata-include Y` args? Maybe helpful to have an example in the help. (Same for exclude)

unstructured/ingest/interfaces.py
```diff
@@ -47,6 +47,8 @@ class BaseConnectorConfig(ABC):
     # where to write structured data outputs
     output_dir: str
     re_download: bool = False
+    metadata_include: str = ""
+    metadata_exclude: str = ""
```
ryannikolaidis, line 51: This will mean that the mutual exclusivity is only actually defined at the click() interface level. By testing `if metadata_exclude` first in the BaseIngestDoc, I think your logic still works, but I wonder if it'd be helpful to be more explicit here? Make these say Optional(str) with default None to distinguish between the user specifying an empty string versus not setting the value. You could then also validate exclusivity there by actually testing `is None`. Thoughts?

ryannikolaidis:
This is looking really good! Just dropped a few questions.

unstructured/ingest/interfaces.py
natygyoon: Yeah, that's a really good idea to make them optional string and more explicitly signal the user that at most one of them should be set. Just pushed a fix.

ryannikolaidis: awesome! taking a look now.

unstructured/ingest/main.py
```diff
@@ -35,6 +36,35 @@
...
+
+    def handle_parse_result(self, ctx, opts, args):
+        if self.mutually_exclusive.intersection(opts) and self.name in opts:
+            raise UsageError(
+                "Illegal usage: `{}` is mutually exclusive with "
+                "arguments `{}`.".format(
+                    self.name,
+                    ", ".join(self.mutually_exclusive),
+                ),
+            )
```
ryannikolaidis, line 59: I just tried:
```
python3 -m unstructured.ingest.main --gitlab-url https://gitlab.com/gitlab-com/content-sites/docsy-gitlab --git-file-glob '*.md,*.txt' --reprocess --metadata-exclude page_number --metadata-include page_number
```

This succeeded. I thought I should have seen this error message, right?

unstructured/ingest/interfaces.py
```diff
@@ -94,7 +99,16 @@ def process_file(self):
         self.isd_elems_no_filename = []
         for elem in isd_elems:
             # type: ignore
-            elem["metadata"].pop("filename", None)  # type: ignore[attr-defined]
+            if self.config.metadata_exclude is not None:
+                ex_list = self.config.metadata_exclude.split(",")
+                for ex in ex_list:
+                    elem["metadata"].pop(ex, None)  # type: ignore[attr-defined]
+            elif self.config.metadata_include is not None:
```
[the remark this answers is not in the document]
natygyoon: I agree that adding additional check here ensures the mutual exclusiveness even though we are checking it on a higher level. Just pushed the fix.

ryannikolaidis (approved):
Looks good, nice work!

PR #373, opened 2023-03-16
chore: add tests for docker

amanda103:


Dockerfile
```diff
@@ -52,6 +59,7 @@ COPY requirements/local-inference.txt requirements-local-inference.txt
 
 RUN python3.8 -m pip install pip==${PIP_VERSION} \
   && pip install --no-cache -r requirements-base.txt \
+  && pip install --no-cache -r requirements-test.txt \
```
ryannikolaidis, line 62: did adding this add much time to install and size to the image? probably easiest for now, but wondering as we try to slim this down what kind of impact this has.

Dockerfile
```diff
@@ -60,5 +68,6 @@ RUN python3.8 -m pip install pip==${PIP_VERSION} \
   && pip install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@v0.6#egg=detectron2"
 
 COPY unstructured unstructured
+COPY test_unstructured test_unstructured
```
ryannikolaidis, line 71: we probably don't want to include these in the docker image right? maybe just a copy step before running the tests on the container in CI?

Dockerfile
```diff
@@ -8,6 +8,12 @@ ARG UNSTRUCTURED
 RUN yum -y update && \
   yum -y install poppler-utils xz-devel which
 
+# Enable the EPEL repository
+RUN yum install -y epel-release && yum clean all
+
```
ryannikolaidis, line 13: guessing centos dependency fun here? curios: was this needed for issues fetching pandoc? or something else?

Dockerfile
awalker4: I'd suggest mounting the test dir in during `make docker-test`. To ryan's point on the pip install, we could even get fancy and mount the local .venv to bring the test deps in at runtime. But yeah, certainly not a priority.

awalker4:
LGTM! (I think the workflow file indentation is off. YAML is fun :) )

ryannikolaidis:
> (I think the workflow file indentation is off. YAML is fun 😃 )

Good eye, hmmm, seems like we should be linting these... 🤨. Writing a ticket.

Dockerfile
amanda103: Yes this is a pandoc necessity, but the tests using pandoc are still complaining. Going to keep investigating!

Dockerfile
amanda103: Not a ton, but as Austin mentioned below maybe there's a better way to bring in test dependencies! Going to play around with that some more.

Dockerfile
ryannikolaidis: Argh! 😞 I definitely feel your pain there!

awalker4 (commented):
LGTM! Just one question on the build step.

scripts/test.sh
```diff
@@ -0,0 +1,5 @@
+##!/bin/bash
+
+docker run --platform linux/amd64 -ti --rm \
+  -v ~work/unstructured/unstructured/test_unstructured/:/home/test_unstructured unstructured-dev:latest \
```
awalker4, line 4: In order to be generic I think we'll want some form of $(PWD) here.

awalker4 (approved):

.github/workflows/ci.yml
```diff
@@ -128,3 +128,40 @@ jobs:
...
+        uses: mxschmitt/action-tmate@v3
+        with:
+          connection-method: 'https'
+      - name: Test Dockerfile
+        run: |
+          source .venv/bin/activate
+          make install
+          make install-detectron2
+          sudo apt-get update
+          sudo apt-get install -y libmagic-dev poppler-utils tesseract-ocr libreoffice pandoc
```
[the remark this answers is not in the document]
amanda103: definitely not, good call!

PR #376, opened 2023-03-17
feat: add --fields-include to unstructured-ingest

natygyoon:
Currently, unstructured-ingest writes all top-level fields in an element with the exception of “coordinates.” This is a reasonable default. Update the command to accept a “--fields-include” in [main.py](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/main.py) so that the default is “element_id,text,type,metadata”.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Enhancements
 
+* `contains_english_word()`, used heavily in text processing, is 10x faster.
+
 ### Features
 
 * Add `--fields-include` parameter to `unstructured-ingest`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Enhancements
 
+* Refactored codebase using `exactly_one`
+
 ### Features
 
 * Add `--fields-include` parameter to `unstructured-ingest`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@
 
 ### Features
 
-* Add `--fields-include` parameter to `unstructured-ingest`
 * Add `--metadata-include` and `--metadata-exclude` parameters to `unstructured-ingest`
 * Add `clean_non_ascii_chars` to remove non-ascii characters from unicode string
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.5-dev1
+## 0.5.5-dev2
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.5.5-dev0
+
+### Enhancements
+
+### Features
+
+* Add `clean_non_ascii_chars` to remove non-ascii characters from unicode string
+
+### Fixes
+
 ## 0.5.4
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.7-dev0
+## 0.5.7-dev1
 
 ### Enhancements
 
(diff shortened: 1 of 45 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
-## 0.5.5-dev0
+## 0.5.5-dev2
```
ryannikolaidis, line 1: why are we going from dev0->dev2?

natygyoon: Fixed, thanks for catching it out!

unstructured/ingest/interfaces.py
```diff
@@ -94,8 +100,24 @@ def process_file(self):
...
+                    if k not in in_list:
+                        elem["metadata"].pop(k, None)  # type: ignore[attr-defined]
+
+            in_list = self.config.fields_include.split(",")
+            ex_list = []
+            for k in elem:
+                if k not in in_list:
+                    ex_list.append(k)
+            for k in ex_list:
+                elem.pop(k)  # type: ignore[attr-defined]
```
ryannikolaidis, line 119: Instead of iterating the lists twice maybe just a quick comprehension here? Something like: 
```suggestion
            {k: v for k, v in elem.items() if k in in_list}
```

ryannikolaidis:
it's a little hard to review with the changes from the --metadata_include/exclude branch mixed in...but it doesn't look like there are any tests added to cover the new `fields_include` functionality, right? worth some coverage?

ryannikolaidis (changes requested):
Dropped a note on iteration and should probably add tests. Also, there are changes from the metadata_include/exclude PR mixed in here, so I can't approve that's fixed. Could either wait until that merges and rebase or make these changes against main.

unstructured/ingest/interfaces.py
natygyoon: Thanks for the feedback, this looks much cleaner for sure.

unstructured/ingest/main.py
```diff
@@ -35,6 +36,35 @@
...
+                    ", ".join(self.mutually_exclusive),
+                ),
+            )
+
+        return super(MutuallyExclusiveOption, self).handle_parse_result(
+            ctx,
+            opts,
+            args,
+        )
+
```
ryannikolaidis, line 66: is this meant to be in this PR?

natygyoon: Nope, I must have made a mistake while merging main in.

cragwolfe (approved):
Tested with

https://github.com/Unstructured-IO/unstructured/blob/main/examples/ingest/wikipedia/ingest.sh 

adding 

    --fields-include text,type \
    
 Confirmed that elements in json outputs only had `text` and `type`.
 
Also confirmed that the outputs are equivalent to prior behavior if --fields-include is not specified.

cragwolfe (changes requested):
Please bump version and update CHANGELOG to reflect the change happening under `0.5.7-devX`

CHANGELOG.md
```diff
@@ -8,6 +18,7 @@
 
 ### Features
 
+* Add `--fields-include` parameter to `unstructured-ingest`
```
cragwolfe, line 21: leftover line can be removed

unstructured/ingest/main.py
```diff
@@ -338,7 +344,13 @@ def main(
     verbose,
     metadata_include,
     metadata_exclude,
+    fields_include,
 ):
+    if "metadata" not in fields_include and (metadata_include or metadata_exclude):
+        logger.error(
```
cragwolfe, line 350:
```suggestion
        logger.warning(
```

nit: this is really more of a warning

cragwolfe (approved):
LGTM!

ryannikolaidis (approved):
lgtm!

PR #379, opened 2023-03-18
Exclude empty tags during depth check

mu4farooqi:
I was trying HTML parser on some YC blogs and it didn't work because they were using unclosed `<br>` tags which don't need to be closed in HTML. I have excluded these tags from the depth calculation. Now any html in following format should work:

<p>
<br>Quick<br>
<br>Brown<br>
<br>Fox<br>
</p>.

Example HTML which wasn't working before: https://www.ycombinator.com/blog/changes-at-yc


MthwRobinson (approved):
LGTM! Thanks for the contribution

PR #382, opened 2023-03-20
chore: use only regex for contains_english_word.

cragwolfe:
Updates the characters to split when creating candidate english words. Now uses regex to parse out non-alphabetic characters for each word
 
 Note: This was originally an attempt to speedup `contains_english_word()` but there is no measurable change in performance.


MthwRobinson (approved):
LGTM

PR #386, opened 2023-03-20
feat: add `--flatten-metadata` to unstructured-ingest

natygyoon:
For some use cases, json output is often translated into a single row in a table, but having an extra level of indirection with a metadata nested dictionary makes inserting into such a table awkward.

Therefore, we want to add a `--flatten-metadata` parameter to https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/main.py, which simply results in all elements in the metadata field moving up a level and removing the metadata field. I.e.:
```
{
    "element_id": "d551bbfc9477547e4dce6264d8196c7b",
    "text": "More info available at the Github Project Page",
    "type": "Title",
    "metadata": {
      "page_number": 1
    }
},
```
->
```
{
    "element_id": "d551bbfc9477547e4dce6264d8196c7b",
    "text": "More info available at the Github Project Page",
    "type": "Title",
    "page_number": 1
},
```

### DIFF
--- a/.github/workflows/create_issue.yml
+++ b/.github/workflows/create_issue.yml
@@ -1,0 +1,31 @@
+name: create_jira_issue
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  create:
+    runs-on: ubuntu-latest
+    name: Create JIRA Issue
+    steps:
+    - name: Login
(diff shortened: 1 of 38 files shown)


natygyoon:
opened a new PR for this.

Issue #374, opened 2023-03-17
PDF partitioner duplicates text


ivan-p92:
Hi, while using `UnstructuredPDFLoader`, we noticed that the text content gets duplicated in the PDF partitioner. First, all elements are added one by one with the page's metadata, and then again, without metadata:

https://github.com/Unstructured-IO/unstructured/blob/f6d787d95ba7abea0fa14c089dbd7f18a8541b87/unstructured/partition/pdf.py#L253-L257

The last line in this snippet should probably be removed, right?

cragwolfe:
Thanks for reporting this @ivan-p92!

PR #388, opened 2023-03-21
Fix: duplicated addition to elements list

benjats07:
This PR addresses the issue https://github.com/Unstructured-IO/unstructured/issues/374 with the suggested fix.

In order to test is possible to use

```
results = pdf.partition_pdf("example-docs/layout-parser-paper-fast.pdf")
all_text =[r1.text for r1 in r]
print(all_text)
```
And the result should be the complete text from the file without any missing or duplicated text block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 0.5.5-dev1
+## 0.5.6
+
+* Fix problem with PDF partition (duplicated test)
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


cragwolfe (approved):

cragwolfe:
Tested with:

```
results = pdf.partition_pdf("example-docs/layout-parser-paper-fast.pdf, strategy="fast")
all_text =[r1.text for r1 in results]
print(all_text)
```

Double elements seen on main, but not with this PR.

PR #391, opened 2023-03-22
feat: update paragraph pattern regex

cragwolfe:
Simplifies the regex expression a touch and allows the pattern to expand to additional whitespace/newlines.


ryannikolaidis (approved):
lgtm!

PR #394, opened 2023-03-23
Adding content_type and file_filename to autopartition

amanda103:
Adding content_type and file_filename to auto_partition function. By accepting content_type as a string, we can bypass filetype detection in a lot of cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
-## 0.5.7-dev3
+## 0.5.7-dev4
 
 ### Enhancements
 
 * Refactored codebase using `exactly_one`
 * Adds ability to pass headers when passing a url in partition_html()
+* Added optional `content_type` and `file_filename` parameters to `partition()` to bypass file detection
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.7-dev4
+## 0.5.7
 
 ### Enhancements
 
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -30,6 +30,7 @@ In cases where ``libmagic`` is not available, filetype detection will fall back 
 As shown in the examples below, the ``partition`` function accepts both filenames and file-like objects as input.
 ``partition`` also has some optional kwargs.
 For example, if you set ``include_page_breaks=True``, the output will include ``PageBreak`` elements if the filetype supports it.
+Additionally you can bypass the filetype detection logic with the optional  ``content_type`` argument.
 You can find a full listing of optional kwargs in the documentation below.
 
 .. code:: python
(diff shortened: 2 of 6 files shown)


unstructured/file_utils/filetype.py
```diff
@@ -138,18 +156,32 @@ def __lt__(self, other):
...
+            _, extension = os.path.splitext(filename)  # type: ignore
+        else:
+            _, extension = os.path.splitext(file_filename)  # type: ignore
         extension = extension.lower()
         if LIBMAGIC_AVAILABLE:
-            mime_type = magic.from_file(filename, mime=True)
+            if filename:
+                mime_type = magic.from_file(filename, mime=True)
+            else:
+                mime_type = magic.from_file(file_filename, mime=True)  # type: ignore
```
cragwolfe, line 182:
```suggestion
                mime_type = magic.from_file(filename or file_filename, mime=True)  # type: ignore
```

cragwolfe (commented):
The logic generally makes sense to me.

Probably worth updating https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/bricks.rst letting the user know they can bypass mime type detection, and add a file_filename which may be important metadata to include in outputs for downstream applications.

docs/source/bricks.rst
```diff
@@ -38,7 +39,7 @@ You can find a full listing of optional kwargs in the documentation below.
 
 
   filename = os.path.join(EXAMPLE_DOCS_DIRECTORY, "layout-parser-paper-fast.pdf")
-  elements = partition(filename=filename)
+  elements = partition(filename=filename, content_type="application/pdf")
```
cragwolfe, line 42: the first example the user sees should probably be without `content_type="application/pdf"` as the easiest way to get started (assuming developer's eyes often just scan to the first code to copy and paste). maybe describe `content_type` in a comment below?

docs/source/bricks.rst
```diff
@@ -30,6 +30,7 @@ In cases where ``libmagic`` is not available, filetype detection will fall back
 As shown in the examples below, the ``partition`` function accepts both filenames and file-like objects as input.
 ``partition`` also has some optional kwargs.
 For example, if you set ``include_page_breaks=True``, the output will include ``PageBreak`` elements if the filetype supports it.
+Additionally you can bypass the filetype detection logic with the optional  ``content_type`` argument.
```
cragwolfe, line 33: ... which may be specified with either the ``filename`` or file-like object, ``file``.
?

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
-## 0.5.7-dev3
+## 0.5.7-dev4
```
cragwolfe, line 1: Probably a good time for a release, let's make this `0.5.7`

cragwolfe (approved):
LGTM!

PR #399, opened 2023-03-24
feat: add local file system connector for unstructured-ingest

natygyoon:
The following parameters have been added to unstructured-ingest to support bulk processing a directory of files in the local file system.
* `--local-input-path` 
* `--local-recursive`
* `--local-file-glob`


examples/ingest/local/ingest.sh
```diff
@@ -0,0 +1,29 @@
...
+#   1) --local-input-path  : path in the local file system which is to be processed
+#   2) --local-file-glob   : types of local files that are accepted,
+#                            provided as a comma-separated list
+# before running.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "$SCRIPT_DIR"/../../.. || exit 1
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --local-input-path "<path in local file system>" \
```
cragwolfe, line 19:
```suggestion
    --local-input-path . \
```
it's ok to be opinionated here

examples/ingest/local/ingest.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# Processes the Unstructured-IO/unstructured repository
+# through Unstructured's library in 2 processes.
```
cragwolfe, line 4: Let's just have this example script process everything under `example-docs/` without the user having to edit it.
Please update this comment accordingly.

examples/ingest/local/ingest.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Structured outputs are stored in local-ingest-output/
+
+# NOTE, this script is not ready-to-run!
```
cragwolfe, line 8: Let's change that :)

examples/ingest/local/ingest.sh
```diff
@@ -0,0 +1,29 @@
...
+# You must enter
+#   1) --local-input-path  : path in the local file system which is to be processed
```
cragwolfe, line 10:
```suggestion
#   1) --local-input-path  : path in the local file system which is to be processed
# The following CLI args are optional:
```

examples/ingest/local/ingest.sh
```diff
@@ -0,0 +1,29 @@
...
+#   2) --local-file-glob   : types of local files that are accepted,
+#                            provided as a comma-separated list
```
cragwolfe, line 12:
```suggestion
#                            provided as a comma-separated list
#   3) --local-recursive      : if specified, the contents sub-directories are processed as well
```

test_unstructured_ingest/test-ingest-local.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+cd "$SCRIPT_DIR"/.. || exit 1
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --metadata-exclude filename \
+    --local-input-path ./example-docs/ \
+    --local-file-glob layout-parser-paper-fast.* \
```
cragwolfe, line 9: nit: let's do *.html as it should be faster, and also more inline with a typical glob pattern.
that should also be in quotes or bash will try not to expand it and also be clear as such, so:

```
    --local-file-glob "*.html"
```

cragwolfe:
Seems pretty close. Some comments:

Some comments:

* Please remove `test_unstructured_ingest/expected-structured-output/local-ingest-output/.gitkeep` . This directory should get created by the connector if not present as it writes files.

* `--local-recursive` does not appear to be working from the command line.

* it should be possible to process a single file that is not a dir. For example:
```
  --metadata-exclude filename \
  --local-input-path example-docs/fake-html.html \
  --structured-output-dir local-ingest-output2 \
  --verbose \
  --reprocess
```

cragwolfe (approved):
LGTM

PR #402, opened 2023-03-27
feat: add --max-docs parameter to unstructured-ingest

natygyoon:
We want to add an optional integer parameter `--max-docs` to [unstructured-ingest](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/main.py) so that a maximum of N docs may be processed if `--max-docs` is specified.


cragwolfe (changes requested):
Rather than change each connector, the doc list can be truncated in the `main` loops: https://github.com/Unstructured-IO/unstructured/blob/a4394f6/unstructured/ingest/main.py#L66 . Please update the PR to do just that.

unstructured/ingest/main.py
```diff
@@ -63,7 +65,7 @@ def cleanup(self):
     def _filter_docs_with_outputs(self, docs):
         num_docs_all = len(docs)
         docs = [doc for doc in docs if not doc.has_output()]
-        num_docs_to_process = len(docs)
+        num_docs_to_process = len(docs) if self.max_docs is None else min(len(docs), self.max_docs)
```
cragwolfe, line 68: the change should be something right before this line, something like:
```
if self.max_docs is not None:
    docs = docs[self.max_docs]
```
then no change to `num_docs_to_process` is needed

unstructured/ingest/main.py
```diff
@@ -103,6 +107,11 @@ def run(self):
 
 
 @click.command()
+@click.option(
+    "--max-docs",
+    default=None,
```
cragwolfe, line 112:
```suggestion
    default=None,
    type=int,
```

unstructured/ingest/main.py
```diff
@@ -63,6 +65,8 @@ def cleanup(self):
     def _filter_docs_with_outputs(self, docs):
         num_docs_all = len(docs)
         docs = [doc for doc in docs if not doc.has_output()]
+        if self.max_docs is not None:
+            docs = docs[: min(len(docs), self.max_docs)]
```
cragwolfe, line 69:
```suggestion
            docs = docs[:self.max_docs]
```
this works. also, for this to not be a stack trace, you need to add `type=int`

cragwolfe: i think there should be another line here:
```
            num_docs_all = self.max_docs if num_docs_all > self.max_docs
```
to avoid misleading output like:
```
2023-03-28 17:07:43,057 MainProcess INFO     Skipping processing for 1 docs out of 3 since their structured outputs already exist
```

when no prior structured outputs exist.

cragwolfe (approved):
LGTM (once CI is clean)

cragwolfe (changes requested):
actually, please bump version again

PR #403, opened 2023-03-27
enhancement: update elements_to_json to potentially return a string

natygyoon:
`elements_to_json` has a filename kwarg that is must not be None for the function call not to result in a stack trace.

Instead, we want to allow `elements_to_json` to return the json string when filename is None (which would be the default value).

Also, `elements_from_json` may take a string instead of a filename with the `text` kwarg.


unstructured/staging/base.py
```diff
@@ -38,11 +39,22 @@ def convert_to_dict(elements: List[Element]) -> List[Dict[str, str]]:
...
+    """
     element_dict = convert_to_dict(elements)
-    with open(filename, "w") as f:
-        json.dump(element_dict, f, indent=indent)
+    if filename is not None:
+        with open(filename, "w") as f:
+            json.dump(element_dict, f, indent=indent)
+            return None
+    else:
+        return str(element_dict)
```
cragwolfe, line 57: this should be:

    return json.dumps(element_dict, indent=indent)

cragwolfe:
Please update the `Serializing Elements` section here, adding the following right after: https://github.com/Unstructured-IO/unstructured/blame/09b52b4/docs/source/getting_started.rst#L136

```
    # alternatively, one can also serialize/deserialize to/from a string with:
    serialized_elements_json = elements_to_json(elements)
    new_elements = elements_from_json(text=serialized_elements_json)
````

File: docs/source/getting_started.rst [hunk 2]
```diff
@@ -133,7 +133,10 @@ serializing and deserializing an ``Element`` list.
 
     elements_to_json(elements, filename=filename)
     new_elements = elements_from_json(filename=filename)
-
+    
+	# alternatively, one can also serialize/deserialize to/from a string with:
```
cragwolfe, line 137: spacing is off. replace tabs with spaces?

cragwolfe (approved):

PR #404, opened 2023-03-27
ci: publish amd and arm images

ryannikolaidis:
* Build, test, and publish both AMD and ARM images
* Publish manifest so that docker pull quay.io/unstructured-io/unstructured will automatically pull the right image based on the machine calling it
* Update `make docker-test` so that parameters are exposed as variables (so we can call it for images as tagged in CI build flow)
* Add .dockerignore

## Note
In running the arm and amd build+test flows as parallel jobs I tried using github artifacts, but this takes a lot of run time. I also tried caching via GHA, but results were inconsistent, likely due to this: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy. Ultimately I landed on pushing them up to the registry and using the registry as a cache, this seems to be the best option.

## Testing 
I ran a test push to quay.io/unstructured-io/unstructured-multiarch. `unstructured-multiarch` being a placeholder so I didn't overwrite our actual image repository. This is visible in the registry [here](https://quay.io/repository/unstructured-io/unstructured-multiarch?tab=tags).
Here is the run: https://github.com/Unstructured-IO/unstructured/actions/runs/4549028194
Test the artifact from it with:
* docker pull quay.io/unstructured-io/unstructured-multiarch:latest
* validate with `docker image inspect quay.io/unstructured-io/unstructured-multiarch:latest | grep Arch` to verify the right image was pulled for your machine


cragwolfe:
should the README be updated with updated run instructions, or saving that for a follow on PR?

ryannikolaidis:
> should the README be updated with updated run instructions, or saving that for a follow on PR?

Yea, my only thinking there being that it'd be nice to see it run once on main end-to-end and actually push up the artifact first. Definitely still tracking this.

cragwolfe:
Couple of questions:

* Why do the `build` jobs do a `docker push` as their last step, when there is a separate `push-images` job below below `build jobs`?
* Just to make sure I'm following, the `:arm` or `:amd` tag convention would be equivalent to `:arm-latest` or `:amd-latest`, right? (not suggesting to change it)

Thanks for pushing to  quay.io/unstructured-io/unstructured-multiarch, makes it a lot easier to track!

cragwolfe (approved):
LGTM! Validated the images in the test repo.

amd64:  (x86_64)
```
$ docker image inspect quay.io/unstructured-io/unstructured-multiarch:latest | grep Arch
        "Architecture": "amd64",
```
arm64:  (mac m1)
```
$ docker image inspect quay.io/unstructured-io/unstructured-multiarch:latest | grep Arch
	"Architecture": "arm64",
```

Also verified am able to do the following on both Intel and Apple silicon after running `python3`:

```
elements = partition_pdf(filename="example-docs/layout-parser-paper-fast.pdf")
print(elements[3])
```

🎉 🎉 🎉

ryannikolaidis:
> * Why do the `build` jobs do a `docker push` as their last step, when there is a separate `push-images` job below below `build jobs`?
Ahh, yea, so the docker push that happens in build is to amd-<commit hash> and arm-<commit hash> as a way to leverage the registry as a cache (I tried a few other options, but this was the cheapest and most reliable). The difficulty here was trying to store it somewhere without overwriting latest until they both pass the tests.

> * Just to make sure I'm following, the `:arm` or `:amd` tag convention would be equivalent to `:arm-latest` or `:amd-latest`, right? (not suggesting to change it)
Yep, that's exactly it. So once they have both passed testing then we push up each as the new "latest" `:arm` and `:amd`. 

Maybe it's simpler if for each case if they pass their own test we push straight to `:arm` and `:amd` (rather than pushing to the `<platform>-<commit hash>` tag as a cache? This would mean that one could fail and one could pass, but since we wouldn't publish the final manifest unless they both do and will only guide folks to pull from that, that's probably okay?(meaning in either platform `docker pull unstructured-io/unstructured` would still point to the previous version that passed both tests) Any feelings?

.github/workflows/docker-publish.yml
```diff
@@ -6,88 +6,117 @@ on:
...
 env:
-  DOCKER_REGISTRY: "quay.io"
-  DOCKER_NAMESPACE: "unstructured-io"
-  IMAGE_PLATFORMS: linux/amd64
-  PACKAGE: "unstructured"
+  DOCKER_REPOSITORY: quay.io/unstructured-io/unstructured
   PIP_VERSION: "22.2.1"
   PYTHON_VERSION: "3.8"
-  NLTK_DATA: ${{ github.workspace }}/nltk_data
   TEST_IMAGE_NAME: "unstructured-dev"
```
cragwolfe, line 12: no longer need `TEST_IMAGE_NAME`?

ryannikolaidis: nope, good eye! will remove

PR #413, opened 2023-03-28
fix: text kwargs no longer fail with empty string

MthwRobinson:
### Summary

Adds a fix to partition functions that accepts a `text` kwarg. They no longer fail if an empty string is passed, but rather return an empty list of elements. Additional, `partition_json` was updated so that it does not fail if the input is an empty list. Instead, an empty list of elements is returned.

### Testing

```python

partition_text(text="")
```

```python
from unstructured.partition.json import partition_json

partition_json(text="[]")
```

### DIFF
--- a/unstructured/partition/json.py
+++ b/unstructured/partition/json.py
@@ -6,7 +6,7 @@ from unstructured.documents.elements import Element
 from unstructured.partition.common import exactly_one
 from unstructured.staging.base import dict_to_elements
 
-LIST_OF_DICTS_PATTERN = r"\A\s*\[\s*{"
+LIST_OF_DICTS_PATTERN = r"\A\s*\[\s*{?"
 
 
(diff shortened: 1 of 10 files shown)


qued (approved):
LGTM!

unstructured/partition/json.py
```diff
@@ -6,7 +6,7 @@
 from unstructured.partition.common import exactly_one
 from unstructured.staging.base import dict_to_elements
 
-LIST_OF_DICTS_PATTERN = r"\A\s*\[\s*{"
+LIST_OF_DICTS_PATTERN = r"\A\s*\[\s*{?"
```
cragwolfe, line 9: why this change?

MthwRobinson: It raised an error on empty lists before

PR #414, opened 2023-03-29
build(deps): bump requirements

cragwolfe:
Mainly to pull in https://github.com/Unstructured-IO/unstructured-inference/pull/73 .


ryannikolaidis (approved):

requirements/cache.txt
```diff
@@ -0,0 +1 @@
+a
```
MthwRobinson, line 1: Is this right? Just double checking since it looked a little funny.

test_unstructured/partition/test_json.py
```diff
@@ -44,6 +44,8 @@ def test_partition_json_from_filename(filename: str):
 
     assert len(elements) == len(test_elements)
     for i in range(len(elements)):
+        print(elements[i].coordinates)
```
MthwRobinson, line 47: Do we still need these `print` statements in the test?

MthwRobinson (approved):
Updated data structured for coordinates is good to go from my perspective. Left a couple of minor/optional inline comments.

requirements/cache.txt
qued: That was me. There was a case where the action cache was essentially "stuck". The cache key had been created without storing the image, and so it couldn't be deleted, recalled, or overwritten. I created this file to change the hash key.

PR #416, opened 2023-03-29
feat: add --download-only parameter to unstructured-ingest

natygyoon:
Add `--download-only` parameter so that files may be downloaded if they are not already present (as usual, in either --download-dir or the default download ~/.cache/... location if --download-dir is not specified) and skip processing them through unstructured.


unstructured/ingest/connector/biomed.py
```diff
@@ -170,19 +175,20 @@ def get_file(self):
 
     def write_result(self):
         """Write the structured json result for this doc. result must be json serializable."""
-        output_filename = self._output_filename()
-        output_filename.parent.mkdir(parents=True, exist_ok=True)
-        with open(output_filename, "w") as output_f:
-            output_f.write(json.dumps(self.isd_elems_no_filename, ensure_ascii=False, indent=2))
-        logger.info(f"Wrote {output_filename}")
+        if not self.config.download_only:
```
cragwolfe, line 178: you can save the excess indentation underneath the `if` statement by doing the following instead:

```
    if self.config.download_only:
        return
    outpu_filename = ...
````

File: unstructured/ingest/interfaces.py [hunk 2]
```diff
@@ -93,40 +94,42 @@ def write_result(self):
...
-
-            if self.config.flatten_metadata:
-                for k, v in elem["metadata"].items():  # type: ignore[attr-defined]
-                    elem[k] = v
-                elem.pop("metadata")  # type: ignore[attr-defined]
-
-            self.isd_elems_no_filename.append(elem)
-
-        return self.isd_elems_no_filename
+        if not self.config.download_only:
```
cragwolfe, line 97: similar to the previous comment which is also applicable in other places but especially worthwhile here: short-circuit all this indentation by doing:

```
if self.config.download_only:
    return None
```

unstructured/ingest/main.py
```diff
@@ -405,10 +414,15 @@ def main(
...
-    if not preserve_downloads and download_dir:
+    if (not preserve_downloads and not download_only) and download_dir:
         logger.warning(
             "Not preserving downloaded files but --download_dir is specified",
         )
+    if local_input_path is not None and download_dir:
+        logger.warning(
+            "Files should already be in local file system: there is nothing to download, "
+            "but --download-dir is specified.",
+        )
```
cragwolfe, line 425: actually, can add a `sys.exit(1)` here. there is no point in the script continuing.

awalker4 (approved):
Tested locally, lgtm!

PR #419, opened 2023-03-30
chore: install all extras in Dockerfile

ryannikolaidis:
* Adds step to install all extras
* Adds smoke test of wikipedia ingest to validate in CI

## Testing
I did a run that triggered the build and tests [here](https://github.com/Unstructured-IO/unstructured/actions/runs/4566877060), all of which succeeded.

The build from that test is available in the `test-unstructured` image repository for commit tag `6fdba0e`.
To test:
```
docker pull quay.io/unstructured-io/test-unstructured:6fdba0e
docker run -td --name unstructured-all-extras quay.io/unstructured-io/test-unstructured:6fdba0e
docker cp test_unstructured_ingest/ unstructured-all-extras:/home/
docker exec -it unstructured-all-extras bash

# in container
# run the ingest tests. note they currently have known minor diffs with the validation set, 
# which is why I recommend running each individually
./test_unstructured_ingest/test-ingest-<TEST>.sh
```


cragwolfe:
> in container
 run the ingest tests. note they currently have known minor diffs with the validation set, 
 which is why I recommend running each individually

Yes, that is highly annoying. Only an ubunutu install matches the existing text fixtures, which starting happening a couple of weeks ago :(

cragwolfe (approved):
Awesome test instructions. Validated! 🎉

PR #420, opened 2023-03-30
!fix: return a list of elements in `stage_for_transformers`

MthwRobinson:
### Summary

Updates the `stage_for_transformers` function to return a list of elements, like the other staging bricks. Users can continue to use `chunk_by_attention_window` if they need to operate directly on strings. Having `stage_for_transformers` enables you to run operations like getting a pandas dataframe with pre-chunked segments by running `convert_to_dataframe(stage_for_transformers(elements, tokenizer))`.

Also fixed a bug that was causing the last word of the last segment to get cut off in `chunk_by_attention_window`.

### Testing

```python
from unstructured.staging.huggingface import stage_for_transformers
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("google/flan-t5-base")
# Copy paste text from this url: https://lite.cnn.com/2023/03/30/us/mlb-pitch-clock-batters-dg
text = <text>
elements = [Title("A baseball article"), NarrativeText(text=text)]
# You should see a title and several narrative text elements
stage_for_transformers(elements, tokenizer)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@
 * Partitioning functions that accept a `text` kwarg no longer raise an error if an empty
   string is passed (and empty list of elements is returned instead).
 * `partition_json` no longer fails if the input is an empty list.
+
+### BREAKING CHANGES
+
 * `stage_for_transformers` now returns a list of elements, making it consistent with other
   staging bricks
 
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM!

PR #421, opened 2023-03-30
fix: correct order of kwargs in pandoc

MthwRobinson:
### Summary

Corrects the order of the kwargs in the pandocs function. This was previously causing `partition_epub` to fail. This was not caught in CI because the EPUB tests were marked `xfail` due to requirements issues in the docker tests.

### Testing

```python
from unstructured.partition.epub import partition_epub

filename = "example-docs/winter-sports.epub"
partition_epub(filename=filename)
```


cragwolfe (approved):

cragwolfe:
time to cut a new release?

MthwRobinson:
Yep think we're in a good spot for a release

PR #422, opened 2023-03-30
build: only skip docker tests in docker

cragwolfe:


cragwolfe:
Already merged in a recent PR.

PR #439, opened 2023-04-03
feat: add sender and receive info to element metadata for emails

MthwRobinson:
### Summary

Adds `sent_from`, `sent_to`, `subject`, and `date` to the element metadata for emails

### Testing

The following code should now show the above metadata fields populated:

```python

elements = partition_email(filename="example-docs/fake-email.eml")
elements[0].metadata
```

```python

elements = partition_msg(filename="example-docs/fake-email.msg")
elements[0].metadata
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.5.10-dev0
+
+### Enhancements
+
+* Add sender, recipient, date, and subject to element metadata for emails
+
+### Features
+
+### Fixes
+
 ## 0.5.9
 
 ### Enhancements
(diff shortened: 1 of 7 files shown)


mallorih:
Is the 'build_email_metadata ` function doing something similar to `partition_email_header`?

MthwRobinson:
Yeah, main difference is it's outputting an `ElementMetadata` object rather than `List[Element]`. Thought about reusing `partition_email_header`, but thought it was cleaner to separate since they have different output types.

mallorih (approved):
lgtm

PR #443, opened 2023-04-04
feat: add --partition-by-api and --partition-host to unstructured-ingest

awalker4:
Testing...

### DIFF
--- a/unstructured/ingest/interfaces.py
+++ b/unstructured/ingest/interfaces.py
@@ -4,6 +4,8 @@ through Unstructured."""
 from abc import ABC, abstractmethod
 from typing import Optional
 
+import requests
+
 from unstructured.ingest.logger import logger
 from unstructured.partition.auto import partition
 from unstructured.staging.base import convert_to_dict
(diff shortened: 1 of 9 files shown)


cragwolfe:
@awalker4 , there should be one .sh file under `test_unstructured_ingest/` that processes 1 quick-to-process file against api.unstructured.io. Yes, https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-local.sh is a good reference. Might as well process a different file, though.

Yes, config needs to be refactored outside of this PR -- there basically should be a DocProcessorConfig class for values that connectors do not need to be aware of. But, that should be a standalone PR (please don't add to this one).

unstructured/ingest/interfaces.py
```diff
@@ -93,12 +97,35 @@ def write_result(self):
...
 
+    def partition_file(self):
+        if not self.config.partition_by_api:
+            logger.debug("Using local partition")
+            elements = partition(filename=str(self.filename))
+            return convert_to_dict(elements)
+
+        else:
+            hostname = self.config.partition_host
+            path = "general/v0/general"
```
cragwolfe, line 108: can this not be hardcoded?
perhaps change partition_host to partition_endpoint?

cragwolfe (approved):
LGTM!

PR #449, opened 2023-04-05
fix: Guard against style attribute being None

siddartha-RE:
Some document elements may have a null style element which triggers an exception when trying to access the name of the style.


MthwRobinson:
@siddartha-RE - Thanks for this contribution! Could you update  the version in `unstructured/__version__.py` to `0.5.10-dev2` and add an entry to the CHANGELOG? Will approve and merge once that's done.

MthwRobinson (approved):
LGTM!

PR #451, opened 2023-04-05
fix: change url to html_url

qued:
I realized that `github.event.issue.url` gives a url to the domain `api.github.com`. `html_url` (which doesn't seem to be documented in the `issues` interface) gives the url we want.


MthwRobinson (approved):
LGTM

PR #452, opened 2023-04-05
fix: fix html encoding to support foreign characters

### LINKED ISSUE #436: Is this an unnecessary encode?
When I tried to use partition_epub to process a Chinese book, I found that the output was incorrect. Through debugging, I found that this encode step caused the final error. However, I'm not sure if it's effective in other situations, so I dare not make a PR rashly.

[document_tree = etree.fromstring(content.encode(), self.parser)
](https://github.com/Unstructured-IO/unstructured/blame/b746cb0ab4c1693b3ee2bd8ddc25e6b5fa68b57e/unstructured/documents/xml.py#L58)

MthwRobinson:
### Summary

Closes #436. Updates the string encoding logic in `documents/xml.py` to better support reading foreign language characters.

### Testing

The following should now show `"每日新闻"` as the output. Previously the output was `"æ¯æ¥æ°é»`".

```python

html_text = "<html><div><p>每日新闻</p></div></html>"
str(elements[0])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.11-dev0
+## 0.5.11
 
 ### Enhancements
 
@@ -7,6 +7,7 @@
 ### Fixes
 
 * Guard against null style attribute in docx document elements
+* Update HTML encoding to better support foreign language characters
 
 ## 0.5.10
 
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM!

PR #453, opened 2023-04-05
fix: allow `replace_mime_encodings` to accept and `encoding` kwarg

### LINKED ISSUE #444: UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 19: invalid start byte
**Describe the bug**
When I attempt to process an EML messages using the html.py  i get this error

UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 19: invalid start byte

from

File ~\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\unstructured\cleaners\core.py:137, in replace_mime_encodings(text)
130 def replace_mime_encodings(text: str) -> str:
131 """Replaces MIME encodings with their UTF-8 equivalent characters.
132
133 Example
134 -------
135 5 w=E2=80-99s -> 5 w’s
136 """
--> 137 return quopri.decodestring(text.encode()).decode("utf-8")

Im actually calling it from Langchain, but my guess is that its using HTML.py

**To Reproduce**
You would need an email file. I can provide if required.

**Expected behavior**
Not throw an exception 

**Screenshots**

 - OS: [e.g. windows, mac, linux]
 - Python version [e.g. 3.8.15]
 
OS: windows 
Python: 3.10

**Additional context**
My guess is this is related to 

TODO(robinson) - There's likely a cleaner was to accomplish this and get all of the
unicode characters instead of just the quotes. Doing this for now since quotes are
an issue that are popping up in the SEC filings tests

One way to fix be to allow other decoder arguments other utf-8 

return quopri.decodestring(text.encode()).decode("utf-8")

I believe that 

https://stackoverflow.com/questions/46000191/utf-8-codec-cant-decode-byte-0x92-in-position-18-invalid-start-byte

says to  use ('ISO-8859-1')

MthwRobinson:
### Summary

Closes #444. Updates `replace_mime_encodings` to accept an `encoding` kwarg. Also passes in the `encoding` kwarg from `partition_email`.

### Testing

The following work now. With utf-8 encoding, this failed with a `UniDecodeError`.

```python
from unstructured.cleaners.core import replace_mime_encodings

text = '5 w=E2=80-92s'
replace_mime_encodings(text, encoding="latin-1")
```

### DIFF
--- a/unstructured/cleaners/core.py
+++ b/unstructured/cleaners/core.py
@@ -140,7 +140,7 @@ def clean_trailing_punctuation(text: str) -> str:
 
 
 def replace_mime_encodings(text: str, encoding: str = "utf-8") -> str:
-    """Replaces MIME encodings with their UTF-8 equivalent characters.
+    """Replaces MIME encodings with their equivalent characters in the specified encoding.
 
     Example
     -------
--- a/unstructured/cleaners/core.py
+++ b/unstructured/cleaners/core.py
@@ -139,14 +139,14 @@ def clean_trailing_punctuation(text: str) -> str:
     return text.strip().rstrip(".,:;")
 
 
-def replace_mime_encodings(text: str) -> str:
+def replace_mime_encodings(text: str, encoding: str = "utf-8") -> str:
     """Replaces MIME encodings with their UTF-8 equivalent characters.
 
     Example
     -------
     5 w=E2=80-99s -> 5 w’s
     """
-    return quopri.decodestring(text.encode()).decode("utf-8")
+    return quopri.decodestring(text.encode()).decode(encoding)
 
 
 def clean_prefix(text: str, pattern: str, ignore_case: bool = False, strip: bool = True) -> str:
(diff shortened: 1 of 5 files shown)


unstructured/cleaners/core.py
```diff
@@ -139,14 +139,14 @@ def clean_trailing_punctuation(text: str) -> str:
     return text.strip().rstrip(".,:;")
 
 
-def replace_mime_encodings(text: str) -> str:
+def replace_mime_encodings(text: str, encoding: str = "utf-8") -> str:
     """Replaces MIME encodings with their UTF-8 equivalent characters.
```
qued, line 143: Should probably update the docstring.

qued (approved):
One nit about the docstring but otherwise LGTM!

unstructured/cleaners/core.py
MthwRobinson: Thanks. Updated the doc string and added a test.

PR #455, opened 2023-04-06
build: add mimetypes DB to docker image

cragwolfe:
The `mailcap` centos7 package provides the file `/etc/mime.types`, which is used by the `mimetypes` python package. That said, the `unstructured` code base does not make much use of this but the upstream `unstructured-api` does.

Bonus: `docx` mimetype added in lookup table.


ryannikolaidis (approved):
LGTM! Tested this by running `yum -y install mailcap` in a container and validating it resolved the `File type not supported: fake.docx` issue when the docx is passed as `application/octet-stream`

PR #456, opened 2023-04-06
feat: enable grouping broken paragraphs in `partition_text`

### LINKED ISSUE #406: Group together blocks of text separated by line breaks in `partition_text`
Text files often contains blocks of text that look like the following:

```python
Hello my name is
Matt and I live in
the USA.

I love to extract
text from documents.
```

Currently, `partition_text` processes the file line-by-line, so you'll wind up with `[NarrativeText(text="Hello my name is"), NarrativeText("Matt and I live in") ...']`. The goal of this issue is to write a cleaning brick called `group_paragraphs` or similar that groups together text separated by a single line break, so we wind up with:

```python
[
NarrativeText(text="Hello my name is Matt and I live in the USA."),
NarrativeText(text="I love to extract text from documents."),
]
```

This cleaning brick should be applied automatically in the `partition_text` function so that text files with line breaks are processed appropriately.

MthwRobinson:
### Summary

Closes #406. Adds a `group_broken_paragraphs` cleaning brick for grouping together sentences that have a page break for formatting purposes. This is common in `.txt` files. Also add a `paragraph_grouper` kwarg to `partition_text` to allow for passing in a paragraph grouping callable.

### Testing

The following code should produce two `NarrativeText` elements. Without the paragraph grouper, `partition_text` produce four elements due to the line break.

```python
from unstructured.cleaners.core import group_broken_paragraphs


text = """The big brown fox
was walking down the lane.

At the end of the lane, the
fox met a bear."""

partition_text(text=text, paragraph_grouper=group_broken_paragraphs)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## 0.5.12-dev1
+## 0.5.12-dev2
 
 ### Enhancements
 
 * Use the image registry as a cache when building Docker images.
+* Adds the ability for `partition_text` to group together broken paragraphs.
 
 ### Features
 
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -414,7 +414,7 @@ Examples:
   At the end of the lane, the
   fox met a bear."""
 
-  partition_text(text=text, paragraph_group=group_broken_paragraphs)
+  partition_text(text=text, paragraph_grouper=group_broken_paragraphs)
 
 
 ########
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -567,6 +567,46 @@ Examples:
   clean_trailing_punctuation("ITEM 1A: RISK FACTORS.")
 
 
+``group_broken_paragraphs``
+---------------------------
+
+Groups together paragraphs that are broken up with line breaks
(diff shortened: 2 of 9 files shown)


cragwolfe:
Couple of comments:

I'll try this out on some real docs that tend to have blank lines between paragraphs and see there are reasonable elements that come out. It's not obvious to me they would. E.g., https://www.ietf.org/rfc/rfc1918.txt and https://www.ietf.org/rfc/rfc1918.txt .

Also, it'd be nice if `parition_txt()` could automatically do the right thing in most cases, allowing the user to choose a certain splitting pattern if desired. I.e., paragraphs in examples above are NarrativeText items, and in the case of a phone book (a bunch of single lines), each line is its own element (not worrying about the type of element too much). From an implementation note, it seems like it would be more straightforward to choose between to regex splitters, adding a "blank line" one after https://github.com/Unstructured-IO/unstructured/blob/3467a27/unstructured/nlp/patterns.py#L71 when partitioning, rather than relying on a cleaning brick to do so after the fact (though the cleaning brick is a nice to have regardless).

qued (approved):
LGTM once the formatting issues are resolved. See also @cragwolfe 's comments though.

PR #460, opened 2023-04-07
fix(ingest): import connector-specific modules on demand

cragwolfe:
As most connectors have additional dependencies that may not be installed locally, only import those connectors' modules when they are being used.


amanda103 (approved):
lgtm pending passing builds 👍🏻

Issue #252, opened 2023-02-22
Create a data connector for Slack


mallorih:
Create a data connector that:

fetches posts from a specific slack channel with the option to add a data filter.
stores the extracted data locally as text files (at least temporarily for processing)
processes them using [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py).
See [Adding Data Connectors](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#adding-data-connectors) for details on how to get started. Make sure to include a link to this issue when submitting a PR.

Definition of Done

The [checklist](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist) has been completed.
The connector is able to pull all conversations in a channel or take in a date range and extract conversations from those dates.
The connector takes in credentials for private channels.
The connector is able to process a single text file.
The connector is able to process all text files extracted from slack, recursively.
For now, it is OK to process only doc types that [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py) is capable of processing.
Bonus points: the ability to filter by document type.

MthwRobinson:
Addressed in #462

PR #462, opened 2023-04-09
Slack connector

tabossert:
This connector takes a slack channel id, token and other options to pull conversation history for a channel and store it as a text file that is then processed by unstructured into expected output.

Related to https://github.com/Unstructured-IO/unstructured/issues/252


CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+
+* Added method to utils to allow date time format validation
+
+### Features
+
+* Add Slack connector to pull messages for a specific channel
+
+### Fixes
+
+
```
cragwolfe, line 13: note on procedure: when it is a -devX bump, just add the bullet points to the existing -dev section (in this case under 0.5.12-dev3) , then bump the devX version (so -dev3 becomes -dev4).

unstructured/ingest/connector/slack.py
```diff
@@ -0,0 +1,211 @@
...
+    token: str
+    oldest: str
+    latest: str
+
+    # Standard Connector options
+    download_dir: str
+    output_dir: str
+    re_download: bool = False
+    preserve_downloads: bool = False
+    verbose: bool = False
```
cragwolfe, line 36: there are more standard options needed here:
https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/connector/git.py#L30
...
(we'll fix this pattern soon)

unstructured/utils.py
```diff
@@ -50,3 +51,26 @@ def dependency_exists(dependency):
     except ImportError:
         return False
     return True
+
+
+# Copied from unstructured/ingest/connector/biomed.py
```
cragwolfe, line 56: cool, might as well update biomed connector to call this one to de-dupe.

unstructured/utils.py
```diff
@@ -50,3 +51,26 @@ def dependency_exists(dependency):
...
+def validate_date_args(date):
+    date_formats = ["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%dT%H:%M:%S%z"]
```
cragwolfe, line 58: can make this is a constant Tuple outside the function

unstructured/utils.py
```diff
@@ -50,3 +51,26 @@ def dependency_exists(dependency):
...
+
+    valid = False
+    if date:
+        for format in date_formats:
+            try:
+                datetime.strptime(date, format)
+                valid = True
```
cragwolfe, line 65: should just return `True` at this point

unstructured/utils.py
```diff
@@ -50,3 +51,26 @@ def dependency_exists(dependency):
     except ImportError:
         return False
     return True
+
+
+# Copied from unstructured/ingest/connector/biomed.py
+def validate_date_args(date):
```
cragwolfe, line 57: worth adding a type hint for the arg:
```
def validate_date_args(date: str):
```
or might have to do:
```
def validate_date_args(date: Optional[str] = None)
```

unstructured/utils.py
```diff
@@ -50,3 +51,26 @@ def dependency_exists(dependency):
...
+    date_formats = ["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%dT%H:%M:%S%z"]
+
+    valid = False
+    if date:
```
cragwolfe, line 61: i'd do `if not date` and just raise a ValueError here, and avoid the extra nesting.

unstructured/utils.py
```diff
@@ -50,3 +51,26 @@ def dependency_exists(dependency):
...
+        for format in date_formats:
+            try:
+                datetime.strptime(date, format)
+                valid = True
+                break
+            except ValueError:
+                pass
+
+        if not valid:
```
cragwolfe, line 70: if we return per above comment, we don't need the `if not valid`, can just raise the `ValueError`

cragwolfe: (sorry for the nits on the copied function, probably better if functions in utils.py have a bit of a higher bar than one-off connector functions)

unstructured/ingest/connector/slack.py
```diff
@@ -0,0 +1,226 @@
...
+            # don't traverse symlinks, not that there every should be any
+            if os.path.isdir(sub_dir) and not os.path.islink(sub_dir):
+                self.cleanup(sub_dir)
+        os.chdir("..")
+        if len(os.listdir(cur_dir)) == 0:
+            os.rmdir(cur_dir)
+
+    def initialize(self):
+        """Verify that can get metadata for an object, validates connections info."""
+        os.mkdir(self.config.download_dir)
```
cragwolfe, line 214: actually can remove this line since when writing a file on line 160, parent dirs are created if needed. (also fwiw if this dir already exists, it results an error)

unstructured/ingest/connector/slack.py
```diff
@@ -0,0 +1,226 @@
...
+                )
+                messages.extend(result["messages"])
+        except SlackApiError as e:
+            logger.error(f"Error: {e}")
+
+        with open(self._tmp_download_file(), "w") as f:
+            for message in messages:
+                f.write(message["text"] + "\n")
+
+        f.close()
```
cragwolfe, line 155: can remove this `f`?

unstructured/ingest/connector/slack.py
```diff
@@ -0,0 +1,226 @@
...
+                    channel=self.channel,
+                    oldest=oldest,
+                    latest=latest,
+                    cursor=result["response_metadata"]["next_cursor"],
+                )
+                messages.extend(result["messages"])
+        except SlackApiError as e:
+            logger.error(f"Error: {e}")
+
+        with open(self._tmp_download_file(), "w") as f:
```
cragwolfe, line 151: total nit, but something a bit more verbose than f for the file handle would be preferred

unstructured/ingest/connector/slack.py
cragwolfe, line 214: (2nd attempt at the same comment if you notice a deleted comment)
this `mkdir` should not be necessary since directories under download_dir (including download_dir) are created on line 103.

examples/ingest/slack/ingest.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Processes the Unstructured-IO/unstructured repository
+# through Unstructured's library in 2 processes.
+
+# Structured outputs are stored in slack-ingest-output/
+
+# oldest, latest arguments are optional
+
+# Ingests a slack text channel into a file.
```
cragwolfe, line 10: Maybe add a hint that the token associated with the app must be in a channel to be able to ingest from it. And, it is channel id, not channel name.

unstructured/ingest/main.py
```diff
@@ -321,6 +321,26 @@ def run(self):
     "the default download ~/.cache/... location in case --download-dir is not specified and "
     "skip processing them through unstructured.",
 )
+@click.option(
+    "--slack-channels",
+    default=None,
+    help="Comma separated list of Slack channels to pull messages from, can be a public or private channel",
```
cragwolfe, line 327:
```suggestion
    help="Comma separated list of Slack channel IDs to pull messages from, can be a public or private channel",
```

test_unstructured_ingest/test-ingest-slack.sh
```diff
@@ -0,0 +1,29 @@
...
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+cd "$SCRIPT_DIR"/.. || exit 1
+
+if [ -z "$SLACK_CHANNELS" ] || [ -z "$SLACK_TOKEN" ]; then
+   echo "Skipping Slack ingest test because SLACK_CHANNELS and SLACK_TOKEN env vars are not set."
+   exit 0
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+        --slack-channels "${SLACK_CHANNELS}" \
```
cragwolfe, line 12: actually, SLACK_CHANNELS can be hardcoded in this script since the name is visible in the expected output file anyway (not a secret, unlike the token).

test_unstructured_ingest/test-ingest-slack.sh
```diff
@@ -0,0 +1,29 @@
...
+        --slack-token "${SLACK_TOKEN}" \
+        --download-dir slack-ingest-download \
+        --structured-output-dir slack-ingest-output \
+        --start-date 2023-04-01 \
+        --end-date 2023-04-08T12:00:00-08:00
+
+if ! diff -ru slack-ingest-output test_unstructured_ingest/expected-structured-output/slack-ingest-channel; then
```
cragwolfe, line 19: Brand new pattern to follow to make updating fixtures easier:

https://github.com/Unstructured-IO/unstructured/blob/7b44bcd/test_unstructured_ingest/test-ingest-azure.sh#L17

Ingest is a moving target! :)

cragwolfe:
Lots of nits, but functionality is all there. Works great when tested locally.

unstructured/utils.py
trevor-primer: @cragwolfe My thinking for putting this in utils is that we have a single set of input args --start-date --end-date, that this method would validate the args, and then each connector that takes a date range would then translate date time into the format the connector would use.

unstructured/utils.py
```diff
@@ -1,9 +1,12 @@
 import importlib
 import json
+from datetime import datetime
 from functools import wraps
 from typing import Dict, List, Optional, Union
 
 
+date_formats = ("%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d+%H:%M:%S", "%Y-%m-%dT%H:%M:%S%z")
```
cragwolfe, line 8:
```suggestion
DATE_FORMATS = ("%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d+%H:%M:%S", "%Y-%m-%dT%H:%M:%S%z")
```

unstructured/ingest/connector/slack.py
```diff
@@ -0,0 +1,226 @@
...
+        for sub_dir in sub_dirs:
+            # don't traverse symlinks, not that there every should be any
+            if os.path.isdir(sub_dir) and not os.path.islink(sub_dir):
+                self.cleanup(sub_dir)
+        os.chdir("..")
+        if len(os.listdir(cur_dir)) == 0:
+            os.rmdir(cur_dir)
+
+    def initialize(self):
+            """Verify that can get metadata for an object, validates connections info."""
```
cragwolfe, line 214: whitespace/indentation off? also should have a `pass` in function body

cragwolfe:
please bump __version__.py and the CHANGELOG again (moving target)

cragwolfe:
LGTM! I think the PR just needs to be taken out of Draft mode

cragwolfe (approved):
LGTM!

PR #467, opened 2023-04-11
build: skip biomed ingest tests 90% of time due to ftp connectivity

cragwolfe:
The Biomed connectors have been flaky with ftp connections lately.


ryannikolaidis:
> > is the ftp connection issue particular to this source (can we host somewhere else)?
> 
> i believe so, that this is a biomed server issue. but this connector is specific to biomed (more than just ftp), so to put it somewhere else would mean we would just be testing the PDF's, not the connector per-se.

ah got it, yea fair. hm, does immediately retrying succeed? (maybe we need an option for retry attempts in the pipeline or just in our test) or are there just blanket periodic outages? (maybe we can detect that and skip the test)

PR #473, opened 2023-04-12
feat: allow headers in `partition`

MthwRobinson:
### Summary

Allows for passing in headers to `partition` if a URL is passed.

### Testing

```python

url = "https://www.understandingwar.org/sites/default/files/Russian%20Offensive%20Campaign%20Assessment%2C%20April%2011%2C%202023.pdf"
elements = partition(url=url, headers={"Accept": "application/pdf"}, strategy="fast")
```

### DIFF
--- a/unstructured/partition/auto.py
+++ b/unstructured/partition/auto.py
@@ -1,5 +1,5 @@
 import io
-from typing import IO, Callable, Optional, Tuple
+from typing import IO, Callable, Dict, Optional, Tuple
 
 import requests
 
@@ -31,6 +31,7 @@ def partition(
     strategy: str = "hi_res",
     encoding: str = "utf-8",
     paragraph_grouper: Optional[Callable[[str], str]] = None,
+    headers: Dict[str, str] = {},
 ):
     """Partitions a document into its constituent elements. Will use libmagic to determine
     the file's type and route it to the appropriate partitioning function. Applies the default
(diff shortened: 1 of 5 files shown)


unstructured/partition/auto.py
```diff
@@ -31,6 +31,7 @@ def partition(
     strategy: str = "hi_res",
     encoding: str = "utf-8",
     paragraph_grouper: Optional[Callable[[str], str]] = None,
+    headers: Dict[str, str] = {},
```
qued, line 34: Should probably add this to the parameters in the docstring.

unstructured/partition/auto.py
```diff
@@ -62,7 +63,11 @@ def partition(
...
 
     if url is not None:
-        file, filetype = file_and_type_from_url(url=url, content_type=content_type)
+        file, filetype = file_and_type_from_url(
+            url=url,
+            content_type=content_type,
+            headers=headers,
+        )
     else:
         filetype = detect_filetype(
```
qued, line 72: Should we emit a warning if we end up in the `else` block and `headers` is being ignored?

qued (approved):
LGTM other than some suggestions. The testing code returns a list of elements.

unstructured/partition/auto.py
MthwRobinson: Oh yeah good idea, will add that in.

Issue #474, opened 2023-04-12
bug/Encoding error for markdown with emojis


MthwRobinson:
**Describe the bug**
The following error occurs when you when try to process a markdown file that contains emojis.

```python
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9c in position 12: invalid start byte
```

**To Reproduce**
```python
from unstructured.partition.auto import partition

url = "https://raw.githubusercontent.com/hwchase17/langchain/master/README.md"
partition(url=url, content_type="text/markdown")
```

**Expected behavior**
The file should be processed without an exception.

**Desktop (please complete the following information):**
 - OS: Mac
 - Python version 3.8
 - `unstructured` version 0.5.12

MthwRobinson:
Partially addressed in #477 but the emojis still aren't retained in the output. We get `Hello ð\x9f\x98\x80` for example instead of `Hello 😀`. Going to keep this issue open until that's addressed.

PR #477, opened 2023-04-12
fix: encode xml string if document_tree is `None` in `_read_xml`

jonvet:
I had trouble reading a markdown document in langchain which contained an emoji. Currently in that case `etree.fromstring(content, self.parser)` returns None. So it doesn't raise the ValueError and doesn't try to encode to string. I'm raising the ValueError in that case to ensure we enter the except block and encode the string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.5.13-dev0
+
+### Enhancements
+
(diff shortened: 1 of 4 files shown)


MthwRobinson (approved):
Thanks for the contribution! This closes #474

PR #478, opened 2023-04-13
fix: fix text_type.py exceeds_cap_ratio() returns

jay-ylee:
There are cases when function is_possible_narrative_text receives an incorrect return from function exceeds_cap_ratio and does an incorrect classification, so some of the return values of exceeds_cap_ratio are corrected

### DIFF
--- a/unstructured/partition/text_type.py
+++ b/unstructured/partition/text_type.py
@@ -265,15 +265,17 @@ def exceeds_cap_ratio(text: str, threshold: float = 0.5) -> bool:
     # NOTE(robinson) - Currently limiting this to only sections of text with one sentence.
     # The assumption is that sections with multiple sentences are not titles.
     if sentence_count(text, 3) > 1:
-        logger.debug(f"Text does not contain multiple sentences:\n\n{text}")
+        # logger.debug(f"Text contains multiple sentences:\n\n{text}")
         return False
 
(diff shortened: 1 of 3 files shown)


unstructured/partition/text_type.py
```diff
@@ -265,15 +265,17 @@ def exceeds_cap_ratio(text: str, threshold: float = 0.5) -> bool:
     # NOTE(robinson) - Currently limiting this to only sections of text with one sentence.
     # The assumption is that sections with multiple sentences are not titles.
     if sentence_count(text, 3) > 1:
-        logger.debug(f"Text does not contain multiple sentences:\n\n{text}")
+        # logger.debug(f"Text contains multiple sentences:\n\n{text}")
```
MthwRobinson, line 268: Can we get rid of this old debug statement?

MthwRobinson (changes requested):
Thanks for this! Could you get add an entry to the changelog under the fixes header and remove the commented out debug statement that I pointed out inline? Will approve and merge once that's in.

unstructured/partition/text_type.py
jay-ylee: Of course, gonna change as you say

MthwRobinson (approved):
LGTM! Thanks for the contribution!

PR #481, opened 2023-04-13
feat: cleaning brick for normalizing bytes string output

### LINKED ISSUE #474: bug/Encoding error for markdown with emojis

MthwRobinson:
### Summary

Closes #474, expanding on the fix from #477. Adds a cleaning brick for converting bytes string output to properly encoded strings. This type of output occurs when there's a character like an emoji in HTML or Markdown that the HTML parser does not understand. Note that this is not required if the HTML document uses [UTF-8 codes](https://www.w3schools.com/html/html_emojis.asp) for emojis.

### Testing

The following two snippets should result in outputs that contain the emoji.

```python
  from unstructured.cleaners.core import bytes_string_to_string

  text = """\n<html charset="utf-8"><p>Hello 😀</p></html>"""
  elements = partition_html(text=text)
  elements[0].apply(bytes_string_to_string)
  # The output should be "Hello 😀"
  elements[0].text
```

```python

  text = """\n<html charset="utf-8"><p>Hello &#128512;</p></html>"""
  elements = partition_html(text=text)
  # The output should be "Hello 😀"
  elements[0].text
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
-## 0.5.13-dev4
+## 0.5.13-dev5
 
 ### Enhancements
 
 * Allow headers to be passed into `partition` when `url` is used.
 
 ### Features
+
+* `bytes_string_to_string` cleaning brick for bytes string output.
 
 ### Fixes
 
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -801,6 +801,37 @@ Examples:
   # Returns "Look at me, I'm flying!"
   extract_text_after(text, r"SPEAKER \d{1}:")
 
+
+``bytes_string_to_string``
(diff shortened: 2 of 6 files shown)


qued (approved):
LGTM!

PR #482, opened 2023-04-14
build(CI): add access token for github-ingest test

cragwolfe:
Avoids the occaisonal CI test failures  in test-ingest-github.sh that were due to rate-limited non-auth'ed requests against a GitHub repo.


ryannikolaidis (approved):
LGTM!

PR #486, opened 2023-04-14
fix: update ingest test fixtures, disable biomed test

cragwolfe:
* Update test fixtures that should have been updated in prior commit
* Disable biomed ingest tests for now, the fail more often than not
* Bonus: echo `tesseract --version` in the update script, since that is a key thing that influences fixture outputs.


amanda103 (approved):
lgtm 👍🏻

PR #488, opened 2023-04-16
test: avoid misleading output in ingest tests

cragwolfe:
Previously, if there was an error (non-zero exit code) in an ingest test script, the script would still complete and echo a warning about mismatched outputs and how to regenerate the fixtures. However, this statement is irrelevant and misleading: if the ingest failed with a non-zero exit code in the first place, that is the failure that should be debugged -- don't confuse the user with a comment about outputs.


awalker4 (approved):
LGTM! I was going to try verifying by breaking one of the tests locally, but of course the differing output bug is getting in the way.

cragwolfe:
@awalker4 , i added test instructions. the goal here is to fail before getting to differing output in the first place.

what "differing output bug is getting in the way"?

awalker4:
That was the issue that @qued chased with different ocr results per tesseract version. In any case, I tried running `test-ingest-against-api.sh` with a bogus `partition-endpoint` and it did stop right at the fail point.

PR #489, opened 2023-04-17
fix: correct return types in `exceeds_caps_ratio`

jay-ylee:
I found one minor error in a previous pull request that was merged. It was already added to the ChangeLog, and I was wondering if it would be possible to merge this fix as well.
Again in partition.text_type.py, as same as last time, I changed one of the returns from False to True.


MthwRobinson (approved):
LGTM! Just added a changelog entry and a small tweak to fix a failing test.

Issue #492, opened 2023-04-17
auto partitioning a json file with the file= param fails


cragwolfe:
**Describe the bug**
Partitioning a .json file (that has a list Unstructured elements) with the `file=` parameter in `unstructured.partition.auto::partition()` results in treating the file like a TXT file instead of a JSON one,
whereas the `filename=` parameter works fine.

**To Reproduce**

```
(unst) bash-5.1$ cat failure-case.py
import json
from unstructured.partition.auto import partition

filename = "test_unstructured_ingest/expected-structured-output/azure-blob-storage/spring-weather.html.json"
with open(filename, "r", encoding="utf-8") as partition_f:
    elems = partition(file=partition_f)
for elem in elems[:4]:
    print(f"elem text: {str(elem)}")

(unst) bash-5.1$ PYTHONPATH=. python failure-case.py
elem text: [
elem text: {
elem text: "element_id": "41f6e17bf5e9a407fcca74e902f802a0",
elem text: "text": "News Around NOAA",

```

**Expected behavior**

Unstructured elements are properly deserialized from the .json file, as is the case when calling:

    partition(filename="test_unstructured_ingest/expected-structured-output/azure-blob-storage/spring-weather.html.json")

So, the first element should have a `text` value of "News Around NOAA".

PR #493, opened 2023-04-17
fix: workaround .json file detection with old libmagic installs

cragwolfe:
Fixes issue where .json files were recognized as "text/plain" rather than "application/json on the Unstructured image (and other installs that may have an older libmagic).

Also adds missing json auto partition tests.

Including an xfail test for https://github.com/Unstructured-IO/unstructured/issues/492 .


amanda103 (approved):
lgtm 👍🏻

PR #508, opened 2023-04-20
Sync version-sync.sh across repos

natygyoon:
There are currently a few versions of the `version-sync.sh` script floating around in various repos. This issue provides for updating them all to the same script.


MthwRobinson (approved):
LGTM assuming the build passes!

PR #510, opened 2023-04-20
feat: add msg attachment support

mallorih:
This PR adds functionality to extract and save the attachments of msg files.

Files changed:
> unstructured>partition>msg.py
> test_unstructured>partition>test_msg.py

### Testing

```python
from unstructured.partition.msg import extract_attachment_info

attachments = extract_msg_attachment_info("example-docs/fake-email-attachment.msg", output_dir="example-docs")

# try with multiple attachments
attachments = extract_msg_attachment_info("example-docs/fake-email-multiple-attachments.msg", output_dir="example-docs")

### CODE REVIEW

File: unstructured/partition/msg.py [hunk 1]
```diff
@@ -67,3 +67,32 @@ def build_msg_metadata(msg_obj: msg_parser.MsOxMessage) -> ElementMetadata:
...
         date=email_date,
     )
+
+
+def extract_attachment_info(
+    filename: str,
+    output_dir: Optional[str] = None,
+) -> List[Dict[str, str]]:
+    if filename is not None:
+        msg_obj = msg_parser.MsOxMessage(filename)
```
MthwRobinson, line 77: I think  🐂 ✉️ whenever I see `MsOxMessage` 🤣

unstructured/partition/msg.py
```diff
@@ -67,3 +67,32 @@ def build_msg_metadata(msg_obj: msg_parser.MsOxMessage) -> ElementMetadata:
         subject=getattr(msg_obj, "subject", None),
         date=email_date,
     )
+
+
+def extract_attachment_info(
+    filename: str,
```
MthwRobinson, line 73: Can we update this to accept `file` as well? And then you can use the `exactly_one` helper function instead of raising the `ValueError` in this function.

unstructured/partition/msg.py
```diff
@@ -67,3 +67,32 @@ def build_msg_metadata(msg_obj: msg_parser.MsOxMessage) -> ElementMetadata:
         subject=getattr(msg_obj, "subject", None),
         date=email_date,
     )
+
+
+def extract_attachment_info(
```
MthwRobinson, line 72: Should we name this `extract_msg_attachment_info` to deconflict with the `.eml` version if both of these will be public functions?

MthwRobinson (changes requested):
Looks great! Example code worked for me. Just two small inline suggestions. May also be worth testing on something with a PDF or other non plain text attachment.

MthwRobinson (approved):
LGTM once you fix the one test. Could you also update the commit message to `feat: add msg attachment support` for consistency with our other messages?

PR #511, opened 2023-04-21
ci: separate job for ingest tests

MthwRobinson:
### Summary

Separates ingest and unit tests into separate CI jobs. This will help the tests to run faster (since they can run in parallel) and make it more apparent which tests have failed in the case of a failure.


ryannikolaidis (approved):
This is a great idea. LGTM!

.github/workflows/ci.yml
```diff
@@ -75,7 +75,44 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/cache@v3
+      id: virtualenv-cache
+      with:
+        path: |
+          .venv
+          nltk_data
+        key: unstructured-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('requirements/*.txt') }}
+    # NOTE(robinson) - This is a fallback in case the lint job does not find the cache.
```
ryannikolaidis, line 95:
```suggestion
    # NOTE(robinson) - This is a fallback in case the job does not find the cache.
```

PR #513, opened 2023-04-21
chore: change table param name

qued:
Updated parameter name that controls whether we try to infer table structure.

#### Testing:

This doesn't change any functionality so as long as CI passes it should be good.


cragwolfe (approved):
LGTM! Thanks for polishing up the kwargs and docstrings.

PR #514, opened 2023-04-21
enhancement: check for copy protection on PDFs and fallback to hi res when necessary

### LINKED ISSUE #469: Fallback to OCR if a PDF has copy/paste protection
Some PDFs have copy/paste protection that obscure text that is copy and pasted from the document. For example, if you copy `Important Title` from the document it could come out `a;dklja adlkfad`. The goal of this issue is to fall back to converting the PDF to image and OCRing if the PDF is copy/paste protected.

MthwRobinson:
### Summary

Closes #469. Updates `partition_pdf` to detect when PDFs have copy protection and falls back to using the hi res strategy when necessary. If `detectron2` is not installed _and_ the text is not extractable, `partition_pdf` raises an exception because there are no available methods to process the PDF.

### Testing

```python

# This will process without issue
elements = partition_pdf("example-docs/copy-protected.pdf", strategy="hi_res")

# This will output a warning and fall back to hi_res
elements = partition_pdf("example-docs/copy-protected.pdf", strategy="fast")
```


docs/source/bricks.rst
```diff
@@ -302,6 +302,25 @@ Examples:
   elements = partition_pdf("example-docs/layout-parser-paper-fast.pdf", ocr_languages="eng+swe")
 
 
+If a PDF is copy protected, ``partition_pdf`` can process the document with the ``"hi_res"`` strategy (which)
+will treat it like an image, but cannot process the document with the ``"fast"`` strategy. If the user
```
qued, line 306:
```suggestion
If a PDF is copy protected, ``partition_pdf`` can process the document with the ``"hi_res"`` strategy (which
will treat it like an image), but cannot process the document with the ``"fast"`` strategy. If the user
```

qued (approved):
LGTM, with one doc formatting suggestion.

PR #518, opened 2023-04-25
feat: `partition_via_api` helper function

MthwRobinson:
### Summary

Adds a helper function for partitioning documents through the hosted REST API. Implemented in response to feedback from users who were interested in more convenient ways to use the hosted API. 

Notes, this function processes a single document and mimics the `partition` and `partition_{doc_type}` functions. The API also accepts multiple files. There will be a follow on PR that calls the API with multiple files.

### Testing

Note the API key here is obviously fake. API keys are not enforced yet but will be soon. For future humans, if the test code doesn't work, please sign up for an API key :-)

```python
from unstructured.partition.api import partition_via_api

elements = partition_via_api(filename=filename, api_key="MY_API_KEY", content_type="message/rfc822")

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.6.2-dev1
+## 0.6.2-dev2
 
 ### Enhancements
 
@@ -6,6 +6,8 @@
   to the hi res strategy when necessary.
 
 ### Features
+
+* Add `partition_via_api` for partitioning documents through the hosted API.
 
 ### Fixes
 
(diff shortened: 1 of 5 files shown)


awalker4:
Awesome! Note that this can replace the `requests` call in ingest - https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L100

cragwolfe:
> Awesome! Note that this can replace the requests call in ingest - https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L100

not sure, actually. why not stick with requests, which is more "pure"?

PR #520, opened 2023-04-26
fix: update `detect_filetype` for JSONs with text/plain MIME type

### LINKED ISSUE #492: auto partitioning a json file with the file= param fails

MthwRobinson:
### Summary

Closes #492. Updates the `detect_filetype` logic to properly detect JSON files even when the MIME type is `text/plain`. Enables to the `partition` function to process JSONs when they are passed in as a file.

### Testing

The output here should show "News Around NOAA"

```python

filename = "test_unstructured_ingest/expected-structured-output/azure-blob-storage/spring-weather.html.json"

    elements = partition(file=f)
elements[0].text

elements[0].text
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.6.2-dev2
+## 0.6.2-dev3
 
 ### Enhancements
 
@@ -12,6 +12,7 @@
 ### Fixes
 
 * Fix how `exceeds_cap_ratio` handles empty (returns `True` instead of `False`)
+* Updates `detect_filetype` to properly detect JSONs when the MIME type is `text/plain`.
 
 ## 0.6.1
 
--- a/example-docs/spring-weather.html.json
+++ b/example-docs/spring-weather.html.json
@@ -1,0 +1,226 @@
+[
+  {
+    "element_id": "41f6e17bf5e9a407fcca74e902f802a0",
+    "text": "News Around NOAA",
+    "type": "Title",
+    "metadata": {
+      "page_number": 1
+    }
+  },
+  {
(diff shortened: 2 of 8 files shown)


qued (approved):
LGTM if CI passes. I'll note that the second output of the testing code provided was already `'News Around NOAA'` without this change (i.e. on `main` branch), but the first output changed after this fix.

cragwolfe (approved):
LGTM. Though, eventually it may be worth refactoring so that there is a not a double .read() of the (potential) json file between detecting the filetype and then partitioning it.

PR #536, opened 2023-05-01
docs: add download badges to README

MthwRobinson:
### Summary

Adds badges to the REAME showing total downloads and downloads per month.

Looks like this in dark mode.

image


qued (approved):
LGTM, I see the download badges in my preview.

PR #538, opened 2023-05-02
ci: only build docs on tags

MthwRobinson:
### Summary

Updates the documentation job to only build on tags. Currently, it builds on every push to `main`, which can result in documentation for new features becoming public before the features themselves are available. Also builds on a `docs-build` branch if we need to do a docs only update between releases. `docs-build` is now a protected branch.

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -5,7 +5,8 @@ name: Sphinx build
 
 on:
   push:
-    branches: [ main ]
+    tags:
+      - '*'
 
 jobs:
   docs:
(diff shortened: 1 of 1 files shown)


.github/workflows/sphinx.yml
```diff
@@ -5,7 +5,9 @@ name: Sphinx build
 
 on:
   push:
-    branches: [ main ]
+    branches: [ docs-build ]
```
qued, line 8: What does this do, is it so there's a special branch to which we can push to force doc rebuilds, but also docs will be rebuilt when a new tag is created?

MthwRobinson: Yeah, I figured if there's a typo in the docs or something like that, we should have a way to update that in between tags. There's a protected branch created for that.

qued (approved):
LGTM!

PR #539, opened 2023-05-02
feat: add `partition_multiple_via_api` function

MthwRobinson:
### Summary

Adds a function for partitioning multiple files through the API. The new function batches multiple files into a single request.

### Testing

```python
  from unstructured.partition.api import partition_multiple_via_api

  filenames = ["example-docs/fake-email.eml", "example-docs/fake.docx"]

  documents = partition_multiple_via_api(filenames=filenames)
```

```python
  from contextlib import ExitStack

  from unstructured.partition.api import partition_multiple_via_api

  filenames = ["example-docs/fake-email.eml", "example-docs/fake.docx"]
  files = [open(filename, "rb") for filename in filenames]

  with ExitStack() as stack:
      files = [stack.enter_context(open(filename, "rb")) for filename in filenames]
      documents = partition_multiple_via_api(files=files, file_filenames=filenames)
```

### DIFF
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -166,7 +166,15 @@ Examples:
 ------------------------------
 
 ``partition_multiple_via_api`` is similar to ``partition_via_api``, but allows you to partition
-multiple documents in a single REST API call. The result has the type ``List[List[Element]]``.
+multiple documents in a single REST API call. The result has the type ``List[List[Element]]``,
+for example:
+
+.. code:: python
+
+  [
+    [NarrativeText("Narrative!"), Title("Title!")],
+    [NarrativeText("Narrative!"), Title("Title!")]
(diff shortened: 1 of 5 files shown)


docs/source/bricks.rst
```diff
@@ -162,6 +162,36 @@ Examples:
     elements = partition_via_api(file=f, file_filename=filename, api_key="MY_API_KEY")
 
 
+``partition_multiple_via_api``
+------------------------------
+
+``partition_multiple_via_api`` is similar to ``partition_via_api``, but allows you to partition
+multiple documents in a single REST API call. The result has the type ``List[List[Element]]``.
```
cragwolfe, line 169: it would be worth mentioning what the return value looks like if one or more of the calls to the API fails.

PR #542, opened 2023-05-03
fix: group together text from the same bounding box in `partition_pdf` with fast strategy

### LINKED ISSUE #541: Check the `partition_pdf` output on the attached PDF
The attached PDFs has document that are broken up subotimally (i.e. a title that spans two lines getting broken up into two separate elements) when using the fast strategy. The goal of this issue is to take a look at the output for this document, do a quality check, and see if there are any improvements we can make to the fast strategy output.

[chevron-sustainability-report-2022.pdf](https://github.com/Unstructured-IO/unstructured/files/11384775/chevron-sustainability-report-2022.pdf)

MthwRobinson:
### Summary

Closes #541. Updates the text extraction logic in `partition_pdf` using the `"fast"` strategy to group together text that is in the same bounding box. Helps account for text such as in the example below that contains multiple line breaks but really belongs in the same grouping.

image

### Testing

If you run the following code:

```python

filename = "example-docs/chevron-page.pdf"

for element in elements[:4]:
    print(f"<<{element.category}>> {element}")
```

You should see output that looks like the following:

```python
<<Title>> eastern mediterranean
<<NarrativeText>> We’re investing to grow in the Eastern Mediterranean, offshore Israel and Egypt. Natural gas production in the region is helping to reduce Israel’s greenhouse gas emissions and improve air quality. In addition to supplying Israel, Chevron-operated Tamar and Leviathan fields are exporting natural gas to neighboring Jordan and Egypt.
<<UncategorizedText>> 2.5
<<Title>> kilograms CO₂e/boe carbon intensity from our Eastern Mediterranean operations in 2022
```


qued (approved):
LGTM!

PR #551, opened 2023-05-05
chore: pin inference version, bump deps, and update openssl

ryannikolaidis:
- pin unstructured-inference to 0.4.4 in setup.py
- bump dependencies
- update image build to install openssl 1.1.1 and python to compile with it
- bonus: disable a labelstudio test that relies on vcr which fails with usrllib3==2.0.2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,7 @@ RUN yum -y update && \
   make && \
   make install && cd .. && \
   ldconfig && \
+  rm -rf openssl-1.1.1k && rm openssl-1.1.1k.tar.gz && \
   $sudo yum -y remove perl-core pcre-devel && \
   yum clean all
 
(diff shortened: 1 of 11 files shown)


cragwolfe (approved):

Dockerfile
```diff
@@ -46,13 +46,31 @@ RUN set -ex && \
...
+  yum install -y perl-core pcre-devel && \
+  wget https://ftp.openssl.org/source/openssl-1.1.1k.tar.gz && \
+  tar -xzvf openssl-1.1.1k.tar.gz && \
+  cd openssl-1.1.1k && \
+  ./config shared --prefix=/usr/local/ssl --openssldir=/usr/local/ssl && \
+  make && \
+  make install && cd .. && \
+  $sudo yum -y remove perl-core pcre-devel && \
+  ldconfig
+
```
tabossert, line 60: could you add the yum cleanup steps here.

ryannikolaidis: yep, good call!

tabossert (approved):

PR #552, opened 2023-05-05
Fixes issue where detectron2 would not install on OSX

tabossert:
Tested on Apple silicon based MacBook Pro.  This installs tensorboard which is required on OSX and arm based cpu’s for detectron2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PACKAGE_NAME := unstructured
 PIP_VERSION := 22.2.1
 CURRENT_DIR := $(shell pwd)
+OS := $(shell uname)
 
 
 .PHONY: help
@@ -87,8 +88,14 @@ install-ingest-wikipedia:
 install-unstructured-inference:
 	python3 -m pip install -r requirements/local-inference.txt
 
+.PHONY: install-tensorboard
+install-tensorboard:
+	@if [ ${OS} = "Darwin" ]; then\
+		python3 -m pip install tensorboard;\
+	fi
+
 .PHONY: install-detectron2
-install-detectron2:
+install-detectron2: install-tensorboard
 	python3 -m pip install "detectron2@git+https://github.com/facebookresearch/detectron2.git@e2ce8dc#egg=detectron2"
 
 ## install-local-inference: installs requirements for local inference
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -75,7 +75,7 @@ installation. NOTE: We do not currently support python 3.11, please use an older
     - `libreoffice` (MS Office docs)
 - If you are parsing PDFs, run the following to install the `detectron2` model, which
   `unstructured` uses for layout detection:
-    - `pip install "detectron2@git+https://github.com/facebookresearch/detectron2.git@e2ce8dc#egg=detectron2"`
+    - make install-detectron2
```
ryannikolaidis, line 78: ahh, like funneling them to the make command and much cleaner than having that commit-specific url duped.

Makefile
```diff
@@ -87,8 +88,14 @@ install-ingest-wikipedia:
 install-unstructured-inference:
 	python3 -m pip install -r requirements/local-inference.txt
 
+.PHONY: install-tensorboard
+install-tensorboard:
+	@if [ ${OS} = "Darwin" ]; then\
+		python3 -m pip install tensorboard;\
+	fi
```
ryannikolaidis, line 95: I think technically this is breaking on any aarch64 device, but probably Linux ARM are mostly raspberry pi folks atm?

ryannikolaidis:
testing with a fresh virtual env now..

Makefile
tabossert: Probably, and there are Linux arm based instances on AWS, I have a better way, will push it up in a minute

ryannikolaidis (approved):
lgtm! tested with:

>>> pyenv virtualenv 3.8.15 unstructured-test-detectron-fix
>>> pyenv activate unstructured-test-detectron-fix
>>> make install-local-inference

Makefile
tabossert: Ok updated to use CPU arch instead of os

README.md
cragwolfe, line 78: for this part of the quickstart, the intention is the commands the user need to run *without* cloning this repo (which one would need to do `make install-___` ).

cragwolfe (changes requested):
Nit on the README: should avoid `make` commands in the "unstructured as a library" case.

README.md
tabossert: Makes sense, pushed those changes

cragwolfe (approved):

PR #569, opened 2023-05-10
chore: add support for SpooledTemporaryFiles

ryannikolaidis:
Add support for SpooledTemporaryFile file arguments when partitioning pdf, docx, and pptx. These all rely on various functionality not supported by SpooledTemporaryFile. This PR adds a `spooled_to_bytes_io_if_needed` function so that we can change filetypes in place and get the functionality expected by these partition modules.

### DIFF
--- a/test_unstructured/partition/test_pdf.py
+++ b/test_unstructured/partition/test_pdf.py
@@ -1,4 +1,5 @@
 import os
+from tempfile import SpooledTemporaryFile
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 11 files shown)


test_unstructured/partition/test_pdf.py
```diff
@@ -172,6 +173,24 @@ def test_partition_pdf(url, api_called, local_called, monkeypatch):
...
+    filename="example-docs/layout-parser-paper-fast.pdf",
+):
+    # Test that the partition_pdf function can handle a SpooledTemporaryFile
+    with open(filename, "rb") as test_file:
+        spooled_temp_file = SpooledTemporaryFile()
+        spooled_temp_file.write(test_file.read())
+        spooled_temp_file.seek(0)
+        result = pdf.partition_pdf(file=spooled_temp_file, strategy=strategy)
+        # validate that the result is a non-empty list of dicts
+        assert len(result) > 0
```
cragwolfe, line 191: only nit here is i think this could be `> 10`

cragwolfe (approved):
LGTM with one nit.
also could be worth a release.

test_unstructured/partition/test_pdf.py
ryannikolaidis: yep!

PR #574, opened 2023-05-11
fix: check json and eml decode ignore error

Mestrace:
First time contributor. Please help review if there are any problems with the style.

This commit fixed the problem when reading the first 4096 bytes results in a chunked character at the end. Decoding these characters will result in `UnicodeDecodeError`. 

In my case, I have a character that is `b'\xe2\x80\x94'` (non-standard dash connector) and reading the first 4096 byte will not read the last `\x94`. Since these two functions only checking json and eml, ignoring the character will fix the problem while not affecting the core logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.6.6-dev1
+
+### Enhancements
+
+### Features
+
+### Fixes
+
+* fix: fileutils/file_type check json and eml decode ignore error
+
 ## 0.6.6-dev0
 
 ### Enhancements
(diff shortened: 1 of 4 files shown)


cragwolfe:
hi @Mestrace , thank you for the contribution! please update https://github.com/Unstructured-IO/unstructured/blob/main/CHANGELOG.md , incrementing the `-devX` version on the first line and adding a bullet for the bug fix. then, please update https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/__version__.py to match.

Mestrace:
@cragwolfe bumped the version to 0.6.6-dev1 as suggested.

cragwolfe (approved):
LGTM!

PR #575, opened 2023-05-11
enhancement: add method for getting datetime; cleanup filename attribute

MthwRobinson:
### Summary

Updates the `ElementMetadata` class to include a `get_date()` method that converts the string `date` attribute to a `datetime` object. The `date` object will remain an ISO 8601 formatted string to support JSON serialization. `get_date()` will help make it easier to perform time-based operations on document elements.

Additionally, updates the `filename` attribute to only include the base filename (i.e. `fake.docx` instead of `../example-docs/fake.docx`) to make the metadata field cleaner. The outputs will change relative to previous versions but the interfaces remain backward compatible.

### Testing

```python

# Should be "fake-email.eml"
elements[0].metadata.filename

elements[0].metadata.get_date()
```


qued (approved):
LGTM!

qued (approved):
LGTM!

PR #581, opened 2023-05-12
fix: make `pytesseract` a function level import

### LINKED ISSUE #562: bug/handle case where `pytesseract` is not installed in `partition.pdf`
**Describe the bug**
Currently, you get a `ModuleNotFound` error if `pytesseract` is not installed in `partition.pdf`, even though there are functions in that module that should work without `pytesseract`.

**To Reproduce**
Run `pip uninstall pytesseract`, then

```python
import unstructured.partition.pdf as pdf
```

**Expected behavior**
The module should import successfully if `pytesseract` is not installed. 

 - `unstructured==0.6.3`
 - Python version 3.8

MthwRobinson:
### Summary

Closes #562. Makes `pytesseract` a function level import so you can use strategies that don't require `pytesseract` without `pytesseract` being installed. Also added the `required_dependencies` decorator for the `"hi_res"` and `"ocr_only"` strategies.

This PR also includes a small docs formatting fix.

### Testing

If you uninstall `pytesseract`, the following should work:

```python

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## 0.6.7-dev0
+
+### Enhancements
+
+### Features
+
+### Fixes
+
+* Makes `pytesseract` a function level import in `partition_pdf` so you can use the `"fast"`
+  or `"hi_res"` strategies if `pytesseract` is not installed. Also adds the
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM!

PR #585, opened 2023-05-15
enhancement: add `file_directory` to element metadata

MthwRobinson:
### Summary

Adds a `file_directory` attribute alongside `filename` in the element metadata.

### Testing

```python

elements = partition_pdf("../unstructured/example-docs/layout-parser-paper-fast.pdf")

# Should be layout-parser-paper-fast.pdf
print(elements[0].metadata.filename)

# Should be ../unstructured/example-docs
print(elements[0].metadata.file_directory)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
-## 0.6.7-dev0
+## 0.6.7-dev1
 
 ### Enhancements
+
+* Add `file_directory` to metadata
 
 ### Features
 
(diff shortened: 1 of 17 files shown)


qued (approved):
LGTM as far as the code and the testing. I'm a little worried that this seemed too easy and it took me a few minutes to understand why it worked with no changes anywhere else. My only suggestion would be maybe some tests that use the `partition_x` path rather than the `partition` path? Clearly that's working right now but if it broke for only that path in the future we might miss it.

MthwRobinson:
Thanks, added some additional tests using the `partition_x` code paths.

PR #590, opened 2023-05-16
docs: Updated the installation command

nickscamara:
Seems like there was a small typo on the main installation command.


awalker4 (approved):
Awesome, thank you!

PR #591, opened 2023-05-16
refactor: update detect_filetype() to use hashmap for mime type return

marcusyatim:
This PR helps to refactor and clean up `unstructured/file_utils/filetype.py`. I really like the STR_TO_FILETYPE hashmap and feel that we could use it for the mime type return in `detect_filetype()` method. This will allow us to remove a lot of duplicates and unnecessary if-else statements. And as such, hopefully this refactor can make the script shorter and cleaner.

Let me know if you like this change and I am happy to receive any feedback and/or edits :) 

Will update the CHANGELOG towards the end after receiving feedback.

Thanks and looking forward to contributing!


MthwRobinson:
@marcusyatim - Thanks for this refactor! This is a great update. Will approve this once CI passes and the changelog is updated.

MthwRobinson:
You can run `make tidy` to fix that error in the linting job.

marcusyatim:
Thanks for the review @MthwRobinson ! Have made some changes. Hopefully this works :)

PR #599, opened 2023-05-17
chore: refactor for ingest standard_config options

ryannikolaidis:
This moves the common config options out of the connector-specific configs and into a new `standard_config`. This leaves all the command line options as-is and only updates organization under the hood.


unstructured/ingest/interfaces.py
```diff
@@ -11,12 +12,41 @@
...
+    partition_endpoint: str = "https://api.unstructured.io/general/v0/general"
+    preserve_downloads: bool = False
+    re_download: bool = False
+
+
+class BaseConnectorConfig(ABC):
+    """Abstract definition on which to define connector-specific attributes."""
+
+
+@dataclass
```
ryannikolaidis, line 38: we make the git connectors dataclasses. to respect that (and assuming others might do the same in the future) we can make the base a dataclass so that way standard_config doesn't need to be redefined on every child (otherwise the init function won't have it as an argument).

ryannikolaidis: alternatively we could just let the git connectors redefine it and not make the base a dataclass

unstructured/ingest/interfaces.py
```diff
@@ -11,12 +12,41 @@
...
-        """Expects a config object that implements BaseConnectorConfig."""
-        pass
+    standard_config: StandardConnectorConfig
+    config: BaseConnectorConfig
+
+    def __init__(self, standard_config: StandardConnectorConfig, config: BaseConnectorConfig):
+        """Expects a standard_config object that implements StandardConnectorConfig
+        and config object that implements BaseConnectorConfig."""
+        self.standard_config = standard_config
+        self.config = config
```
ryannikolaidis, line 49: I think we could potentially ditch this given this is a dataclass. though maybe helpful to see the definitions here and have a space carved  out if we want to add base init functionality in the future. in the subclasses, adopted the pattern of calling `super().__init__(standard_config, config)` rather than duplicating `self.standard_config = standard_config` and `self.config = config` everywhere, but this also could mean that we can leverage in the future if we need anything to happen on init across all connectors.

unstructured/ingest/interfaces.py
```diff
@@ -11,12 +12,41 @@
...
 
 
+@dataclass
+class StandardConnectorConfig:
+    """Common set of config options passed to all connectors."""
+
+    # where raw documents are stored for processing, and then removed if not preserve_downloads
+    download_dir: str
+    # where to write structured data outputs
+    output_dir: str
```
ryannikolaidis, line 22: these being required probably made more sense when they were standard but could be omitted. maybe we make this Optional[str] = None so you don't have to define it. also noting, these are in the weird "standard" place where they are standard to everything BUT local (local is actually the biggest outlier). tempted to carve out as some sort of "output" mixin, but this would literally just be carving around the local connector afaict.

cragwolfe: yeah, seems fine as is right now, but could definitely revisit.

cragwolfe (approved):
LGTM! 🙏

PR #608, opened 2023-05-18
Issue/unicode error

christinestraub:
This PR adds functionality to try other common encodings if an error related to the encoding is raised and the user has not specified an encoding.

Files changed:
> unstructured>file_utils>encoding.py (created)
> unstructured>partition>text.py (updated)
> test_unstructured>partition>test_text.py (updated)

### Testing

```

# try with filename
elements = partition_text(filename="example-docs/fake-text-utf-16.txt")

# try with file-like object
with open("example-docs/fake-text-utf-16.txt") as f:
    elements = partition_text(file=f)

with open("example-docs/fake-text-utf-16.txt", "rb") as f:
    elements = partition_text(file=f)
```

```
@pytest.mark.parametrize(
    "filename",
    ["fake-text-utf-16.txt", "fake-text-utf-16-le.txt", "fake-text-utf-32.txt"],
)
def test_partition_text_from_filename_without_encoding(filename):
    elements = partition_text(filename=filename)
    assert len(elements) > 0
    assert elements == EXPECTED_OUTPUT
```


cragwolfe (approved):
It's a bit awkward that [auto.py](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py) still has the default encoding of `utf-8`, though that applies to other document types besides text.

What are your thoughts about the other file types? (html, pdf, etc)

For `partition_text`, LGTM!

cragwolfe (changes requested):
Missed a changelog issue.

If cutting a new release, the next version is actually 0.6.9.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.6.10
+
+### Enhancements
+
+### Features
+
+### Fixes
+
+* Adds functionality to try other common encodings if an error related to the encoding is raised and the user has not specified an encoding.
```
cragwolfe, line 9: so, two options on this changelog. cut a new pypi release at `0.6.9`, or just increment `0.6.9-dev1` to `0.6.9-dev2` and add the bullet point inline. in either case, the block under what is currently line 11 is the block to add the bullet item.

christinestraub: > It's a bit awkward that [auto.py](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py) still has the default encoding of utf-8, though that applies to other document types besides text.

> What are your thoughts about the other file types? (html, pdf, etc)


For now, the auto.py applies the default encoding of utf-8 to 5 types of files (txt, pdf, eml, html, xml)

`txt`: already implemented

`pdf`: The encoding kwarg is passed to partition_pdf, but is not used. (https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L267) so I think the encoding kwarg should be removed from partition_pdf. 

`eml`: The .eml files have encoding information in the metadata. The encoding information is typically specified in the Content-Type header of the email message. (https://github.com/Unstructured-IO/unstructured/blob/main/example-docs/email-with-image.eml#L13). 

I think the encoding information should be parsed from the metadata if the user has not specified an encoding.

`html`: two options for filename and file-like object, except url
	option1: parse encoding information from html file metadata (<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">)
	option2: implement same pattern as partition_text

	factor to consider: whether to accept html files don't have charset information in the metadata

`xml`: similar to hmtl
	option1: parse encoding information from html file metadata (<?xml version="1.0" encoding="UTF-8"?>)
	option2: implement same pattern as partition_text

	factor to consider: whether to accept xml files don't have encoding information in the metadata

christinestraub: I'm not sure if I should do all the work for the 5 types of files in the same 'PR' or if I should create separate 'PRs'.

christinestraub:
> It's a bit awkward that [auto.py](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py) still has the default encoding of utf-8, though that applies to other document types besides text.

> What are your thoughts about the other file types? (html, pdf, etc)


For now, the auto.py applies the default encoding of utf-8 to 5 types of files (txt, pdf, eml, html, xml)

`txt`: already implemented

`pdf`: The encoding kwarg is passed to partition_pdf, but is not used. (https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L267) so I think the encoding kwarg should be removed from partition_pdf. 

`eml`: The .eml files have encoding information in the metadata. The encoding information is typically specified in the Content-Type header of the email message. (https://github.com/Unstructured-IO/unstructured/blob/main/example-docs/email-with-image.eml#L13). 

I think the encoding information should be parsed from the metadata if the user has not specified an encoding.

`html`: two options for filename and file-like object, except url
	option1: parse encoding information from html file metadata (<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">)
	option2: implement same pattern as partition_text

	factor to consider: whether to accept html files don't have charset information in the metadata

`xml`: similar to hmtl
	option1: parse encoding information from html file metadata (<?xml version="1.0" encoding="UTF-8"?>)
	option2: implement same pattern as partition_text

	factor to consider: whether to accept xml files don't have encoding information in the metadata
	
I'm not sure if I should do all the work for the 5 types of files in the same PR or if I should create separate PRs.

cragwolfe (approved):
LGTM!

Issue #600, opened 2023-05-17
Add `page_name` to metadata for Excel documents


MthwRobinson:
Currently we're tracking `page_number` as the sheet number for Excel documents. However, Excel documents also have a sheet name that is informative. The goal of this issue is to add `page_name` to `ElementMetadata` and track the sheet name as `page_name` for Excel documents.

PR #609, opened 2023-05-18
feat: add page_name to metadata for Excel documents

meirdev:
Added tracking for the sheet name of Excels documents as a `page_name` field in the `ElementMetadata` class.

#600


MthwRobinson (approved):
LGTM once the linting job passes - thanks for the contribution!

PR #614, opened 2023-05-18
Chore(ingest) : add tests on PDFs with fast strategy

yuming-long:
DRAFT


cragwolfe (approved):

unstructured/partition/pdf.py
```diff
@@ -286,7 +287,16 @@ def _process_pdfminer_pages(
                 element = element_from_text(_text)
                 element.coordinates = ((x1, y1), (x1, y2), (x2, y2), (x2, y1))
                 element.metadata = metadata
-                elements.append(element)
+                page_elements.append(element)
+
+        sorted_page_elements = sorted(
+            page_elements,
+            key=lambda el: (
+                el.coordinates[0][1] if el.coordinates else float("inf"),
```
cragwolfe, line 295: i think in the future we'd want to not strictly sort by Y coordinate if 2 points are very close to each other on the Y.

E.g., if there are 2 horizontally aligned blocks where the right block begins one pixel higher than the left, it would be ordered before the left block.

This can be tracked in a follow-on issue, though.

qued: Seems related to the Slack conversation [here](https://unstructuredw-kbe4326.slack.com/archives/C057R3F8F7A/p1686172773173559)

qued (approved):
LGTM, no changes when I ran the ingest update.

PR #619, opened 2023-05-19
feat: add `partition_csv` function

### LINKED ISSUE #570: feat: Add support for CSV filetype
Some users want to process CSV filetypes (example use case [here](https://github.com/hwchase17/langchain/pull/3906)) and we don't currently support it.

The easiest implementation would be to format it as one element per cell. Alternatively if `extract_tables` is `True` we could possibly extract a single element (the whole csv) with the table structure.

MthwRobinson:
### Summary 

Closes #570. Adds `partition_csv` for processing CSV files.

### Testing

```python
from unstructured.partition.csv import partition_csv

elements = partition_csv(filename="example-docs/stanley-cups.csv")
```


unstructured/partition/csv.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Parameters
+    ----------
+    filename
+        A string defining the target filename path.
+    file
+        A file-like object using "rb" mode --> open(filename, "rb").
+    metadata_filename
+        The filename to use for the metadata. Relevant because partition_doc converts the
+        document to .csv before partition. We want the original source filename in the
```
qued, line 29: I think this is copy pasted from `partition_docx` and isn't actually true.

qued (approved):
LGTM other than one doc issue!

unstructured/partition/csv.py
MthwRobinson: Oooops! Fixed that here and for `partition_xlsx` too.

PR #620, opened 2023-05-19
fix: add more mime types for csv

MthwRobinson:
### Summary

Adds more MIME types for CSVs because not every system uses the standard `text/csv`.


unstructured/file_utils/filetype.py
```diff
@@ -93,6 +93,11 @@ def __lt__(self, other):
     "image/jpeg": FileType.JPG,
     "image/png": FileType.PNG,
     "text/plain": FileType.TXT,
+    "text/x-csv": FileType.CSV,
+    "application/csv": FileType.CSV,
+    "application/x-csv": FileType.CSV,
+    "text/comma-separated-values": FileType.CSV,
+    "text/x-comma-separated-values": FileType.CSV,
```
qued, line 100: These will need to be excluded when we reverse the dict for `FILETYPE_TO_MIMETYPE`.

MthwRobinson: Excluded! Though they're sad they didn't get an invite

qued (approved):
LGTM!

Issue #468, opened 2023-04-11
Add configurable retry logic to biomed connector


cragwolfe:
The biomed connector encounters spurious ftp errors. Update the connector to retry (after exponential backoff) on failure, but make the number of retries, max retry time, and decay configurable.

Relates to https://github.com/Unstructured-IO/unstructured/pull/467

PR #623, opened 2023-05-19
feat(biomed connector): added additional params (#468)

kravetsmic:
## Summary

Added max retries, max request time with backoff and decay for Biomed connector.

### New Parameters

<details open>
<summary>Max retries</summary>

Not required parameter. Accepts number which indicates the amount of attempts to connect to Biomed.
Default value: `1`

```sh
--biomed-max-retries 5
```
</details>

<details open>
<summary>Max request time</summary>

Not required parameter. Accepts seconds which indicates request timeout for the connection.
Default value: `45`

```sh
--biomed-max-request-time 30
```
</details>

<details open>
<summary>Decay</summary>

Not required parameter. Accepts float which indicates a factor to multiply the delay between retries.
Default value: `0.3`

```sh
--biomed-decay .3
```
</details>

---

Closes issue #468


cragwolfe (commented):
please remove the `exit 0` in test_unstructured_ingest/test-ingest-biomed-api.sh so we can see if it passes CI. (actually fixtures will probably need to be updated too, see https://github.com/Unstructured-IO/unstructured/blob/main/scripts/ingest-test-fixtures-update.sh

cragwolfe (commented):
LGTM once CI is clean.

Please add an entry to the CHANGELOG.md and increment the -devX version on the first line (implies also incrementing the devX version of unstructured/\__version__.py)

cragwolfe:
@kravetsmic , a few things:

2. [test-ingest-biomed-path.sh](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-biomed-path.sh) should be updated in a similar manner to `-api.sh`.
2. it looks like the biomed .json fixtures need to be updated which is not surprising given these PDF's have been skipped for awhile. 
3. you'll need to run https://github.com/Unstructured-IO/unstructured/blob/main/scripts/ingest-test-fixtures-update.sh on intel-compatible arch to update the fixtures so that the ingest tests can pass.

cragwolfe (approved):

PR #627, opened 2023-05-21
Update filetype.py

wesleysanjose:
I'd like to parse the .log file of the ubuntu system, but found .log file is not supported. I think the quickest fix is to let it to be processed same as txt file.


MthwRobinson (approved):
LGTM - thanks!

PR #630, opened 2023-05-22
Chore: Support epub tests in docker image

yuming-long:
### Summary

Can support more filetype: `epub`, `odt` and `rtf` since the new base image now have a newer version of pandoc

### Test
I would suggest to run this test on docker image/ec2 instance since the pandoc might not be outdated locally.
```

file_path = "example-docs/winter-sports.epub"
partition(filename=file_path)

file_path = "example-docs/fake.odt"
partition(filename=file_path)

file_path = "example-docs/fake-doc.rtf"
partition(filename=file_path)
```


awalker4:
Looks good! Can you paste a quick snippet for testing? I suppose all we'd need to do is try to call partition() with these file types, correct?

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:experimental
 
-FROM quay.io/unstructured-io/base-images:centos7.9
+FROM quay.io/unstructured-io/base-images:centos7.9-1
```
awalker4, line 3: Does this correspond to a change in [this repo](https://github.com/Unstructured-IO/base-images)? I would think we don't need a `install-pandoc` target in here if the installation happens in the base image.

yuming-long: the centos 7.9-1 is supposed to be built once this https://github.com/Unstructured-IO/base-images/pull/1 is merged by ci, but it might fail because of how long it takes on the slow GHA runners (and i believe Trevor will have a solution later). And currently, the new image is built on large ec2 instance manually. you can find it at:https://quay.io/repository/unstructured-io/base-images?tab=tags&tag=centos7.9

yuming-long: the `install-pandoc` is for `test-unit` on CI Ubuntu system, the command in ci before is also out-dated:https://quay.io/repository/unstructured-io/base-images?tab=tags&tag=centos7.9

awalker4: Ah, makes sense!

yuming-long: i know it sounds repeating the installation here but since the pandoc is installed in base-image, i could't share it here with ci

test_unstructured/partition/test_auto.py
```diff
@@ -432,15 +429,13 @@ def test_auto_with_page_breaks():
     assert PageBreak() in elements
 
 
-@pytest.mark.skipif(is_in_docker, reason="Skipping this test in Docker container")
 def test_auto_partition_epub_from_filename():
```
yuming-long, line 432: @awalker4 I don't think my tests in instruction would make any difference, here is an unittest with `partition`

awalker4 (approved):
Yep, passing tests should verify it. LGTM

PR #632, opened 2023-05-23
feat: add xls support

cragwolfe:
Add support for older .XLS files from the partition function in `unstructured.partition.auto`.

Note, this should also work on the centos7 unstructured image (with the requirements/*txt updates in this PR).


Makefile
```diff
@@ -123,7 +123,7 @@ pip-compile:
 	cp requirements/build.txt docs/requirements.txt
 	pip-compile --upgrade --extra=s3        --output-file=requirements/ingest-s3.txt        requirements/base.txt setup.py
 	pip-compile --upgrade --extra=azure     --output-file=requirements/ingest-azure.txt     requirements/base.txt setup.py
-	pip-compile --upgrade --extra=discord   --output-file=requirements/ingest-azure.txt     requirements/base.txt setup.py
+	pip-compile --upgrade --extra=discord   --output-file=requirements/ingest-discord.txt   requirements/base.txt setup.py
```
ryannikolaidis, line 126: woops, nice catch!

Issue #631, opened 2023-05-23
bug/mismatched library versions in some requirements/*.txt files


cragwolfe:
**Describe the bug**
Some extras libraries potentially have different versions of libraries installed.

For instance, currently, requirements/ingest-s3.txt has 
[requests==2.28.2](https://github.com/Unstructured-IO/unstructured/blob/a78719666a745fb63a389e2c82cf5fa431109366/requirements/ingest-s3.txt#L190), as do some others, while other .txt files have 2.30.2

```
$ grep 'requests==' *txt
base.txt:requests==2.30.0
build.txt:requests==2.30.0
huggingface.txt:requests==2.30.0
ingest-azure.txt:requests==2.28.2
ingest-discord.txt:requests==2.28.2
ingest-github.txt:requests==2.28.2
ingest-gitlab.txt:requests==2.28.2
ingest-google-drive.txt:requests==2.28.2
ingest-reddit.txt:requests==2.28.2
ingest-s3.txt:requests==2.28.2
ingest-slack.txt:requests==2.28.2
ingest-wikipedia.txt:requests==2.28.2
local-inference.txt:requests==2.30.0
test.txt:requests==2.30.0
test.txt:types-requests==2.30.0.0
```

This leads to potentially unpredictable behaviour depending what extras are installed

There should be a CI job that fails for such mismatches to prevent these inconsistencies from occurring in the first place. (that's the enhancement)

* Note: it is conceivable there could be 2 mutually exclusive extras that require different versions that should not be installed at the same time, but this doesn't currently exist in the library.

qued:
Closing as addressed by #636

PR #636, opened 2023-05-23
build(deps): avoid version conflicts

qued:
Addresses [#631](https://github.com/Unstructured-IO/unstructured/issues/631).

- Uses constraints to keep dependency versions more consistent.
- Moves all dependencies to `.in` files which are then ingested by `setup.py`.
- Adds script to check consistency of all extras.
- Adds consistency check to CI.

I should note that while it shouldn't be possible to cause a conflict between `base.txt` and any of the extras (because `base.txt` constrains all the extras) it is possible to get a conflict between two of the extras files. There are ways of trying to avoid that (like constraining each file by all the files that have already been processed before it in the order given in the `make pip-compile` target) but the ones I could think of seemed a little overwrought, and come with problems of their own. If a conflict arises, it should be flagged by CI or locally with `make check-deps`. When/if that happens, you can resolve the conflict by adding appropriate global constraints in `requirements/constraints.txt`.

Also note that if `fileA.in` is constrained by `fileB.txt`, then `fileB.in` should be compiled _before_ `fileA.in` in the `make pip-compile` target. Otherwise `fileA.in` will be compiled with the old version of `fileB.txt` which can cause conflicts or keep dependencies from being updated properly.

#### Testing:
- Verify `make check-deps` says deps are consistent.
- Change line 59 of `requirements/ingest-s3.txt` to:
```
urllib3==1.27.1
```
- Verify `make check-deps` produces an error.


cragwolfe:
Flagging @ryannikolaidis in case he is aware of any potential downstream issues with the urllib<2 pin. But, really great to see the dep issues sorted out and a way of preventing the mismatches in the future. Passing CI with Dockerfile is also promising 🤞 .

ryannikolaidis:
> Flagging @ryannikolaidis in case he is aware of any potential downstream issues with the urllib<2 pin. But, really great to see the dep issues sorted out and a way of preventing the mismatches in the future. Passing CI with Dockerfile is also promising 🤞 .

Thanks! Nope, I don't think we had anything explicitly forcing our hand. iirc the main push to urllib3 was because the latest for a bunch of our dependencies were requiring urllib3>2. As long as we're okay with those not getting bumps, then I don't think there are any other immediate concerns.

Thanks for this @qued! 🙏

cragwolfe (approved):
LGTM!

PR #638, opened 2023-05-25
test: add benchmark script

ryannikolaidis:
Adds a script which provides a method for reproducibly benchmarking partitioning performance across various document types. The script allows the user to conduct the test either directly on the calling host or in a Docker container built at test time. Tests can be run a number of iterations (NUM_ITERATIONS) while tests specified in SLOW_FILES are only ever run once. For tracking purposes, results are optionally stored in S3 (with the PUBLISH_RESULTS option).

## Environment Variables:
- DOCKER_TEST: Set to "true" to run benchmark inside a Docker container (default: "false")
- NUM_ITERATIONS: Number of iterations for benchmark (e.g., 100) (default: 3)
- INSTANCE_TYPE: Type of benchmark instance (e.g., "c5.xlarge") (default: "unspecified")
- PUBLISH_RESULTS: Set to "true" to publish results to S3 bucket (default: "false")

## Example Usage

```
PUBLISH_RESULTS=true INSTANCE_TYPE=t3.large NUM_ITERATIONS=2 DOCKER_TEST=true ./scripts/performance/benchmark.sh
```

---

bonus: removes syncing from S3 from profiling code
bonus: mimetype detection now follows symlinks to determine type


scripts/performance/benchmark-local.sh
```diff
@@ -0,0 +1,63 @@
...
+    total_execution_time=0
+    for ((i = 1; i <= num_iterations; i++)); do
+        echo "Iteration $i"
+        strategy="fast"
+        if [[ " ${HI_RES_STRATEGY_FILES[*]} " =~ $(basename "$file") ]]; then
+            echo "Testing with hi_res strategy"
+            strategy="hi_res"
+        fi
+        
+        if ! response=$( { time process_file "$file" "$strategy"; } 2>&1 ); then
```
cragwolfe, line 45: with this method, a new python process is spun up and many modules are imported as a result of calling `partition` (the warmup call above does force the issue for downloading needed modules, but there is still overhead for each fresh process after that).

the best way to reduce the noise is to make one call to `partition` call for the given filetype before measuring the next one, which implies doing more in the python execution side. E.g., the python script could return the time for the call after in-process warmup. The python process also could do the N iterations, rather doing one per process.

this may not be a terrible refactor, but it could also get punted to a future PR.

ryannikolaidis: yep, good call! getting this in. I think since we'll want to have some initial benchmarks to track progress against, worth having this in place now. from quick testing, module loading etc is clearly a a non-insignificant chunk of time spent.

ryannikolaidis: okay, should now have changes as per both of your recommendations. it will attempt to find an inexpensive warmup file of the same type or otherwise just use a pass on the test doc as the warm up. it also does all the iterations and timing inside of python.

scripts/performance/warmup.pdf
cragwolfe, line 1: nit: could the docs in performance/ be symbolic links to docs in example-docs/ ?

cragwolfe:
Running on mac with `DOCKER_TEST=false` worked great.

Running on intel hardware with `DOCKER_TEST=true`:

```
PIP_VERSION=23.1.2 DOCKER_IMAGE_NAME=unstructured:perf-test ./scripts/docker-build.sh
#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 1.68kB done
#1 DONE 0.0s

<snip>

#5 [internal] load metadata for quay.io/unstructured-io/base-images:centos7.9-1
#5 DONE 0.5s

<snip>

#16 naming to docker.io/library/unstructured:perf-test done
#16 DONE 0.0s

#17 exporting cache
#17 preparing build cache for export done
#17 DONE 0.0s
Starting benchmark test...
Testing file: *
  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/scripts/performance/time_partition.py", line 37, in <module>
    warm_up_process(filename)
  File "/home/scripts/performance/time_partition.py", line 15, in warm_up_process
    partition(filename, strategy="fast")
  File "/home/unstructured/partition/auto.py", line 206, in partition
    raise ValueError(f"{msg}. The {filetype} file type is not supported in partition.")
ValueError: Invalid file /home/scripts/performance/docs/*. The FileType.UNK file type is not supported in partition.
error: 
Error: Results filename value not found in the benchmark logs.
```

Wonder if it is a magic issue that may longer exist with Rocky linux?

scripts/performance/warmup.pdf
ryannikolaidis: ohh yea, totally. when I started wasn't sure how many net-new files I was going to be adding and how large they would be, but I don't think the damage is too bad just adding the new ones to example-docs and then sym linking everything

ryannikolaidis:
> Wonder if it is a magic issue that may longer exist with a Rocky linux base image?
Okay, merged up to latest (including Rocky Linux)  and tested with DOCKER_TEST=true and everything passed, if you want to give it a shot again?

scripts/performance/warmup.pdf
ryannikolaidis: okay, should all be symlinked now. ran into an issue where mime type detection was failing with symlinks, so there's a small bonus to resolve that issue.

cragwolfe (approved):
Nice! 🙏

PR #639, opened 2023-05-25
Issue/encoding error eml

christinestraub:
This PR adds functionality to try other common encodings for email (`.eml`) files if an error related to the encoding is raised and the user has not specified an encoding.

Files changed:
> unstructured>cleaners>core.py (updated)
> unstructured>partition>email.py(updated)
> test_unstructured>partition>test_text.py (updated)

### Testing
```

elements = partition_email(filename="example-docs/fake-email-utf-16.eml")
```
```
@pytest.mark.parametrize(
    "filename",
    ["fake-email-utf-16.eml", "fake-email-utf-16-be.eml", "fake-email-utf-16-le.eml"],
)
def test_partition_email_from_filename_default_encoding(filename):
    assert len(elements) > 0
    assert elements == EXPECTED_OUTPUT
```


unstructured/partition/email.py
```diff
@@ -258,13 +251,20 @@ def partition_email(
...
         elements = partition_html(text=content, include_metadata=False)
         for element in elements:
             if isinstance(element, Text):
                 _replace_mime_encodings = partial(replace_mime_encodings, encoding=encoding)
-                element.apply(_replace_mime_encodings)
+                try:
+                    element.apply(_replace_mime_encodings)
+                except UnicodeDecodeError:
+                    # If decoding fails, try decoding with default encoding (utf-8)
+                    _replace_mime_encodings = partial(replace_mime_encodings)
```
cragwolfe, line 266: missing arg in the partial?

cragwolfe:
> This PR adds functionality to try other common encodings for email (.eml) files if an error related to the encoding is raised and the user has not specified an encoding.

However, in auto.py, if the user does not specify an encoding they get the default `utf-8` which gets passed down so other common encodings would not be tried.

I think it is OK to change `auto.py` now to have a `None` default for encoding, and have that mean different default args (`None` or `utf-8`) for depending on the `partition_` call.

cragwolfe:
Just curious, how did you generate the test emails with different encodings?

christinestraub:
> However, in auto.py, if the user does not specify an encoding they get the default `utf-8` which gets passed down so other common encodings would not be tried.
> 
> I think it is OK to change `auto.py` now to have a `None` default for encoding, and have that mean different default args (`None` or `utf-8`) for depending on the `partition_` call.

  I already have two working Branches for `html` and `xml` that are ready to create PRs. I also expect to create another working branch to remove the `encoding` parameter from `partition_pdf` and finally update `auto.py`.

christinestraub:
> how did you generate the test emails with different encodings?

I generated test emails with different encodings using the following code.

```
def convert_encoding(file_path, output_encoding='utf-16', filetype='html'):
    with open(file_path, 'r', encoding='utf-8') as file:
        # Read the content of the file with UTF-8 encoding
        content = file.read()

    # Convert the content to output encoding
    converted_content = content.encode(output_encoding)

    # Write encoded content to a new file
    new_file_path = file_path.replace(f'.{filetype}', f'-{output_encoding}.{filetype}')
    with open(new_file_path, 'wb') as new_file:
        new_file.write(converted_content)
  
output_encoding = 'utf-16'
input_file = "example-docs/fake-email.eml"
output_file = f"example-docs/fake-email-{output_encoding}.eml"
convert_encoding(input_file, output_file, output_encoding)
```

unstructured/partition/email.py
christinestraub: If decoding fails, it will try decoding with the default value (UTF-8) of the `encoding` parameter of the `replace_mime_encodings` function. I will fix the missing arg in the partial.

https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/cleaners/core.py#L190

cragwolfe (approved):
LGTM! (when CI is clean)

Issue #543, opened 2023-05-04
bug/parsing an html doc results in zero elements


cragwolfe:
**Describe the bug**
`partition_html` does not extract elements for https://www.congress.gov/107/plaws/publ56/PLAW-107publ56.htm .

**To Reproduce**
Save the link locally.

The following code:
```
from unstructured.partition.auto import partition
from unstructured.partition.html import partition_html

elements = partition(filename="PLAW-107publ56.htm")
print(len(elements))
# see if different extension makes a difference
elements = partition(filename="PLAW-107publ56.html")
print(len(elements))
elements = partition_html(filename="PLAW-107publ56.htm")
print(len(elements))
```

has outputs of:
```
0
0
0
```

**Expected behavior**
Many elements are returned with all the text of the document represented.

**Desktop (please complete the following information):**
 - OS: mac
 - Python version: 3.8.15
 - Unstructured version: 0.6.3-dev1

mu4farooqi:
All of the text is in <pre> tag which this library does not support which makes sense too. For 99% of the cases we don't put text in <pre> tags.

I can submit a quick fix if other agrees.

cragwolfe:
ideally, the library would do the right automatically when parsing a \<pre> tag and extract natural language from the block if it looks like natural language. in any case, the user should also have a way overriding[ text tags](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/documents/html.py#L32) in addition to "auto" \<pre>-text detection.

PR #642, opened 2023-05-25
feature(html partition): parse pre tag

kravetsmic:
### Summary
Added parsing text in the html tag pre. 

### Steps to reproduce

```python3

elements = partition(filename="PLAW-107publ56.html")
print(len(elements)) # not equals 0
elements = partition_html(filename="PLAW-107publ56.html")
print(len(elements)) # not equals 0
```

---

Closes issue #543

### DIFF
--- a/unstructured/documents/xml.py
+++ b/unstructured/documents/xml.py
@@ -4,6 +4,10 @@ from lxml import etree
 
 from unstructured.documents.base import Document, Page
 from unstructured.logger import logger
+from unstructured.partition.text import (
+    element_from_text,
+    split_by_paragraph,
+)
 
 VALID_PARSERS = Union[etree.HTMLParser, etree.XMLParser, None]
 
@@ -64,8 +68,18 @@ class XMLDocument(Document):
         if self.document_tree is None:
             try:
                 document_tree = etree.fromstring(content, self.parser)
+                if "<pre>" and "</pre>" in content:
+                    from bs4 import BeautifulSoup
+                    soup = BeautifulSoup(content, "html.parser")
+                    for element in soup.find_all("pre"):
+                        text_content = split_by_paragraph(element.get_text())
+                        for text in text_content:
+                            element = etree.Element("span")
+                            element.text = str(element_from_text(text=text))
+                            document_tree.append(element)
                 if document_tree is None:
                     raise ValueError("document_tree is None")
+
             # NOTE(robinson) - The following ValueError occurs with unicode strings. In that
             # case, we call back to encoding the string and passing in bytes.
             #     ValueError: Unicode strings with encoding declaration are not supported.


MthwRobinson (changes requested):
@kravetsmic - Can you add a test for the new functionality?

test_unstructured/partition/test_auto.py
```diff
@@ -792,3 +792,13 @@ def test_auto_partition_csv_from_file(filename="example-docs/stanley-cups.csv"):
     assert isinstance(elements[0], Table)
     assert elements[0].metadata.text_as_html == EXPECTED_XLSX_TABLE
     assert elements[0].metadata.filetype == "text/csv"
+
+
+@pytest.mark.skipif(is_in_docker, reason="Skipping this test in Docker container")
```
MthwRobinson, line 797: Does this test fail in the docker container? I think this one should be able to run.

unstructured/documents/xml.py
```diff
@@ -64,8 +68,19 @@ def _read_xml(self, content):
         if self.document_tree is None:
             try:
                 document_tree = etree.fromstring(content, self.parser)
+                if "<pre>" and "</pre>" in content:
+                    from bs4 import BeautifulSoup
```
MthwRobinson, line 72: Can we do this with `lxml` instead of `bs4` to avoid adding a new dependency?

MthwRobinson (changes requested):

CHANGELOG.md
```diff
@@ -120,6 +120,8 @@
 
 ### Fixes
 
+* Fix pre tag parsing
```
cragwolfe, line 123: this bullet needs to be moved higher up in the CHANGELOG.

test_unstructured/partition/test_auto.py
```diff
@@ -775,6 +775,15 @@ def test_auto_partition_csv_from_file(filename="example-docs/stanley-cups.csv"):
     assert elements[0].metadata.filetype == "text/csv"
 
 
+def test_auto_partition_html_pre_from_file(filename="example-docs/fake-html-pre.htm"):
+    elements = partition(filename=filename)
+
+    assert len(elements) > 0
```
cragwolfe, line 781: maybe slightly more sanity testing on the `.text` content of the element(s) here?

unstructured/documents/xml.py
```diff
@@ -65,8 +69,19 @@ def _read_xml(self, content):
         if self.document_tree is None:
             try:
                 document_tree = etree.fromstring(content, self.parser)
+                if "<pre>" and "</pre>" in content:
```
MthwRobinson, line 72: Can we move the conditional logic for `<pre>` outside of the `try`/`except`? Mainly worried about masking legitimate exceptions if they occur within this block.

The `try`/`except` was a hacky fix to a double encoding issue a while back, wouldn't mind refactoring that at some point if we every get time to.

MthwRobinson (changes requested):
Looking good, just one small request

unstructured/documents/xml.py
christinestraub: Sounds reasonable to me!

MthwRobinson (approved):
LGTM from my end! And looks like you already resolved @cragwolfe request on the tests.

christinestraub:
> LGTM from my end! And looks like you already resolved @cragwolfe request on the tests.

Yes, I already resolved @cragwolfe request on the tests.

PR #643, opened 2023-05-25
fix(ci):  prevent gha caching conflicts

ryannikolaidis:
Jobs occasionally fall in to a competing state to set the cache where they don't all install the same things in the venv (which is cached), but they all use the same cache key. So if say `check-deps` wins, it will only install base-pip-packages [here](https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ci.yml#L65). Then any time anyone else fetches they'll get -that- version of the venv and not have the modules they necessarily need.

To resolve, this makes the setup job solely responsible for writing to the cache. All other jobs require setup to complete and then only pull the cache (not try to write to it). 

To make matters slightly more odd, afaict it seems the path argument for the gha cache module will impact whether or not a key matches. So when we were matching again `path: .venv` vs `path: .venv nldk_data` these were actually retrieving different caches. For now, just always specifying the same path every time regardless of the needs and this works reliably. Pulling the nldk_data is pretty quick so seems like a fine compromise for stability (and not managing multiple caches).

bonus: adds a version to the gha cache key so it's easier to invalidate/bump in the future. also only sets this in one place as an env var.
bonus: creates NLTK directory at specified location which should actually allow it to be downloaded and cached as expected
bonus: removes note on CI limits (not relevant)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
 
 env:
-  GHA_CACHE_KEY: unstructured-${{ runner.os }}-${{ matrix.python-version }}-v1-${{ hashFiles('requirements/*.txt') }}
+  GHA_CACHE_KEY_SUFFIX: -v1-${{ hashFiles('requirements/*.txt') }}
 
 jobs:
   setup:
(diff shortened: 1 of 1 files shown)


cragwolfe (approved):

qued (approved):
This looks great! If this works consistently, maybe we should get rid of the backup setup steps on jobs like `check-deps`, `lint`, and `test_unit`?

PR #644, opened 2023-05-26
feat: add --api-key parameter to unstructured-ingest

kravetsmic:
### Summary

Added `--api-key` parameter for connectors. Updated StandardConnectorConfig, added partition_api_key string value.


cragwolfe:
Note that this API key should be passed as as header here: https://github.com/Unstructured-IO/unstructured/blob/7942bc9/unstructured/ingest/interfaces.py#L126
where the header name is `unstructured-api-key`.

Also, please rename the argument name to `--api-key` from `--unstructured-api-key` so that it doesn't get confused as a "true" partition arguments that get passed along in non-api mode here: https://github.com/Unstructured-IO/unstructured/blob/7942bc9/unstructured/ingest/interfaces.py#L117 .

unstructured/ingest/interfaces.py
```diff
@@ -126,6 +127,7 @@ def partition_file(self, **partition_kwargs):
                 response = requests.post(
                     f"{endpoint}",
                     files={"files": (str(self.filename), f)},
+                    headers={"UNSTRUCTURED-API-KEY": self.standard_config.api_key},
```
cragwolfe, line 130: let's not send the API key header at all if `api_key` is empty

cragwolfe (approved):

cragwolfe (approved):

PR #646, opened 2023-05-27
fix: Fix metadata merging strategy

ytoml:
Hi!
I noticed that the current implementation in `_add_element_metadata` (called from the `add_metadata_with_filetype` decorator, etc.) is overriding metadata instead of 'adding' it, which leads to a bug where all elements with `page_number=1` are returned when parsing PDFs with multiple pages...

### DIFF
--- a/unstructured/partition/common.py
+++ b/unstructured/partition/common.py
@@ -93,14 +93,14 @@ def _add_element_metadata(
         )
         if isinstance(element, list):
             for _element in element:
-                _element.metadata = metadata.merge(_element.metadata)
+                _element.metadata = _element.metadata.merge(metadata)
             elements.extend(element)
         elif isinstance(element, PageBreak):
             page_number += 1
             if include_page_breaks:
                 elements.append(element)
         else:
-            element.metadata = metadata.merge(element.metadata)
+            element.metadata = element.metadata.merge(metadata)
             elements.append(element)
     return elements
 


qued:
Hi @ytoml , thanks for the contribution! The situation where `_add_element_metadata` is overriding keys in the metadata is intentional. There is a pattern that is used for some filetypes that goes like:

```
partition_aaa called
- convert file to bbb
- partition_bbb called
  - partition content
  - set metadata for bbb
- set metadata for aaa
```
The metadata is set by `_add_element_metadata`. If the outer partitioning brick is not allowed to override metadata, then things like the file name and file type will be set by the inner brick, and won't be correct. You can see when you run the tests on your branch several related to filename and filetype are failing.

I haven't investigated this issue but I'd be curious whether the page number bug is observed when using the `strategy="fast"` parameter, or if it only occurs using the `hi_res` strategy. There was a recent [fix](https://github.com/Unstructured-IO/unstructured-inference/pull/118) over in `unstructured-inference` that might fix the issue in the `hi_res` strategy as soon as we bump `unstructured-inference` to the latest version in this repo.

On the other hand if the issue also occurs in the `fast` strategy (which doesn't use `unstructured-inference`), maybe the page number is being set too early in the `partition_pdf` pipeline, and we just have to ensure it gets set late enough that it take precedence over the page numbers set by other steps in the workflow.

MthwRobinson:
Closing this one, the page number issue was fixed in #657 .

Issue #641, opened 2023-05-25
bug/unstructured-ingest with no args throws error


cragwolfe:
**Describe the bug**
`unstructured-ingest` with no args throws an error. instead, it should show same output as --help (or brief usage)

**To Reproduce**

PYTHONPATH=. python unstructured/ingest/main.py
PYTHONPATH=. python unstructured/ingest/main.py
Traceback (most recent call last):
  File "unstructured/ingest/main.py", line 748, in <module>
    main()
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/cragwolfe/.pyenv/versions/unst/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "unstructured/ingest/main.py", line 519, in main
    raise ValueError(
ValueError: This connector does not support saving downloads to ~/.cache/  , --download-dir must be provided

PR #649, opened 2023-05-30
feat: if no params show help

kravetsmic:
## Summary

Added validation for parameters. If no parameters passed, show help

---

Closes issue #641


unstructured/ingest/main.py
```diff
@@ -94,8 +94,11 @@ def run(self):
 
         self.cleanup()
 
+# 1, 2, 3, 6, 7, 10
```
yuming-long, line 97: could you remove the comments

yuming-long (changes requested):
The decorator looks good but i don't think the help echo is working as expected because other commands are broken now. I tested with
```
    --metadata-exclude filename,file_directory \
    --local-input-path example-docs \
    --local-file-glob "*.html" \
    --structured-output-dir local-ingest-output \
    --partition-strategy hi_res \
    --verbose \
    --reprocess
  ```
  
and it just returned echo help without actually running the command.
I would suggest fix broken ci first and you can test with the `main` commands from our [ingest test file ](https://github.com/Unstructured-IO/unstructured/tree/main/test_unstructured_ingest) (they are expected to run).

yuming-long (approved):

Issue #592, opened 2023-05-16
bug/`page_number` still appears in `partition_html` metadata if `include_metadata` is set to `False`


MthwRobinson:
**Describe the bug**
If you use `partition_html` with `include_metadata=False`, `page_number` is still included in the metadata.

**To Reproduce**
```python
from unstructured.partition.html import partition_html

filename = "example-docs/example-10k.html"
elements = partition_html(filename=filename, include_metadata=False
```

**Expected behavior**
`page_number` should not be in the metadata.

**Desktop (please complete the following information):**
 - `unstructured==0.6.6`
 - Python version 3.8

Coniferish:
Hey @MthwRobinson. I'm starting to look into this and want to make sure that you're saying that the `page_number` attribute shouldn't exist at all in the metadata, not just that it should be `None`. I am only seeing `None` values when I run the steps to reproduce, but realize I might be using a more recent version of unstructured and that could be creating the difference.

MthwRobinson:
Hey @Coniferish - yeah the attribute should exist but be None. If all of the attributes are None when include_metadata=False, that’s the correct behavior. Possible this might have been corrected already. If that’s the case I think it would still be worthwhile to add in the code you used to verify as a unit test.

Coniferish:
Sounds good. I can write a test for that

Coniferish:
@MthwRobinson I must have been looking at `page_name` instead of `page_number` because I see the bug you mentioned. I'll look into it and write a test for disabling metadata

Coniferish:
Hey @MthwRobinson 

I've found the issue, but am unsure of the best way to address it.

`partition_html` is decorated by `add_metadata_with_filetype`, which currently just returns the elements as-is if `include_metadata = False`. The page number already exists in the element, so that's why we're still seeing it.

I'm wondering if it'd make sense to create a `_remove_element_metadata` method to call. I'm looking at the `_add_element_metadata` (when `include metadata = True`) and wondering what all needs to be included if I'm creating a new method:
```
 if hasattr(layout_element, "text_as_html"):
            text_as_html: Optional[str] = layout_element.text_as_html
```
One question I have is what this^ block is doing. If it does have `"text_as_html"`, that variable isn't being used anywhere it seems.


Here's the full method:
`unstructured/partition/common.py`:
```
def _add_element_metadata(
    layout_elements,
    include_page_breaks: bool = False,
    filename: Optional[str] = None,
[...]

PR #658, opened 2023-05-30
fix: page_number appears in partition_html metadata if include_metadata=False

Coniferish:
## Summary
Fixes bug from #592
Creates test for partitioning html from filename when `metadata=False`
Creates `_remove_element_metadata` method

## Testing
```

filename = "example-docs/example-10k.html"
elements = partition_html(filename=filename, include_metadata=False)
elements[-1].metadata
```

### DIFF
--- a/unstructured/partition/common.py
+++ b/unstructured/partition/common.py
@@ -79,11 +79,11 @@ def _add_element_metadata(
     elements: List[Element] = []
     page_number: int = 1
     for layout_element in layout_elements:
-        element = normalize_layout_element(layout_element)
-        if hasattr(layout_element, "text_as_html"):
-            text_as_html: Optional[str] = layout_element.text_as_html
-        else:
+        try:
+            text_as_html = layout_element.text_as_html
+        except AttributeError:
             text_as_html = None
+        element = normalize_layout_element(layout_element)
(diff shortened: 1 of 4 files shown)


unstructured/partition/common.py
```diff
@@ -79,12 +79,12 @@ def _add_element_metadata(
     elements: List[Element] = []
     page_number: int = 1
     for layout_element in layout_elements:
-        element = normalize_layout_element(layout_element)
-        if hasattr(layout_element, "text_as_html"):
-            text_as_html: Optional[str] = layout_element.text_as_html
-        else:
+        try:
```
MthwRobinson, line 82: Was there a reason for switching this to the `try`/`except` instead of checking if the attribute exists?

MthwRobinson (commented):
Implementation looks good. Just one small question. Could you also add a CHANGELOG entry?

unstructured/partition/common.py
MthwRobinson: All good! It's functionally the same here, just wanted to check if there was something that broke with `hasattr`.

MthwRobinson (approved):
LGTM - thanks for the contribution!

PR #660, opened 2023-05-30
Issue: fix encoding/decoding error with default utf-8 encoding for html, xml, and auto

christinestraub:
This PR is for the following
- Add functionality to try other common encodings for html, xml files if an error related to the encoding is raised and the user has not specified an encoding.
- Change auto.py to have a `None` default for encoding
- Remove the unused parameter `encoding` from `partition_pdf`
- Add functionality to the `read_txt_file` utility function to handle file-like object from URL

### Testing
```

elements = partition_html(filename="example-docs/example-10k-utf-16.html")

@pytest.mark.parametrize(
    ("filename", "encoding", "error"),
    [
        ("example-10k-utf-16.html", "utf-8", UnicodeDecodeError),
        ("example-steelJIS-datasheet-utf-16.html", "utf-8", UnicodeDecodeError),
    ],
)
def test_partition_html_from_filename_raises_encoding_error(filename, encoding, error):
    with pytest.raises(error):
        with open(filename) as f:
            partition_html(file=f, encoding=encoding)

@pytest.mark.parametrize(
    "filename",
    ["example-10k-utf-16.html", "example-steelJIS-datasheet-utf-16.html"],
)
def test_partition_html_from_filename_default_encoding(filename):
    elements = partition_html(filename=filename)
    assert len(elements) > 0
```

```
from unstructured.partition.xml import partition_xml

elements = partition_xml(filename="example-docs/factbook-utf-16.xml", xml_keep_tags=True)

@pytest.mark.parametrize(
    ("filename", "encoding", "error"),
    [("factbook-utf-16.xml", "utf-8", UnicodeDecodeError)],
)
def test_partition_xml_from_filename_with_tags_raises_encoding_error(filename, encoding, error):
    with pytest.raises(error):
        file_path = os.path.join(DIRECTORY, "..", "..", "example-docs", filename)
        partition_xml(filename=file_path, xml_keep_tags=True, encoding=encoding)


@pytest.mark.parametrize(
    "filename",
    ["factbook-utf-16.xml"],
)
def test_partition_xml_from_filename_with_tags_default_encoding(filename):
    file_path = os.path.join(DIRECTORY, "..", "..", "example-docs", filename)
    elements = partition_xml(filename=file_path, xml_keep_tags=True)

    assert elements[5].text == "<name>United States</name>"
    assert elements[5].metadata.filename == filename
```


unstructured/partition/xml.py
```diff
@@ -71,13 +71,11 @@ def partition_xml(
...
         if filename:
-            with open(filename) as f:
-                raw_text = f.read()
+            _, raw_text = read_txt_file(filename=filename, encoding=encoding)
         elif file:
-            f = spooled_to_bytes_io_if_needed(  # type: ignore
-                cast(Union[BinaryIO, SpooledTemporaryFile], file),
-            )
-            raw_text = f.read().decode(encoding)  # type: ignore
+            _, raw_text = read_txt_file(file=file, encoding=encoding)
```
cragwolfe, line 76: is there harm in keeping `spooled_to_bytes_io_if_needed`? in practice, this was introduced to work around a `SpooledTemporaryFile`  that could have originated, say as an `UploadFile` in FastApi

christinestraub: I updated the `read_txt_file` utility function to keep using `spooled_to_bytes_io_if_needed` for xml

PR #661, opened 2023-05-31
test: adds profiling script

ryannikolaidis:
The profiling script allows a user to inspect how time time and memory are spent across called functions when performing partitioning on a given document. There are a few different ways to profile and visualize the data and being able to quickly iterate any of these while debugging performance seems entirely reasonable. To account for this I've made this an interactive script, allowing the user to quickly iterate profiling on custom or fixed set test docs and quickly navigate various slices/views of the profiled data.

bonus: only mkdir for ci's cached nltk content if it doesn't already exist (I suspect this was necessary because there was a version of the ci cache floating out there before the changes made last week).

## Install
Profiling has a few dependencies which can be installed with: 
`pip install -r scripts/performance/requirements.txt`

Before running a test either populate the `docs` directory with test documents of interest or set environment variable SYNC_S3_DOCS=true

Export / assign desired environment variable settings:
- SYNC_S3_DOCS: Set to true to sync test documents from S3 (default: false)
- DOCKER_TEST: Set to true to run profiling inside a Docker container (default: false)

NOTE: the DOCKER_TEST option won't work on M1 Mac because there isn't an ARM-Linux build of memray, but this should work from an x86/AMD environment.

## Usage
`./scripts/performance/profile.sh`
- Run the script and choose the profiling mode: 'run' or 'view'.
- In the 'run' mode, you can profile custom files or select existing test files.
- In the 'view' mode, you can view previously generated profiling results.
- The script supports time profiling with cProfile and memory profiling with memray.
- Users can choose different visualization options such as flamegraphs, tables, trees, summaries, and statistics.
- Test documents are synced from an S3 bucket to a local directory before running the profiles


cragwolfe (approved):
This works wonderfully! 🙏 🙏 🙏

ryannikolaidis:
> This works wonderfully! 🙏 🙏 🙏

Awesome! Thanks! PS: snuck in another option to specify the strategy. Small change, but just wanted to give heads up.

tabossert:
This looks amazing.  Great job @ryannikolaidis

ryannikolaidis:
> This looks amazing. Great job @ryannikolaidis

Thanks! 🙏

PR #662, opened 2023-05-31
build(deps): update inference version

qued:
Updated to the the latest version of `unstructured-inference`. Xfailed tests that fail as a result until we can address the model degradation.

`detectron2` now gets implemented with `onnxruntime`, yay!


MthwRobinson:
As part of this should we update the installation docs to indicate that you don't need to install `detectron2` from source any longer?

Detectron2 installation is also in a couple spot in the CI workflow
- https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ci.yml#L135

test_unstructured/partition/test_auto.py
```diff
@@ -342,8 +345,11 @@ def test_auto_partition_pdf_from_file(pass_file_filename, content_type):
     assert isinstance(elements[0], Title)
     assert elements[0].text.startswith("LayoutParser")
 
+    # NOTE(alan): Xfail since new model misses the first word Zejiang
+    request.applymarker(pytest.mark.xfail)
```
MthwRobinson, line 349: Do we need both the `xfail` and the update to the test below? If we plan in addressing the bounding box issue in inference I think just the `xfail` make sense.

test_unstructured_ingest/test-ingest-s3.sh
```diff
@@ -5,11 +5,15 @@ set -e
...
-if [[ "$(find test_unstructured_ingest/expected-structured-output/s3-small-batch/ -type f -size +20k | wc -l)" -ne 3 ]]; then
-    echo "The test fixtures in test_unstructured_ingest/expected-structured-output/ look suspicious. At least one of the files is too small."
-    echo "Did you overwrite test fixtures with bad outputs?"
-    exit 1
-fi
+# NOTE(robinson) - uncomment the section below and regenerate test fixtures once
+# the ONNXRuntimeError issue in unstructured-inference is resolved
+# https://github.com/Unstructured-IO/unstructured-inference/issues/116
+
+# if [[ "$(find test_unstructured_ingest/expected-structured-output/s3-small-batch/ -type f -size +20k | wc -l)" -ne 3 ]]; then
```
MthwRobinson, line 12: We can revert this change now. This was a temporary work around from when we hit the blank page issue with ONNX

MthwRobinson (changes requested):
🎉 - looks good overall, just noticed a few things sneak over from the original PR that can be removed.

.gitignore
```diff
@@ -151,7 +151,6 @@ tramp
 
 ## https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore
 .vscode/*
-!.vscode/settings.json
```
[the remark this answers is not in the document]
qued: The reason was... I was tired of my `settings.json` showing up on my list of changed files. I don't think we want to specify a `settings.json` for the repo, do we?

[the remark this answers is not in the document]
qued: If desired, I could commit mine which has the settings to configure VSCode's integrated testing.

test_unstructured/partition/test_auto.py
qued: Oops, I meant to keep the old one with "Zejiang". I must have missed this when merging.

test_unstructured_ingest/test-ingest-s3.sh
qued: Reverted in [9db31f2](https://github.com/Unstructured-IO/unstructured/pull/662/commits/9db31f2dca3ac5b0982f8c5e6c4835c9888d499f).

test_unstructured/partition/test_auto.py
qued: Restored Zejiang in [988a7dd](https://github.com/Unstructured-IO/unstructured/pull/662/commits/988a7dda6f5a4a14769b2e5496853209cf844b41).

MthwRobinson (approved):
LGTM with the updates!

PR #663, opened 2023-05-31
fix: raise `ValueError` in `partition_via_api` if filename not present

### LINKED ISSUE #533: bug/`partition_via_api` does not work if `file_filename` is not passed
**Describe the bug**
Currently, `partition_via_api` fails for file-like objects if `file_filename` is not passed.

**To Reproduce**
```python
from unstructured.partition.api import partition_via_api

  elements = partition_via_api(file=f, api_key="MY_API_KEY")
```

**Expected behavior**
The above snippet should work the same as

```python
```

**Screenshots**

 - Unstructured version 0.6.2
 - Python version 3.8.13

MthwRobinson:
### Summary

Closes #533. Raises a `ValueError` if you use `partition_via_api` with a file-like object but do not specify `file_filename`. Currently this fails with a `400`, making it unclear what the problem was. We can remove the `ValueError` if the API supports files without `file_filename` in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.6.12-dev0
+## 0.6.12-dev1
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM!

PR #665, opened 2023-06-01
Build from rocky base image

tabossert:
This modifies the docker image that will be built.  This builds off a base image of Rocky Linux 8, replacing CentOS 7.  This has been done to get latest security fixes and reduce the number of CVE's.  We will be rebuilding the base images on a regular basis to keep the image updated.

This will have the added benefit of reduced build times as the base image contains the system packages and compiled libraries that are required.

Base images can be found at quay.io/repository/unstructured-io/base-images:rocky8.7
Dockerfile: https://github.com/Unstructured-IO/base-images/blob/main/dockerfiles/base-images%3Arocky8.7

Screenshot 2023-05-31 at 4 59 01 PM


ryannikolaidis (approved):
LGTM! 

Checked out on ARM and AMD architecture and on both ran `make docker-build; make docker-start-bash` and then `python3 -c 'from unstructured.partition.auto import partition; elements = partition("example-docs/layout-parser-paper.pdf");print(elements)'` inside the bash terminal.

Ran `docker inspect unstructured:dev | grep Arch` on the built images to confirm both built images for their respective architectures.

cragwolfe:
IMO worth a `-devX` version bump for the sake of image tags

tabossert:
After talking with @cragwolfe we have decided to bump it a minor version instead of a -dev

PR #672, opened 2023-06-01
feat: Add `stage_for_weaviate` and schema creation function

### LINKED ISSUE #566: Add a `stage_for_weaviate` staging brick and add Weaviate to the integrations page
The goal of this issue is to add a `stage_for_weaviate` staging brick to the `unstructured` library and add a Weaviate section to the ecosystem page in the Unstructured docs. Check out [this notebook](https://github.com/weaviate/how-to-ingest-pdfs-with-unstructured/blob/master/notebooks/01_blog.ipynb) for an example of how to load `unstructured` outputs into Weaviate.

MthwRobinson:
### Summary

Closes #566. Adds a `stage_for_weaviate` function to prepare outputs for upload to Weaviate and a `create_unstructured_weaviate_class` helper function for defining the schema.

### Testing

Run the example Jupyter notebook to test the Weaviate workflow locally.

### DIFF
--- a/examples/weaviate/requirements.txt
+++ b/examples/weaviate/requirements.txt
@@ -1,0 +1,3 @@
+tqdm
+weaviate-client
+unstructured
--- a/examples/weaviate/requirements.txt
+++ b/examples/weaviate/requirements.txt
@@ -1,3 +1,4 @@
+jupyter
 tqdm
 weaviate-client
 unstructured
--- a/examples/weaviate/requirements.txt
+++ b/examples/weaviate/requirements.txt
@@ -1,4 +1,4 @@
 jupyter
 tqdm
 weaviate-client
-unstructured
+unstructured[local-inference]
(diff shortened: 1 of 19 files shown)


qued (approved):
For some reason the logic isn't properly skipping the weaviate test in the container. 

Side-note: `base.txt` is out of sync with `base.in`, maybe through a merge, but it is allowing tests to pass when `base.in` does not actually require `tabulate` which *is* required for some of the table extraction functions. Users will get errors when `pip install`ing and trying to use this functionality. I'll create a PR to address this.

Otherwise, a few doc typos I've highlighted, and assuming the docker issue gets fixed, LGTM!

examples/weaviate/requirements.txt
```diff
@@ -0,0 +1,4 @@
+jupyter
+tqdm
+weaviate-client
+unstructured
```
qued, line 4: Actually I think we need this?
```suggestion
unstructured[local-inference]
```

MthwRobinson:
Good catch on the base dependencies!

PR #673, opened 2023-06-01
build(deps): add tabulate dependency

qued:
`tabulate` is used by functions that extract tables from Microsoft documents, but there is nothing explicitly requiring the library. This was not caught by tests, because for some reason, `tabulate` is in `base.txt`. 

This PR adds the dependency to `base.in` (which also puts it in `setup.py`), and recompiles the dependencies.


MthwRobinson (approved):
LGTM!

Issue #666, opened 2023-06-01
bug/partition_pptx : page_number values


Fanchouille:
All elements extracted from partition_pptx have the same page_number that is the number of slides in the pptx file.

**To Reproduce**
```python
from unstructured.partition.pptx import partition_pptx
elements = partition_pptx(filename="my_pptx.pptx") #pptx file with 10 slides
{elt.metadata.page_number for elt in elements} # {10} here
```

**Expected behavior**
```python
from unstructured.partition.pptx import partition_pptx
elements = partition_pptx(filename="my_pptx.pptx") #pptx file with 10 slides
elements[0].metadata.page_number # 1 here, first element should be on first page
elements[-1].metadata.page_number # 10 here
```

**Setup**
 - OS: mac
 - Unstructured version : 0.7.0
 - Python version  : 3.9

PR #675, opened 2023-06-02
fix: metadata.page_number of pptx files

meirdev:
#666

The reason for the problem is that the metadata of all the elements refer to the same ElementMetadata object.


MthwRobinson (approved):
LGTM - thanks for the PR!

PR #677, opened 2023-06-02
fix: recognize code files with auto

### LINKED ISSUE #624: bug/cannot-read-python-files-with-auto
I have a bug when importing because `py` is not an extension in the `file_types.py` module. I run into the problem on the code path while using Langchain. 

(https://github.com/hwchase17/langchain/issues/5025)

Here's a screenshot in a notebook. 
image

I believe the issue can be fixed by adding something related to a python plain-text file to lines 71 and 133, https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/file_utils/filetype.py. I do understand that this is not sustainable and perhaps there may be cleaner ways to open things up to code or something different altogether.

This is the line that errors out: https://github.com/Unstructured-IO/unstructured/blob/fda51d6ead85af49cc38133da71065375c30bb94/unstructured/partition/auto.py#L206

MthwRobinson:
### Summary

Closes #624. Updates the filetype detection logic so that code files are detected as plain text. In the future, we may add a filetype for `CODE` or individual programming languages if we have code/language specific partitioning logic.

### Testing

```python

# This should return FileType.TXT
detect_filetype(filename="unstructured/logger.py")
```


unstructured/file_utils/filetype.py
```diff
@@ -363,6 +383,29 @@ def document_to_element_list(
...
+    "java",
+    "c++",
+    "cpp",
+    "csharp",
+    "c#",
+    "php",
+    "ruby",
+    "swift",
+    "typescript",
+    "go",
```
qued, line 398: This one seems susceptible to misclassification, I can find a few mime types with the letters `go` in them. There are mime types related to google applications that I think would be triggered, as well as one `application/x-seelogo`.

qued (approved):
Everything looks good, testing works, but I do have one concern that I've left a comment for.

unstructured/file_utils/filetype.py
MthwRobinson: Yeah good call, I'll do an explicit check against `text/x-go` for that one.

PR #688, opened 2023-06-06
test: revive ingest unit tests

ryannikolaidis:
At some point [tests were created](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test_interfaces.py) to explicitly validate the features of the ingest interface definitions. These do not currently pass and aren't being processed in CI. This PR updates/fixes the tests (now pass). It also refactors such that they behave as unit tests and don't rely on external services. It also adds them to the CI workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,7 +167,7 @@ jobs:
         source .venv/bin/activate
         mkdir "$NLTK_DATA"
         make install-ci
-    - name: Test (unit)
+    - name: Test Ingest (unit)
       run: |
         source .venv/bin/activate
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -167,7 +167,11 @@ jobs:
         source .venv/bin/activate
         mkdir "$NLTK_DATA"
         make install-ci
-    - name: Test
+    - name: Test (unit)
```
cragwolfe, line 170:
```suggestion
    - name: Test Ingest (unit)
```
(just to make it clearer in github actions UI)

cragwolfe (approved):

PR #690, opened 2023-06-06
feat: adds data_source metadata to ElementMetadata

ryannikolaidis:
We need to add a few abstractions to the ingest base classes to support querying data and populating it when ingesting files. To do so we need to add a few optional fields in Element metadata to support tracking data source metadata:

- version (to know when something is out of date, last modified could work for this field in some cases)
- source url (where applicable, so an end user can view original source content)
- record locator (for querying for the document, which could be used to determine if it exists). It may make sense to make this a dictionary since in some cases there may be multiple values that are used to uniquely identify a document.
- date_created, date_modified, date_processed (for general query of the document state)

This PR adds these fields to a new data_source field in ElementMetadata. It additionally an exists() property to the base IngestDoc definition. Subclasses should leverage record_locator to determine whether a given IngestDoc exists on the source.

bonus: bump dependencies (adds freezegun to validate date_processed functionality)
bonus: adds functionality to exclude nested metadata by dot notation
bonus: updates ingest fixtures to account for data_source field which is empty with addition of date_processed exclusion

## Testing
Adds unit test coverage for data source metadata fields


cragwolfe:
These make abstractions make sense to me! ✅

cragwolfe (approved):
LGTM!

PR #691, opened 2023-06-06
fix:  Filetype detection if a CSV has a text/plain MIME type

### LINKED ISSUE #621: Filetype detection if a CSV has a `text/plain` MIME type
Sometimes the detected MIME type for CSV files is `text/plain`. This causes `partition` (and our docker tests) to fail because it treats the file like a plain text file instead of a CSV. The goal of this issue is to add logic that looks at the first few lines of a file to determine if it's a CSV. One possibility is to look for the same comma count for the first `n` rows. This would be similar to the checks we do for JSON and EML.

Coniferish:
## Summary

Creates filetype detection if a CSV has a text/plain MIME type. 
Closes #621.
Also adds some parameters for `test_is_text_file_a_json`, removes unnecessary empty lines, and improves regex pattern for JSON matching.

## Testing
```
from unstructured.file_utils.filetype import _is_text_file_a_csv
from io import BytesIO

# [invalid, invalid, valid]
files = [b"d\xe2\x80", b'[{"key": "value"}]', b"column1,column2,column3\nvalue1,value2,value3\n"]

for file in files:
    with BytesIO(file) as f:
        print(_is_text_file_a_csv(file=f))
```

### DIFF
--- a/unstructured/file_utils/filetype.py
+++ b/unstructured/file_utils/filetype.py
@@ -7,7 +7,7 @@ from functools import wraps
 from typing import IO, Callable, List, Optional
 
 from unstructured.documents.elements import Element, PageBreak
-from unstructured.nlp.patterns import LIST_OF_DICTS_PATTERN
+from unstructured.nlp.patterns import JSON_PATTERN
 from unstructured.partition.common import (
     _add_element_metadata,
     _remove_element_metadata,
@@ -334,6 +334,27 @@ def _is_text_file_a_json(
 ):
     """Detects if a file that has a text/plain MIME type is a JSON file."""
     exactly_one(filename=filename, file=file)
+    if file is not None:
+        file.seek(0)
+        file_content = file.read(4096)
+        if isinstance(file_content, str):
+            file_text = file_content
+        else:
+            file_text = file_content.decode(errors="ignore")
+        file.seek(0)
+    elif filename is not None:
+        with open(filename) as f:
+            file_text = f.read()
+    return re.match(JSON_PATTERN, file_text) is not None
+
+
+def _is_text_file_a_csv(
+    filename: Optional[str] = None,
+    content_type: Optional[str] = None,
+    file: Optional[IO] = None,
+):
+    """Detects if a file that has a text/plain MIME type is a CSV file."""
+    exactly_one(filename=filename, file=file)
 
     if file is not None:
         file.seek(0)
@@ -347,7 +368,17 @@ def _is_text_file_a_json(
         with open(filename) as f:
             file_text = f.read()
 
-    return re.match(LIST_OF_DICTS_PATTERN, file_text) is not None
+    lines = file_text.strip().splitlines()
+    if len(lines) < 2:
+        return False
+
+    header = lines[0].split(",")
+    columns = len(header) if len(header) < 5 else 5
+
+    if any("," not in line for line in lines[1:columns]):
+        return False
+
+    return all(len(line.split(",")) == len(header) for line in lines[1:columns])
 
 
 def _check_eml_from_buffer(file: IO) -> bool:
@@ -359,7 +390,6 @@ def _check_eml_from_buffer(file: IO) -> bool:
         file_head = file_content.decode("utf-8", errors="ignore")
     else:
         file_head = file_content
-
     return EMAIL_HEAD_RE.match(file_head) is not None
 
 
--- a/unstructured/file_utils/filetype.py
+++ b/unstructured/file_utils/filetype.py
@@ -7,7 +7,7 @@ from functools import wraps
 from typing import IO, Callable, List, Optional
 
 from unstructured.documents.elements import Element, PageBreak
-from unstructured.nlp.patterns import JSON_PATTERN
+from unstructured.nlp.patterns import LIST_OF_DICTS_PATTERN
 from unstructured.partition.common import (
     _add_element_metadata,
     _remove_element_metadata,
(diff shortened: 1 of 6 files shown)


MthwRobinson:
Right now it looks like it's returning `True` on non-CSV text files. Haven't looked at the test failures yet but might be related to that.

```python
In [3]: from unstructured.file_utils.filetype import _is_text_file_a_csv

In [4]: _is_text_file_a_csv(filename="example-docs/book-war-and-peace-1p.txt")
Out[4]: True

In [5]: _is_text_file_a_csv(filename="example-docs/fake-text.txt")
Out[5]: True
```

unstructured/file_utils/filetype.py
```diff
@@ -347,7 +368,17 @@ def _is_text_file_a_json(
         with open(filename) as f:
             file_text = f.read()
 
-    return re.match(LIST_OF_DICTS_PATTERN, file_text) is not None
+    lines = file_text.strip().splitlines()
```
MthwRobinson, line 371: In case the file is really big, it might be good to just read the first 10 or 20 lines of the file instead of the entire file.

unstructured/file_utils/filetype.py
```diff
@@ -334,6 +334,27 @@ def _is_text_file_a_json(
...
+        file.seek(0)
+        file_content = file.read(4096)
+        if isinstance(file_content, str):
+            file_text = file_content
+        else:
+            file_text = file_content.decode(errors="ignore")
+        file.seek(0)
+    elif filename is not None:
+        with open(filename) as f:
+            file_text = f.read()
```
MthwRobinson, line 347: Since this chunk is reused between `_is_text_file_a_csv` and `_is_text_file_a_json`, we could pull that out into a `_read_file_start_for_type_check` function. As noted below, we probably don't need to read the entire file either for CSV or JSON.

unstructured/file_utils/filetype.py
```diff
@@ -347,7 +368,17 @@ def _is_text_file_a_json(
         with open(filename) as f:
             file_text = f.read()
 
-    return re.match(LIST_OF_DICTS_PATTERN, file_text) is not None
+    lines = file_text.strip().splitlines()
+    if len(lines) < 2:
+        return False
+
+    header = lines[0].split(",")
+    columns = len(header) if len(header) < 5 else 5
```
MthwRobinson, line 376: Is there a reason for making columns having a minimum length of 5?

MthwRobinson (changes requested):
Added a couple of comments. Could you also add a section to the PR description that looks like the following? That'll make it quicker to test the changes.

### Testing

```python
from unstructured.file_utils.filetype import _is_text_file_a_csv

<example of positive case>
<example of negative case>
```

unstructured/file_utils/filetype.py
Coniferish: No. I wasn't thinking straight and was thinking that was the first 5 lines of the file (so we aren't iterating over the whole file). I'll change that.

MthwRobinson:
Don't think they'll need to take `content_type` as an input. They're only ever called if they file is detected as plain text, which means it would have a `text/plain` (or similar) content type. So `content_type` sits upstream of where both of those are called.

https://github.com/Unstructured-IO/unstructured/blob/f0306b57e790b93901b6636b7e5e5b1316690ed4/unstructured/file_utils/filetype.py#L260-L261

Coniferish:
@MthwRobinson  I just thought of an edge case for reading the lines from the files that I'm unsure how to handle:
```
def _read_file_start_for_type_check(
    file: Optional[IO] = None,
) -> str:
    """Reads the start of the file and returns the text content."""
    if file is not None:
        file.seek(0)
        file_content = file.read(4096)
        if isinstance(file_content, str):
            file_text = file_content
        else:
            file_text = file_content.decode(errors="ignore")
        file.seek(0)
    elif filename is not None:
        with open(filename) as f:
            file_text = f.read(4096)
    else:
        file_text = ""
    return file_text
```

If this is the `_read_file_start_for_type_check` function and it only reads a limited number of bytes, there's the potential that it wouldn't read a full line and when the file is checked if it's valid csv format (or json I'd think), then there's the possibility that the closing brackets/characters won't match up with the opening ones. What should I do about this?

MthwRobinson:
@Coniferish - What do you think about throwing away the last line? Hopefully it will read in enough lines that that will be okay

Coniferish:
> Added a couple of comments. Could you also add a section to the PR description that looks like the following? That'll make it quicker to test the changes.
> ### Testing
> 
> ```python
> from unstructured.file_utils.filetype import _is_text_file_a_csv
> 
> <example of positive case>
> <example of negative case>
> ```

MthwRobinson (approved):
LGTM! Tested on a few of the example files. Looks like it's currently failing on non UTF-8 encoded text files, but that's a separate issue from what's being addressed in this PR so I think we can spin off a follow on issue for that.

```python
In [9]: _is_text_file_a_csv(filename="example-docs/fake-text-utf-32.txt")
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[9], line 1
----> 1 _is_text_file_a_csv(filename="example-docs/fake-text-utf-32.txt")

File ~/repos/forks/coniferish-unstructured/unstructured/file_utils/filetype.py:364, in _is_text_file_a_csv(filename, file)
    359 def _is_text_file_a_csv(
    360     filename: Optional[str] = None,
    361     file: Optional[IO] = None,
    362 ):
    363     """Detects if a file that has a text/plain MIME type is a CSV file."""
--> 364     file_text = _read_file_start_for_type_check(file=file, filename=filename)
    365     lines = file_text.strip().splitlines()
    366     if len(lines) < 2:

File ~/repos/forks/coniferish-unstructured/unstructured/file_utils/filetype.py:346, in _read_file_start_for_type_check(filename, file)
    344 if filename is not None:
    345     with open(filename) as f:
--> 346         file_text = f.read(4096)
    347 return file_text

File ~/.pyenv/versions/3.8.13/lib/python3.8/codecs.py:322, in BufferedIncrementalDecoder.decode(self, input, final)
    319 def decode(self, input, final=False):
    320     # decode input (taking the buffer into account)
    321     data = self.buffer + input
--> 322     (result, consumed) = self._buffer_decode(data, self.errors, final)
    323     # keep undecoded input until the next call
    324     self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

PR #702, opened 2023-06-08
ci: update ingest fixtures from gh workflow

ryannikolaidis:
Adds a ingest-test-fixtures-update-pr workflow. It will generate a PR that targets the triggering branch, but includes any changes as a result of running the test ingest fixtures script (with OVERWRITE=true). It can be triggered on [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) (which, once merged to main, should add a dispatch button to the workflow in the Actions menu). It can also be triggered with any commit containing the message: "ingest-test-fixtures-update". So one can also trigger with: `git commit -m "ingest-test-fixtures-update" --allow-empty`. 

The created PR includes the original author as assignee and as reviewer. It is authored by Unstructured-DevOps.

Note: because we trigger on commit push and workflow dispatch we have to guess at the original PR name by looking all PRs from the target branch (hopefully each branch only ever has one PR, but theoretically a known edge case).

## Testing
I made a change to the azure metadata-exclude (to trigger ingest expected result changes) and pushed a commit with the name `ingest-test-fixtures-update` to trigger creation of this PR: https://github.com/Unstructured-IO/unstructured/pull/731


.github/workflows/ingest-test-fixtures-update-pr.yml
```diff
@@ -0,0 +1,110 @@
+name: Ingest Test Fixtures Update PR
+
+on:
+  push:
```
cragwolfe, line 4: does this mean it gets triggered for every push?

cragwolfe (approved):

.github/workflows/ingest-test-fixtures-update-pr.yml
ryannikolaidis: it does, but it will get cancelled if the commit message doesn't contain "ingest-test-fixtures-update". I couldn't find a better way to do trigger this on commit, but totally open if you have other thoughts?

cragwolfe: ah, i see. works for now!

PR #703, opened 2023-06-08
fix: suppress file conversion warnings in `convert_office_doc`

### LINKED ISSUE #647: Libreoffice subprocess dumps output to stdout
At the moment, ingesting files can produce output like 
```
convert /home/user/blah.ppt -> /tmp/tmpi864kzzm/blah.pptx using filter : Impress MS PowerPoint 2007 XML
libpng warning: iCCP: known incorrect sRGB profile
```
which I believe is dumped to stdout by a subprocess call.
I think the call is here:
https://github.com/Unstructured-IO/unstructured/blob/66058e76bf69f87c561aeb57e5bb4efd265dbd67/unstructured/partition/common.py#L115
but I'm not familiar with the codebase enough to be sure.

In my opinion, this output should be captured and sent to `logging.info` or `logging.warning`.

MthwRobinson:
### Summary

Closes #647. Updates `convert_office_doc` so that informational message about file conversion are not printed to stdout. Output is now captured and routed to `logger.info` or `logger.error` as appropriate.

### Testing

You should no longer see a file conversion message when you run this code:

```python

filename = "example-docs/fake.doc"
elements = partition_doc(filename=filename)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.3-dev0
+## 0.7.3-dev1
 
 ### Enhancements
 
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+* `convert_office_doc` no longers prints file conversion info messages to stdout.
 * `partition_via_api` reflects the actual filetype for the file processed in the API.
 
 ## 0.7.2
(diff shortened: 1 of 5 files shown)


qued (approved):
LGTM! On `main` branch I get the warning, on this branch no output, turn on logging, I get the warning over `INFO`.

MinnyKuan:
hi,
I'm using version 0.11.8.
I use the following code to execute partition_doc :

```

filename = ""D:\\Testcase\\test.doc""
elements = partition_doc(filename=filename)

```

However, I encountered a UnicodeDecodeError.
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xaa in position 38: invalid start byte`

 Could you assist me in resolving this issue?

MthwRobinson:
Hi @MinnyKuan - could you open an new issue for that error?

MinnyKuan:
hi @MthwRobinson ,
Addressed in https://github.com/Unstructured-IO/unstructured/issues/2449

Thank you

PR #708, opened 2023-06-09
Fix: Pass `strategy` parameter down from `partition` for `partition_image`

yuming-long:
### Summary

Fix the problem that `strategy`  is not passed down from `partition` to`partition_image`. Add some doc string nit.

### Test
```

filename = "example-docs/english-and-korean.png"
elements = partition(filename=filename, ocr_languages="eng+kor", strategy="ocr_only")
print(elements[3].text)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.3-dev3
+## 0.7.3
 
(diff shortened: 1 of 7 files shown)


MthwRobinson (approved):
LGTM - good catch!

PR #710, opened 2023-06-09
enhancement: handling for empty files in `detect_filetype` and `partition`

### LINKED ISSUE #612: bug/detect_filetype hits error on an empty file
**Describe the bug**
When a 0 byte file is sent to `detect_filetype()`, the mimetype is returned as `inode/x-empty` which is not a supported mimetype.

**To Reproduce**
* Start a shell with `make docker-start-bash`
* Create an empty file: `touch empty.txt`
* Try the following in ipython:
```
In [2]: from unstructured.partition.auto import partition
   ...: filename = "empty.txt"
   ...: with open(filename, "rb") as f:
   ...:     partition(filename, f)
   ...: 
The MIME type of 'empty.txt' is 'inode/x-empty'. This file type is not currently supported in unstructured.
```

**Expected behavior**
An empty file should be handled at some level. This is probably best done before `detect_filetype`.

MthwRobinson:
### Summary

Closes #612. Adds handling for empty files in `detect_filetype` and `partition`. Empty files will now return an empty list from `partition`.

### Testing

```python

filename = "example-docs/empty.txt"

detect_filetype(filename=filename) # Should be FileType.Empty
partition(filename=filename) # Should be []

    filetype = detect_filetype(filename=filename) # Should be FileType.Empty

    elements = partition(filename=filename) # Should be []
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.7.4-dev0
+
+### Enhancements
+
+* Adds handling for empty files in `detect_filetype` and `partition`.
+
+### Features
+
+### Fixes
+
 ## 0.7.3
 
 ### Enhancements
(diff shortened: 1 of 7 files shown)


qued (approved):
LGTM!

PR #714, opened 2023-06-09
fix: Enable MIME type detection if libmagic is not available

### LINKED ISSUE #682: Enable MIME type detection if libmagic is not available
The goal of this issue is to enable MIME type detection if `libmagic` is not available. We can use [`filetype`](https://github.com/h2non/filetype.py/tree/master) as a fallback if `libmagic` is not present. `libmagic` is still preferred if it's available since it's more accurate and covers more file types.

If `libmagic` is not available, you'll get the following error if your run `import magic`:

```python
ImportError: failed to find libmagic.  Check your installation
```

Coniferish:
## Summary
Closes #682
Enables MIME type detection if libmagic is not available. 
Uses [filetype](https://github.com/h2non/filetype.py/tree/master) as a fallback if libmagic is not present.
Updates associated `ImportError` test to check for `warning`
Adds comments

## Testing
```
pytest test_unstructured/file_utils/test_filetype.py
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.7.4-dev0
+
+### Enhancements
+
+* Enable MIME type detection if libmagic is not available
+
+### Features
+
+### Fixes
+
 ## 0.7.3
 
 ### Enhancements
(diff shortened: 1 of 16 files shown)


MthwRobinson:
Think you need `filetype` in the dependencies. You can add that to `setup.py` then do `pip-compile --upgrade requirements/base.in` (you'll need `pip-tools` installed)

MthwRobinson (approved):
LGTM!

PR #725, opened 2023-06-12
feat: `partition_rst` for ReStructured Text documents

### LINKED ISSUE #720: feat/rst support (ReStructuredText)
Please add support to parse RST files. https://en.wikipedia.org/wiki/ReStructuredText

RST is a common file format for documentation in the Python ecosystem. It predates Markdown. It is supported by Sphinx ( https://www.sphinx-doc.org/en/master/ ) and many other tools.

I would love to be able to use RST files with Langchain, via unstructured. I have some dev guides written as many such files. Thanks for listening.

MthwRobinson:
### Summary

Closes #720. Adds support for processing `.rst` documents. This feature requires users to have `pandoc` installed on their system.

### Testing

```python
from unstructured.partition.rst import partition_rst

elements = partition_rst(filename="example-docs/README.rst")
```


qued (approved):
LGTM overall, a few nits. One general comment/question though, the element types that come back are `HTMLTitle`, `HTMLNarrativeText`, etc. Is that what we want?

MthwRobinson:
One `HTMLTitle`, etc, that's an artifact of converting to HTML before we process and probably not what we want. It's a sub-class of `Title` though and JSON serialized representation has `Title`, which is what we want. That also impacts EPUB and RTF. I can spinoff an issue to fix that.

```python
{'element_id': '9bd312a379b45e09744daab96ca5f881',
 'coordinates': None,
 'text': 'Example Docs',
 'type': 'Title',
 'metadata': {'filename': 'README.rst',
  'file_directory': 'example-docs',
  'filetype': 'text/x-rst',
  'page_number': 1}}
```

MthwRobinson:
Created #726 for the `HTMLTitle` issue

PR #734, opened 2023-06-13
feature: Enhance leaf element string check in XML parsing

MalteHB:
This PR introduces an enhancement to the `get_leaf_elements` function in our XML parsing logic. Previously, this function was adding the text of an element to the `leaf_elements` list if that element was a leaf (i.e., it didn't have any children), regardless of whether the element's text was a string or None.

With the changes in this PR, we add an additional check to ensure that the text of an element is only added to the `leaf_elements` list if it is a string. This is done by introducing a new `is_string` function that checks if the `.text` attribute of an element is of type `string` and not `None`. This function is used in conjunction with the existing `is_leaf` function to decide whether an element's text should be added to the `leaf_elements` list.

This change will make our XML parsing more robust, as it will prevent None values from being added to the leaf_elements list, which would cause issues when joining by newline in the return statement.

**Changes:**

Added a new `is_string` function that checks if the `.text` attribute of an element is of type `string`.
Updated the `get_leaf_elements` function to use the `is_string` function when deciding whether to add an element's text to the `leaf_elements` list.


MthwRobinson (commented):
Thanks for the contribution! Could you take a look at these [test failures](https://github.com/Unstructured-IO/unstructured/actions/runs/5252520830/jobs/9495947400?pr=734) in CI and also add a CHANGELOG entry?

```python
FAILED test_unstructured/partition/test_xml_partition.py::test_partition_xml_from_file[factbook.xml] - AttributeError: 'str' object has no attribute 'text'
```

MthwRobinson (approved):
Fixed the broken tests and updated the CHANGELOG. LGTM!

PR #737, opened 2023-06-13
ci: add cache version to ingest-test-fixtures-update-pr workflow

ryannikolaidis:
Without a version specified this added one more cache, set to v1 so this re-uses the same one in the CI workflow


yuming-long (approved):
LGTM!

PR #739, opened 2023-06-13
Bug/635 unicode decode error eml

christinestraub:
This PR is for the following

- Adds functionality to extract charset info from `eml` files
- Adds missed file-like object handling in `detect_file_encoding`
- Adds functionality to replace the `MIME` encodings for `eml` files with one of the common encodings if a `unicode` error occurs
- Organize the `eml` example files in the `example-docs/eml` directory

### Testing

```
f_path = "example-docs/eml/email-no-utf8-2008-07-16.062410.eml"

elements = partition_email(filename=f_path)

with open(f_path) as file:
    elements = partition_email(file=file)

with open(f_path, "rb") as file:
    elements = partition_email(file=file)
```


unstructured/partition/email.py
```diff
@@ -219,12 +240,15 @@ def partition_email(
     exactly_one(filename=filename, file=file, text=text)
 
     if filename is not None:
-        encoding, file_text = read_txt_file(filename=filename, encoding=encoding)
-        msg = email.message_from_string(file_text)
-
+        encoding, msg = parse_email(filename=filename)
```
cragwolfe, line 243: this is odd in that it is overwriting the `encoding` kwarg from line 227

cragwolfe (commented):
please add a test for the case that would have failed without the change

unstructured/partition/email.py
cragwolfe: oh, nvm, i see `encoding` is only applicable in the `text` case. maybe renaming the variable here would make that a bit more obvious?

unstructured/partition/email.py
```diff
@@ -182,6 +182,27 @@ def find_embedded_image(
...
+def parse_email(
+    filename: Optional[str] = None,
+    file: Optional[IO] = None,
+) -> Tuple[Optional[str], Message]:
+    if filename is not None:
+        with open(filename, "rb") as f:
+            msg = email.message_from_binary_file(f)
+    elif file is not None:
+        with open(file.name, "rb") as f:
+            msg = email.message_from_binary_file(f)
```
cragwolfe, line 194: any reason why not `email.message_from_binary_file(file)` rather than re-opening the `file` object?

unstructured/partition/email.py
MthwRobinson: Yeah I agree it would be good to rename there variable here to make that clear.

kafkasl:
@christinestraub hello, I was testing your PR because I was facing the same issue. This PR is clearly an improvement. I have a sample of 1000 emails and the amount of errors got down from 661 to just 80. I am attaching here an example of the errors.
This is the stacktrace for one
```
  File "/tmp/ipykernel_737/1662947039.py", line 22, in <module>
    data = loader.load()
  File "/usr/local/lib/python3.9/dist-packages/langchain/document_loaders/unstructured.py", line 71, in load
    elements = self._get_elements()
  File "/usr/local/lib/python3.9/dist-packages/langchain/document_loaders/email.py", line 24, in _get_elements
    return partition_email(filename=self.file_path, **self.unstructured_kwargs)
  File "/usr/local/lib/python3.9/dist-packages/unstructured/file_utils/filetype.py", line 483, in wrapper
  File "/usr/local/lib/python3.9/dist-packages/unstructured/partition/email.py", line 287, in partition_email
    element.apply(_replace_mime_encodings)
  File "/usr/local/lib/python3.9/dist-packages/unstructured/documents/elements.py", line 188, in apply
    cleaned_text = cleaner(cleaned_text)
  File "/usr/local/lib/python3.9/dist-packages/unstructured/cleaners/core.py", line 203, in replace_mime_encodings
    raise e
  File "/usr/local/lib/python3.9/dist-packages/unstructured/cleaners/core.py", line 200, in replace_mime_encodings
    return quopri.decodestring(text.encode(encoding)).decode(encoding)
UnicodeEncodeError: 'ascii' codec can't encode character '\u2019' in position 15: ordinal not in range(128)
```

The encoding getting into the `cleaners/core.py` function is `us-ascii` and the text  `As discussed, I’m putting in some time for an introductory call.`.  I have also other examples were the failing encoding is `utf-8`. Is there any way I can help with this? Should I just create another ticket?

unstructured/partition/email.py
christinestraub: @cragwolfe That sounds like a good point! I'll be working on renaming the variable.

christinestraub:
@kafkasl Thanks for your feedback! Can you share the sample of 1,000 emails you used for your test?

unstructured/partition/email.py
cragwolfe: ah, that makes sense. i think it is ok to just reopen for now, it won't be open for long. thanks for the detailed explanation.

MthwRobinson:
@kafkasl - Thanks for the input on this issue! Added #755 for `.mbox` files and also #756 for `.pst`, which is the MS Outlook equivalent.

christinestraub:
> please add a test for the case that would have failed without the change

@cragwolfe At this point, it's quite difficult to add a test for the case that would have failed without the change because the errors have been fixed and no longer appear. Can you please guide me on how to do this?

christinestraub:
@kafkasl The errors in the files you shared have been fixed except for `email-3.eml`. There are no HTML tags in the file, in which case the argument `content_source` needs to be set to `text/plain`.

Can you please test this PR against your 1000 sample emails and give us feedback? Thanks!

unstructured/file_utils/encoding.py
```diff
@@ -27,6 +27,10 @@
 ]
 
 
+def format_encoding_str(encoding: str) -> str:
+    return encoding.lower().replace("_", "-")
```
MthwRobinson, line 31: Could you add a quick comment here showing what the encoding looks like before and after?

cragwolfe:
> @cragwolfe At this point, it's quite difficult to add a test for the case that would have failed without the change because the errors have been fixed and no longer appear. Can you please guide me on how to do this?

i just mean, ideally there would be a test where `partition_email` succeeds, whereas before this commit, it'd fail (protect against regressions). i think it i just about identifying a file that currently breaks on `main` and adding a test for that that passes in the current branch.

unstructured/partition/email.py
```diff
@@ -264,12 +301,24 @@ def partition_email(
                 _replace_mime_encodings = partial(replace_mime_encodings, encoding=encoding)
                 try:
                     element.apply(_replace_mime_encodings)
-                except UnicodeDecodeError:
-                    # If decoding fails, try decoding with default encoding (utf-8)
-                    element.apply(replace_mime_encodings)
+                except (UnicodeDecodeError, UnicodeError):
```
MthwRobinson, line 304: It might be useful to pull this logic into a function in `encoding.py` in case we need to do something similar in another partitioning function (like `partition_text` or `partition_msg`).

MthwRobinson (commented):
Looks good from my end, just two small suggestions. Since @cragwolfe was reviewing this prior I'll defer to him for a final ✅

cragwolfe (approved):
Looks good to me, once Matt's two comments are addressed
🙏

unstructured/partition/email.py
christinestraub: @MthwRobinson @cragwolfe There is already similar logic in the `detect_file_encoding` function in `encoding.py` (https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/file_utils/encoding.py#L55-L62). The above logic was added separately to handle Unicode errors from the `replace_mime_encodings` function in `unstructed/cleaners/core.py`. Other partitioning functions (like partition_text or partition_msg) don't seem to use this `replace_mime_encodings` function. 

Do you mean to pull this logic into the `replace_mime_encodings` function in `unstructed/cleaners/core.py`?

unstructured/file_utils/encoding.py
christinestraub: @MthwRobinson @cragwolfe Added comments showing what the encoding looks like before and after.

kafkasl:
> @kafkasl The errors in the files you shared have been fixed except for `email-3.eml`. There are no HTML tags in the file, in which case the argument `content_source` needs to be set to `text/plain`.
> 
> Can you please test this PR against your 1000 sample emails and give us feedback? Thanks!

I've tested it and still get decoding errors very similar to the ones I shared in the previous message. Not sure why it would be fixed only for some emails. Maybe the errors are coming from different parts of the message? not just the content. Still it's much better thanks for the effort!

christinestraub:
> > @kafkasl The errors in the files you shared have been fixed except for `email-3.eml`. There are no HTML tags in the file, in which case the argument `content_source` needs to be set to `text/plain`.
> > Can you please test this PR against your 1000 sample emails and give us feedback? Thanks!
> 
> I've tested it and still get decoding errors very similar to the ones I shared in the previous message. Not sure why it would be fixed only for some emails. Maybe the errors are coming from different parts of the message? not just the content. Still it's much better thanks for the effort!

Can you please provide any emails that are still failing? It seems like a bunch is fixed, but maybe there are lurking edge cases. Thanks!

kafkasl:
> > I've tested it and still get decoding errors very similar to the ones I shared in the previous message. Not sure why it would be fixed only for some emails. Maybe the errors are coming from different parts of the message? not just the content. Still it's much better thanks for the effort!
> 
> @kafkasl Can you please provide any emails that are still failing? It seems like a bunch is fixed, but maybe there are lurking edge cases. Thanks!

@christinestraub I've tried again and all the errors are gone (+12k emails). I assume when I tried last time, the install from git was still not using the latest changes. I will keep using it with more emails and report any error. Thanks so much for your hard work!

PR #742, opened 2023-06-13
enhancement: extract headers and footers in `partition_docx`

MthwRobinson:
### Summary

Updates `partition_docx` to grab headers and footers. Headers are added to the beginning of a section and footers to the end. There is now also a `header_footer_type` for MS Word documents indicating where the header and footers apply. We're not able to identify pages in `.docx` files yet. Once we are, we can apply the headers and footers to the appropriate page.

### Testing

```python

elements = partition_docx(filename=filename)
elements[0] # Should be a header
elements[-1] # Should be a footer
```


qued (approved):
LGTM!

PR #743, opened 2023-06-13
ci: add set up python to test_unit job

ryannikolaidis:
Otherwise if the cache is evicted before the unit tests, they will fail due to not having the correct version of python and they will populate the cache without the python binary, so they will always fail when using the cache after that point.

Here I'm following the convention of all the other jobs and just always running Setup Python in the unit tests job.


yuming-long (commented):
could we add the set up python for test_ingest as well?

ryannikolaidis:
> could we add the set up python for test_ingest as well?

agh, I could have sworn I checked. yes, good eye!

yuming-long (approved):
LGTM!

tabossert (approved):

PR #746, opened 2023-06-14
feat: add google cloud storage connector

potter-potter:
* Adds google cloud storage connector
* Adds recursive folder walking for gcs, s3 and azure
* Adds ingest test for google cloud storage

Note: properties supporting date_modified, date_processed, exists, record_locator, source_url, and version will be handled in a follow-on PR.


unstructured/ingest/main.py
```diff
@@ -555,9 +565,20 @@ def main(
...
                 ),
             )
+        elif protocol in ("gs", "gcs"):
+            from unstructured.ingest.connector.gcs import GcsConnector, SimpleGcsConfig
+
+            doc_connector = GcsConnector(  # type: ignore
+                standard_config=standard_config,
+                config=SimpleGcsConfig(
+                    path=remote_url,
+                    recursive=recursive,
```
ryannikolaidis, line 579: how does this handle cases where we need any authorization to access the gcs bucket? do we need to pass through access_kwargs?

unstructured/ingest/main.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         reprocess,
         verbose,
         max_docs,
+        recursive,
```
ryannikolaidis, line 35: why do we need to pass recursive here? we ultimately set/get this via SimpleFsspecConfig, which has already been assigned on the connector at this point, right?

unstructured/ingest/main.py
```diff
@@ -392,6 +395,12 @@ def run(self):
     show_default=True,
     help="Number of parallel processes to process docs in.",
 )
+@click.option(
+    "--recursive",
```
ryannikolaidis, line 399: we have local-recursive and drive-recursive flags currently, so it might be misleading to have a non-specific recursive flag that impacts specifically some of the fsspec connectors. that said, wondering, should we elevate recursive to a StandardConnectorConfig field?

unstructured/ingest/main.py
```diff
@@ -392,6 +395,12 @@ def run(self):
...
+    is_flag=True,
+    default=False,
+    help="Will recurse directories for S3, GCS and Azure storage.",
```
ryannikolaidis, line 402: Obviously bump the types included if we go for the suggestion above. Maybe slight tweak to follow description pattern like remote-url to something like:
```suggestion
    help="Recursively download files in their respective folders"
    "otherwise stop at the files in provided folder level."
    " Supported protocols are: `gcs`, `gs`, `s3`, `s3a`, `abfs` "
    "and `az`.",
```

unstructured/utils.py
```diff
@@ -35,7 +35,7 @@ def wrapper(*args, **kwargs):
                 raise ImportError(
                     f"Following dependencies are missing: {', '.join(missing_deps)}. "
                     + (
-                        f"Please install them using `pip install unstructured[{extras}]`."
+                        f"""Please install them using `pip install "unstructured[{extras}]"`."""
```
ryannikolaidis, line 38: was there a reason for the triple quote here? or some auto-formatting thing happen?

potter-potter: Yes. `unstructured[blabla]` needs to have double quotes around it in the error message for the user to install correctly. If the quotes are not there, then it does not pip install.

unstructured/ingest/main.py
potter-potter: I'm not sure how we are handling authorization yet. I'll look into it.

unstructured/ingest/main.py
potter-potter: Yes. I've been thinking about this. I will look into this.

unstructured/utils.py
potter-potter: I guess you could probably use single quotes instead of triple double...

ryannikolaidis: ohhh totally didn't even see the quotes around `unstructrued[{extras}]`. YES, 100% nice catch!

ryannikolaidis:
oh, also, [some additions](https://github.com/Unstructured-IO/unstructured/pull/690) to the base class came in very recently.

we haven't added support in any of the connectors yet so there aren't good references, but want to blaze the trail on adding [these](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L99-L136)? Specifically, properties supporting: date_modified, date_processed, exists, record_locator, source_url, and version?

potter-potter:
> oh, also, [some additions](https://github.com/Unstructured-IO/unstructured/pull/690) to the base class came in very recently. these will be really important for future cases where we need to be thoughtful about how/when things get reprocessed as well as giving users more access to that information.
> 
> we haven't added support in any of the connectors yet so there aren't good references, but want to blaze the trail on adding [these](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L99-L136)? Specifically, properties supporting: date_modified, date_processed, exists, record_locator, source_url, and version?

I can take a look at these. But I want to finish this PR first.

unstructured/ingest/main.py
potter-potter: Truth!

unstructured/ingest/main.py
potter-potter: Looks like we can just use one flag
--recursive
for fsspec, local, and google drive

Doesn't seem like it belongs in StandardConnectorConfig until we can confirm that it works for all connectors, since StandardConnectorConfig seems to be more related to the second step after connecting (like where to download, etc...)

Would be nice to see if we can get it working on all connectors, but that seems out of scope for right now. Though some connectors would never need a recursive (like wikipedia...)

Thoughts?

ryannikolaidis: Yea, I don't know that all of the StandardConnectorConfig options always pertain to _all_ of the connectors (for instance none of the download fields impact the local connector). That said, probably `remote_url` is a closer parallel for what we're trying to do here (overlap a collection of connectors) and it's more as you suggest (generic parameter that gets mapped to each connector-specific config). So totally good with that, as long as we only end up with one flag --recursive.

unstructured/ingest/main.py
potter-potter: Authorization token added. By default it should be none. But it is available now if needed.

```
if token=None (default), GCSFS will attempt to use your default gcloud credentials or, attempt to get credentials from the google metadata service, or fall back to anonymous access. This will work for most users without further action.

File: unstructured/ingest/main.py (hunk 3 above)
Comment (potter-potter) [reply unstructured/ingest/main.py:389]: One flag to rule them all.

File: unstructured/ingest/main.py (hunk 4 above)
Comment (potter-potter) [reply unstructured/ingest/main.py]: Done.

File: examples/ingest/google_cloud_storage/ingest.sh [hunk 6]
```diff
@@ -0,0 +1,16 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "$SCRIPT_DIR"/../../.. || exit 1
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+   --remote-url gs://unstructured_public/ \
+   --structured-output-dir gcs-output \
+   --num-processes 2 \
+   --recursive \
+   --verbose
```
ryannikolaidis, line 16: I pushed up a gcs test bucket with private and public paths, we should update to use that. I'll follow up with details directly.

unstructured/ingest/connector/dropbox.py
```diff
@@ -0,0 +1,78 @@
+from dataclasses import dataclass
```
ryannikolaidis, line 1: this will be a separate pr, right? assuming just some extra in-progress stuff in here?

unstructured/ingest/connector/fsspec.py
```diff
@@ -177,7 +188,25 @@ def initialize(self):
...
+            # fs.find will recursively walk directories
+            # "size" is a common key for all the cloud protocols with fs
+            return [
+                k
+                for k, v in self.fs.find(
+                    self.config.path_without_protocol,
+                    detail=True,
+                ).items()
+                if v.get("size") > 0
+            ]
```
ryannikolaidis, line 209: awesome!

unstructured/ingest/connector/dropbox.py
potter-potter: Damn. How did that turkey get in there. That file is for another PR.

potter-potter: Same with main_for_dropbox.py. I will remove this afternoon!

unstructured/ingest/main.py
ryannikolaidis: yea, probably not individuals, but definitely suspect think there will be enterprise users that will need this (and ...we need it for CI)

unstructured/ingest/main.py
```diff
@@ -172,7 +166,13 @@ def run(self):
...
     default=None,
     help="Remote fsspec URL formatted as `protocol://dir/path`, it can contain both "
-    "a directory or a single file. Supported protocols are: `s3`, `s3a`, `abfs`, and `az`.",
+    "a directory or a single file. Supported protocols are: `gcs`, `gs`, `s3`, `s3a`, `abfs` "
+    "and `az`.",
+)
+@click.option(
+    "--gcs-token",
+    default=None,
+    help="Token used to access Google Cloud. By default shouldn't be needed.",
```
ryannikolaidis, line 175: rather than "By default shouldn't be needed" might drop more info that you had in comment below. around using default gcloud credentials - attempt to get credentials from the google metadata service - fall back to anonymous access.

test_unstructured_ingest/test-ingest-google-cloud.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [[ "$(find test_unstructured_ingest/expected-structured-output/google-cloud-storage/ -type f -size +2 | wc -l)" -ne 5 ]]; then
+    echo "The test fixtures in test_unstructured_ingest/expected-structured-output/ look suspicious. At least one of the files is too small."
+    echo "Did you overwrite test fixtures with bad outputs?"
+    exit 1
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --metadata-exclude filename,file_directory,metadata.data_source.date_processed \
+    --remote-url gs://unstructured_public/ \
```
ryannikolaidis, line 16: let's use this test folder: gs://utic-test-ingest-fixtures
it has two folders `public` and `private`, where private will require that the auth token is used. hopefully can leverage what [was done in the google drive test](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-google-drive.sh#L8-L27) that I just added?

examples/ingest/google_cloud_storage/ingest.sh
ryannikolaidis: let's use gs://utic-test-ingest-fixtures

ryannikolaidis (commented):
just a few comments about switching over the test source and a nit about a description, but otherwise this is looking great!

test_unstructured_ingest/test-ingest-google-cloud.sh
potter-potter: Worked. But had to `cat` the cred info into the tempfile instead of `echo`

unstructured/ingest/main.py
potter-potter: Good point.

examples/ingest/google_cloud_storage/ingest.sh
potter-potter: This would require the user to have authentication token. Is that what we really want? Or do we want a no credits example?

ryannikolaidis: oh sorry, this is for the example. yes, no auth. I'll see if I can make a public bucket for you

unstructured/ingest/connector/dropbox.py
ryannikolaidis: Sweet, thanks!

test_unstructured_ingest/test-ingest-google-cloud.sh
ryannikolaidis: huh, weird. yea, as long as it works locally and in CI, that works for me.

ryannikolaidis: looking at CI, it doesn't look like linux is happy about the `cat` change. what's weird is that the way I had it before worked for me locally, maybe worth double checking?:

`cat: '***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''  ***'$'\n''***'$'\n': File name too long`

https://github.com/Unstructured-IO/unstructured/actions/runs/5330890729/jobs/9658103678

ryannikolaidis: ah, right, I see the confusion here. (should have noticed before). I think this wasn't working for you locally because you were setting GCP_INGEST_SERVICE_KEY to a filepath. GCP_INGEST_SERVICE_KEY is the actual key itself. This should works with echo on any platform.

ryannikolaidis (approved):
Pushed some changes to support updated fixture location, test key handling, expected output, and addressed some bugs.

Everything looks good!

PR #750, opened 2023-06-14
enhancements: add page numbers for word docs when available

### LINKED ISSUE #689: bug/<short-name>
Similar issue that #666 but for doc/docx files.

All elements extracted from partition_docx have the same page_number that is 1 (not the total number of pages)

**To Reproduce**

```python
elements = partition_docx(filename="my_docx.docx") #docx file with 10 pages
{elt.metadata.page_number for elt in elements} # {1} here
```

**Expected behavior**

```python
elements = partition_docx(filename="my_docx.docx") #docx file with 10 pages
elements[0].metadata.page_number # 1 here, first element should be on first page
elements[-1].metadata.page_number # 10 here
```

**Setup**

OS: mac
Unstructured version : 0.7.0
Python version : 3.9

MthwRobinson:
### Summary

Closes #689. Adds page numbers for Word documents when they are available. Note that the "page breaks" that you see when a Word document is rendered are inserted by the renderer and are not always present in the document. Some renderers insert `lastRenderedPageBreak` element when you save, which is what this looks for.

Additionally, in this PR we:

- Change the default page number from 1 to 0. Page number 0 is used in cases where page numbers are available. Changed from 1 to avoid having users think that page numbers were detected versus being added as a default value.
- Fixes errors in the Slack invite button and corrects information in the LangChain integration section

### Testing

```python
from unsturctured.partition.docx import partition_docx

elements = partition_docx(filename=filename)
# Should see page numbers. The 0 is for the header which is not associated with a page
[el.metadata.page_number for el in elements]

filename = "example-docs/fake.docx"
elements = partition_docx(filename=filename)
# Should be 0, no page numbers available
[el.metadata.page_number for el in elements]
```

### DIFF
--- a/test_unstructured_ingest/expected-structured-output/discord-ingest-channel/1099442333440802930.json
+++ b/test_unstructured_ingest/expected-structured-output/discord-ingest-channel/1099442333440802930.json
@@ -7,8 +7,7 @@
       "data_source": {},
       "filename": "1099442333440802930.txt",
       "file_directory": "discord-ingest-download",
-      "filetype": "text/plain",
-      "page_number": 1
+      "filetype": "text/plain"
     }
   },
   {
@@ -19,8 +18,7 @@
       "data_source": {},
       "filename": "1099442333440802930.txt",
       "file_directory": "discord-ingest-download",
-      "filetype": "text/plain",
-      "page_number": 1
+      "filetype": "text/plain"
     }
   },
   {
@@ -31,8 +29,7 @@
       "data_source": {},
       "filename": "1099442333440802930.txt",
       "file_directory": "discord-ingest-download",
-      "filetype": "text/plain",
-      "page_number": 1
+      "filetype": "text/plain"
     }
   },
   {
@@ -43,8 +40,7 @@
       "data_source": {},
       "filename": "1099442333440802930.txt",
       "file_directory": "discord-ingest-download",
-      "filetype": "text/plain",
-      "page_number": 1
+      "filetype": "text/plain"
     }
   }
 ]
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,13 +1,19 @@
-## 0.7.6-dev0
+## 0.7.6-dev1
 
 ### Enhancements
 
+* Adds support for page numbers in `.docx` and `.doc` numbers when user or renderer
+  created page braks are present.
```
qued, line 6:
```suggestion
* Adds support for page numbers in `.docx` and `.doc` when user or renderer
  created page breaks are present.
```

CHANGELOG.md
```diff
@@ -1,13 +1,19 @@
...
+
 ### Features
 
 * Updates `partition_docx` to include headers and footers in the output.
 
 ### Fixes
 
+* Page number defaults to `0` instead of `1` when page number is not present in the metadata.
+  A page number of `0` indicates that page numbers are not being tracked for the document.
```
qued, line 15: I wonder if `None` would be better than 0, but not a big deal either way.

unstructured/partition/docx.py
```diff
@@ -194,6 +208,21 @@ def _paragraph_to_element(paragraph: docx.text.paragraph.Paragraph) -> Optional[
         return element_class(text)
 
 
+def _element_contains_pagebreak(element) -> bool:
+    """Detects if an element contains a page break. Checks for both "hard" page breaks
+    (page breaks inserted by the user) and "soft" page breaks, which are somtimes
```
qued, line 213:
```suggestion
    (page breaks inserted by the user) and "soft" page breaks, which are sometimes
```

qued (approved):
LGTM other than nits and a comment.

CHANGELOG.md
MthwRobinson: Yeah I agree on `None`, I'll make that update

MthwRobinson: Alright got that in there. Turned out to be a little harder than expected because `_all_element_metadata` serves multiple purposes (it's the actual page incrementor when you use `partition_pdf` with `strategy="hi_res"`). May be worth a refactor at some point but seems to be working for now.

PR #757, opened 2023-06-15
Update Dockerfile to use multistage build and cache layers

rbiseck3:


Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN mkdir ${HOME}/.ssh && chmod go-rwx ${HOME}/.ssh \
 ENV PYTHONPATH="${PYTHONPATH}:${HOME}"
 ENV PATH="/home/usr/.local/bin:${PATH}"
 
+FROM base as deps
 # Copy and install Unstructured
 COPY requirements requirements
```
yuming-long, line 17: looks like the `pip install` below this didn't get cached from my end, I always have a step #14 runnning in my docker build log

yuming-long: okay its because `--no-cache-dir `

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:experimental
-FROM quay.io/unstructured-io/base-images:rocky8.7-1
+FROM quay.io/unstructured-io/base-images:rocky8.7-1 as base
```
cragwolfe, line 2: now `rocky8.7-2` :)

PR #761, opened 2023-06-15
fix malformed pptx issue

arms3:
Added a new test to check for the ability to partition a malformed PowerPoint file. Modified the `partition_pptx` function to skip processing shapes that are not on the actual slide, but only if they have top and left positions. Also modified `_order_shapes` function to handle cases where shapes do not have top or left positions.


MthwRobinson:
@arms3 - Thanks for the PR! Could you add a changelog entry under the "Fixes" heading? Otherwise LGTM

arms3:
Added a changelog entry, hopefully in the correct place.

MthwRobinson (approved):
LGTM!

PR #767, opened 2023-06-16
chore: rm old detectron2 install from makefile

cragwolfe:
Since the move to onnxruntime'd version of detetron2 in https://github.com/Unstructured-IO/unstructured/pull/662 , these makefile targets are no longer serving a purpose.


MthwRobinson (approved):
LGTM

PR #769, opened 2023-06-16
chore: DRY ingest connectors

ryannikolaidis:
This PR mostly doesn't change functionality but rather is a refactor of Connector/IngestDoc classes to minimize the amount of repetition. In some places (e.g. `write_result`) we were repeating the exact same code block across all connectors; these were moved 1:1 to the respective base class. In some cases we were _mostly_ repeating code (same intent, but varied approaches) (e.g. `has_output` where we were checking file size in different ways, with the same goal of proving this is a non-empty file on disk), in these cases a single approach was chosen and moved to the base class. In other cases, code was repeated across most but not all connectors (e.g. cleaning functions for IngestDocs and Connectors were repeated in all but the local connector); in these cases mixins were leveraged to optionally re-use these definitions.

- add skip_if_file_exists decorator to remove duplicate check/log/return calls
- define `write_result` on BaseIngestDoc (we are repeating this consistently across all subclasses)
- declare `_output_filename` as an abstract property on BaseIngestDoc (required internally by write_result)
- define `has_output` on BaseIngestDoc (we were effectively do the same thing across all subclasses with subtle but unnecessary differences)
- add IngestDocCleanupMixin mixin such that IngestDocs that need to call `cleanup_file` (all but local doing so the same way) can add this mixin instead
- add ConnectorCleanupMixin mixin such that Connectors can use this mixin rather than defining a `cleanup` function (all but local were doing so the same way)
- bonus: move Slack-specific imports into the function that requires respective dependencies


cragwolfe (approved):
Very nice!

PR #772, opened 2023-06-16
fix: enable `partition_html` to grab content outside of `<article>` tags

### LINKED ISSUE #770: bug/`partition_html` misses deeply nested content in HTML documents
**Describe the bug**
For the following website, `partition_html` does not extract important content such as names and titles in the "our leaders" section.

- https://www.abbvie.com/who-we-are/our-leaders.html

**To Reproduce**
```python

url = "https://www.abbvie.com/who-we-are/our-leaders.html"
```

**Expected behavior**
`partition_html` should extract the elements mentioned above.

 - `unstructured==0.7.6`

**Additional context**
- This may have to do with the content in question being deeply nested within `<div>`s.

MthwRobinson:
### Summary

Closes #770. Adds a `html_assemble_articles` flag to `partition_html` that allows `partition_html` to grab content from outside of `<article>` tags when `<article>` tags are present. By default, if `<article>` tags are present, `partition_html` only grabs content from with the `<article>` tags to avoid processing things like advertisements on news webpages.

### Testing

Using the html from [this page](https://www.abbvie.com/who-we-are/our-leaders.html), the following code should return several elements matching a name that was previously missing in the output.

```python

filename = "abbvie.html"
elements = partition_html(filename=filename, html_assemble_articles=False)
[x for x in elements if "Gonzalez" in x.text]
```


unstructured/documents/html.py
```diff
@@ -407,9 +416,12 @@ def _find_main(root: etree.Element) -> etree.Element:
     return main_tag_elem if main_tag_elem is not None else root
 
 
-def _find_articles(root: etree.Element) -> List[etree.Element]:
+def _find_articles(root: etree.Element, assemble_articles: bool = True) -> List[etree.Element]:
```
cragwolfe, line 419: could there be an auto mode which does 
`root.findall(".//article")` if `article`s exist, otherwise use the root?

cragwolfe: i left this comment on the wrong function, but idea generally holds. Ideally, `partition_html` does the opinionated right thing by default for news-page html pages vs. plain old html pages.

qued (approved):
LGTM!

cragwolfe (changes requested):
Requesting changes mainly to make sure my comment is not overlooked.

theZaX:
Would like to add +1 to this. Thanks for working on this. Seems the broad majority of websites of interest for us don't follow the <article> convention so this will be a great improvement. I'm not sure how your CI works but any estimate on when this will be deployed to the python package? I'm using the Unstructured-API package.

unstructured/documents/html.py
MthwRobinson: That's what it does now actually, it first looks for `.//article`, then `.//div[@itemprop='articleBody']` and if it doesn't find either then the returns `root`.

MthwRobinson:
@theZaX - Updated the default to `html_assemble_article=False` since, as you point out, the majority of websites don't use the `<article>` contention. We'll likely get this released in the `unstructured` package later today.

@cragwolfe - Do you know how long this will take to make it into the API?

cragwolfe (approved):

cragwolfe:
> @cragwolfe - Do you know how long this will take to make it into the API?

It generally takes 1-3 business days. We can prioritize the next release bump, CC @yuming-long , @awalker4

awalker4:
For reference - this is making it out to `unstructured-api` via [this pr](https://github.com/Unstructured-IO/unstructured-api/pull/131).

PR #774, opened 2023-06-17
feat: coordinate systems

qued:
Added the `CoordinateSystem` class for tracking the system in which coordinates are represented, and changing the system if desired.

#### Testing:

```python
from unstructured.documents.coordinates import PixelSpace, RelativeCoordinateSystem

coordinates = ((10, 10), (10, 100), (200, 100), (200, 10))
coordinate_system = PixelSpace(width=850, height=1100)
element = Element(coordinates=coordinates, coordinate_system=coordinate_system)
print(element.coordinate_system)
print(element.coordinates)
element.convert_coordinates_to_new_system(RelativeCoordinateSystem(), in_place=True)
# Should now be in terms of new coordinate system
print(element.coordinate_system)
print(element.coordinates)

```


unstructured/file_utils/filetype.py
```diff
@@ -489,13 +499,16 @@ def wrapper(*args, **kwargs):
             include_metadata = params.get("include_metadata", True)
             if include_metadata:
                 metadata_kwargs = {
-                    kwarg: params.get(kwarg) for kwarg in ("include_page_breaks", "filename", "url")
+                    kwarg: params.get(kwarg) for kwarg in ("filename", "url", "text_as_html")
```
MthwRobinson, line 502: Do we not need `include_page_breaks` in here any longer? Just double checking that was intentional

MthwRobinson (commented):
Implementation and test code looks good to me. Main question is on the page break changes that were in there. Do the coordinate system updates impact the page break logic at all, or did you stumble into something there that wasn't working quite right?

MthwRobinson (approved):
LGTM from my end. Caught up on the page break comments offline.

unstructured/file_utils/filetype.py
qued: Since `_add_element_metadata` doesn't handle page breaks anymore and `include_page_breaks` isn't actually stored in the metadata, we don't need to pass it.

PR #778, opened 2023-06-20
fix: check for xml attribute when identifying pagebreaks

### LINKED ISSUE #775: error in docx partition
After updating from 0.7.5 to 0.7.6, I'm getting an error when trying to parse a docx file.

```bash
'lxml.etree._Element' object has no attribute 'xml'
``` 

I tracked the error to the following line. Apparently not every element has an `xml` attribute:

https://github.com/Unstructured-IO/unstructured/blob/743482b6d3ea8dd03bf029c9c7f62ceb095e0dfc/unstructured/partition/docx.py#L224-L225

MthwRobinson:
### Summary

Closes #775. First checks for the `xml` attribute on the document elements before checking for pagebreaks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.7-dev1
+## 0.7.7
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


qued (approved):
LGTM!

PR #790, opened 2023-06-21
enhancement: clean pdf elements (bump unstructured-inference)

qued:
* More deterministic element ordering when using `hi_res` PDF parsing strategy (from unstructured-inference bump to 0.5.4)
* Make large model available (from unstructured-inference bump to 0.5.3)
* Combine inferred elements with extracted elements (from unstructured-inference bump to 0.5.2)


unstructured/partition/pdf.py
```diff
@@ -32,6 +33,8 @@
 from unstructured.partition.text import element_from_text, partition_text
 from unstructured.utils import requires_dependencies
 
+RE_MULTISPACE = re.compile(r"\s+")
```
cragwolfe, line 36: it looks like this is only used in the context of multiline replace below (line 236).

so to make the intent more clear, update to:

```
    RE_MULTISPACE_INCLUDING_NEWLINES = re.compile(r"\s+", re.DOTALL)
```
and use that instead?

unstructured/partition/pdf.py
```diff
@@ -226,8 +229,12 @@ def _partition_pdf_or_image_local(
             ocr_languages=ocr_languages,
             extract_tables=infer_table_structure,
         )
+    elements = document_to_element_list(layout, include_page_breaks=include_page_breaks)
+    for el in elements:
+        if isinstance(el, Text):
+            el.text = re.sub(RE_MULTISPACE, " ", el.text.replace("\n", " ")).strip()
 
-    return document_to_element_list(layout, include_page_breaks=include_page_breaks)
+    return [el for el in elements if not isinstance(el, Text) or el.text]
```
cragwolfe, line 237: can this code be rewritten to not require staring it a bit to understand the intent?

since we're already looping through `elements` in line 233 anyway, why not just append to `return_elements` there and then the logic of what is getting returned should be nice and clear?

cragwolfe:
Comment on the .json output for (so far):
```
test_unstructured_ingest/expected-structured-output/s3-small-batch/small-pdf-set/recalibrating-risk-report.pdf.json
```

This PR is introducing a bunch of `Image` tags that are all noise on Page 1. Page 3 looks a little better in that there is now:
```
    "type": "UncategorizedText",
    "text": "Executive Summary"
```

which was not there before (though it'd probably be better characterized as a `Title`).

unstructured/partition/pdf.py
rbiseck3: In general, if we're dropping anything without text, won't this also remove all page breaks being injected by `document_to_element_list`?

cragwolfe: ah, not if `text` is not empty, which i think it wouldn't be if it is `"<Page Break>"` (which it probably shouldn't be in the first place).

i think this will also be clearer to discuss if the logic is in the loop and not list comprehension to begin with.

CC @qued , soliciting comments on this thread and my other comment

rbiseck3: Ah didn't realize the text of the PageBreak element was itself `<PAGE BREAK>`. I looked at the constructor and saw ```
text: Optional[str] = None,
```
Thinking there would be no text associated with that element by default.

File: unstructured/partition/pdf.py [hunk 3]
```diff
@@ -322,18 +329,19 @@ def _process_pdfminer_pages(
                     element.metadata = metadata
                     page_elements.append(element)
 
+        if include_page_breaks:
+            elements.append(PageBreak())
+
```
rbiseck3, line 334: Do we not want the page break to be appended after the content of the page is sorted? Otherwise this can be modified from being at the end of the page.

qued: You're right, updated.

test_unstructured_ingest/expected-structured-output/azure-blob-storage/IRS-form-1987.png.json
```diff
@@ -267,7 +267,7 @@
       "filetype": "image/png",
       "page_number": 1
     },
-    "text": "We ask for this information to carry out the\nInternal Revenue laws of the United States. We\nneed it to ensure that taxpayers are complying\nwith these laws an¢ to allow us to figure and\n\ncollect the right amount of tax. You are required\nto give us this information,\n"
+    "text": "We ask for this information to carry out the Internal Revenue laws of the United States. We need it to ensure that taxpayers are complying with these laws an¢ to allow us to figure and collect the right amount of tax. You are required to give us this information,"
```
cragwolfe, line 270: it looks like we're overly aggresive stripping out newlines:
image

cragwolfe: although some use cases may prefer this, hmm

rbiseck3:
LGTM

rbiseck3 (approved):

PR #796, opened 2023-06-22
fix: enable `partition_pdf` to recursively grab text with fast strategy

### LINKED ISSUE #784: bug/fast strategy skips pages on some PDFs
**Describe the bug**
We've noticed that `partition_pdf` with the `"fast"` strategy skips pages on some PDFs;

**To Reproduce**
Using the attached example document, do the following:

```python

filename = "RELIANCE.pdf"
```

**Expected behavior**
`partition_pdf` should process all of the pages.

 - `unstructured==0.7.7`
[RELIANCE.pdf](https://github.com/Unstructured-IO/unstructured/files/11818617/RELIANCE.pdf)

MthwRobinson:
### Summary

Closes #784. Updates the `"fast"` strategy for `partition_pdf` to recursively grab text in PDFs that's nested in non-text containers. Noted one potential future enhancement, which is to grab text from images (potentially through OCR), but we'll leave that for a future PR.

Note, in the example PDF (the first several pages of a user provided example doc), I'm seeing some duplication of content between pages. I get the same behavior when I use `pdfminer` directly. Can't tell whether this PDF just has wonky formatting or if there's an issue with the text extraction code. 

### Testing

The following PDF should have multiple pages. Previously the fast strategy only extracted content from the first page.

```python

filename = "example-docs/reliance.pdf"
elements[-1].metadata.page_number
```


qued (approved):
LGTM!

PR #804, opened 2023-06-24
ci: fix amd build issue

ryannikolaidis:
AMD images were building but failing to load in CI (meaning there weren't able to test and publish). This seems to be an issue with buildx. The solution is to use the "docker" driver. 

[Here is a reference ](https://github.com/Unstructured-IO/unstructured/actions/runs/5366686113/jobs/9736247833)to test commit with build+test steps enabled on push.


cragwolfe (approved):
🙌 🙌

PR #807, opened 2023-06-25
enhancement: clean pdf elements (bump unstructured-inference) <- Ingest test fixtures update

Unstructured-DevOps:
This pull request includes updated ingest test fixtures.
Please review and merge if appropriate.


cragwolfe (commented):
Looks like the delta here is removing `PageBreak` types like:
 ```
  {
    "type": "PageBreak",
    "element_id": "5ea24028ea5addabb8f07dfff681501d",
    "metadata": {
      "data_source": {},
      "filetype": "application/pdf"
    },
    "text": "<PAGE BREAK>"
  },
``` 
Perhaps it is an upstream (unstructured-inference) issue that the text in those `PageBreaks` had text of "`<PAGE BREAK`>" to begin with rather than ""?

qued (approved):

PR #810, opened 2023-06-26
ci: fix arm build issue with docker driver

ryannikolaidis:
[This PR](https://github.com/Unstructured-IO/unstructured/pull/804) introduced a change to use the "docker" driver instead of the "docker-container" driver that docker/setup-buildx-action@v1 defaults to. This intermittently hits an issue with arm builds where the build fails with `exec /bin/sh: exec format error`. Because the original issue only affected AMD builds, this PR reverts the change specifically for ARM builds by making the driver condition on the architecture specified.

[Here](https://github.com/Unstructured-IO/unstructured/actions/runs/5383408561/jobs/9770135678) is a run showing this conditional driver logic working and passing the image build steps for both architectures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,246 +1,246 @@
-# name: CI
+name: CI
 
-# on:
-#   push:
-#     branches: [ main ]
-#   pull_request:
-#     branches: [ main ]
+on:
+  push:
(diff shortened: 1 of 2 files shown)


cragwolfe (approved):
🙌

PR #814, opened 2023-06-27
chore: refactor ingest tests

ryannikolaidis:
- Adds reusable validation scripts (check-x.sh) to minimize repeated (or near-repeated) code and create one source of truth
- Restructures the location of download and output folders such that they are nested in the test_unstructured_ingest directory
- Adds gitignore for output folders / files to avoid them accidentally getting checked into the repository
- Construct paths as reusable variables declared at top of scripts
- Sort order of flag for ingest calls, across all tests (this makes it easier to parse at a glance)
- OVERWRITE_FIXTURES removes all old fixtures for path to guarantee no stale results are left behind
- Bonus: don't check/exit on expected number of expected outputs when OVERWRITE_FIXTURES is true
- Bonus: exclude file_directory from Slack and Discord test scripts (match convention in all others)

Reference for new tree:

* test_unstructured_ingest
  * download
    * biomed-api
    * biomed-path
    * ...
  * expected-structured-output
    * biomed-api 
    * biomed-path
    * ...
  * structured-output
    * biomed-api
    * biomed-path
    * ...
  * ...
  
  Tested restructure by regenerating the fixtures ([here](https://github.com/Unstructured-IO/unstructured/pull/829)) with the new locations using the update test fixtures PR and validating the file changes (they are all recognized as file moves)


test_unstructured_ingest/test-ingest-github.sh
```diff
@@ -2,61 +2,35 @@
...
 elif [[ "$CI" == "true" ]]; then
     echo "Warning: GH_READ_ONLY_ACCESS_TOKEN is not defined in the CI environment."
     echo "This can lead to intermittent failures in test-ingest-github.sh, as non-auth'ed"
     echo "requests are severely rate limited by GitHub."
     echo
-    if [ "$(( RANDOM % 10))" -lt 1 ] ; then
-        # NOTE(crag): proper fix is being tracked here: https://github.com/Unstructured-IO/unstructured/issues/306
-        echo "Skipping ingest 90% of github ingest tests to avoid rate limiting issue."
-        exit 0
-    fi
```
ryannikolaidis, line 23: we resolved the issue by adding the needed key, so this _shouldn't_ ever be missing in CI. tempted to just fail here now (would rather know if that key was missing in CI?

test_unstructured_ingest/check-num-files-expected-output.sh
```diff
@@ -0,0 +1,24 @@
...
+set +e
+
+EXPECTED_NUM_FILES=$1
+OUTPUT_FOLDER_NAME=$2
+EXPECTED_SIZE=$3
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+EXPECTED_OUTPUT_DIR=$SCRIPT_DIR/expected-structured-output/$OUTPUT_FOLDER_NAME
+NUM_FILES=$(find "$EXPECTED_OUTPUT_DIR" -type f -size +"$EXPECTED_SIZE" | wc -l)
+
+if [[ "$NUM_FILES" -ne "$EXPECTED_NUM_FILES" && "$OVERWRITE_FIXTURES" == "false" ]]; then
```
ryannikolaidis, line 19: adds "$OVERWRITE_FIXTURES" == "false" (we weren't doing this before, but if we're overwriting I don't think we need to exit here, in fact it's probably more problematic if we do)

test_unstructured_ingest/test-ingest-against-api.sh
```diff
@@ -2,22 +2,18 @@
 
 set -e
 
-SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
```
ryannikolaidis, line 5: I ran into an issue with this when called from check-X scripts. The updated form does work everywhere, so updated across the board for consistency.

test_unstructured_ingest/expected-structured-output/local-single-file/example-docs/english-and-korean.png.json
```diff
@@ -27,13 +27,13 @@
...
-    "type": "NarrativeText",
-    "element_id": "577b0ea5cd4f62a19b9d14dd0bd7272e",
+    "type": "UncategorizedText",
+    "element_id": "23a37aa2d5f39d5e2275dec011be76be",
     "metadata": {
       "data_source": {},
       "filetype": "image/png"
     },
-    "text": "안녕하세요, 저 희 는 YGEAS 그룹 TREASUREWH HARUTOM|2] 팬 입니다. 팬 으 로서, HARUTO 씨 받 는 대 우 에 대해 의 구 심 과 불 공 평 함 을 LRU, 이 메 일 을 통해 저 희 의 의 혹 을 전 달 하여 귀 사 의 진지한 고 민 과 적극적인 답 변 을 받을 수 있 기 를 바랍니다."
+    "text": "StS ofAl2, AS|E YGEAS 1B TREASUREMH HARUTOM| 2] BHEYLICH, BHO AY, HARUTO M BE = WSO Hol Wat SSGstsS LRU, O| Wil BS SH ASP ASS AZopO} HAS] TAI St a Bat FSAel GAS HS + U7/S HLICh."
```
jasonbot, line 36: Was it supposed to lose the Korean text?

ryannikolaidis: ohh, good eye! I don't think this was caused by my pr but clearly something has broken along the way

ryannikolaidis: will be resolved once we bring in [this fix](https://github.com/Unstructured-IO/unstructured/pull/838)

awalker4:
The diffs are a lot to digest but the refactor makes sense to me! I'm getting an output change when I run locally (in the container), but I suppose I should get back in the habit of running on x86. I'll give that a shot, but otherwise LGTM.

awalker4 (approved):

ryannikolaidis:
> The diffs are a lot to digest but the refactor makes sense to me! I'm getting an output change when I run locally (in the container), but I suppose I should get back in the habit of running on x86. I'll give that a shot otherwise LGTM.

Awesome, thanks! Yea, I know it's a _lot_ of diffs. Hopefully for the better :). Yea, output change to be expected on ARM architecture (even in a container). I think there might also be a bug where there's an issue with ordering outside of CI. These expected results were generated by the GHA update workflow so hopefully at least valid against the same mechanism that checks them. 🤷

Issue #809, opened 2023-06-26
Connector for Elasticsearch


ahmetmeleq:
Create a data connector that pulls documents from Elasticsearch, stores them locally (at least temporarily for processing), and runs them through [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py).

See [Adding Data Connectors](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#adding-data-connectors) to get started.

Make sure to include a link to this issue when submitting a PR.

Definition of Done

- The [data connector checklist](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist) has been completed.
- [PR checklist 1](https://github.com/Unstructured-IO/docs-and-arch/blob/main/Pull-Requests.md#before-creating-pr) and [PR checklist 2](https://github.com/Unstructured-IO/docs-and-arch/blob/main/Pull-Requests.md#pr-guidelines) have been completed.
- The connector is able to process a single document.
- The connector is able to process all documents in an Elasticsearch index.
- CI tests are able to automatically create a local elasticsearch index is and fill it for testing.
- Values in the JSON structured elasticsearch document are serialised to text, and values are able to be custom-selected based on their keys, via querying such as with jq.
- For now, it is OK to process only doc types that [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py) is capable of processing. Blob files are left out for now.
- Bonus points: multiprocessing optimisations such as one python client per process, and other kinds of optimisations.

PR #817, opened 2023-06-27
feat: elasticsearch connector

ahmetmeleq:
This connector: 
- takes an Elasticsearch cluster URL, an index name, and other options,
- to pull specific fields from all documents in an index, 
- and store each document as a text file,
- which are then processed by unstructured auto into the expected output format.

Related to: https://github.com/Unstructured-IO/unstructured/issues/809

To test the changes, make the necessary setups and run _test-ingest-elasticsearch.sh_ within _test_unstructured_ingest_ .


unstructured/ingest/main.py
```diff
@@ -86,14 +86,15 @@ def run(self):
...
+
         with mp.Pool(
             processes=self.num_processes,
             initializer=ingest_log_streaming_init,
             initargs=(logging.DEBUG if self.verbose else logging.INFO,),
         ) as pool:
             pool.map(self.doc_processor_fn, docs)
 
-        self.cleanup()
+        # self.cleanup()
```
ryannikolaidis, line 97: assume we should re-enable this?

unstructured/ingest/main.py
```diff
@@ -385,6 +386,22 @@ def run(self):
...
+@click.option(
+    "--elasticsearch-index-name",
+    default=None,
+    help="Name for the elasticsearch index to pull data from",
+)
+@click.option(
+    "--jq-query",
+    default=None,
+    help="JQ query to get a subset of the fields from a JSON document, before processing it. \
+        For a group of JSON documents, it assumes that all of the documents have the same schema",
```
ryannikolaidis, line 403: do we imagine jq-query being used in other connectors? (I'm guessing so) if _not_ we should prefix with elasticsearch. but assuming so, let's go ahead and indicate that this is currently supported only by the elasticsearch connector (so folks don't try to use with other connectors)

examples/ingest/elasticsearch/ingest.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# If cluster has not got live, exit after a certain number of tries
+if [ "$status_code" -ne 200 ]; then
+  echo "Cluster took an unusually long time to create (>25 seconds). Expected time is around 10 seconds. Exiting."
+fi
+
+# Kill the container so the script can be repeatedly run using the same ports
+docker stop "$container_id"
+# Kill even when there's an error from the previous commands
+trap 'docker stop "$container_id"' ERR
```
ryannikolaidis, line 58: should probably add this before the commands that might error, right?

examples/ingest/elasticsearch/wiki_movie_plots_small.csv
```diff
@@ -0,0 +1,100 @@
+Release Year,Title,Origin/Ethnicity,Director,Cast,Genre,Wiki Page,Plot
+1901,Kansas Saloon Smashers,American,Unknown,,unknown,https://en.wikipedia.org/wiki/Kansas_Saloon_Smashers,"A bartender is working at a saloon, serving drinks to customers. After he fills a stereotypically Irish man's bucket with beer, Carrie Nation and her followers burst inside. They assault the Irish man, pulling his hat over his eyes and then dumping the beer over his head. The group then begin wrecking the bar, smashing the fixtures, mirrors, and breaking the cash register. The bartender then sprays seltzer water in Nation's face before a group of policemen appear and order everybody to leave.[1]"
```
ryannikolaidis, line 2: I think we could chop this down to say the top 10 or so?

examples/ingest/elasticsearch/ingest.sh
```diff
@@ -0,0 +1,58 @@
...
+  # Process the files only when the ES cluster is live
+  if [ "$status_code" -eq 200 ]; then
+    echo "Cluster is live."
+
+    python examples/ingest/elasticsearch/elasticsearch_cluster.py
+
+    PYTHONPATH=. ./unstructured/ingest/main.py \
+            --elasticsearch-url http://localhost:9200 \
+            --elasticsearch-index-name movies \
+            --jq-query '{ethnicity, director}' \
```
ryannikolaidis, line 39: how about we add plot in here? since that's the beefiest field?

unstructured/ingest/main.py
ahmetmeleq: That's right, I missed this

unstructured/ingest/main.py
ahmetmeleq: Makes sense. If there are other popular stores that represent documents in JSON format I think it'd be beneficial, or even necessary for them. I'll add the note that this is currently only supported in elasticsearch.

examples/ingest/elasticsearch/wiki_movie_plots_small.csv
ahmetmeleq: Sounds good 👍, I was also having doubts on the exact number of these

examples/ingest/elasticsearch/ingest.sh
ahmetmeleq: Sounds good 👍

test_unstructured_ingest/test-ingest-elasticsearch.sh
```diff
@@ -0,0 +1,76 @@
...
+
+  # Send a GET request to the cluster health API
+  response=$(curl -s -o /dev/null -w "%{http_code}" "$url")
+  status_code="$response"
+
+  # Process the files only when the ES cluster is live
+  if [ "$status_code" -eq 200 ]; then
+    echo "Cluster is live."
+
+    python examples/ingest/elasticsearch/elasticsearch_cluster.py
```
ryannikolaidis, line 26: can we move all of this code to launch the es cluster into its own shell script? 
I think it'll be cleaner, but extra benefit that then we don't repeat this in the example ingest.sh as well

ryannikolaidis, line 26: it's kind of awkward that we rely on a module that lives in `examples/` in order to run a test.
that said, I don't have great alternatives. I think the best I can think of is: create a new folder in `scripts/`, that folder should contain the shell script that launches / waits / retries (that I mentioned in the comment above this), `es_cluster_config.py`, and `elasticsearch_cluster.py`. then both the examples/../ingest.sh and `test-ingest-elasticsearch.sh` can simply call that one script in one line and afterwards have a running instance to test against.

ryannikolaidis: EDIT: actually, include all the way up to line 6 (so, all logic needed to launch it)

test_unstructured_ingest/test-ingest-elasticsearch.sh
```diff
@@ -0,0 +1,76 @@
...
+fi
+
+OVERWRITE_FIXTURES=${OVERWRITE_FIXTURES:-false}
+
+# Kill the container so the script can be repeatedly run using the same ports
+docker stop "$container_id"
+
+# # Kill even when there's an error from the previous commands
+trap 'docker stop "$container_id"' ERR
+
```
ryannikolaidis, line 55: I think you'll want to trap earlier, if we've gotten here we've already attempted to stop the docker container

unstructured/ingest/main.py
ryannikolaidis: yep, totally agree

ryannikolaidis (commented):
looking good! dropped some notes. another heads up, you'll want to rebase/merge with main to pull in changes from [this PR](https://github.com/Unstructured-IO/unstructured/pull/769) which refactors some of the connector code (largely to avoid repeating as much across connectors)

test_unstructured_ingest/test-ingest-elasticsearch.sh
ryannikolaidis: EDIT: and code lower down where we exit on retries and stop the docker container. basically everything to encapsulate launching the test es cluster

ahmetmeleq: I isolated all code in another script, except the line shutting the container down

Had to trap the shut-down in this script since we only want it shut down after this script exits

If we'd like to isolate that line as well, I can create another script solely for shutting the container down

unstructured/ingest/main.py
ahmetmeleq: Changes applied!

unstructured/ingest/main.py
ahmetmeleq: Changes applied! (help message updated, saying this is currently only supported for elasticsearch)

examples/ingest/elasticsearch/ingest.sh
ahmetmeleq: Changes applied! (signal code changed to EXIT, the line moved to the earliest place possible)

examples/ingest/elasticsearch/wiki_movie_plots_small.csv
ahmetmeleq: Changes applied! (chopped down to 31 examples, which leaves 10 docs after processing. [that reduction in numbers should be due to dropna() in pandas])

test_unstructured_ingest/test-ingest-elasticsearch.sh
ahmetmeleq: Changes applied! (all isolated code moved into `scripts` folder, both  `examples/../ingest.sh` and `test-ingest-elasticsearch.sh` calling the isolated code from there. shellcheck ignored for those particular script calls since it gave a SC1091 warning.)

test_unstructured_ingest/test-ingest-elasticsearch.sh
ahmetmeleq: Changes applied! (signal code changed to EXIT, the line moved to the earliest place possible)

ahmetmeleq:
All requested changes were applied! Should be ready to go after a review.

examples/ingest/elasticsearch/wiki_movie_plots_small.csv
ryannikolaidis: perfect, thanks!

test_unstructured_ingest/test-ingest-elasticsearch.sh
ryannikolaidis: nope, I think that makes perfect sense!

ryannikolaidis:
Heads up, pushed a few changes:

* add --name to docker container (this allows us to not need the id_log_filepath juggling)
* update test to leverage refactored test flow
* removed unnecessary chmod and call with `sh` rather than sourcing the script
* add query hash to filename (because we currently use filename to determine whether to reprocess a file and the output will be different depending on the query used, I wanted to guarantee this would be unique to the query)
* add "_" prefix to internal use functions
* removed functions handled by base and mixin class definitions
* replace es.search with es.get
* update expected results using PR flow (it looks like they had cleaning applied that wouldn't match the diff in CI)

ryannikolaidis (approved):
LGTM!

PR #818, opened 2023-06-27
enhancement: `max_partition` kwarg for limiting element size

### LINKED ISSUE #808: bug/very large text partitions
**Describe the bug**
I am using `partition_text` with this file and getting only one `narrative_text` segment back. https://storage.googleapis.com/mindsmith/uploads%2Ff4403857-5e6d-41b7-83d1-cb1c68db516a%2F369378ff-20bf-470f-ba47-809fbe9ff98d.txt
This breaks our downstream embedding process, and is not a very good partitioning of the underlying content.

**To Reproduce**
It seems like the partition_text is simply using \n\n to determine where to partition the text. So any text document that doesn't use that as a way to separate content will have very large partitions / only one partition.

**Expected behavior**
Some sort of partitioning that breaks apart lines that are long or some sort of max partition size. Looking at the document it seems like one could have a more intelligent way of partitioning it, maybe something to do with line length.

In general since partitioning often takes place in same context as embedding it might make sense to have something like a `max_partition` size. Since embedding models can only take a maximum size of content and the semantic search works better with smallish partitions.

MthwRobinson:
### Summary

Closes #808. Adds a `max_partition` parameter to `partition_text`, `partition_pdf`, `partition_email`,  `partition_msg` and `partition_xml` that sets a limit for the size of an individual document elements. Defaults to `1500` for everything except `partition_xml`, which has a default value of `None`.

### Testing

The following should show multiple elements. Previously everything was one giant `NarrativeText` element.

```python

filename = "example-docs/norwich-city.txt"
elements = partition_text(filename=filename)
len(elements)
```


docs/source/bricks.rst
```diff
@@ -752,6 +781,12 @@ If ``xml_keep_tags=True``, the function returns tag information in addition to t
 
   elements = partition_xml(filename="example-docs/factbook.xml", xml_keep_tags=False)
 
+``partition_xml`` includes a ``max_partition`` parameter that indicates the maximum character length for a document element.
+The default value is ``None``, meaning no maximum character count is applied and
+``partition_xml`` will process the whole XML file as a single element.
```
qued, line 786: Wait, is it true that it will definitely partition the document as a single element, or is that just a possibility?

unstructured/partition/email.py
```diff
@@ -241,6 +242,9 @@ def partition_email(
         other: "text/plain"
     encoding
         The encoding method used to decode the text input. If None, utf-8 will be used.
+    max_partition
+        The maximum number of characters to include in a partition. If None is passed,
+        no maximum is applied. Only applies if processing the text/plain content.
```
qued, line 247:
```suggestion
        no maximum is applied. Only applies if processing text/plain content.
```

qued (approved):
LGTM, one question and a few suggestions.

docs/source/bricks.rst
MthwRobinson: Yeah you're right, if there are line breaks in there it definitely could get split up into different elements. They might just be weird elements with it being XML and all. Maybe we default that one to 1500 too then?

MthwRobinson: Updated XML to default to 1500 as well.

PR #824, opened 2023-06-27
feat: choose model

qued:
Added the ability to select the `hi_res` model via the environment variable `UNSTRUCTURED_HI_RES_MODEL_NAME`. Variable must be a string that matches up with a model name defined in `unstructured_inference`.

Also removed code related to old `unstructured_inference` API which has been removed from currently pinned version of `unstructured-inference` and is no longer running as a service.

#### Testing:

```python
import os
# Default model
out = partition_pdf("example-docs/layout-parser-paper-fast.pdf", strategy="hi_res")
print(len(out))

os.environ["UNSTRUCTURED_HI_RES_MODEL_NAME"] = "yolox"
out_yolox = partition_pdf("example-docs/layout-parser-paper-fast.pdf", strategy="hi_res")
# Length should be different than above, indicating inference was done with a different model
print(len(out_yolox))

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
-## 0.7.10-dev1
+## 0.7.10-dev2
 
 ### Enhancements
 
 * DRY connector refactor
 
 ### Features
+
+* `hi_res` model for pdfs and images is selectable via environment variable.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.10-dev2
+## 0.7.10-dev3
 
 ### Enhancements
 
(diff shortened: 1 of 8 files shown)


cragwolfe (approved):
Tested, works!

Also worth mentioning the extra stuff being removed for sake of completeness in the description.

cragwolfe (commented):
please update https://github.com/Unstructured-IO/unstructured/blob/7f0f5fa/unstructured/ingest/doc_processor/generalized.py#L14 as well

cragwolfe (changes requested):
requesting changes per prior comment

qued:
@cragwolfe Good catch, updated.

cragwolfe (approved):

PR #834, opened 2023-06-28
feat: Adds OneDrive connector

rvztz:


test_unstructured_ingest/test-ingest-onedrive.sh
```diff
@@ -0,0 +1,46 @@
...
+OVERWRITE_FIXTURES=${OVERWRITE_FIXTURES:-false}
+
+set +e
+
+# to update ingest test fixtures, run scripts/ingest-test-fixtures-update.sh on x86_64
+if [[ "$OVERWRITE_FIXTURES" != "false" ]]; then
+
+    cp google-drive-output/* test_unstructured_ingest/expected-structured-output/google-drive-output/
+
+elif ! diff -ru test_unstructured_ingest/expected-structured-output/google-drive-output google-drive-output ; then
```
ryannikolaidis, line 30: update to point to one-drive instead of google-drive everywhere

unstructured/ingest/connector/one_drive.py
```diff
@@ -0,0 +1,211 @@
...
+            raise ValueError(
+                f"Extension not supported. "
+                f"Value MUST be one of {', '.join([k for k in EXT_TO_FILETYPE if k is not None])}.",
+            )
+        self._set_download_paths()
+
+    def _set_download_paths(self) -> str:
+        """Sets download and output directories"""
+        dpath = Path(f'{self.standard_config.download_dir}')
+        opath = Path(f'{self.standard_config.output_dir}')
```
ryannikolaidis, line 76: nit: I would just call these download_path and output_path to be consistent with convention elsewhere

unstructured/ingest/connector/one_drive.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if pref := self.file.get_property('parentReference','').path.split(':')[-1]:
```
ryannikolaidis, line 78: nit: just call this parent_ref or parent_reference (pref could easily be read as preference)

unstructured/ingest/main.py
```diff
@@ -377,6 +377,31 @@ def run(self):
...
+)
+@click.option(
+    "--ms-client-cred",
+    default=None,
+    help="Microsoft App client secret"
+)
+@click.option(
+    "--ms-authority-url",
+    default='https://login.microsoftonline.com',
+    help="Authentication token provider for MS apps, default is https://login.microsoftonline.com"
```
ryannikolaidis, line 393: nit
```suggestion
    help="Authentication token provider for Microsoft apps, default is https://login.microsoftonline.com"
```

unstructured/ingest/connector/one_drive.py
```diff
@@ -0,0 +1,211 @@
...
+        
+        if self.ext not in EXT_TO_FILETYPE.keys():
+            raise ValueError(
+                f"Extension not supported. "
+                f"Value MUST be one of {', '.join([k for k in EXT_TO_FILETYPE if k is not None])}.",
+            )
+        self._set_download_paths()
+
+    def _set_download_paths(self) -> str:
+        """Sets download and output directories"""
```
ryannikolaidis, line 74: looks like some juggling in here, maybe a docstring that describes more specifically what we're doing and/or some inline comments along the way?

unstructured/ingest/connector/one_drive.py
```diff
@@ -0,0 +1,211 @@
...
+            not self.standard_config.preserve_downloads and not self.standard_config.download_only
+        )
+        self._set_client()
+
+    @requires_dependencies(['office365'])
+    def _set_client(self):
+        from office365.graph_client import GraphClient
+        self.client = GraphClient(self.config.token_factory)
+    
+    def _list_objects(self, folder, recursive) -> list:
```
ryannikolaidis, line 178: nit: maybe typing to indicate this is specifically a list of `driveItem` right?

unstructured/ingest/connector/onedrive.py
```diff
@@ -0,0 +1,155 @@
...
+        """Parses the folder structure from the source and creates the download and output paths"""
+        download_path = Path(f"{self.standard_config.download_dir}")
+        output_path = Path(f"{self.standard_config.output_dir}")
+
+        if parent_ref := self.file.get_property("parentReference", "").path.split(":")[-1]:
+            odir = parent_ref[1:] if parent_ref[0] == "/" else parent_ref
+            download_path = download_path if odir == "" else (download_path / odir).resolve()
+            output_path = output_path if odir == "" else (output_path / odir).resolve()
+
+        dname = f'{self.file.get_property("id")}-{self.file.name}'
```
ryannikolaidis, line 85: why are we prefixing the id here? is this because we're flattening the file structure? any way we could keep the file structure and ditch the need for an id added to the filename?

ryannikolaidis:
Pushed some changes:

* exclude date_processed from test run (otherwise the test will never pass the diff)
* add onedrive test to test-ingest.sh (so it gets run in CI)
* add makefile install command
* add make install to ci and `ingest test fixtures` workflows
* refactor filenames to *onedrive* for consistency
* update ingest test to match new refactored test flow
* update connector to utilize new Cleanup mixins and base definitions (DRY)
* fix broken typing imports

ryannikolaidis:
I think this is very close. Last blocker is that question on file naming and hierarchy.

unstructured/ingest/connector/onedrive.py
ryannikolaidis: Got it, I think if we preserve the file structure there should never be two colliding documents though, right? If so, safe to remove that prefix.

examples/ingest/onedrive/onedrive.sh
```diff
@@ -0,0 +1,30 @@
...
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --ms-client-id "<Azure AD app client-id>" \
+    --ms-client-cred "<Azure AD app client-secret>" \
+    --ms-authority-url "<Authority URL, default is https://login.microsoftonline.com>" \
+    --ms-tenant "<Azure AD tenant_id, default is 'common'>" \
+    --ms-user-pname "<Azure AD principal name, in most cases is the email linked to the drive>" \
+    --structured-output-dir onedrive-ingest-output \
+    --num-processes 2 \
+    --verbose
```
potter-potter, line 30: While trying this out. 
1) had to chmod the script to execute - maybe thats normal
2) got `This connector does not support saving downloads to ~/.cache/  , --download-dir must be provided` so maybe just need to add --download-dir

ryannikolaidis: 1. good catch, I hit that too
2. this should always have been the case when your connector doesn't define an elif block like [this one here](https://github.com/Unstructured-IO/unstructured/blob/4b827f07932c9c18df08e75a2615b01f0ab0bc74/unstructured/ingest/main.py#L589-L593). So, that said, maybe let's define one for this connector?

ryannikolaidis (changes requested):
discussed outside this thread, remaining piece: we need a way to list and specify drive ids.

ryannikolaidis:
@rvztz looks like it needs linting fixes and a bump to the changelog and __version__.py

ryannikolaidis (approved):
lgtm!

PR #837, opened 2023-06-28
Avoid setting metadata in constructor signature for elements

rbiseck3:
Avoid setting metadata in constructor signature for elements because that can lead to unexpected object reuse (and modification).

Bonus refactor for PageBreak to have text values of "".


test_unstructured/partition/test_pdf.py
```diff
@@ -119,7 +119,7 @@ def test_partition_pdf_with_spooled_file(
         # validate that the result is a non-empty list of dicts
         assert len(result) > 10
         # check that the pdf has multiple different page numbers
-        assert {element.metadata.page_number for element in result} == {1, 2}
+        assert len({element.metadata.page_number for element in result}) > 1
```
rbiseck3, line 122: @qued , These tests were breaking due to the page break param not being passed through for `hi_res` which was fixed [here](https://github.com/Unstructured-IO/unstructured/pull/837/files#diff-cefa2d296ae7ffcf5c28b5734d5c7d506fbdb225c05a0bc27c6b755d5424ffdaR116) so these tests shouldn't need to be changed. If anything they should stay the same in case something about page numbers breaks that causes this to be more than what's expected.

unstructured/documents/elements.py
```diff
@@ -377,9 +380,9 @@ def __init__(
         element_id: Union[str, NoID] = NoID(),
         coordinates: Optional[List[float]] = None,
         coordinate_system: Optional[CoordinateSystem] = None,
-        metadata: ElementMetadata = ElementMetadata(),
+        metadata: Optional[ElementMetadata] = None,
     ):
-        super().__init__(text="<PAGE BREAK>")
```
rbiseck3, line 382: @qued , I thought this was a clean solution here, I'm usually a fan of explicitly stating which kwarg is being passed in in case the position ever changes: 
```python
    def __init__(self, **kwargs):
        super().__init__(text="<PAGE BREAK>")
```

qued: I think this would break in the case someone using it passed positional arguments.

rbiseck3: Can I update it to `def __init__(self, *args, **kwargs):` then since none of them are actually used.

test_unstructured/partition/test_pdf.py
qued: That's fair, I just decided to take the preceding comment at its word that we were supposed to be checking whether the pdf came back with multiple page numbers. Feel free to revert my change.

unstructured/documents/elements.py
qued: Agree on passing `text=...` though.

qued: Per our offline conversation, it's fine if we go back to the args/kwargs pattern if we aren't passing on the arguments.

cragwolfe (approved):

PR #838, opened 2023-06-28
fix: parameterized ingest test overwriting

ryannikolaidis:
`test-ingest-local-single-file.sh` does not set `OVERWRITE_FIXTURES` to false by default, meaning that it always succeeds and always overwrites the fixtures. This PR assigns the default value.

* sets OVERWRITE_FIXTURES to default to false in `test-ingest-local-single-file.sh`
* fixes incorrect expected results
* bonus: installs language pack for Korean in CI and ingest fixture workflows


yuming-long (approved):
LGTM!

.github/workflows/ingest-test-fixtures-update-pr.yml
```diff
@@ -69,6 +69,7 @@ jobs:
           sudo apt-get install -y libmagic-dev poppler-utils libreoffice pandoc
           sudo add-apt-repository -y ppa:alex-p/tesseract-ocr5
           sudo apt-get install -y tesseract-ocr
+          sudo apt-get install -y tesseract-ocr-kor
```
six5532one, line 72: IIUC this runs before generating these types of PRs: https://github.com/Unstructured-IO/unstructured/pull/776

For education, what triggers these PRs?

six5532one (approved):

.github/workflows/ingest-test-fixtures-update-pr.yml
ryannikolaidis: > IIUC this runs before generating these types of PRs: https://github.com/Unstructured-IO/unstructured/pull/776

Yes, that's correct.

> For education, what triggers those PRs?

Great question! So [here](https://github.com/Unstructured-IO/unstructured/blob/44411ecc59eddbc5d3859800e91bfa43bb25c1ae/.github/workflows/ingest-test-fixtures-update-pr.yml#L5) we trigger on: `workflow_dispatch`. More info on workflow_dispatch [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch), but tldr; it will add a trigger-able button in the Actions menu in Github, where you can manually launch the workflow.

PR #843, opened 2023-06-28
fix: use api key where needed in tests

ryannikolaidis:
Api keys are now required for the hosted unstructured-api. We have several tests that validate against the remote service. This updates them to use a valid api key to make these requests.

- passes api key for unstructured-api to unit and ingest tests as needed.
- adds check for env var CI to otherwise skip tests that require an api key


test_unstructured/partition/test_api.py
```diff
@@ -98,7 +100,8 @@ def test_partition_via_api_raises_with_bad_response(monkeypatch):
 def test_partition_via_api_valid_request_data_kwargs():
```
yuming-long, line 100: could you remove the tests that are skipped with reason `@pytest.mark.skip(reason="Temporary skip until since API key is now required")` like this one

ryannikolaidis: done

yuming-long: i think there are four those lines in total

yuming-long: okay saw them thanks

yuming-long (approved):
LGTM! 🙏

PR #844, opened 2023-06-29
feat: add Dropbox connector

potter-potter:
Adds Dropbox connector.
Notice the structure of the path for the root folder for Dropbox. Not pretty, but the best I could come up with.
Definitely open for any feedback.


examples/ingest/dropbox/dropbox_token.py
```diff
@@ -0,0 +1,66 @@
+# Make sure this is importing the official dropbox python package, not unstructured/ingest/connector/dropbox
+import dropbox
+from dropbox import DropboxOAuth2FlowNoRedirect
+
+"""
+This example walks through a basic oauth flow using the existing long-lived token type
+Populate your app key and app secret in order to run this locally
+https://www.dropboxforum.com/t5/Dropbox-API-Support-Feedback/Get-refresh-token-from-access-token/m-p/596755
+https://www.dropboxforum.com/t5/Dropbox-API-Support-Feedback/Oauth2-refresh-token-question-what-happens-when-the-refresh/td-p/486241
+"""
```
ryannikolaidis, line 10: this is very nice. going to step through it in a bit.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.7.???????????????
```
ryannikolaidis, line 1:
```suggestion
## 0.7.10-dev1
```

examples/ingest/dropbox/ingest.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+# Processes several files in a nested folder structure from dropbox://utic-test-ingest-fixtures/ 
+# through Unstructured's library in 2 processes.
+# Due to Dropbox's interesting sdk, to list and get the root folders/files we use "dropbox:// /"
```
ryannikolaidis, line 5: do they have a page that describes this pattern (that we can link to)?

what does getting a nested path look like? `dropbox:// /some/other/path`?
is there ever a time when we don't have the `dropbox://` prefix or when that turns into a longer page `dropbox://something/else`?

potter-potter: Yes. If we are not going from the root folder (the app folder) we can just use a simple path

`dropbox://nested-1`

potter-potter: The whole root as " " really put a wrench into the simplicity of this thing.

potter-potter: I will rewrite the comment to make more sense.

ryannikolaidis: Ahh, okay, got it. Totally makes sense. I mean objectively confusing on their side, but am tracking now. Thanks!

ryannikolaidis (commented):
Bumped with lint and test refactor.

ryannikolaidis (approved):
LGTM!

PR #852, opened 2023-06-29
Strategy comparison script

rbiseck3:
### Description
Add bash script to:
* generate inference results across strategies
* produce text results from json files
* upload everything to s3

### DIFF
--- a/scripts/validate-models-s3.sh
+++ b/scripts/validate-models-s3.sh
@@ -6,7 +6,7 @@ input_dir=s3://utic-dev-tech-fixtures/partition-strategy-evaluation/pdf/lang/
 function list_s3_dirs() {
   # s3 doesn't have a notion of directories, need to do some parsing to get them from a particular s3 path
   path=$1
-  aws s3 ls $path | sed \$d | awk '{print $2}' | grep -E '/' | awk '{ print substr( $0, 1, length($0)-1 ) }'
+  aws s3 ls "$path" | sed \$d | awk '{print $2}' | grep -E '/' | awk '{ print substr( $0, 1, length($0)-1 ) }'
 }
 
 function run_inference() {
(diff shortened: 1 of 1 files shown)


scripts/validate-models-s3.sh
```diff
@@ -0,0 +1,100 @@
...
+  output_dir=partition-strategy-evaluation/output/json/$strategy
+  echo "writing output content to local dir $output_dir"
+  PYTHONPATH=. ./unstructured/ingest/main.py \
+    --metadata-exclude filename,file_directory,metadata.data_source.date_processed \
+    --remote-url "$input_path" \
+    --structured-output-dir "$output_dir" \
+    --preserve-downloads \
+    --partition-strategy "$strategy" \
+    --partition-ocr-languages "$lang" \
+    --download-dir partition-strategy-evaluation/input/
```
six5532one, line 26: I don’t think we need to keep the downloaded PDFs locally so I don’t think `--preserve-downloads` and `--download-dir` are necessary.

scripts/validate-models-s3.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
+
+strategies=${STRATEGIES:=fast,hi_res,ocr_only}
```
six5532one, line 3: How about passing these as arguments so we can reuse this script for other strategies? If this sounds good, documenting how to use the script would help too.

scripts/validate-models-s3.sh
```diff
@@ -0,0 +1,100 @@
...
+    done
+  done
+
+}
+
+echo "------------ Running model validation ------------"
+process_files_s3
+generate_txt_outputs
+upload_s3
+echo "------------ Model validation complete ------------"
```
six5532one, line 100: "validation" implies examining the models' outputs. I think "Running model" and "Model output upload complete" more accurately describe what the script does. Also, I think renaming the script to something like "generate_model_outputs_to_s3.sh" makes it easier to understand what it does.

six5532one (commented):
Awesome. This is very comprehensive. Thanks for taking care of reorganizing the input files to streamline this workflow.

scripts/validate-models-s3.sh
rbiseck3: Since the script can take a while, this helps in case something breaks along the way.

scripts/validate-models-s3.sh
rbiseck3: In essence this solves the same thing by allowing you to set the env var to the values you want. This just sets the default in case one is not set.

six5532one: That makes sense. Commenting or documenting in the script that the user can specify strategies by setting the STRATEGIES environment variable makes it easier to use for someone who isn't familiar with shell scripting.

PR #853, opened 2023-06-29
feat: Add include_metadata across all partition functions

### LINKED ISSUE #820: feat/Add `include_metadata` across all partition functions
Some partitions have a `include_metadata` kwarg that allows the user to specify whether or not to include element metadata. The goal of this issue is to add support for this kwarg across the remaining partitioning functions.

Coniferish:
## Summary
Closes #820
Adds `include_metadata` kwarg to all partition functions except for `partition_image`. 
Adds associated tests for excluding metadata for all partitions.
Adds functionality to `add_metadata_with_filetype` decorator so `include_metadata` info is automatically added to a decorated function's docstring

## Testing
```

filename="example-docs/layout-parser-paper-fast.pdf"
elements = pdf.partition_pdf(filename=filename, strategy="auto", include_metadata=False)
elements[0].metadata
```


test_unstructured/partition/test_org.py
```diff
@@ -15,3 +15,18 @@ def test_partition_org_from_file(filename="example-docs/README.org"):
 
     assert elements[0] == Title("Example Docs")
     assert elements[0].metadata.filetype == "text/org"
+
+
+def test_partition_org_from_filename_exclude_metadata(filename="example-docs/README.org"):
+    elements = partition_org(filename=filename, include_metadata=False)
+
+    for i in range(len(elements)):
+        assert any(elements[i].metadata.to_dict()) is False
```
MthwRobinson, line 24: This should be an empty dictionary, asserting that it's equal to `{}` might have been better, but since this pattern is already added everywhere I say keep it.

unstructured/file_utils/filetype.py
```diff
@@ -502,6 +502,25 @@ def _is_code_mime_type(mime_type: str) -> bool:
 
 def add_metadata_with_filetype(filetype: FileType):
     def decorator(func: Callable):
+        if func.__doc__:
```
MthwRobinson, line 505: Could you add this to `process_metadata` instead?

MthwRobinson (changes requested):
One minor request on where the doc strings are getting added, otherwise looks good

test_unstructured/partition/test_org.py
Coniferish: I can do a quick search for that assert and change them to dicts

MthwRobinson (approved):
LGTM once linting is passing!

PR #855, opened 2023-06-29
feat: attachment processing for emails

### LINKED ISSUE #604: Add ability to partition attachments in `partition_email` and `partition_msg`
Currently, `partition_email` and `partition_msg` partition the content of the email message, and there are separate functions for extracting and storing attachments. The goal of this issue is to add a `partition_attachments` flag. If it's `True`, `partition_email` and `partition_msg` should partition the attachments in addition to the email and append the results to `elements`. For attachments, the `filename` attribute should be the name of the attached file, and there should be an additional `attached_to` attribute indicating what email the file was attached to. `partition_attachments` should default for `False`.

MthwRobinson:
### Summary

Closes #604. Adds the ability to process attachments to `partition_email` and `partition_msg`. The partitioning function to use with the attachments is set with `attachment_partitioner`. We can't set that automatically to `partition` since that would create a circular import. Adds an `attached_to_filename` to track the file the attached document was originally attached to.

### Testing

```python

filename = "example-docs/eml/fake-email-attachment.eml"
elements = partition_email(
  filename=filename, process_attachments=True, attachment_partitioner=partition
)
```

```python

filename = "example-docs/fake-email-attachment.msg"
elements = partition_msg(
  filename=filename, process_attachments=True, attachment_partitioner=partition
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
-## 0.7.11-dev0
+## 0.7.11-dev1
 
 ### Enhancements
+
+* `partition_email` and `partition_msg` will now process attachments if `process_attachments=True`
+  and a attachment partitioning functions is passed through with `attachment_partitioner=partition`.
 
 ### Features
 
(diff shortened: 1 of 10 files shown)


qued (approved):
LGTM!

Issue #441, opened 2023-04-03
Update `partition_pdf` and `partition_image` to use GPUs when available and strategy is set to "hi_res"


DoctorSlimm:
The goal of this issue is to update `partition_pdf` and `partition_image` to use GPUs when available. Currently, the model only runs against CPUs. This will entail updates to `unstructured-inference`.

MthwRobinson:
@DoctorSlimm - `partition_pdf` and `partition_image` do not use GPUs at the moment. Updated this issue to add GPU support for `partition_pdf` and `partition_image`. Thanks for adding this issue! cc @qued .

Balearica:
@SimranJha2408 Was this issue closed as completed properly?  I don't see evidence that this feature was added. 

Per the title, this issue appears to track whether GPUs are used in `partition_pdf` and `partition_image` when `strategy` is set to `HI_RES`.  The only PR that is linked (#865) was closed (not merged).  Additionally, that PR does not appear to be directly related to this issue.  It looks like that PR would change the default `strategy` to `HI_RES` if a GPU is detected and `strategy` is not set explicitly, which is distinct from whether or not GPU is used when `strategy` is set to `HI_RES` by the user.

MthwRobinson:
Hi @Balearica - this feature was not added. We do plan to add GPU support for our API in spring or summer.

ChiNoel-osu:
At the moment `tesseract` in `unstructured` is only using 1 CPU core and it's so slow. Would be great if we could have GPU accellerated OCR.
Even better, split document by page and run multiple `tesseract` at the same time.

MthwRobinson:
Thanks for the comment @ChiNoel-osu . If you're interested in trying out our API, the [`unstructured-python-client`](https://github.com/Unstructured-IO/unstructured-python-client) splits up long PDFs so you can parallelize across pages within the doc.

PR #865, opened 2023-06-30
feat(partition): auto gpu detection

kravetsmic:
## Summary

* Added GPU detection
* Updated strategy detection and default value

---
Closes issue #441


unstructured/partition/strategies.py
```diff
@@ -72,6 +73,13 @@ def determine_pdf_or_image_strategy(
     logic if some dependencies are not available."""
     pytesseract_installed = dependency_exists("pytesseract")
     unstructured_inference_installed = dependency_exists("unstructured_inference")
+    if not strategy:
+        if torch.cuda.is_available() or (
```
MthwRobinson, line 77: @kravetsmic - For this one, the `torch` library is only available if you install the `local-inference` extra, so I think this would fail with just base `unstructured`. Could you wrap this in a `_is_gpu_available` function, and also check first if `torch` is installed?

unstructured/partition/strategies.py
```diff
@@ -1,6 +1,7 @@
 from tempfile import SpooledTemporaryFile
 from typing import BinaryIO, Dict, List, Optional, Union, cast
 
+import torch
```
MthwRobinson, line 4: Same comment here, I don't think we should import `torch` at the module level in case the `local-inference` extra isn't installed.

unstructured/partition/pdf.py
```diff
@@ -92,11 +92,11 @@ def partition_pdf(
 
 
 def partition_pdf_or_image(
-    filename: str = "",
+    filename: Optional[str] = "",
     file: Optional[Union[bytes, BinaryIO, SpooledTemporaryFile]] = None,
     is_image: bool = False,
     include_page_breaks: bool = False,
-    strategy: str = "auto",
+    strategy: Optional[str] = None,
```
MthwRobinson, line 99: Let's keep the default "auto", and the logic should route to "hi_res" if "auto" is selected and GPUs are available.

unstructured/partition/pdf.py
```diff
@@ -43,7 +43,7 @@ def partition_pdf(
     filename: str = "",
     file: Optional[Union[BinaryIO, SpooledTemporaryFile]] = None,
     include_page_breaks: bool = False,
-    strategy: str = "auto",
+    strategy: Optional[str] = None,
```
MthwRobinson, line 46: Same for the rest of the strategy defaults.

MthwRobinson (changes requested):
Added a few suggestions, but generally looks good.

kravetsmic:
Will be re-opened after brought up to date

muazhari:
Damn, I need GPU support for partitioning large files. Roughly, when the GPU support feature is implemented?

PR #876, opened 2023-07-03
fix: Fix KeyError in isd_to_elements

awalker4:
I don't have a reproducible example, but we continue to get API errors with the following stack. I suppose the safest thing is to skip an item that has no type, but happy to consider other behavior if we don't want to be silently ignoring data. Any suggestions for how to track down where these items are coming from?

Screenshot 2023-06-09 at 11 48 50 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


MthwRobinson (approved):
LGTM!

PR #890, opened 2023-07-06
Add optional param for model name when partitioning pdfs

rbiseck3:
### Description
To support unstructured-api in choosing between the default detectron model and the latest large model that was introduced, an optional parameter was added to use first before checking the environment variable which is the current behavior. 

To also support multiple models being, the `initialize()` method was also updated to check a new env var `UNSTRUCTURED_HI_RES_SUPPORTED_MODEL` which can be a comma-delimited string of all models to initialize explicitly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## 0.8.0-dev2
+## 0.8.0
 
 ### Enhancements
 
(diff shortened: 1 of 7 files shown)


MthwRobinson (approved):
LGTM!

ryannikolaidis (approved):
lgtm!

PR #900, opened 2023-07-07
fix: better extractable check

### LINKED ISSUE #717: Unstructured suddenly can't parse scanned PDFs
I'm using `Unstructured` for parsing PDFs and have it installed through a docker dev container. It was working a few months ago but after I rebuilt the container for deployment, it suddenly breaks.

Here's my dockerfile:
```
FROM python:3.9-slim-buster

# Update package lists
RUN apt-get update && apt-get install ffmpeg libsm6 libxext6 gcc g++ git build-essential libpoppler-cpp-dev libmagic-dev pkg-config poppler-utils tesseract-ocr libtesseract-dev -y

# Make working directories
RUN  mkdir -p  /app
WORKDIR  /app

# Copy the requirements.txt file to the container
COPY requirements.txt .

# Install dependencies
RUN pip install --upgrade pip

RUN pip install torch torchvision torchaudio

RUN pip install unstructured-inference

RUN pip install -r requirements.txt

RUN pip install 'git+https://github.com/facebookresearch/detectron2.git@e2ce8dc#egg=detectron2'

# Copy the .env file to the container
COPY .env .

# Copy every file in the source folder to the created working directory
COPY  . .

# Expose the port that the application will run on
EXPOSE 8080

# Start the application
CMD ["python3.9", "-m", "uvicorn", "main:app", "--proxy-headers", "--host", "0.0.0.0", "--port", "8080", "--workers", "4"]
```

and `requirements.txt`:
```
fastapi
uvicorn
langchain
python-poppler
pytesseract
unstructured[local-inference]
psycopg2-binary
pgvector
openai
tiktoken
python-dotenv
pypdf
```

I've verified that detectron is in my system but it seems like it's not being used by unstructured as it defaults to using `pdfminer`

qued:
Hopefully closes #717 . `auto` strategy was choosing the `fast` strategy in cases where the pdf contents were just a flat image, resulting in no output. This PR changes the behavior of `auto` so that elements that can be extracted by `fast` are extracted, a cursory examination of the elements is made to see if there are elements with text present, and if so then these elements are used as the output. Otherwise fallback strategies come into play.

#### Testing
```python
elements = partition_pdf("example-docs/loremipsum-flat.pdf", strategy="auto")
print(elements)
```
Output should be `[]` on current `main` branch and should contain elements on this branch. (File doesn't exist on the `main` branch but it can be found in `unstructured-inference` under `sample-docs`.)


MthwRobinson (approved):
Looks good to me! Thanks for updating that

Issue #905, opened 2023-07-09
feat/confluence-connector


ahmetmeleq:
Create a data connector that pulls documents from Confluence, stores them locally (at least temporarily for processing), and runs them through [unstructured.partition.auto](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py).

See [Adding Data Connectors](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#adding-data-connectors) to get started.

Make sure to include a link to this issue when submitting a PR.

Definition of Done

- The [data connector checklist](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md#the-checklist) has been completed.
- [PR checklist 1](https://github.com/Unstructured-IO/docs-and-arch/blob/main/Pull-Requests.md#before-creating-pr) and [PR checklist 2](https://github.com/Unstructured-IO/docs-and-arch/blob/main/Pull-Requests.md#pr-guidelines) have been completed.
- The connector is able to process all documents in all spaces in a Confluence Cloud instance.
- For now, it is OK to process only the body text of the document, excluding comments and images.
- Bonus points: multiprocessing optimisations such as one python client per process, and other kinds of optimisations / user considerate implementation.

PR #906, opened 2023-07-09
feat: confluence connector (cloud)

ahmetmeleq:
This connector:

- takes a Confluence Cloud Instance URL, user email, and api token; to authenticate into Confluence Cloud,
- to process the body text for each document in each space within the particular Confluence Cloud instance,
- and to store each document as a text/html file,
- which are then processed by unstructured auto into the expected output format.
- Related to: https://github.com/Unstructured-IO/unstructured/issues/905

To test the changes, make the necessary setups and run test-ingest-confluence.sh within test_unstructured_ingest .


test_unstructured_ingest/expected-structured-output/confluence/testteamsp/1605942.json
```diff
@@ -0,0 +1,142 @@
...
+  },
+  {
+    "type": "ListItem",
+    "element_id": "e3b0c44298fc1c149afbf4c8996fb924",
+    "metadata": {
+      "data_source": {},
+      "filetype": "text/html",
+      "page_number": 1
+    },
+    "text": ""
```
ryannikolaidis, line 120: what was the deal with these empty string list items?

test_unstructured_ingest/test-ingest-confluence.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -e
+
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+cd "$SCRIPT_DIR"/.. || exit 1
+
+# source "./../../secrets/confluence2.txt"
```
ryannikolaidis, line 7: ?? assuming this was for development?

unstructured/ingest/connector/confluence.py
```diff
@@ -0,0 +1,164 @@
...
+        results = self.confluence.get_all_spaces(start=start, limit=limit, expand=expand)
+        space_ids = [space["key"] for space in results["results"]]
+        return space_ids
+
+    @requires_dependencies(["atlassian"])
+    def _get_all_docs_ids_in_space(
+        self,
+        space_id: str,
+        start: int = 0,
+        limit: int = 100,
```
ryannikolaidis, line 127: does this mean the connector won't fetch more than 100 page ids in a space? what is the risk if this is higher? should we expose to user?

unstructured/ingest/connector/confluence.py
```diff
@@ -0,0 +1,164 @@
...
+            limit=limit,
+            status=status,
+            expand=expand,
+            content_type=content_type,
+        )
+        doc_ids = [(space_id, doc["id"]) for doc in results]
+        return doc_ids
+
+    @requires_dependencies(["atlassian"])
+    def _get_all_doc_ids_in_all_spaces(self):
```
ryannikolaidis, line 144: so we always return all spaces? should we allow the user to specify a space? either by name or id? and then maybe default to all spaces if they don't specify?

ryannikolaidis (changes requested):
This is looking really good! Just a few questions and thoughts.

test_unstructured_ingest/expected-structured-output/confluence/testteamsp/1605942.json
ahmetmeleq: They correspond to the empty rows in the Confluence template docs, which are formed by default (image attached)

We have a few options here:

- Modify these documents to fill the rows so the text fields are not empty
- Delete this kind of template documents and create a few more custom test documents to replace them
- Leave them as they are

I think we can leave them as is, if it won't cause confusion. We have other documents with bulkier text. But if it'd (deceivingly) look like a bug in the future, it might be good to replace them now to avoid confusion later.



image

unstructured/ingest/connector/confluence.py
```diff
@@ -0,0 +1,164 @@
...
+    @requires_dependencies(["atlassian"])
+    def initialize(self):
+        self.confluence = Confluence(
+            url=self.config.url,
+            username=self.config.user_email,
+            password=self.config.api_token,
+        )
+
+    @requires_dependencies(["atlassian"])
+    def _get_all_space_ids(self, start: int = 0, limit: int = 500, expand=None):
```
ahmetmeleq, line 116: Might be good to expose the `limit` variable to the user

unstructured/ingest/connector/confluence.py
ahmetmeleq: I think a list of spaces makes sense then (in case they'd like a specific set of spaces)

Agreed on defaulting to all spaces if they don't specify 👍

test_unstructured_ingest/expected-structured-output/confluence/testteamsp/1605942.json
ryannikolaidis: yep, totally makes sense and agreed!

unstructured/ingest/connector/confluence.py
ryannikolaidis: great idea!

test_unstructured_ingest/test-ingest-confluence2.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
ryannikolaidis, line 1: maybe we name these test files with a suffix other than `1` and `2`. something like `-diff` and `-large`?

test_unstructured_ingest/test-ingest-confluence2.sh
```diff
@@ -0,0 +1,32 @@
...
+set -e
```
ryannikolaidis, line 2: maybe a docstring at the top of each test to explain the purpose (given that we have two for confluence)?

unstructured/ingest/main.py
```diff
@@ -409,6 +409,44 @@ def run(self):
...
+)
+@click.option(
+    "--confluence-list-of-spaces",
+    default=None,
+    help="A list of confluence space ids to be fetched. From each fetched space, \
+        --confluence-num-of-docs-from-each-space number of docs will be ingested. \
+        --confluence-list-of-spaces and --confluence-num-of-spaces cannot be used at the same time",
+)
+@click.option(
+    "--confluence-num-of-spaces",
```
ryannikolaidis, line 437:
```suggestion
    "--confluence-max-num-of-spaces",
```

?

unstructured/ingest/main.py
```diff
@@ -409,6 +409,44 @@ def run(self):
...
+    default=500,
+    help="Number of confluence space ids to be fetched. From each fetched space, \
+        --confluence-num-of-docs-from-each-space number of docs will be ingested. \
+        --confluence-list-of-spaces and --confluence-num-of-spaces cannot be used at the same time",
+)
+@click.option(
+    "--confluence-num-of-docs-from-each-space",
```
ryannikolaidis, line 444:
```suggestion
    "--confluence-max-num-of-docs-from-each-space",
```

test_unstructured_ingest/test-ingest-confluence2.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --download-dir "$DOWNLOAD_DIR" \
+    --confluence-url https://unstructured-ingest-test.atlassian.net \
+    --confluence-user-email "$CONFLUENCE_USER_EMAIL" \
+    --confluence-api-token "$CONFLUENCE_API_TOKEN" \
+    --confluence-num-of-spaces 10 \
+    --confluence-list-of-spaces testteamsp1 \
+    --confluence-num-of-docs-from-each-space 250 \
```
ryannikolaidis, line 25: maybe instead we have this test use the defaults and validate it grabs everything?

test_unstructured_ingest/test-ingest-confluence2.sh
```diff
@@ -0,0 +1,32 @@
...
+   exit 0
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --download-dir "$DOWNLOAD_DIR" \
+    --confluence-url https://unstructured-ingest-test.atlassian.net \
+    --confluence-user-email "$CONFLUENCE_USER_EMAIL" \
+    --confluence-api-token "$CONFLUENCE_API_TOKEN" \
+    --confluence-num-of-spaces 10 \
+    --confluence-list-of-spaces testteamsp1 \
```
ryannikolaidis, line 24: I thought per your text in main.py that one is not supposed to include both num-of-spaces and and list-of-spaces args?

examples/ingest/confluence/ingest.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Processes all the documents in all spaces within a confluence domain, using the `unstructured` library.
```
ryannikolaidis, line 3: maybe a comment here that one can target specific spaces by arg?

ryannikolaidis (changes requested):
A few comments, but looking really close!

test_unstructured_ingest/test-ingest-confluence2.sh
ahmetmeleq: Sounds good, now added with [736e729](https://github.com/Unstructured-IO/unstructured/pull/906/commits/736e729d069e792adfcd8c5c015739645036af17) 👍

test_unstructured_ingest/test-ingest-confluence2.sh
ahmetmeleq: Yes, this part in the test checks if the connector can act the right way even when it is being used in the wrong way 
(checks if it can ignore the redundant arg and go with the preferred one)

If we'd like to keep each test checking exactly one aspect rather than testing multiple aspects with one test, I can remove the redundant arg there

I also added an explanation to avoid confusion if we keep the part:

https://github.com/Unstructured-IO/unstructured/blob/541497408523c799544b4ccdad87be61b35478a5/test_unstructured_ingest/test-ingest-confluence-large.sh#L20

test_unstructured_ingest/test-ingest-confluence2.sh
ahmetmeleq: Agreed that we need to test that aspect (grabbing every doc). I now changed confluence1 (-diff) to use the default, and get all the docs.

In confluence2 (-large) I wanted to test if the scroll functionality can actually honor `--confluence-num-of-docs-from-each-space 250` rather than just going 3 scrolls and ingesting 300 docs, that's why I specifically mentioned 250 there

ahmetmeleq: What do you think?

examples/ingest/confluence/ingest.sh
ahmetmeleq: Agreed, added with [169cafd](https://github.com/Unstructured-IO/unstructured/pull/906/commits/169cafdd078ad42f712980e7ed495ec5abb7b268) now 👍

test_unstructured_ingest/test-ingest-confluence2.sh
ryannikolaidis: I think with explanation this is gtg

test_unstructured_ingest/test-ingest-confluence2.sh
ryannikolaidis: perfect!

test_unstructured_ingest/expected-structured-output/confluence/testteamsp/1867777.json
```diff
@@ -0,0 +1,11 @@
...
+  {
+    "type": "NarrativeText",
+    "element_id": "d2cfe2b37317a51a54267f553e9b1247",
+    "metadata": {
+      "data_source": {},
+      "filetype": "text/plain"
+    },
+    "text": "<p>Testdoc1 has only this text for 10 times: 1</p><p>Testdoc1 has only this text for 10 times: 2</p><p>Testdoc1 has only this text for 10 times: 3</p><p>Testdoc1 has only this text for 10 times: 4</p><p>Testdoc1 has only this text for 10 times: 5</p><p>Testdoc1 has only this text for 10 times: 6</p><p>Testdoc1 has only this text for 10 times: 7</p><p>Testdoc1 has only this text for 10 times: 8</p><p>Testdoc1 has only this text for 10 times: 9</p><p>Testdoc1 has only this text for 10 times: 10</p>"
+  }
+]
```
ryannikolaidis, line 11: any idea why some of these are coming through as text/plain rather than text/html?

unstructured/ingest/connector/confluence.py
```diff
@@ -0,0 +1,208 @@
...
+    config: SimpleConfluenceConfig
+    file_meta: ConfluenceFileMeta
+
+    # TODO: remove one of filename or _tmp_download_file, using a wrapper
+    @property
+    def filename(self):
+        return (
+            Path(self.standard_config.download_dir)
+            / self.file_meta.space_id
+            / f"{self.file_meta.document_id}.txt"
```
ryannikolaidis, line 94: does this always return valid html? maybe we just make this a .html file so we don't miss the on the auto detection?

test_unstructured_ingest/expected-structured-output/confluence/testteamsp/1867777.json
ahmetmeleq: I'll check this to see why this happens

unstructured/ingest/connector/confluence.py
ahmetmeleq: File naming might indeed be the reason for the above problem. I'll check if it always returns html, and if so I'll change this part 👍

ahmetmeleq: Checking from [atlassian rest client](https://github.com/atlassian-api/atlassian-python-api/blob/5d9cd15a91eea0384333d408ae0e68148c7dc302/atlassian/rest_client.py#L285), which gets called from [a method within the confluence module](https://github.com/atlassian-api/atlassian-python-api/blob/5d9cd15a91eea0384333d408ae0e68148c7dc302/atlassian/confluence.py#L320)

- Rest client makes a request to the Confluence page using `requests`
- tries to decode the response as json, 
- and if it can't decode as json, it returns `response.text `

I think it is safe to say confluence pages return html responses, given our results; which means `response.text` would be html. I've also checked responses from multiple websites around the web to verify this. (in addition to the verification from our ingest download results)

JSON responses should be for other types of content (such as content served on API endpoints).

test_unstructured_ingest/expected-structured-output/confluence/testteamsp/1867777.json
ahmetmeleq: Yes, `partition.auto` doesn't default to html parsing due to the file extension. Since it's safe to say [requests to confluence pages return valid html](https://github.com/Unstructured-IO/unstructured/pull/906#discussion_r1265222659), I'm changing file extensions to html as proposed.

unstructured/ingest/connector/confluence.py
ryannikolaidis: beautiful!

ryannikolaidis (approved):
looks GREAT! nice work! 🎉

PR #827, opened 2023-06-28
chore: Move coordinate details from Element model to a metadata model

six5532one:
`unstructured.documents.elements.Element` used to have these attributes:
- coordinates
- _coordinate_system

This PR adds a coordinates metadata attribute to `unstructured.documents.elements.Element.ElementMetadata`. It moves the two coordinates attributes mentioned above to this new metadata model.

```
(unstructured) emilychen@emilys-MacBook-Pro-3 unstructured % python3
Python 3.8.17 (default, Jun 20 2023, 16:35:11) 
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from unstructured.partition.auto import partition
>>> filename = "/Users/emilychen/Downloads/How to ask Sage.pdf"
>>> elements = partition(filename=filename, strategy="hi_res")
>>> elements[0].to_dict()
{'type': 'Title', 'element_id': 'da4ec9889840dcff91303d563be40534', 'metadata': {'coordinates': {'points': ((545.0150947570801, 226.5191650390625), (545.0150947570801, 254.7656043600921), (704.879451751709, 254.7656043600921), (704.879451751709, 226.5191650390625)), 'system': 'PixelSpace', 'layout_width': 1700, 'layout_height': 2200}, 'filename': 'How to ask Sage.pdf', 'file_directory': '/Users/emilychen/Downloads', 'filetype': 'application/pdf', 'page_number': 1}, 'text': '1.2. Dashboard'}
>>> 
>>> elements_fast = partition(filename=filename, strategy="fast")
>>> elements_fast[0].to_dict()
{'type': 'Title', 'element_id': '053bc65874ad6098e58c41c57b378a2f', 'metadata': {'coordinates': {'points': ((0.0, 0.0), (0.0, 792.0), (612.0, 792.0), (612.0, 0.0)), 'system': 'PixelSpace', 'layout_width': 612.0, 'layout_height': 792.0}, 'filename': 'How to ask Sage.pdf', 'file_directory': '/Users/emilychen/Downloads', 'filetype': 'application/pdf', 'page_number': 1}, 'text': 'Discord'}
```

Because this is a breaking change, this PR updates the minor version.

Other changes:
* Maintains the existing, expected outputs for the unstructured-ingest tests by ensuring they do not request coordinates metadata
* Maintains the existing functionality of weaviate module by continuing to omit an element's coordinates info when staging it for weaviate
[...]

PR #912, opened 2023-07-10
Don't instantiate an element with a coordinate system when there isn't a way to get its location

six5532one:
https://github.com/Unstructured-IO/unstructured/pull/827 introduced an optional `CoordinatesMetadata` for each Element. As a safeguard, its constructor raises a ValueError on attempts to instantiate it with coordinate points and no coordinate system, or vice versa.

As a result, the following code raises a ValueError:
```
s.environ["UNSTRUCTURED_HI_RES_MODEL_NAME"] = "chipper"
os.environ["UNSTRUCTURED_HF_TOKEN"] = "redacted"
filename = "/path/to/chevron-page.pdf"
elements = partition(filename=filename, strategy="hi_res") 
```

In `unstructured_inference`, `process_file_with_model` returns a `PageLayout` object that has an image attribute. Therefore, the unstructured library resolves a coordinate system here:
https://github.com/Unstructured-IO/unstructured/blob/b3936893b89743bfdf011e97e08074fbea5966ce/unstructured/file_utils/filetype.py#L456

But then `document_to_element_list` proceeds to call `normalize_layout_element` with `layout_element` being a LocationlessLayoutElement, so this library calls the Text constructor with a coordinate system but no points and the ValueError is raised when attempting to instantiate CoordinatesMetadata.

If we don’t have a way of getting the location data for the elements (as is the case in the code snippet above), we don’t need a coordinate system either. This PR does not resolve a coordinate system when a page's element does not have its bounding box coordinates.

### DIFF
--- a/unstructured/file_utils/filetype.py
+++ b/unstructured/file_utils/filetype.py
@@ -8,6 +8,8 @@ from enum import Enum
 from functools import wraps
 from typing import IO, TYPE_CHECKING, Callable, List, Optional
 
+from unstructured_inference.inference.layoutelement import LocationlessLayoutElement
+
 from unstructured.documents.coordinates import PixelSpace
 from unstructured.documents.elements import Element, PageBreak
 from unstructured.file_utils.encoding import detect_file_encoding, format_encoding_str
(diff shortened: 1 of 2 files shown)


unstructured/file_utils/filetype.py
```diff
@@ -8,6 +8,8 @@
 from functools import wraps
 from typing import IO, TYPE_CHECKING, Callable, List, Optional
 
+from unstructured_inference.inference.layoutelement import LocationlessLayoutElement
```
six5532one, line 11: I confirmed this causes `test_unstructured_ingest/unit/test_interfaces.py::test_process_file_fields_include_default` to fail.

six5532one: Due to the observation in the comment above, closing in favor of https://github.com/Unstructured-IO/unstructured/pull/913

PR #913, opened 2023-07-11
Don't instantiate an element with a coordinate system when there isn't a way to get its location

six5532one:
https://github.com/Unstructured-IO/unstructured/pull/827 introduced an optional CoordinatesMetadata for each Element. As a safeguard, its constructor raises a ValueError on attempts to instantiate it with coordinate points and no coordinate system, or vice versa.

As a result, the following code raises a ValueError:

```
s.environ["UNSTRUCTURED_HI_RES_MODEL_NAME"] = "chipper"
os.environ["UNSTRUCTURED_HF_TOKEN"] = "redacted"
filename = "/path/to/chevron-page.pdf"
elements = partition(filename=filename, strategy="hi_res") 
```

It goes through this code path:
In unstructured_inference, `process_file_with_model` returns a PageLayout object that has an image attribute. Therefore, the unstructured library resolves a coordinate system here:
https://github.com/Unstructured-IO/unstructured/blob/b3936893b89743bfdf011e97e08074fbea5966ce/unstructured/file_utils/filetype.py#L456

But then `document_to_element_list` proceeds to call `normalize_layout_element` with layout_element being a LocationlessLayoutElement, so this library calls the Text constructor with a coordinate system but no points and the ValueError is raised when attempting to instantiate CoordinatesMetadata.

If we don’t have a way of getting the location data for the elements (as is the case in the code snippet above), we don’t need a coordinate system either. This PR does not resolve a coordinate system when a page's element does not have its bounding box coordinates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * Fixed `auto` strategy detected scanned document as having extractable text and using `fast` strategy, resulting in no output.
 * Fix list detection in MS Word documents.
+* Don't instantiate an element with a coordinate system when there isn't a way to get its location data.
 
 ## 0.8.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.8.1-dev1
+## 0.8.1-dev2
 
 ### Enhancements
 
--- a/unstructured/partition/pdf.py
+++ b/unstructured/partition/pdf.py
@@ -326,12 +326,16 @@ def _process_pdfminer_pages(
                 _text = clean_extra_whitespace(_text)
                 if _text.strip():
                     text_segments.append(_text)
-                    element = element_from_text(_text)
                     coordinate_system = PixelSpace(
                         width=width,
                         height=height,
                     )
                     points = ((x1, y1), (x1, y2), (x2, y2), (x2, y1))
+                    element = element_from_text(
+                        _text,
+                        coordinates=points,
+                        coordinate_system=coordinate_system,
+                    )
                     coordinates_metadata = CoordinatesMetadata(
                         points=points,
                         system=coordinate_system,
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 * Fixed `auto` strategy detected scanned document as having extractable text and using `fast` strategy, resulting in no output.
 * Fix list detection in MS Word documents.
+* Don't instantiate an element with a coordinate system when there isn't a way to get its location data.
```
qued, line 13: Could you also change the version in line one to 0.8.1-dev2, and change the version in `__version__.py` to match?

qued (approved):
LGTM! Just one minor nit.

unstructured/partition/pdf.py
```diff
@@ -326,12 +326,16 @@ def _process_pdfminer_pages(
...
                     coordinate_system = PixelSpace(
                         width=width,
                         height=height,
                     )
                     points = ((x1, y1), (x1, y2), (x2, y2), (x2, y1))
+                    element = element_from_text(
+                        _text,
+                        coordinates=points,
+                        coordinate_system=coordinate_system,
+                    )
```
six5532one, line 338: @qued IIUC this change doesn't make a difference when partitioning with Chipper but I don't see the harm of making it. Let me know if you think differently.

qued: If we construct the element this way, should we get rid of the lines that follow that create coordinate metadata and store it in the metadata, since presumably it should already be there?

qued: Actually, I'm fine if we just merge this. I think there's some inconsistency around how we're handling coordinates vs other metadata, but I'll create an issue to track that.

six5532one: I made these changes locally and unit tests passed:
change_havent_committed

I didn't push them because I saw [your comment about creating a new issue](https://github.com/Unstructured-IO/unstructured/pull/913#discussion_r1259167066).

Issue #886, opened 2023-07-05
bug/LangChain's `UnstructuredFileLoader` detects .rtf file format instead of .txt


six5532one:
**Describe the bug**
A LangChain user used the `DirectoryLoader` in LangChain's Python library. This uses LangChain's `UnstructuredFileLoader` class, which uses the unstructured library to load files.
https://github.com/hwchase17/langchain/blob/d669b9ece9df4485db88d7e8722caf1d5f0b6da4/langchain/document_loaders/directory.py#L36

When using `DirectoryLoader` with an input of a directory of .txt files, they encountered this warning:
`Invalid input format! Got "rtf" but expected one of these common...'

https://discord.com/channels/1038097195422978059/1126158067948601397

**To Reproduce**
The warning appears when using LangChain's `DirectoryLoader` but not when using their `TextLoader`.

**Expected behavior**
Documents should be loaded with plain text format instead of rich text format.

**Screenshots**
<img width="881" alt="langchain_dirloader_rff" src="https://github.com/Unstructured-IO/unstructured/assets/1435725/37197be0-3843-49d4-9a4e-42b9897d44db">

**Desktop (please complete the following information):**
 - OS: [e.g. windows, mac, linux]
 - Python version [e.g. 3.8.15]

**Additional context**

chancharikmitra:
To add some context to reproduce this. The issue occurs on Google Colab on Chrome. You can create a directory in your drive, create some test .txt files, and the following code will work (after necessary imports, of course):

Imports:
```
!pip -q install langchain

from langchain.document_loaders import TextLoader
from langchain.document_loaders import DirectoryLoader
```

Working Code:
```loader = TextLoader(data_dir + '/specific_file.txt') 
documents = loader.load()
len(documents)
```

Failing Code:

```loader = DirectoryLoader(data_dir, '**/*.txt', silent_errors=True) 
documents = loader.load()
len(documents)
```

MthwRobinson:
@chancharikmitra - Thanks for test code, will take a look at this as soon as we can.

MthwRobinson:
@chancharikmitra - I tried on Google Colab with a new text file I created and wasn't able to reproduce the issue. I did open #916 that may help. Could you also try the following with those files on Colab and let me know what MIME type gets returned?

```python
import magic

mime_type1 = magic.from_file("specific_file.txt", mime=True)

with open("specific_file.txt", "rb") as f:
[...]

PR #916, opened 2023-07-11
fix: adds to list of extensions to check if a file has a plain text MIME type

MthwRobinson:
### Summary

Related to #886. Adds `.txt`, `.text` and `.tab` to the list of file extension to check if a file has a `text/plain` MIME type.


qued (approved):
LGTM!

PR #918, opened 2023-07-12
feat: Sharepoint connector

rvztz:
Initial working version for the sharepoint connector. Feature to pull Sharepoint page remains pending.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 0.9.2-dev3
+=======
+
+### Enhancements
+
+### Features
+
+* Adds Sharepoint connector.
+
+### Fixes
+
 ## 0.9.2-dev2
 =======
 
--- a/unstructured/__version__.py
+++ b/unstructured/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.9.2-dev2"  # pragma: no cover
+__version__ = "0.9.2-dev3"  # pragma: no cover


ryannikolaidis:
Thanks for pushing up work in progress! Just bumped this to Draft since it looks like it's still in progress, but just bump the ready for review and assign to me when ready.

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,190 @@
...
+    def filename(self):
+        return Path(self.download_filepath).resolve()
+
+    @property
+    def _output_filename(self):
+        return Path(self.output_filepath).resolve()
+
+    def _get_page(self):
+        try:
+            content_labels = ["CanvasContent1"]
```
ryannikolaidis, line 84: what is the hardcoded CanvasContent1 here? is this something that will exist for every page? or is this just something you're testing as you deveop?

ryannikolaidis: (okay, I'm guessing the latter, but confirm)

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,190 @@
...
+            content = self.file.listItemAllFields.select(content_labels).get().execute_query()
+            pld = ""
+            for label in content_labels:
+                c = content.properties.get(label, "")
+                pld = pld if not c else pld + c
```
ryannikolaidis, line 89: nit: prefer using full words (easier to read)

unstructured/ingest/connector/sharepoint.py
ryannikolaidis: oh guessing this is something internal to define where the content of the document is. is there ever a CanvasContent2? or anything else we need to look for here?

ryannikolaidis: either way, definitely good to have a docstring and comments that explain what's going on throughout this

unstructured/ingest/main.py
```diff
@@ -423,10 +423,36 @@ def run(self):
...
+    default="Shared Documents",
+    help="Path to start parsing files from. If the connector is to process all sites \
+    within the tenant this filter will be applied to all sites document libraries. \
+    Default 'Shared Documents'",
+)
+@click.option(
+    "--ms-sharepoint-all",
+    is_flag=True,
+    default=False,
+    help="Process all sites within the tenant. Requires for the app to be registered at a tenant \
```
ryannikolaidis, line 447:
```suggestion
    help="Process all sites within the tenant. Requires the app to be registered at a tenant \
```

unstructured/ingest/main.py
```diff
@@ -423,10 +423,36 @@ def run(self):
...
+@click.option(
+    "--ms-sharepoint-site",
+    default=None,
+    help="Sharepoint site url. Process either base url e.g https://[tenant].sharepoint.com \
+        or relative sites https://[tenant].sharepoint.com/sites/<site_name>.",
+)
+@click.option(
+    "--ms-sharepoint-path",
+    default="Shared Documents",
+    help="Path to start parsing files from. If the connector is to process all sites \
```
ryannikolaidis, line 439:
```suggestion
    help="Path from which to start parsing files. If the connector is to process all sites \
```

unstructured/ingest/main.py
```diff
@@ -423,10 +423,36 @@ def run(self):
...
+    default=False,
+    help="Process all sites within the tenant. Requires for the app to be registered at a tenant \
+        level and a site url in https://[tenant]-admin.sharepoint.com.",
+)
+@click.option(
+    "--ms-sharepoint-pages",
+    is_flag=True,
+    default=False,
+    help="Process site pages.",
+)
```
ryannikolaidis, line 455: I suspect users would want pages by default. Maybe we invert this logic and call it `--ms-sharepoint-files-only`?

test_unstructured_ingest/test-ingest-sharepoint.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [ -z "$MS_CLIENT_ID" ] || [ -z "$MS_CLIENT_CRED" ]; then
+   echo "Skipping Sharepoint ingest test because the MS_CLIENT_ID or MS_CLIENT_CRED env var is not set."
+   exit 0
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --ms-client-id "$SHAREPOINT_CLIENTID" \
+    --ms-client-cred "$SHAREPOINT_CRED" \
+    --ms-sharepoint-site "$SHAREPOINT_SITE" \
```
ryannikolaidis, line 19: these need to be passed in from CI. I can get the secrets loaded in, but then see the changes we do in `.github/workflows` in [this PR](https://github.com/Unstructured-IO/unstructured/pull/939/files) for reference for how these need to be passed through (both in ci and in the ingest-test-fixtures workflows)

ryannikolaidis (changes requested):
looking good. dropped a few notes. can you also update the test page content? maybe a couple of pages (nested if that's a thing) and add some actual content to the pages: bulleted lists, narrative text, etc)

test_unstructured_ingest/test-ingest-sharepoint.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -e
+
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+cd "$SCRIPT_DIR"/.. || exit 1
+OUTPUT_FOLDER_NAME=Sharepoint
+OUTPUT_DIR=$SCRIPT_DIR/structured-output/$OUTPUT_FOLDER_NAME
+DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
+
+if [ -z "$MS_CLIENT_ID" ] || [ -z "$MS_CLIENT_CRED" ]; then
```
ryannikolaidis, line 11: should not check for MS_CLIENT_ID and MS_CLIENT_CRED here. instead we need to validate that: SHAREPOINT_CLIENTID, SHAREPOINT_CRED, and SHAREPOINT_SITE are set

test_unstructured_ingest/test-ingest-sharepoint.sh
```diff
@@ -0,0 +1,31 @@
...
+   echo "Skipping Sharepoint ingest test because the MS_CLIENT_ID or MS_CLIENT_CRED env var is not set."
+   exit 0
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    --ms-client-id "$SHAREPOINT_CLIENTID" \
+    --ms-client-cred "$SHAREPOINT_CRED" \
+    --ms-sharepoint-site "$SHAREPOINT_SITE" \
+    --ms-sharepoint-folder "Shared Documents" \
```
ryannikolaidis, line 20:
```suggestion
    --ms-sharepoint-pages "Shared Documents" \
```

ryannikolaidis, line 20:
```suggestion
    --ms-sharepoint-path "Shared Documents" \
```

ryannikolaidis (changes requested):
I'm getting `No connector-specific option was specified!` when running the test.

unstructured/ingest/main.py
ryannikolaidis: we probably should be consistent both between ENV vars and args, so yes. but also we should probably be consistent across connectors. wondering if we should drop the ms prefix altogether and refactor onedrive and outlook to just use their own connector specific variables rather than sharing ms- creds. don't need to solve for this pr, but maybe we bump that in a soon to follow pr.

test_unstructured_ingest/test-ingest-onedrive.sh
```diff
@@ -27,7 +27,6 @@ PYTHONPATH=. ./unstructured/ingest/main.py \
     --client-id "$MS_CLIENT_ID" \
     --tenant "$MS_TENANT_ID" \
     --user-pname "$MS_USER_PNAME" \
-    --onedrive-folder '/utic-test-ingest-fixtures' \
```
ryannikolaidis, line 30: should we update this to specify the `--path`?

ryannikolaidis:
You'll need to add the three SHAREPOINT_X test environment variables (from the same name stored as a GH Secret) to CI, as is done [in the CI ingest tests](https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ci.yml#L188-L201) and [in the update fixture flow](https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ingest-test-fixtures-update-pr.yml#L60-L74)

test_unstructured_ingest/expected-structured-output/sharepoint/Shared Documents/ideas-page.json
```diff
@@ -0,0 +1,55 @@
...
+          "ServerRelativeUrl": "/Shared Documents/ideas-page.html",
+          "TimeCreated": "2023-06-16T05:04:47Z",
+          "TimeLastModified": "2023-06-16T05:04:47Z",
+          "Title": "How to Get New Ideas",
+          "UIVersion": 512,
+          "UIVersionLabel": "1.0",
+          "UniqueId": "0dfe3d76-00c0-42db-ae1b-8cf22d4b3f10",
+          "ServerRelativePath": {
+            "DecodedUrl": "/Shared Documents/ideas-page.html"
+          }
```
ryannikolaidis, line 32: Looks like a lot of data we don't need in here. record_locator should container only data necessary to uniquely identify the document on the source system

unstructured/ingest/cli/cli.py
```diff
@@ -29,14 +29,15 @@ def ingest():
     cli_cmds.local,
     cli_cmds.elasticsearch,
     cli_cmds.confluence,
+    cli_cmds.sharepoint,
 ]
 
 for subcommand in subcommands:
-    ingest.add_command(subcommand())
+    ingest.add_command(subcommand())  # type: ignore
```
ryannikolaidis, line 36: why was that ignore line needed? I don't see this required or flagged in our linting in CI

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,312 @@
...
+    BaseIngestDoc,
+    ConnectorCleanupMixin,
+    IngestDocCleanupMixin,
+    StandardConnectorConfig,
+)
+from unstructured.ingest.logger import logger
+from unstructured.utils import requires_dependencies
+
+if TYPE_CHECKING:
+    from office365.sharepoint.files.file import File
```
ryannikolaidis, line 20: why do we import the File class here but then use the string "File" in the actual typing? can we remove this? benefit that then @requires_dependencies decorators will catch if the library isn't installed

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,312 @@
...
+                logger.info("Caught an error while processing documents %s", e.response.text)
+            return []
+
+    @requires_dependencies(["office365"])
+    def _list_pages(self, site_client) -> list:
+        from office365.runtime.client_request_exception import ClientRequestException
+
+        try:
+            pages = site_client.site_pages.pages.get().execute_query()
+            pfiles = []
```
ryannikolaidis, line 241: lets call these page_files instead of pfiles

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,312 @@
...
+        except ClientRequestException as e:
+            logger.info("Caught an error while processing pages %s", e.response.text)
+            return []
+
+        return pfiles
+
+    def initialize(self):
+        pass
+
+    def _ingest_site_docs(self, site_client) -> List["SharepointIngestDoc"]:
```
ryannikolaidis, line 265: why is SharepointIngestDoc in quotes here? can't we actually use the class in this case?

unstructured/ingest/cli/cmds/sharepoint.py
```diff
@@ -0,0 +1,76 @@
...
+    within the tenant this filter will be applied to all sites document libraries. \
+    Default 'Shared Documents'",
+)
+@click.option(
+    "--process-all",
+    is_flag=True,
+    default=False,
+    help="Process all sites within the tenant. Requires the app to be registered at a tenant \
+        level and a site url in https://[tenant]-admin.sharepoint.com.",
+)
```
ryannikolaidis, line 47: rather than having two parameters how about we process all sites if you don't specify a specific site with the --site flag?

also please call out this logic in the cli help text and the example

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,312 @@
...
+            for page_meta in pages:
+                page_url = page_meta.get_property("Url", None)
+                if page_url is None:
+                    logger.info("Missing site_url. Omitting page... ")
+                    break
+                page_url = f"/{page_url}" if page_url[0] != "/" else page_url
+                file_page = site_client.web.get_file_by_server_relative_path(page_url)
+                site_path = None
+                if (spath := (urlparse(site_client.base_url).path)) and (spath != "/"):
+                    site_path = spath[1:]
```
ryannikolaidis, line 252: I don't like the use of the name spath here. is there a more accurate name? full_site_path?

unstructured/ingest/cli/cmds/sharepoint.py
ryannikolaidis: no, too finicky and hard to discover. 
by "prefix is not present" do you mean if the --site option isn't set? hmm, what are we _currently_ doing if it --site flag is not?

unstructured/ingest/connector/sharepoint.py
ryannikolaidis: that works

unstructured/ingest/cli/cli.py
ryannikolaidis: sure, but note that we don't see this issue in CI. meaning this is an issue with your local setup.

what does `mypy --version` return?
we should see [1.4.1](https://github.com/Unstructured-IO/unstructured/blob/f97300b7a0dc350bc1f21fd661fd90d7fae40684/requirements/test.txt#L55)

test_unstructured_ingest/expected-structured-output/sharepoint/Shared Documents/ideas-page.json
ryannikolaidis: Seems like? If you needed to write a script to pull retrieve this file programmatically, could you do it with those pieces of information? Would you need a tenant ID? Can assume that one has credentials, but does this give you enough data to fetch that document again?

ryannikolaidis: okay, cool, let's have those three then?

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,312 @@
...
+    path: str
+    process_all: bool = False
+    process_pages: bool = False
+    recursive: bool = False
+
+    def __post_init__(self):
+        if not (self.client_id and self.client_credential and self.site_url):
+            raise ValueError(
+                "Please provide one of the following mandatory values:"
+                "\n-ms-client_id\n-ms-client_cred\n-ms-sharepoint-site",
```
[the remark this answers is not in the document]
ryannikolaidis: ...you resolved this, but did not update these keys? they still have the no longer correct ms- prefix

unstructured/ingest/cli/cmds/sharepoint.py
ryannikolaidis: the site url would only have that address if we wanted to process all sites though, correct? if so, can we just detect if the --site url has that form and if so then process all sites?

ryannikolaidis:
hitting this error when I try to run the ingest test locally:

```
  File "/Users/ryannikolaidis/Development/unstructured/unstructured/unstructured/ingest/connector/sharepoint.py", line 236, in _list_files
    for f in items.items:
AttributeError: 'Folder' object has no attribute 'itemss'
Error: 'Folder' object has no attribute 'itemss'```

tried correcting to `items` without the typo, but got the same error.

**Comment by ryannikolaidis** [comment]
wondered why that ^ wasn't tripping up in CI, but it looks like you haven't added the sharepoint test to [test-ingest](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest.sh). I think that is part of the checklist. can you add it?

File: test_unstructured_ingest/test-ingest-sharepoint.sh [hunk 18]
```diff
@@ -0,0 +1,32 @@
...
+
+if [ -z "$SHAREPOINT_CLIENT_ID" ] || [ -z "$SHAREPOINT_CRED" ]; then
+   echo "Skipping Sharepoint ingest test because the MS_CLIENT_ID or MS_CLIENT_CRED env var is not set."
+   exit 0
+fi
+# excluding metadata.last_modified since this will always update as date processed because the Sharepoint connector creates documents on the fly
+PYTHONPATH=. ./unstructured/ingest/main.py \
+    sharepoint \
+    --download-dir "$DOWNLOAD_DIR" \
+    --metadata-exclude file_directory,metadata.data_source.date_processed,meta.last_modified \
```
ryannikolaidis, line 19:
```suggestion
    --metadata-exclude file_directory,metadata.data_source.date_processed,metadata.last_modified \
```

test_unstructured_ingest/expected-structured-output/sharepoint/Shared Documents/fake-text.json
```diff
@@ -0,0 +1,254 @@
...
+          "ServerRelativeUrl": "/Shared Documents/fake-text.txt",
+          "TimeCreated": "2023-06-16T05:04:55Z",
+          "TimeLastModified": "2023-06-16T05:04:55Z",
+          "Title": null,
+          "UIVersion": 512,
+          "UIVersionLabel": "1.0",
+          "UniqueId": "880f80ca-cebf-48d0-b639-aeb671b3c431",
+          "ServerRelativePath": {
+            "DecodedUrl": "/Shared Documents/fake-text.txt"
+          }
```
ryannikolaidis, line 74: why are there still expected results with all of these fields we decided to remove in it?

.github/workflows/ci.yml
```diff
@@ -200,6 +200,10 @@ jobs:
         MS_TENANT_ID: ${{ secrets.MS_TENANT_ID }}
         MS_USER_EMAIL: ${{ secrets.MS_USER_EMAIL }}
         MS_USER_PNAME: ${{ secrets.MS_USER_PNAME }}
+        BOX_APP_CRED: ${{ secrets.BOX_APP_CRED }}
```
ryannikolaidis, line 203: not necessary as this was renamed and is defined [here](https://github.com/Unstructured-IO/unstructured/blob/b4fe40e4847d4517fc9ece58743222bf38ffe633/.github/workflows/ci.yml#L189)

test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/fake-text.json
```diff
@@ -0,0 +1,110 @@
...
+  },
+  {
+    "type": "ListItem",
+    "element_id": "39a3ae572581d0f1fe7511fd7b3aa414",
+    "metadata": {
+      "data_source": {
+        "record_locator": {
+          "site": "https://unstructuredio.sharepoint.com/",
+          "UniqueId": "880f80ca-cebf-48d0-b639-aeb671b3c431",
+          "ServerRelativeUrl": "/Shared Documents/fake-text.txt"
```
ryannikolaidis, line 64: lets make these as snake case per convention. so: `unique_id` and `server_relative_url`

unstructured/ingest/connector/sharepoint.py
```diff
@@ -0,0 +1,326 @@
...
+            record_source = self.meta["page"]
+            resource_url_name = "AbsoluteUrl"
+        else:
+            record_source = self.file
+            resource_url_name = "ServerRelativeUrl"
+
+        return {
+            "site": self.config.site_url,
+            "UniqueId": record_source.get_property("UniqueId", ""),
+            resource_url_name: record_source.get_property(resource_url_name, ""),
```
ryannikolaidis, line 121: see note above on snake case for any keys we store here

ryannikolaidis (commented):
needs a bump to the -dev version

test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/ideas-page.json
```diff
@@ -0,0 +1,37 @@
+[
+  {
+    "type": "NarrativeText",
+    "element_id": "c08fcabe68ba13b7a7cc6592bd5513a8",
+    "metadata": {
+      "data_source": {
+        "record_locator": {
+          "site": "https://unstructuredio.sharepoint.com/",
+          "unique_id": "0dfe3d76-00c0-42db-ae1b-8cf22d4b3f10",
+          "server_relative_url": ""
```
ryannikolaidis, line 10: why is this empty? it's root?

test_unstructured_ingest/expected-structured-output/Sharepoint/SitePages/Home.json
```diff
@@ -0,0 +1,42 @@
...
+  {
+    "type": "Title",
+    "element_id": "b4e929d8bcfe04189801a8ed61496d17",
+    "metadata": {
+      "data_source": {
+        "version": "1.2",
+        "record_locator": {
+          "site": "https://unstructuredio.sharepoint.com/",
+          "unique_id": "2b564fff-e9bb-4b64-9822-64f96a20ea10",
+          "absolute_url": ""
```
ryannikolaidis, line 11: why is this empty?

test_unstructured_ingest/expected-structured-output/sharepoint/Shared Documents/fake-text.json
ryannikolaidis: looks like you addressed this.

test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/fake-text.json
ryannikolaidis: looks like you addressed this.

ryannikolaidis:
let's update the docs [here](https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/connectors.rst)

.github/workflows/ingest-test-fixtures-update-pr.yml
```diff
@@ -72,6 +72,10 @@ jobs:
           MS_TENANT_ID: ${{ secrets.MS_TENANT_ID }}
           MS_USER_EMAIL: ${{ secrets.MS_USER_EMAIL }}
           MS_USER_PNAME: ${{ secrets.MS_USER_PNAME }}
+          BOX_APP_CRED: ${{ secrets.BOX_APP_CRED }}
```
ryannikolaidis, line 75: ahh, same here, this is not needed

ryannikolaidis (approved):
lgtm! I know this was a big one, thanks for all the effort here.

PR #919, opened 2023-07-12
docs: update table and API documentation

### LINKED ISSUE #693: Add more documentation on table extraction
Currently there is not much documentation on table extraction and how it works. The goal of this issue is to add a table extraction section to the documentation. We should also add an example to the `partition_pdf` docs that shows how to use the `infer_table_structure` kwarg.

MthwRobinson:
### Summary

Closes #693. Adds more detailed documentation on `partition_via_api` and table extraction capabilities.

### DIFF
--- a/docs/source/bricks.rst
+++ b/docs/source/bricks.rst
@@ -753,27 +753,56 @@ You can disable ``max_partition`` by setting it to ``None``.
 ``partition_via_api`` allows users to partition documents using the hosted Unstructured API.
 The API partitions documents using the automatic ``partition`` function. Currently, the API
 supports all filetypes except for RTF and EPUBs.
-To use another URL for the API use the ``api_url`` kwarg. This is helpful if you're hosting
+This is helpful if you're hosting
 the API yourself or running it locally through a container. You can pass in your API key
 using the ``api_key`` kwarg. You can use the ``content_type`` kwarg to pass in the MIME
 type for the file. If you do not explicitly pass it, the MIME type will be inferred.
 
-See `here <https://api.unstructured.io/general/docs>`_ for the hosted API swagger documentation
-and `here <https://github.com/Unstructured-IO/unstructured-api#dizzy-instructions-for-using-the-docker-image>`_ for
-documentation on how to run the API as a container locally.
-
-Examples:
 
 .. code:: python
 
   from unstructured.partition.api import partition_via_api
 
-  filename = "example-docs/fake-email.eml"
+  filename = "example-docs/eml/fake-email.eml"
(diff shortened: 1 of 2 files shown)


docs/source/bricks.rst
```diff
@@ -753,29 +753,58 @@ You can disable ``max_partition`` by setting it to ``None``.
 ``partition_via_api`` allows users to partition documents using the hosted Unstructured API.
 The API partitions documents using the automatic ``partition`` function. Currently, the API
 supports all filetypes except for RTF and EPUBs.
```
qued, line 755: We still don't support these in the API?

docs/source/bricks.rst
```diff
@@ -753,29 +753,58 @@ You can disable ``max_partition`` by setting it to ``None``.
...
+  filename = "example-docs/eml/fake-email.eml"
 
   elements = partition_via_api(filename=filename, api_key="MY_API_KEY", content_type="message/rfc822")
 
   with open(filename, "rb") as f:
     elements = partition_via_api(file=f, file_filename=filename, api_key="MY_API_KEY")
 
 
+You can pass additional settings such as ``strategy``, ``ocr_languages`` and ``encoding`` to the
+API through optional ``request_kwargs``. These options get added to the request body when the
```
qued, line 775: Is `request_kwargs` a parameter name? From the example it seems like it's not, but it being written in the monospace font with an underscore suggest it is. Okay I see now, it's a ** parameter. It might be more confusing to make note of the parameter name that the dynamic kwargs are stored in, but maybe it's just me.

docs/source/getting_started.rst
```diff
@@ -110,6 +109,43 @@ The following code shows how you can limit your output to only narrative text wi
 	        print("\n")
 
 
+######
+Tables
+######
+
+For ``Table`` elements, the raw text of the table will be stored in the ``text`` attribute for the Element, and HTML representation
+of the table will be available in the element metadat under ``element.metadata.text_as_html``. For most documents where
```
qued, line 117:
```suggestion
of the table will be available in the element metadata under ``element.metadata.text_as_html``. For most documents where
```

docs/source/bricks.rst
MthwRobinson: I think we do now, @yuming-long bumped the `pandocs` version a while back. Will update that.

docs/source/bricks.rst
MthwRobinson: Yeah I'll remove that, I agree it's more confusing than helpful

PR #923, opened 2023-07-12
docs: Add source code links to bricks' docs

fran-unstructured:
Adding the source code for each brick in bricks.rst, and resorting partition_tsv and partition_org to keep alphabetical order.

I added only the file link for the source code links and not the line of the particular function to keep it maintainable.


MthwRobinson (approved):
LGTM! And thanks for the alphabetical ordering fix.

Issue #521, opened 2023-04-26
Out of memory processing pdf


qued:
Running `partition_pdf` on the following document resulted in a killed process due to running out of memory on Ubuntu. The machine has (only) 8 gigs of memory total.

Since pages are processed individually, only one page should sit in memory at a time. Maybe that's not true? Or is one particular page of this document causing an issue?

[irm02-003-059--2022-12-12.pdf](https://github.com/Unstructured-IO/unstructured/files/11337442/irm02-003-059--2022-12-12.pdf)

qued:
Closing as addressed by [PR #136](https://github.com/Unstructured-IO/unstructured-inference/pull/136). Thanks @christinestraub !

christinestraub:
@qued Created [PR #924](https://github.com/Unstructured-IO/unstructured/pull/924)

PR #924, opened 2023-07-12
Fix/521 pdf2image memory error

christinestraub:
### Summary
Closes [issue #521](https://github.com/Unstructured-IO/unstructured/issues/521). Implements the same logic as [unstructured-inference/PR #136](https://github.com/Unstructured-IO/unstructured-inference/pull/136) for the `ocr_only` strategy.

- Add functionality to convert a PDF in small chunks of pages at a time
- Add functionality to write images to computer storage temporarily instead of keeping them in memory
- Set the file's current position to the beginning after reading the file in `convert_to_bytes`

### Testing
```
filename = "example-docs/pdf2image-memory-error-test-400p.pdf"
elements = partition_pdf(filename, strategy="ocr_only")
```


unstructured/partition/pdf.py
```diff
@@ -363,6 +364,38 @@ def _process_pdfminer_pages(
...
+                first_page=start_page,
+                last_page=end_page,
+            )
+        else:
+            chunk_images = pdf2image.convert_from_path(
+                filename,
+                first_page=start_page,
+                last_page=end_page,
+            )
+        yield chunk_images
```
qued, line 396: The only improvement I can think of would be to `yield` the individual images rather than the chunks, while still handling things as chunks under the hood. However, I think what's here is good to commit rather than delaying. We can come back later and tweak the interface.

unstructured/partition/pdf.py
```diff
@@ -363,6 +364,38 @@ def _process_pdfminer_pages(
     return elements
 
 
+def convert_pdf_to_images(
+    filename: str = "",
+    file: Optional[Union[bytes, BinaryIO, SpooledTemporaryFile]] = None,
+    chunk_size: int = 10,
+) -> Iterator[List[object]]:
+    # Convert a PDF in small chunks of pages at a time (e.g. 1-10, 11-20... and so on)
```
qued, line 372:
```suggestion
    ```Convert a PDF in small chunks of pages at a time (e.g. 1-10, 11-20... and so on)```
```

qued (approved):
LGTM, I like the generator implementation! One formatting tweak, and a general (nonblocking) suggestion for the function interface.

unstructured/partition/pdf.py
christinestraub: @qued I got your point. At first, I also considered yielding individual images rather than chunks so I compared elapsed time when `chunk_size` is 1 and `chunk_size` is 10. The result showed that handling things as chunks is a bit faster than handling things as individual images. Please check the test result below. 

```
- chunk_size: 1
    - elapsed_time: 835.2164249420166
    - memory_peak: 255.3 MB
- chunk_size: 10
    - elapsed_time: 808.2655990123749
    - memory_peak: 277.3 MB
```

unstructured/partition/pdf.py
christinestraub: LGTM!

unstructured/partition/pdf.py
qued: To be clear, I'm saying we should load chunks, but yield individual images from the generator, so that we get the efficiency of chunks but the person using the generator doesn't have to manage the chunks.

christinestraub: Sounds good! I'll get to that soon. Thanks for your suggestion! @qued

unstructured/partition/pdf.py
```diff
@@ -363,6 +364,40 @@ def _process_pdfminer_pages(
     return elements
 
 
+def convert_pdf_to_images(
+    filename: str = "",
+    file: Optional[Union[bytes, BinaryIO, SpooledTemporaryFile]] = None,
+    chunk_size: int = 10,
+) -> Iterator[List[object]]:
```
qued, line 371: Should this change now?
```suggestion
) -> Iterator[PIL.Image.Image]:
```

christinestraub: Fixed it. Thanks!

PR #926, opened 2023-07-13
Add min_partition kwarg to that combines elements below a specified threshold

Coniferish:
## Summary
Add min_partition kwarg to that combines elements below a specified threshold and modifies splitting of strings longer than max partition so words are not split.

## Testing
```
from unstructured.partition.text import split_content_to_fit_min_max
min_max_text = "This is a story. This is a story that doesn't matter because it is just being used as an example. Hi. Hello. Howdy. Hola. The example is simple and repetitive and long and somewhat boring, but it serves a purpose. End."
segments = split_content_to_fit_min_max(content=min_max_text, max_partition=75, min_partition=3)
segments = split_content_to_fit_min_max(content=min_max_text, max_partition=75, min_partition=6)
```


unstructured/partition/text.py
```diff
@@ -26,52 +27,125 @@
...
 
+def combine_paragraphs_less_than_min(
+    split_paragraphs: List[str],
+    max_partition: Optional[int] = 1500,
+    min_partition: Optional[int] = 0,
+) -> List[str]:
+    """Combine paragraphs less than min_partition while not exceeding max_partition."""
+    file_content: List[str] = []
+    tmp_paragraph = ""
+    y = 0
```
MthwRobinson, line 114: Could you give `x`, `y`, and `z` more descriptive variable names?

unstructured/partition/text.py
```diff
@@ -26,52 +27,125 @@
...
-    chunk = ""
-
-    for i, sentence in enumerate(sentences):
-        if len(sentence) > max_partition:
-            chunks.extend(_split_content_size_n(sentence, n=max_partition))
-
-        if len(chunk + " " + sentence) > max_partition:
-            chunks.append(chunk)
-            chunk = sentence
+    tmp = ""
```
MthwRobinson, line 68: Could you change this back to `chunk` (or `tmp_chunk` would be fine too) to make the role of the variable more clear?

unstructured/partition/text.py
```diff
@@ -26,52 +27,125 @@
...
+    for sentence in sentences:
+        if max_partition is not None and len(sentence) > max_partition:
```
MthwRobinson, line 70: Could you add comments to these conditionals since they're not totally self explanatory?

unstructured/partition/text.py
```diff
@@ -26,52 +27,125 @@
...
+    min_partition: Optional[int] = 0,
+) -> List[str]:
+    """Combine paragraphs less than min_partition while not exceeding max_partition."""
+    file_content: List[str] = []
+    tmp_paragraph = ""
+    y = 0
+    for x, paragraph in enumerate(split_paragraphs):
+        if y > x:
+            continue
+        if min_partition is not None and len(paragraph) < min_partition:
```
MthwRobinson, line 118: Same for the conditionals in this function, I think they would benefit from comments

MthwRobinson (changes requested):
Test cases look good. Added a couple of minor inline comments.

test_unstructured/documents/test_xml.py
```diff
@@ -44,6 +44,20 @@ def test_from_string(sample_document):
     assert type_tag.text.strip() == "10-K"
 
 
+def test_from_string_with_pre_tag():
+    sample_document = """
+    <pre>
```
MthwRobinson, line 49: I think the original bug reported for this had HTML in the `<pre>` tag, like this:

```html
<pre>
<p> Hello there! </p>
</pre>
```

@christinestraub already added a test for that in #642, so I don't think there's anything additional we need testing wise there (unless you noticed/fixed an issue with it).

unstructured/documents/xml.py
```diff
@@ -78,13 +78,15 @@ def _read_xml(self, content):
...
             except ValueError:
                 document_tree = etree.fromstring(content.encode(), self.parser)
-
             if "<pre>" and "</pre>" in content:
                 tree = etree.HTML(content)
                 for element in tree.xpath("//pre"):
                     if not element.text:
                         continue
-                    text_content = split_by_paragraph(element.text)
+                    text_content = [
```
MthwRobinson, line 86: I think a regular for loop is more readable here, but the list comprehension is fine too if you want to keep it

MthwRobinson (commented):
If you do `git commit -m "ingest-test-fixtures-update" --allow-empty` and push that, it'll trigger a ingest fixture update to fixed that failed job. Otherwise looks good once CI is passing!

test_unstructured/documents/test_xml.py
Coniferish: Ah, thanks! Didn't see that since the pre tag tests were in the auto and html test files. I think this test is still needed, though, because xml makes a call to that part of code--I forgot the exact reason why, but I was getting a test failure (maybe with the html test?), so I was ensuring things worked for xml partition as well. I can remove it if you don't think it's needed.

MthwRobinson: Yeah this is still a nice test to have too since the other one tests it from the partition function and not from the `XMLDocument` class. Thanks for adding it!

unstructured/documents/xml.py
Coniferish: So, something like this?
```
for element in partition_text(text=element.text, paragraph_grouper=False):
    text_content.append(element.text)
```

MthwRobinson: Yep! That'll do

MthwRobinson:
When I do this is looks like `max_partition` gets respected but not `min_partition`. Is that because there's no way to combine them that satisfies both? 

```python
In [20]: elements = partition_text(text=text, min_partition=50, max_partition=100)

In [21]: print("\n\n".join([str(el) for el in elements]))
This is a story. This is a story that doesn't matter because it is just being used as an example.

Hi. Hello. Howdy. Hola.

The example is simple and repetitive and long and somewhat boring, but it serves a purpose. End.

In [22]: len(elements[0].text)
Out[22]: 97

In [23]: len(elements[1].text)
Out[23]: 23

In [24]: len(elements[2].text)
Out[24]: 96
```

Coniferish:
> 

Yes, but I had the `elif` statement wrong, so the error I included wasn't being raised

MthwRobinson (approved):
LGTM!

PR #928, opened 2023-07-13
docs: Add Unstructured API documentation

fran-unstructured:
Adding something very similar to the unstructured-api Readme to the documentation for visibility purposes.

Special attention to the "Output Format" parameter that was extracted from the local implementation in the Readme, and to the "Using the API Locally" documentation.

Parallel Mode hasn't been added due to doubts about the value of exposing it to users.


docs/source/api.rst
```diff
@@ -0,0 +1,212 @@
...
+  'https://api.unstructured.io/general/v0/general' \
+  -H 'accept: application/json'  \
+  -H 'Content-Type: multipart/form-data' \
+  -H 'unstructured-api-key: <YOUR API KEY>' \
+  -F 'files=@sample-docs/fake-xml.xml' \
+  -F 'xml_keep_tags=true' \
+  | jq -C . | less -R
+
+
+Using the API Locally
```
MthwRobinson, line 163: To me the preferred way to use the API locally is through the Docker container since that requires less manual set up. Can we move this section down and update the title to "Developing with the API Locally" or similar?

docs/source/api.rst
```diff
@@ -0,0 +1,212 @@
...
+  | jq -C . | less -R
+
+
+Generating Python Files from the Pipeline Notebooks
+====================================================
+
+You can generate the FastAPI APIs from your pipeline notebooks by running ``make generate-api``.
+
+
+Using Docker Images
```
MthwRobinson, line 195: Per other comment, can we move this section up? And also add a comment to the effect of "If you are self-hosting the API or running it locally, we strongly suggest running the Docker container."

docs/source/api.rst
```diff
@@ -0,0 +1,212 @@
...
+	
+  curl -X 'POST' \
+  'http://localhost:8000/general/v0/general' \
+  -H 'accept: application/json' \
+  -H 'Content-Type: multipart/form-data' \
+  -F 'files=@sample-docs/family-day.eml' \
+  | jq -C . | less -R
+
+
+Generating Python Files from the Pipeline Notebooks
```
MthwRobinson, line 189: I'd drop this section since it's already in the API repo README and generating Python files from the API is a vestige of our old "pipeline notebook" concept. Most users won't need to do that and the ones who do can see it in the repo.

docs/source/api.rst
```diff
@@ -0,0 +1,212 @@
+Unstructured API
+#################
+
+Try our hosted API! It's freely available to use with any of the file types listed above. This is the easiest way to get started, all you need is an API key. You can get your API key `here <https://www.unstructured.io/api-key/>`_ now and start using it today.
```
MthwRobinson, line 4: Can you also add a link to the API docs page?

https://api.unstructured.io/general/docs

MthwRobinson (changes requested):
Looks great, thanks! Just a couple of minor comments

MthwRobinson (approved):
LGTM with the updates. Thanks!

PR #933, opened 2023-07-14
build: remove docs-build branch

MthwRobinson:
### Summary

Remove the option to build docs off a branch. We can actually just do this with a tag like `0.8.1-docs-rebuild` since the docs build on any tag.


qued (approved):
LGTM!

PR #934, opened 2023-07-14
fix: enable passing filters to `partition_doc` for libreoffice conversion

### LINKED ISSUE #777: bug/docx conversion fails (no export filter provided)
When using unstructured==0.7.4 with LangChain and using a DirectoryLoader that happens to traverse a .doc file, I get this:

Error: no export filter for /private/var/folders/hj/vcvdzqt5037_640tkfxfbdym0000gn/T/tmpdqn7lr0h/mydoc.docx found, aborting.

Here I found important info:

https://stackoverflow.com/questions/52277264/convert-doc-to-docx-using-soffice-not-working
Syntax for LibreOffice 7 seems to be --convert-to "docx:MS Word 2007 XML", with an output filter name supplied after the docx. Options for the output filter name ("MS Word 2007 XML") seem to be here https://git.libreoffice.org/core/+/refs/heads/master/filter/source/config/fragments/filters

And fair enough, this worked:

soffice --headless --convert-to "docx:MS Word 2007 XML"  myFile.doc 
convert myfile.doc -> myfile.docx using filter : MS Word 2007 XML

I wonder if the way Unstructured and LangChain interact can be fixed so that the conversion works.

MthwRobinson:
### Summary

Closes #777. Updates `partition_doc` to accept a `libre_office_filter` arg, which defaults to the syntax required for LibreOffice7 (as pointed out in the issue). You can pass `None` to that arg to disable the filter, which equivalent to the current functionality.

### Testing

First confirm your LibreOffice version with `soffice --version`. It should be 7.x.x. Then run.

```python

filename = "example-docs/fake.doc"
elements = partition_doc(filename=filename)
elements = partition_doc(filename=filename, libre_office_filter=None)
```


qued (approved):
LGTM!

PR #935, opened 2023-07-14
chore: remove outdated error message

qued:
There's [an issue](https://github.com/Unstructured-IO/unstructured-inference/issues/64) in `unstructured-inference` about these blocks trapping unrelated import errors. The fix for that would be to narrow the scope of the traps, but I think this is made redundant by the `requires_dependencies` decorator, so I removed it completely.

#### Testing:
I guess the way to thoroughly test this would be to:
1. Make a call to this code without `unstructured-inference` installed to make sure it raises an error through the decorator. The environment can be set up by creating a fresh environment and then running `pip install -r requirements/base.txt` (we intentionally skip installing `unstructured-inference`). Then the following should generate an error message from the dependency decorator:

```python
from unstructured.partition.pdf import _partition_pdf_or_image_local
elements = _partition_pdf_or_image_local("example-docs/layout-parser-paper.pdf")

```
2. Reproduce the error (or something similar) mentioned in the above issue, to make sure the error message comes through without the misleading message. This can be done by creating an ubuntu instance through the usual terraform, except in the `user_data` omit installing the OS packages `liblzma-dev` and `libbz2-dev`, without which the Python installation will be broken (I can provide a script that does this).

    Then in the instance, create and activate a fresh virtual environment, clone `unstructured`, and `make install` (note this installs `unstructured-inference`). From the `main` branch you can check that the above Python snippet gives a similar misleading error message to that referenced in the original issue, but on this branch the misleading message is absent.


MthwRobinson (approved):
LGTM!

PR #939, opened 2023-07-16
Potter/outlook Outlook Connector

potter-potter:
Adds Outlook connector
Hasn't been tested on email other than the devops one.
Does not download attachments.

changed file name in Onedrive example to match the rest.
changed documentation to correct the pip install.
Also notice change from or to and in main.py line 874. OneDrive and Outlook share credentials and it was triggering the Onedrive connector. Hopefully we can have an `and` instead of an `or` or else we'll need a different approach.


unstructured/ingest/connector/outlook.py
```diff
@@ -0,0 +1,214 @@
...
+    BaseConnectorConfig,
+    BaseIngestDoc,
+    ConnectorCleanupMixin,
+    IngestDocCleanupMixin,
+    StandardConnectorConfig,
+)
+from unstructured.ingest.logger import logger
+from unstructured.utils import requires_dependencies
+
+MAX_NUM_EMAILS = 10000
```
ryannikolaidis, line 22: what drives this max number? does the library trigger paging otherwise? is there a generator or some alternative so we can avoid the upper limit?

unstructured/ingest/connector/outlook.py
```diff
@@ -0,0 +1,214 @@
...
+            self.client.users[self.config.user_email].mail_folders.get().execute_query()
+        )
+
+        for folder in get_root_folders:
+            self.root_folders[folder.display_name].append(folder.id)
+            if folder.get_property("childFolderCount") > 0:
+                root_folders_with_subfolders.append(folder.id)
+
+        for folder in root_folders_with_subfolders:
+            self.recurse_folders(folder, self.root_folders)
```
ryannikolaidis, line 187: do we have folders and subfolders in our example? or does everything get flattened? probably okay if so since the file names are by id anyways, but want to make sure we're testing the nesting logic.

ryannikolaidis (commented):
pushed a tiny change to update onedrive secrets while we're making changes there.

just a couple quick questions, otherwise this looks great!

unstructured/ingest/connector/outlook.py
potter-potter: The library triggers paging as the default. Which is 10. Supposedly there is a timeout at some point. But consider that this call is just getting mail objects (It's not downloading), it should be pretty fast.

At the point that it starts downloading, it does it with separate calls, so there shouldn't be a timeout. Though it would be hard for us to test with just a couple emails...

unstructured/ingest/connector/outlook.py
potter-potter: We do have folders and subfolders. And I have tested double folders, etc...

Everything is returned flattened for simplicity. The way the api lists parent folders was less than ideal, so this was my simple solution. 

I'd be happy to look at subfolders in the partitioned return if you think it's important.

unstructured/ingest/connector/outlook.py
ryannikolaidis: got it, makes sense. is there any reason we need to limit this at all?

potter-potter: If we don't use `.top()` (line 209) it will default to 10. if we use `.top()` with nothing in it. I think it always defaults to 10 also. I will look in the documentation.

potter-potter: Looks like top needs an parameter number. At least there is no default in the documentation. Without .top you only get 10 documents

unstructured/ingest/connector/outlook.py
ryannikolaidis: Nope, I think that's fine. As long as getting the filename is ultimately unique and deterministic, since we currently rely on that to determine if we should process a file on subsequent runs.

unstructured/ingest/connector/outlook.py
ryannikolaidis: that's so bizarre! issue is that's the pre-filtered number, so that will include you're entire inbox (I'm guessing trash too?). I guess I'm tempted to make this something much bigger, like 10M. kind of a bummer we have to query all of the full messages and hold them in memory, unlike other connectors where each doc is responsible for actually fetching content, bc that does have me slightly worried about things blowing up and waiting for a giant call before processing. 

fwiw it [looks like](https://github.com/vgrem/Office365-REST-Python-Client/blob/2b603f5c9643ece12b40cfcaf1e4005cbca141e1/examples/outlook/messages/list_new.py#L12) there's a way to filter in that original query by mailbox, right? if that's correct, we should probably do that for the sake of reducing memory and request demands and skirting around the max fetch a bit.

potter-potter: ooh. let me look into that.

potter-potter: I'll look deeper into this. For some reason I am recalling that a few of the root folders they allow "Inbox" as the ID (and "Sent Items" and a few others). There may have been a reason that this did not work though in the long run for non standard folders. I can't totally remember. But I will definitely do some exploration.

potter-potter: Ok. It will only take folder IDs in that mail_folder. BUT we could get all the folders and subfolders first (and their ids) and then collect the mail after that we get the folders. Then we don't have to do any filtering later. And we will only get the relevant emails.

Glad you thought of doing that.

Should be a reasonably easy fix.

ryannikolaidis: Awesome! Thanks for digging here.

ryannikolaidis (approved):
lgtm! appreciate the catch/fix on the email partitioning!

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.8.3-dev1
+
+### Enhancements
+
+### Features
+
+* Adds Outlook connector.
+
+### Fixes
+
```
ryannikolaidis, line 10:
```suggestion
### Fixes

* Fixes issue with email partitioning where From field was being assigned the To field value.

```

ryannikolaidis:
@potter-potter can you update this to latest, bump version (accordingly), and review quick suggestion for changelog?

Issue #606, opened 2023-05-17
Add element type filters to the partition functions


MthwRobinson:
The goal of this issue is to add an `include_element_types` and `exclude_element_types` kwarg to the partition functions. `include_element_types` that will determine what element types to keep in the outputs. `include_element_types` will indicate "I only want these element types" and `exclude_element_types` will indicate "I want everything except these element types". This will help for downstream users in `langchain` who currently don't have an easy way to filter on element types during the load step.

kravetsmic:
What  if `include_element_types` and `exclude_element_types` overlap, For example:
include = ["A", "B"]
exclude = ["A", "C"]

MthwRobinson:
You could check the intersection and raise an error if it isn't empty. Might be better to say you can only use include or exclude but not both. Either way I think you'd want to raise an exception.

PR #947, opened 2023-07-18
Kravetsmic/606 add element type filters to the partition functions

kravetsmic:
### Summary

Added `exclude_element_types` and `include_element_types` parameters to doc, docx, email, epub, html, image, json, md, msg, odt, org, pdf, ppt, pptx, rst, rtf, text, xml (all of the types except csv, tsv, xlsx).
`include_element_types` - indicates "I only want these element types".
`exclude_element_types` - indicates "I want everything except these element types".

### New Parameters

<details open>
<summary>include_element_types</summary>
Not required parameter. Accepts the list of element types
Default value: []
</details open>

<details open>
<summary>exclude_element_types</summary>
Not required parameter. Accepts the list of element types
Default value: []
</details open>

Bump to 0.8.3

---

Closes issue #606


MthwRobinson (changes requested):
Can we close this PR and open a new PR that just has `filter_element_types` as a function in `staging.base`? This is a simple post-processing step, so I don't think we need the complication of adding handling this through kwargs in partition (though I know the original issue said to do that).

kravetsmic:
> Can we close this PR and open a new PR that just has `filter_element_types` as a function in `staging.base`? This is a simple post-processing step, so I don't think we need the complication of adding handling this through kwargs in partition (though I know the original issue said to do that).

Sorry, but can you explain a little bit more what exactly you want me to do?

MthwRobinson:
@kravetsmic - You can just move the `filter_element_types` you have in `partition/common.py` to `staging/base.py` and write a small test for it. Then we have can use that as a post processing function. We don't need to add a kwarg for filtering.

PR #966, opened 2023-07-24
Roman/expose dpi param

rbiseck3:
### Description
Bump the inference library version to `0.5.6` which exposes the new dpi parameter. Check for this in the kwargs being passed through and set it on the kwargs being used my the method that had that added only if it exists to avoid passing `None` for the value. 

Also update how the image metadata is worked with given the refactor that was implemented in the inference library. 

Update `CHANGELOG` to be ready to cut new release

### DIFF
--- a/test_unstructured/partition/test_pdf.py
+++ b/test_unstructured/partition/test_pdf.py
@@ -386,10 +386,16 @@ def test_partition_pdf_with_copy_protection():
 
 def test_partition_pdf_with_dpi():
     filename = os.path.join("example-docs", "copy-protected.pdf")
-    elements = pdf.partition_pdf(filename=filename, strategy="hi_res", pdf_image_dpi=100)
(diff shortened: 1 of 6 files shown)


test_unstructured/partition/test_pdf.py
```diff
@@ -384,6 +384,14 @@ def test_partition_pdf_with_copy_protection():
     assert {element.metadata.page_number for element in elements} == {1, 2}
 
 
+def test_partition_pdf_with_dpi():
+    filename = os.path.join("example-docs", "copy-protected.pdf")
+    elements = pdf.partition_pdf(filename=filename, strategy="hi_res", pdf_image_dpi=100)
```
cragwolfe, line 389: worth asserting the default `pdf_image_dpi` somewhere?

cragwolfe (approved):

PR #967, opened 2023-07-25
Update `partition_via_api` to not post a strategy value if not user specified

shreyanid:
### Summary 

Removes using `hi_res` as the default strategy value for the API. Does not post any `strategy` form var if not specified by the user, and lets the API default behavior take care of selecting the strategy.

### Test
Wrote two tests in `test_unstructured/partition/test_api.py`:

`test_partition_via_api_with_no_strategy()`
- tests that the default strategy is no longer `hi_res` by comparing that the output is different from the `hi_res` output for an image

`test_partition_via_api_with_image_hi_res_strategy_includes_coordinates()`
- tests that the `hi_res` strategy can output coordinates


MthwRobinson (approved):
LGTM! Don't think this is a breaking change either since `strategy` is coming from `request_kwargs`.

PR #968, opened 2023-07-25
Set default strategy for images to be "hi_res"

shreyanid:
### Summary

Set default strategy for images (not PDFs) to be `hi_res`.

### Test

`test_partition_image_default_strategy_hi_res()` in `test_unstructured/partition/test_image.py` checks that the `hi_res` strategy was used when no strategy was specified by confirming that the first line of the output matches the expected `hi_res` outputted first line

### DIFF
--- a/test_unstructured/partition/test_auto.py
+++ b/test_unstructured/partition/test_auto.py
@@ -380,6 +380,25 @@ def test_partition_pdf_doesnt_raise_warning():
     ("pass_file_filename", "content_type"),
     [(False, None), (False, "image/jpeg"), (True, "image/jpeg"), (True, None)],
 )
+def test_auto_partition_image_default_strategy_hi_res(pass_file_filename, content_type):
+    filename = os.path.join(EXAMPLE_DOCS_DIRECTORY, "layout-parser-paper-fast.jpg")
+    file_filename = filename if pass_file_filename else None
+    elements = partition(
+        filename=filename,
+        file_filename=file_filename,
+        content_type=content_type,
+        strategy="auto",
+    )
+
+    # should be same result as test_partition_image_default_strategy_hi_res() in test_image.py
+    first_line = "LayoutParser: A Unified Toolkit for Deep Learning Based Document Image Analysis"
+    assert elements[0].text == first_line
+
+
+@pytest.mark.parametrize(
+    ("pass_file_filename", "content_type"),
+    [(False, None), (False, "image/jpeg"), (True, "image/jpeg"), (True, None)],
+)
 def test_auto_partition_jpg(pass_file_filename, content_type):
     filename = os.path.join(EXAMPLE_DOCS_DIRECTORY, "layout-parser-paper-fast.jpg")
     file_filename = filename if pass_file_filename else None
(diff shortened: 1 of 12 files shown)


cragwolfe:
To handle the case where `partition()` from the unstructured.partition.auto is called and the file happens to be an image, i think the translation from "auto" to "hi_res" still needs to happen somewhere. here https://github.com/Unstructured-IO/unstructured/blob/bfcc565639c26b210d5c128cdb24efb6fa0d4e45/unstructured/partition/strategies.py#L53 perhas?

shreyanid:
Right, thanks for the pointer. At the line you mentioned, that just ensures that the fast strategy does not get used for images (since it's not supported), and I can go ahead and change the fallback there to be hi_res instead of ocr_only. 

The actual consideration is what to do when "auto" is called - for an image, it leads to this function `_determine_image_auto_strategy` https://github.com/Unstructured-IO/unstructured/blob/bfcc565639c26b210d5c128cdb24efb6fa0d4e45/unstructured/partition/strategies.py#L118, which chooses hi_res if `infer_table_structure` is true (which it seems like it never gets set to true for images) and ocr_only otherwise. Do we want to change this behavior of "auto" to always be hi_res instead?

My thought is that since my initial change already set the default strategy to hi_res instead of auto, users will now only be able to choose `auto` explicitly, and the existing auto decision making in `_determine_image_auto_strategy` should suffice. If instead we wanted the auto strategy for images to result in hi_res in all cases, I can make that change, but I think there probably exist some cases where ocr_only should be chosen over hi_res (maybe very large images). We should either leave the auto image strategy unchanged, or revisit the cases altogether, but not set hi_res as the only option outputted by the auto strategy.

test_unstructured/partition/test_auto.py
```diff
@@ -376,6 +376,25 @@ def test_partition_pdf_doesnt_raise_warning():
...
+    elements = partition(
+        filename=filename,
+        file_filename=file_filename,
+        content_type=content_type,
+        strategy="auto",
+    )
+
+    # should be same result as test_partition_image_default_strategy_hi_res() in test_image.py
+    first_line = "LayoutParser: A Unified Toolkit for Deep Learning Based Document Image Analysis"
+    assert elements[0].text == first_line
```
cragwolfe, line 395: it would also be worth asserting the presence of non-None value for `coordinates`

unstructured/partition/image.py
```diff
@@ -12,39 +12,45 @@ def partition_image(
...
+            model if to identify document elements. When using the "ocr_only" strategy,
+            partition_image simply extracts the text from the document using OCR and processes it.
+    <<<<<<< HEAD
+            The default strategy `auto` will determine when a image can be extracted using
+            `ocr_only` mode, otherwise it will fall back to `hi_res`.
+        metadata_last_modified
+    =======
+            The default strategy is `hi_res`.
+        metadata_date
+    >>>>>>> 2a1e725 (set default image strategy to hi_res, test, update changelog)
```
cragwolfe, line 51: merge artifact here needs cleanup

test_unstructured/partition/test_auto.py
shreyanid: added

cragwolfe (approved):

PR #973, opened 2023-07-25
Chore: Pass table support  param to partition image

yuming-long:
### Summary
* Add parameter `skip_infer_table_types` to enable (skip) table extraction for other doc types (currently support `pdf`, `jpg`, `png`).


unstructured/partition/auto.py
```diff
@@ -193,6 +193,7 @@ def partition(
             file=file,  # type: ignore
             url=None,
             include_page_breaks=include_page_breaks,
+            infer_table_structure=pdf_infer_table_structure,
```
MthwRobinson, line 196: Only thing that's a little weird here is that we're passing `pdf_infer_table_structure` to `partition_image`.

cragwolfe: yeah, i can see the case for wanting `image_infer_table_structure`, `pdf_infer_table_structure`, and `other_infer_table_structure` given that image table inference is still expensive (especially for many docs), but not as expensive as large PDF docs. but that is getting awkward.

one solution to slowly move to a new top level of `docs_types_no_infer_table_structure` that has `pdf` as the default, but could be `pdf,jpg,png,tiff`?

[one subtlety to note is that tiff's tend to be many page, more like PDF's]

MthwRobinson: Yeah I'm game for that, and I like that specify a setting per document type. I'd probably go for a shorter kwarg name though, like `skip_infer_table`

yuming-long: do i need to add `tiff`? i don't think we support it right now.

MthwRobinson: Don't need to add `tiff` in this PR, I can create a follow on issue for that

MthwRobinson: #977

yuming-long: another question: in order to use table extraction for PDF, a user has to set `pdf_infer_table_structure` to true, and also set `skip_infer_table_types` to not include `"pdf"`, is that too much trouble for them to use?

MthwRobinson: What about `pdf_infer_table_structure` take precedence for PDFs if it's set to `True`, and logging out a warning that says that you should use `skip_infer_table_types` to turn on table extraction for PDFs instead?

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
-## 0.8.2-dev6
+## 0.8.2
 
 ### Enhancements
 
+* Pass parameter `pdf_infer_table_structure` in partition_image to support image table extration
```
MthwRobinson: Small nit: partition_image -> `partition_image` (code ticks)

unstructured/partition/auto.py
```diff
@@ -274,3 +284,30 @@ def file_and_type_from_url(
...
     filetype = detect_filetype(file=file, content_type=content_type, encoding=encoding)
     return file, filetype
+
+
+def decide_table_extraction(
+    filetype: Optional[FileType],
+    skip_infer_table_types: List[str],
+    pdf_infer_table_structure: bool,
+) -> bool:
+    FILETYPE_MAPPING = {
```
MthwRobinson, line 294: Instead of this mapping you could do `filetype.name.lower()`, then we won't have to remember to add to this each time.

CHANGELOG.md
yuming-long: this changelog is outdated, if you refresh you could see the latest one

MthwRobinson: Yeah I noticed that after I posted and couldn't figure out how to uncomment 😂

PR #978, opened 2023-07-26
Roman/ingest refactor

rbiseck3:
### Description
As we expand the number of connectors supported, it's going to become unwieldy to fold in cli params for each connector. This PR is meant to pull each connector into its own subcommand which also allows us to enforce certain fields exist using click when that subcommand is called. Also allowed us to remove the prefix on all fields since they're no longer conflicting.


unstructured/ingest/cli/cmds/azure.py
```diff
@@ -0,0 +1,94 @@
...
+    process_documents,
+    update_download_dir_remote_url,
+)
+from unstructured.ingest.logger import ingest_log_streaming_init, logger
+
+
+@click.command()
+@click.pass_context
+@click.option(
+    "--remote-url",
```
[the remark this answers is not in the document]
rbiseck3: I only left that on subcommands that leverage the remote url. Others don't (i.e. local)

unstructured/ingest/process.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # remove docs that have already been processed
+        if not self.reprocess:
+            docs = self._filter_docs_with_outputs(docs)
+            if not docs:
+                return
+
+        # Debugging tip: use the below line and comment out the mp.Pool loop
+        # block to remain in single process
+        self.doc_processor_fn(docs[0])
```
jasonbot, line 74: P{probably want to switch back to uncommenting the `.Pool` option here

ryannikolaidis:
Thanks for posting for early feedback and also for reading my mind on the need for this. Will try try to go through this today.

test_unstructured_ingest/test-ingest-onedrive.sh
```diff
@@ -8,24 +8,26 @@ OUTPUT_FOLDER_NAME=onedrive
 OUTPUT_DIR=$SCRIPT_DIR/structured-output/$OUTPUT_FOLDER_NAME
 DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
 
-if [ -z "$MS_CLIENT_ID" ] || [ -z "$MS_CLIENT_CRED" ]; then
-   echo "Skipping OneDrive ingest test because the MS_CLIENT_ID or MS_CLIENT_CRED env var is not set."
+if [ -z "$MS_CLIENT_ID" ] || [ -z "$MS_CLIENT_CRED" ] || [ -z "$MS_USER_PNAME" ]; then
```
ryannikolaidis, line 11: good catch here.

unstructured/ingest/cli/cmds/azure.py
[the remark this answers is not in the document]
ryannikolaidis: yea, the problem makes sense. is there a re-use paradigm we could use so we don't have to redefine the entire option for some of these that are shared across a lot but not all? e.g. remote-url or recursive? a bit of a bummer to have to copy-paste and update across a bunch of files if we want to make a change in the future.

ryannikolaidis:
I know this is still in progress, so feel to ignore if it's already on your radar, but note you'll also probably need to bump the relevant examples [here](https://github.com/Unstructured-IO/unstructured/tree/main/examples/ingest) and _possibly_ docs: [ingest readme](https://github.com/Unstructured-IO/unstructured/blob/main/Ingest.md) and [published docs for connectors](https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/connectors.rst)

unstructured/ingest/cli/cmds/reddit.py
```diff
@@ -0,0 +1,95 @@
...
+        "params: {}".format(
+            {
+                "subreddit_name": subreddit_name,
+                "client_id": client_id,
+                "client_secret": client_secret,
+                "user_agent": user_agent,
+                "search_query": search_query,
+                "num_posts": num_posts,
+            },
+        ),
```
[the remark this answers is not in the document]
ryannikolaidis: yea, this ^ is a good point, probably shouldn't be logging secrets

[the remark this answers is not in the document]
ryannikolaidis: same for creds or even slightly sensitive across any of these connectors

unstructured/ingest/cli/cli.py
```diff
@@ -0,0 +1,252 @@
...
+import click
+from click import ClickException
+
+import unstructured.ingest.cli.cmds as cli_cmds
+from unstructured.ingest.logger import ingest_log_streaming_init, logger
+
+
+@click.group()
+@click.pass_context
+@click.option(
```
ryannikolaidis, line 12: for lack of a better alternative, should we sort options alphabetically?

ryannikolaidis: (mostly thinking from a readability POV when scanning)

ryannikolaidis:
curious: is there a reasonable way to define those standard/common options as options on each connector rather than options directly defined for the base group? 

i.e. so I can call:
`PYTHONPATH=. ./unstructured/ingest/main.py local --reprocess  --verbose  --file-glob "*.html"`
rather than:
`PYTHONPATH=. ./unstructured/ingest/main.py --reprocess  --verbose  local --file-glob "*.html"`

and only need to run `PYTHONPATH=. ./unstructured/ingest/main.py local` to get a complete list of all options available when making the call; rather than as a user knowing I need to first parse my standard options and then connector-specific options (and knowing the difference between the two)

unstructured/ingest/cli/cmds/azure.py
```diff
@@ -0,0 +1,100 @@
...
+    ingest_log_streaming_init(logging.DEBUG if context_dict["verbose"] else logging.INFO)
+
+    logger.debug(f"parent params: {context_dict}")
+    logger.debug(
+        "params: {}".format(
+            {
+                "remote_url": remote_url,
+                "recursive": recursive,
+                "account_name": account_name,
+                "account_key": account_key,
```
ryannikolaidis, line 64: probably want to scrub all of these for any sensitive or identifying info

unstructured/ingest/connector/onedrive.py
```diff
@@ -22,12 +22,12 @@
 
 @dataclass
 class SimpleOneDriveConfig(BaseConnectorConfig):
-    client_id: str
-    client_credential: str = field(repr=False)
+    client_id: Optional[str]
+    client_credential: Optional[str] = field(repr=False)
```
ryannikolaidis, line 26: I believe these are both required. took a second but I think the post_init message below this needs to be updated (note that they are AND'd) I believe the ValueError message should be "Please provide all of the following...". Also, we need to bump the flag names in that error message.

test_unstructured_ingest/test-ingest-slack.sh
```diff
@@ -8,21 +8,19 @@ OUTPUT_FOLDER_NAME=slack
 OUTPUT_DIR=$SCRIPT_DIR/structured-output/$OUTPUT_FOLDER_NAME
 DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
 
-if [ -z "$SLACK_TOKEN" ]; then
-   echo "Skipping Slack ingest test because the SLACK_TOKEN env var is not set."
-   exit 0
-fi
```
ryannikolaidis, line 14: we still need this, right?

unstructured/ingest/connector/reddit.py
```diff
@@ -21,10 +21,10 @@
 @dataclass
 class SimpleRedditConfig(BaseConnectorConfig):
     subreddit_name: str
-    client_id: str
-    client_secret: str
-    user_agent: str
-    search_query: str
+    client_id: Optional[str]
+    client_secret: Optional[str]
+    user_agent: Optional[str]
```
ryannikolaidis, line 26: these are required

test_unstructured_ingest/test-ingest-slack.sh
rbiseck3: I got it to work without a token. Not sure if the one we're pulling from is public?

unstructured/ingest/connector/slack.py
```diff
@@ -26,9 +26,9 @@ class SimpleSlackConfig(BaseConnectorConfig):
     """Connector config to process all messages by channel id's."""
 
     channels: List[str]
-    token: str
-    oldest: str
-    latest: str
+    token: Optional[str]
```
ryannikolaidis, line 29: I think token is required

unstructured/ingest/process.py
```diff
@@ -0,0 +1,110 @@
...
+    BaseConnector,
+    ProcessorConfigs,
+)
+from unstructured.ingest.logger import ingest_log_streaming_init, logger
+
+with suppress(RuntimeError):
+    mp.set_start_method("spawn")
+
+
+class Process:
```
ryannikolaidis, line 17: I suspect this was originally called MainProcess to not conflate with the individual pool processes. Any thoughts?

test_unstructured_ingest/test-ingest-slack.sh
rbiseck3: Oh it's because the slack token is hard coded into the params that are passed in

ryannikolaidis: yep yep

unstructured/ingest/process.py
rbiseck3: Renamed the class to `Processor` since it's responsible for kicking off how ever many processes are allocated. Hopefully that makes it a bit clearer.

test_unstructured_ingest/test-ingest-slack.sh
ryannikolaidis: double checking we on the same page, the token for the test comes from env var `SLACK_TOKEN` which should be required to run this.

ryannikolaidis: when I try to run the test without the SLACK_TOKEN variable set I get: 
```
2023-07-28 13:33:57,697 SpawnPoolWorker-1 ERROR    Error: The request to the Slack API failed. (url: https://www.slack.com/api/conversations.history)
The server responded with: {'ok': False, 'error': 'not_authed'}```

**Comment by ryannikolaidis** [comment]
this all looks great. one last issue I noticed: I'm seeing duplicate log lines. to easily repro, just add a logger debug line in the wikipedia connector init [here](https://github.com/Unstructured-IO/unstructured/blob/f5b1c4c5dcbfc1f0ad50ebba4946ca6e9d252175/unstructured/ingest/connector/wikipedia.py#L119). then run the test `./test_unstructured_ingest/test-ingest-wikipedia.sh`. note that the init log appears four times.

File: test_unstructured_ingest/test-ingest-slack.sh (hunk 8 above)
Comment (rbiseck3) [reply test_unstructured_ingest/test-ingest-slack.sh:14]: Looks like the latest script on `main` also has that requirement removed: [examples/ingest/slack/ingest.sh](https://github.com/Unstructured-IO/unstructured/blob/main/examples/ingest/slack/ingest.sh). Pretty sure whatever value was being pulled from `SLACK_TOKEN` now gets passed in explicitly via the `--token` parameter of the cli call: 
```python
        slack \
        --channels 12345678 \
        --token 12345678 \
        --download-dir slack-ingest-download \
        --structured-output-dir slack-ingest-output \
        --start-date 2023-04-01T01:00:00-08:00 \
        --end-date 2023-04-02
```

rbiseck3:
@ryannikolaidis, duplicate logs issue was fixed. Added a check to avoid multiple handlers being added to the same logger instance.

ryannikolaidis (approved):
lgtm, nice work! this was a huge lift. should make dev cleaner and the interface more usable. thanks for all the iterations as it came together.

PR #981, opened 2023-07-26
Update pip in makefile

shreyanid:
Update pip version to latest (23.2.1)


cragwolfe (changes requested):
Please update the Changelog / version to 0.8.15-dev1 , so we don't have 2 commits claiming to be 0.8.14

cragwolfe:
> Please update the Changelog / version to 0.8.15-dev1 , so we don't have 2 commits claiming to be 0.8.14

Another commit has since merged. it's still probably worth bumping the -dev version to the next one but not a big deal if a few commits happen to share the same -dev version.

cragwolfe (approved):

PR #985, opened 2023-07-27
fix: only download nltk packages if necessary

### LINKED ISSUE #610: bug/Shouldn't download nltk packages when we don't use related functions.
**Describe the bug**

unstructured will download nltk packages when `unstructured.nlp.tokenize` module is imported.

```
# module unstructured.nlp.tokenize
...
def _download_nltk_package_if_not_present(package_name: str, package_category: str):
    """If the required nlt package is not present, download it."""
    try:
        nltk.find(f"{package_category}/{package_name}")
    except LookupError:
        nltk.download(package_name)


NLTK_PACKAGES = [
    ("tokenizers", "punkt"),
    ("taggers", "averaged_perceptron_tagger"),
]

for package_category, package_name in NLTK_PACKAGES:
    _download_nltk_package_if_not_present(package_name, package_category)
...
```

**Expected behavior**
Trigger the download logic when we really need it.

MthwRobinson:
### Summary

Closes #610. Updates the `tokenize` module so that NLTK packages are only downloaded if they are going to be used rather than downloading them at the top level of the module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.8.5-dev1
+## 0.8.5-dev2
 
 ### Enhancements
 
@@ -8,6 +8,8 @@
 ### Features
 
 ### Fixes
+
+* NLTK now only gets downloaded if necessary.
 
 ## 0.8.4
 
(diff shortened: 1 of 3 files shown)


yuming-long (approved):
LGTM!

PR #988, opened 2023-07-28
chore: remove debug print

ryannikolaidis:


cragwolfe (approved):

cragwolfe (commented):
actually worth cutting a release? that can be a crazy amount of output.

ryannikolaidis:
> actually worth cutting a release? that can be a crazy amount of output.

yea, was trying to debug elsewhere and it was quite noisy. yep, will do!

Issue #544, opened 2023-05-04
feat/unstructured command line to print library information


cragwolfe:
**Is your feature request related to a problem? Please describe.**
When debugging issues, it can help to know the various libraries and dependencies that are installed, including.

* OS version
* python version
* unstructured, unstructured-inference (if installed) versions
* tesseract versions and tesseract lang packages installed
* detectron2 version, torch version
* PaddleOCR versions
* other OS packages: libmagic, libreoffice
* Perhaps entire pip freeze packages output

**Describe the solution you'd like**
A CLI tool in the unstructured lib that a user can type to get the above info to cut and paste when creating / discussing an issue.

**Describe alternatives you've considered**
Doc-only is another solution, but it's easiest to just run a single command rather than multiple. Having it as a python commands helps it be OS agnostic.

ryannikolaidis:
pytorch has a nice solution for this that could be some inspiration. In [their bug reporting flow](https://github.com/pytorch/pytorch/issues/new?assignees=&labels=&template=bug-report.yml)
they ask that you run [this script](https://raw.githubusercontent.com/pytorch/pytorch/main/torch/utils/collect_env.py) to produce the debug info.

PR #989, opened 2023-07-28
Python script to collect environment for debugging issues

ron-unstructured:
The script collect the following information:
- OS version
- Python version
- Unstructured and Unstructured-Inference (if installed) versions
- Tesseract versions and tesseract lang packages installed
- Detectron2 version, torch version
- PaddleOCR versions
- other OS packages: libmagic, libreoffice


ron-unstructured:
resolved issue #544

MthwRobinson:
@ron-unstructured - Could you run `make tidy` to resolve the linting errors?

scripts/collect_env.py
```diff
@@ -0,0 +1,281 @@
...
+    Check if a command exists in the system
+
+    Args:
+        command (str): The command to check
+
+    Returns:
+        bool: True if command exists, False otherwise
+    """
+    try:
+        subprocess.run([command], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
MthwRobinson, line 17: Can we look for the command in `PATH` instead of running the command? This seems like it would execute the command.

scripts/collect_env.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    Get the current operating system version
+
+    Returns:
+        str: The current operating system version
+    """
+    return platform.platform()
+
+
+def is_python_package_installed(package_name):
```
MthwRobinson, line 43: For this one you could use `dependency_exists` from `unstructured.utils`

https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/utils.py#L50

scripts/collect_env.py
MthwRobinson: Alternatively, you could make this `brew_command_exists` and have it do `brew help` since I only see it checking for `brew` in the script.

MthwRobinson (changes requested):
Looks good overall! Just a few minor comments

MthwRobinson:
Could you add documentation (probably in the GitHub issue template) on how to use this? This looks good otherwise.

PR #991, opened 2023-07-28
fix: correct nltk download arg order

MthwRobinson:
### Summary

Closes #990. Corrects the kwarg order for the NLTK download step.

### Testing

Run `rm -rf ~/nltk_data/tokenizers` and `rm -rf ~/nltk_data/taggers` (or wherever you have NLTK DATA stored) to trigger the download. Then run:

```python
from unstructured.nlp.tokenize import pos_tag, word_tokenize, sent_tokenize

sentence = "I am a big brown bear. What are you?"
sent_tokenize(sentence)
word_tokenize(sentence)
pos_tag(sentence)
```

PR #992, opened 2023-07-28
Chore: put back function `split_by_paragraph`

yuming-long:
Put the function back since the notebook in API repo is using it, and is helpful in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 ### Enhancements
 
+* Put back useful function `split_by_paragraph`
+
 ### Features
 
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


MthwRobinson (approved):
Welcome back, `split_by_paragraph`! LGTM as long as tests pass.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 0.8.7
```
MthwRobinson, line 1: Only one request, could you wait to cut the tag til #991 is in? Want to get that one in as soon as possible since it's caused breakages for some users

yuming-long: of course!

PR #996, opened 2023-07-29
Box Connector

potter-potter:
Adds Box connector
Was tricky setting up the folders / access
Were issues with multiprocessing and the session (wouldn't pickle locks). So there is a temporary work around.
Passing in a json string for the cred that gets turned into a JWT.


ryannikolaidis:
I think totally works to have the `box-app-cred` as an argument, but should we also support a user pointing to their credential as a file path? and then can test similar to what we did with google drive over [here](https://github.com/Unstructured-IO/unstructured/blob/adbdcfceb39fc1758fe242e8f3d39fb95b1fa7ae/test_unstructured_ingest/test-ingest-google-drive.sh#L17-L31)? thinking being that passing in that json file as an inline argument could be a bit clumsy and it may be easier for folks to just store it on disk.

potter-potter:
> I think totally works to have the `box-app-cred` as an argument, but should we also support a user pointing to their credential as a file path? and then can test similar to what we did with google drive over [here](https://github.com/Unstructured-IO/unstructured/blob/adbdcfceb39fc1758fe242e8f3d39fb95b1fa7ae/test_unstructured_ingest/test-ingest-google-drive.sh#L17-L31)? thinking being that passing in that json file as an inline argument could be a bit clumsy and it may be easier for folks to just store it on disk.

Agreed. I always think huge text strings in env is clumsy. Should we use the same env var for text and file? Or only use file?. I couldn’t really remember what we did with gcs since I couldn’t see the secret

potter-potter:
Made it read from a file, not json.
Lets just keep it the same as GCS. It will be a path to the json file.

ryannikolaidis:
A few quick tweaks:
- refactoring to match the connector refactor that just merged
- add install steps to docker and make
- add require extras wrapper and moving import into method
- add ability to specify path or value for config in test (easier for folks running locally)

ryannikolaidis (approved):
lgtm! nice work!

PR #997, opened 2023-07-29
chore: rename Element's "date" field to "last_modified"

cragwolfe:
Change the Element's `date` field name to the more specific `last_modified` so there is less room for confusion of what that field represents.


MthwRobinson (approved):
Looks good to me once the merge conflicts are sorted.

Issue #1000, opened 2023-07-31
bug/file_filename from partition is not ending up in metadata


qued:
**Describe the bug**
The docstring for `unstructured.partition.auto.partition` implies that the `file_filename` parameter's purpose is to provide a way to store the filename in metadata for cases when the filename is not otherwise provided (e.g. when the file is passed as a file object). However this doesn't actually seem to happen, and `file_filename` is only used for detecting the filetype.

**To Reproduce**
```python
from unstructured.partition.auto import partition
with open("example-docs/book-war-and-peace-1p.txt") as f:
  elements = partition(file=f, file_filename="book-war-and-peace-1p.txt")
print(elements[0].metadata)
```
(`filename` is `None` in the metadata.)

**Expected behavior**
I would expect `file_filename` to be used to populate the `filename` property of the metadata when the parameter is not `None`.

six5532one:
Related: https://github.com/Unstructured-IO/unstructured/issues/1007

qued:
Closing as address by #1002 . Thanks @ChrisPappalardo !

PR #1002, opened 2023-07-31
fixed filename metadata bug when using file and file_filename

ChrisPappalardo:
addresses #1000


six5532one:
Thanks @ChrisPappalardo ! Would you mind adding a test, perhaps using a test case similar to the code snippet in https://github.com/Unstructured-IO/unstructured/issues/1000#issue-1829359752?

ChrisPappalardo:
Latest will fix failing tests and changelog check, but not sure what to do about:  https://github.com/Unstructured-IO/unstructured/actions/runs/5727711571/job/15520848419?pr=1002

qued:
@ChrisPappalardo The remaining test failures are non-blocking as far as I'm concerned. Could you make one last change and move the description of the change in the changelog up to the latest revision, and increment the dev revision number (i.e. from `dev0` to `dev1` as things stand now)?

ChrisPappalardo:
> @ChrisPappalardo The remaining test failures are non-blocking as far as I'm concerned. Could you make one last change and move the description of the change in the changelog up to the latest revision, and increment the dev revision number (i.e. from `dev0` to `dev1` as things stand now)?

Done

qued (approved):
LGTM! Thanks for the contribution @ChrisPappalardo !

qued:
Oops, I neglected to say the version also needs to be changed in __version__.py. Could you update that as well @ChrisPappalardo , and then we should be good to merge.

ChrisPappalardo:
> Oops, I neglected to say the version also needs to be changed in **version**.py. Could you update that as well @ChrisPappalardo , and then we should be good to merge.

Done!

Issue #726, opened 2023-06-12
`convert_and_partition_html` should return `Title`, etc instead of `HTMLTitle`


MthwRobinson:
Currently `convert_and_partition_html` returns `HTMLTitle` as an artifact of converting the document to HTML prior to processing. We should return `Title` here instead since the source document is not HTML

PR #1008, opened 2023-08-01
Kravetsmic/726 convert and partition html should return `Title` instead of `HTMLTitle`

kravetsmic:
### Summary
Change the behavior of work `convert_and_artition_html`.
Not HTML files (`.rst`, `.rtf`,`.org`,`.epub`) mark title as `Title` instead of `HTMLTitle`

Bump to 0.8.8

---------
Close issue #726


MthwRobinson:
@kravetsmic - Since you have a few other PRs already open, please get those closed out before continuing work on this one or opening any new PRs.

MthwRobinson (changes requested):
Instead of having the `is_html` conditional logic to pass on, what about adding a method to `HTMLText`, etc that returns the appropriate `Text` element (and same for the other element types)? Then we can just call that at the end of `convert_and_partition_html` instead of passing that kwarg all the way through. Something like this (may not be 100% right)

```python
class HTMLText(TagsMixin, Text):
    """Text with tag information."""

   def get_parent_element(self):
        return Text(text=self.text, metadata=self.metadata)
```

kravetsmic:
> Instead of having the `is_html` conditional logic to pass on, what about adding a method to `HTMLText`, etc that returns the appropriate `Text` element (and same for the other element types)? Then we can just call that at the end of `convert_and_partition_html` instead of passing that kwarg all the way through. Something like this (may not be 100% right)
> 
> ```python
> class HTMLText(TagsMixin, Text):
>     """Text with tag information."""
> 
>    def get_parent_element(self):
>         return Text(text=self.text, metadata=self.metadata)
> ```
 
Is it right that `convert_and_partition_html` must always return base elements' types instead of HTML ones? (Text instead of HTMLText, Title instead of HTMLTitle, etc.) Also, you propose to add method to all HTML elements which returns instance of parent class. After that we will call this method for every element returned from `partition_html`. Did I understand you correctly?

kravetsmic:
Will be re-opened after brought up to date

PR #1020, opened 2023-08-01
doc: update API doc to sych with new parameter in prod API

yuming-long:
as of `0.0.33` prod API, we have a new partition parameter `include_page_break`, add this in doc to reflect changes


MthwRobinson (approved):
LGTM - tested the new curl command in the docs locally and it worked.

PR #1024, opened 2023-08-02
chore: custom layout order example notebook

LaverdeS:
Example notebook to visualise and reorder the elements in a `DocumentLayout`. This solves the problem for Federal Regulation documents but is far from being a general solution. It works as an example of customisation for order and visualisation of the elements in a page `layout.pages[i].elements` of a `DocumentLayout` object in `unstructured_inference.inference.layout` and as a base for further work into a general solution for reordering elements in multiple-column textual-images.


benjats07:
Seems very good, will open a PR implementing this strategy.

benjats07 (approved):

PR #1025, opened 2023-08-02
chore: Use VCR to keep cached partition

jasonbot:
Not sure if this is the right approach, but this allows the `test_json` tests to no longer use the vanilla `partition` function in a somewhat roundabout way: by recording the `partition` calls as VCR `partition_by_api` cassettes instead, the tests can run in isolation without needing `partition` and its dependencies installed. The caveat is that the cassettes will need to be periodically re-recorded.


MthwRobinson:
Thanks @jasonbot ! Yeah this approach works for me. Intent was to all `test_json.py` to run without the extra dependencies.

test_unstructured/partition/test_json.py
```diff
@@ -32,10 +35,23 @@
 is_in_docker = os.path.exists("/.dockerenv")
 
 
+# WARNING! You will need to set up API key, remove the cassettes, and re-run if
```
MthwRobinson, line 38: Could you add the path to the cassettes so people know where to find them?

MthwRobinson: Also, would it be helpful to have a small script that updates the cassettes and also scrubs out UNSTRUCTURED-API-KEY to make it less likely someone accidentally commits that?

MthwRobinson (commented):
Pattern looks good to me - thanks! Just one inline suggestion.

cragwolfe (changes requested):
I'd prefer to avoid having to regenerate VCR fixtures.
There must be a place we can stick tests that have all the deps installed that we can enforce the critical invariant of:

(in pseudo code)
partition(foo.pdf) = partition(save_as_json(partition(foo.pdf)))

and likely same for additional file types

PR #1029, opened 2023-08-02
chore: add Dropbox secrets to CI environments

ryannikolaidis:
* add Dropbox environment variable keys to ingest test and create PR Github workflows
* bonus: alphabetically sorts environment variable keys for both jobs
* bonus: fixes box environment variable name and removes last_modified from result


qued (approved):
LGTM once the new dropbox ingest test diffs are checked in!

PR #1032, opened 2023-08-02
fix: simplify `min_partition` logic for `partition_text`

MthwRobinson:
### Summary

Simplifies `min_partition` logic in `partition_text`.  Fixes an issue where segments exceeding the min partition were observed in an example file.

### Testing

The following should show all segments having a length greater than 1000, except for the last one which doesn't have a subsequent segment to merge with.

```python

elements = partition_text(filename=filename, min_partition=1000, max_partition=1500)
filename = "example-docs/norwich-city.txt"
[len(el.text) for el in elements]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.9.1-dev0
+## 0.9.1-dev1
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


qued (approved):
LGTM!

PR #1033, opened 2023-08-03
roman/Add notion connector

rbiseck3:
### Description
Connector added that interacts with the Notion api to pull in contents of a page or database. Code added to add structure to the json response being returned to pull out only the text of the components on the page/database. 

Some things to consider:
* For any element that is structured as a table in the UI, this text now gets pulled in as newline delimited content row-by-row. This can be manipulated though if there's a better format when converting the content to a simple text file. 
* Child pages/databases are only known after the content is pulled in preventing the connector from lazily knowing about them. This breaks the current workflow since it assumes each ingest doc has a single file associated with is. May require further refactoring to allow the Notion connector to pull in content recursively. Currently, ignores any child pages or databases found in parent page. 

### TODO
Need to stage some content in Notion for the CI to use when running the ingest test.


ryannikolaidis:
@rbiseck3 I'm getting `Error: No such command 'notion'.` when trying to run `./unstructured/ingest/main.py notion --api-key <SECRET>

It looks like I'm on latest. Any thoughts? Do you see the same?

unstructured/ingest/processor.py
```diff
@@ -77,15 +77,19 @@ def run(self):
...
-            initializer=ingest_log_streaming_init,
-            initargs=(logging.DEBUG if self.verbose else logging.INFO,),
-        ) as pool:
-            pool.map(self.doc_processor_fn, docs)
-
-        self.cleanup()
+        logger.info(f"Processing {len(docs)} docs")
+        for doc in docs:
+            self.doc_processor_fn(doc)
+        return
```
ryannikolaidis, line 83: was this for testing serially? assuming we need to remove this right?

unstructured/ingest/logger.py
```diff
@@ -14,3 +14,14 @@ def ingest_log_streaming_init(level: int) -> None:
         logger.addHandler(handler)
 
     logger.setLevel(level)
+
+
+def make_default_logger(level: int) -> logging.Logger:
```
ryannikolaidis, line 19: what is the idea here? why is notion the only one using this?

rbiseck3: I found this to be the better pattern but to limit how much is changes in the code base only the new notion code is using this.

ryannikolaidis (approved):
lgtm! nice work, this one required some really deep digging, though hopeful this has some patterns to leverage if we come across data sources with similar structures (I doubt this will be the last)

Note: given the surface area, merging this with test and updates to soon follow. Manually tested with `PYTHONPATH=. python ./unstructured/ingest/main.py notion --api-key <API_KEY> --page-ids <PAGE_ID>`

PR #1036, opened 2023-08-03
fix: remove default encoding for ingest

ryannikolaidis:
test-ingest-local was failing to partition utf-16 encoded files because by default we explicitly assign utf-8. Probably rather than assuming utf-8, we're better off letting the auto encoding detection determine that and otherwise allow users to opt-in with specification.

* sets ingest default for encoding to None (to automatically detect encoding)
* updates number of expected files for local ingest (since it now properly handles non-utf-8 files)
* updates error message on mismatched number of files in ingest tests for more details


test_unstructured_ingest/check-num-files-output.sh
```diff
@@ -12,9 +12,10 @@ EXPECTED_NUM_FILES=$1
 OUTPUT_FOLDER_NAME=$2
 SCRIPT_DIR=$(dirname "$(realpath "$0")")
 OUTPUT_DIR=$SCRIPT_DIR/structured-output/$OUTPUT_FOLDER_NAME
+num_files_created="$(find "$OUTPUT_DIR" -type f -exec printf '.' \; | wc -c | xargs)"
 
-if [[ "$(find "$OUTPUT_DIR" -type f -exec printf '.' \; | wc -c | xargs)" != "$EXPECTED_NUM_FILES" ]]; then
+if [[ num_files_created != "$EXPECTED_NUM_FILES" ]]; then
```
qued, line 17: 👍 on the readability upgrade.

qued (approved):
LGTM!

PR #1039, opened 2023-08-03
Update `make check-version` script to fail if release version is unchanged

shreyanid:
## Summary
We only want one commit to correspond to a release, so this change checks if a non-dev version is identical to a previous commit release version.

## Test
Manual version changes to test the cases below. Print statements included temporarily for visibility.

### __version__ and CHANGELOG synced and dev version
no errors (Unaffected by current PR)
checkversion3

### __version__ and CHANGELOG out of sync and dev version
Error: out of sync (Unaffected by current PR)
checkversion2

### __version__ and CHANGELOG synced and same as current release version
Error: already a commit for this release version
Screenshot 2023-08-03 at 3 32 14 PM

### __version__ and CHANGELOG out of sync and __version__ same as current release version
Error: already a commit for this release version
Screenshot 2023-08-03 at 3 32 38 PM

### __version__ and CHANGELOG out of sync and CHANGELOG version same as current release version
Error: out of sync
checkversion1

### __version__ and CHANGELOG synced and ahead of current release version
no errors
Screenshot 2023-08-03 at 3 33 58 PM

### DIFF
--- a/scripts/version-sync.sh
+++ b/scripts/version-sync.sh
@@ -108,17 +108,19 @@ for i in "${!FILES_TO_CHECK[@]}"; do
     RE_SEMVER=${RE_SEMVERS[$i]}
     UPDATED_VERSION=${UPDATED_VERSIONS[$i]}
     FILE_VERSION=$(grep -o -m 1 -E "${RE_SEMVER}" "$FILE_TO_CHANGE")
-    CURRENT_RELEASE=$(curl -sL https://api.github.com/repos/Unstructured-IO/unstructured/releases/latest | jq -r ".tag_name")
+    MAIN_VERSION=$(git show main:unstructured/__version__.py | grep -o -m 1 -E "${RE_SEMVER_FULL}")
+    MAIN_IS_RELEASE=false
+    [[ $MAIN_VERSION != *"-dev"* ]] && MAIN_IS_RELEASE=true
 
     if [ -z "$FILE_VERSION" ];
     then
         # No match to semver regex in VERSIONFILE, so nothing to replace
         printf "Error: No semver version found in file %s.\n" "$FILE_TO_CHANGE"
         exit 1
-    elif [ "$FILE_VERSION" == "$CURRENT_RELEASE" ]
+    elif [[ "$MAIN_IS_RELEASE" == true && "$FILE_VERSION" == "$MAIN_VERSION" ]];
     then 
-        # Only one commit should be associated with a particular non-dev release
-        printf "Error: there is already a commit associated with release version %s.\n" "$CURRENT_RELEASE"
+        # Only one commit should be associated with a particular non-dev version
+        printf "Error: there is already a commit associated with version %s.\n" "$MAIN_VERSION"
         exit 1
     else
         # Replace semver in VERSIONFILE with semver obtained from SOURCE_FILE
--- a/scripts/version-sync.sh
+++ b/scripts/version-sync.sh
@@ -116,15 +116,10 @@ for i in "${!FILES_TO_CHECK[@]}"; do
         printf "Error: No semver version found in file %s.\n" "$FILE_TO_CHANGE"
         exit 1
     elif [ $FILE_VERSION == $CURRENT_RELEASE ]
-    then
-        # report error
-        # only one commit should be associated with a particular release
-        # echo "sed version must be >= ${REQUIRED_VERSION}" && exit 1
-        printf "Error: there is already a commit associated with release version ${CURRENT_RELEASE}"
+    then 
+        # Only one commit should be associated with a particular non-dev release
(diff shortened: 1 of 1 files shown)


scripts/version-sync.sh
```diff
@@ -108,11 +108,18 @@ for i in "${!FILES_TO_CHECK[@]}"; do
     RE_SEMVER=${RE_SEMVERS[$i]}
     UPDATED_VERSION=${UPDATED_VERSIONS[$i]}
     FILE_VERSION=$(grep -o -m 1 -E "${RE_SEMVER}" "$FILE_TO_CHANGE")
+    CURRENT_RELEASE=$(curl -sL https://api.github.com/repos/Unstructured-IO/unstructured/releases/latest | jq -r ".tag_name")
```
cragwolfe, line 111: rather than using the github API, please use the git command line.
e.g., before the loop begins, you can define:

```
MAIN_VERSION=$(git show main:unstructured/__version__.py | grep -o -m 1 -E "${RE_SEMVER_FULL}")
```

cragwolfe: actually, also need to check if that version is not a dev version but a real release. could do that with another env var:
```
MAIN_VERSION=$(git show main:unstructured/__version__.py | grep -o -m 1 -E "${RE_SEMVER_FULL}")
MAIN_IS_RELEASE=false
[[ $MAIN_VERSION != *"-dev"* ]] && MAIN_IS_RELEASE=true
```

shreyanid: thanks for the feedback - didn't realize I could use the command line here :) 

By using the github api call, I knew I would be getting the latest release, not a dev version. In your suggestion, by checking the __version__.py file, we then have to add the check for it being a dev version, but I thought the point was to make sure that the proposed release version is not the same as the latest release, whether or not there have been subsequent dev versions worked on.

For example, if the latest release is 0.8.8, and there has since been a 0.8.9-dev1 (let's say that's what's currently in __version__.py), and the proposed new release version in a new PR is 0.8.8 (same as the current release, which is the case we want to catch), if we only look at the version in __version__.py file (in this example 0.8.9-dev1), we would ignore the whole check against the current release and not realize that the PR is proposing the same release version as the current. The only other way this wouldn't be the case is if there is some sort of check for chronological versions (i.e., 0.8.8 cannot come after 0.8.9-dev1), but I don't believe that is currently in place.

I'm not saying we have to use the github api, cli is probably fine, but I do think we need the latest actual release and not just the version in __version__.py since we want to be comparing actual releases and not involve dev versions. The api was nice in that we could request the latest release tag and know that it won't be a dev version. How else might we find this information on main without the API?

cragwolfe: > For example, if the latest release is 0.8.8, and there has since been a 0.8.9-dev1 (let's say that's what's currently in version.py), and the proposed new release version in a new PR is 0.8.8 (same as the current release, which is the case we want to catch), 

Actually, i had thought. the primary case we want to catch is two `0.8.8`'s being committed in a row. I.e., it is much more likely to have an accidental follow on commit to a release rather than going backwards in CHANGELOG. 

More generally, there could (should) also be a separate semver check to ensure the CHANGELOG version >= current version. (super simple to do with say https://pypi.org/project/semver/ , not super simple in bash). I.e., one also shouldn't go from `-dev3` to `-dev2`.

The other problem of relying on the github API is human's need to generate releases, and once in awhile they forget to 😐 .

shreyanid: That makes more sense, thanks for the clarification! 

So it seems like to address the primary case:
> an accidental follow on commit to a release

the changes you proposed using the git cli to get the version from `__version__.py` in main are best for now, and then I can make a new ticket for the general case of checking for an increment in version (whether dev or numerical)

scripts/version-sync.sh
```diff
@@ -108,11 +108,20 @@ for i in "${!FILES_TO_CHECK[@]}"; do
...
+
     if [ -z "$FILE_VERSION" ];
     then
         # No match to semver regex in VERSIONFILE, so nothing to replace
         printf "Error: No semver version found in file %s.\n" "$FILE_TO_CHANGE"
         exit 1
+    elif [[ "$MAIN_IS_RELEASE" == true && "$FILE_VERSION" == "$MAIN_VERSION" ]];
+    then 
+        # Only one commit should be associated with a particular non-dev version
+        printf "Error: there is already a commit associated with version %s.\n" "$MAIN_VERSION"
```
cragwolfe, line 123:
```suggestion
        printf "Error: there is already a commit associated with version %s in the main branch.\n" "$MAIN_VERSION"
```

cragwolfe (approved):
Tested by added a fake commit to my local main branch, then doing the same in the local version of this branch (which can then be undone with `git reset --hard HEAD~1`).

PR #1041, opened 2023-08-04
Added two-column ordering from unstructured-inference

benjats07:
This PR adds the `order_two_column_document` from unstructured-inference to PDF processing.
It is expected that the version supporting the new function will be 0.5.10


cragwolfe:
Closing since it is stale.
@benjats07 , feel free to re-open if it seems this ordering strategy ought to be supported in addition to xy-cut introudced in https://github.com/Unstructured-IO/unstructured/pull/1161. If so should, it should be indicated by the [sort_mode kwarg](https://github.com/Unstructured-IO/unstructured/blob/2777313/unstructured/partition/pdf.py#L378) rather than the new`two_columns` kwarg introduced in this PR.

PR #1045, opened 2023-08-04
Dynamically import modules for ingest cli

rbiseck3:
### Description
To ease developer experience with adding new connectors and therefore new cli subcommands, have them dynamically added in the `__init__.py` file. This does remove IDE support since it no longer has concrete information on the dynamically added modules/methods and mypy needed to be ignored in all places these are imported since it can't check typing information either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.9.1-dev6
+## 0.9.1-dev7
 
 ### Enhancements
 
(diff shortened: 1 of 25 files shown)


rbiseck3:
As part of this exploratory effort, it was deemed that the lack of IDE support made this not a suitable change. Looking into other ways to aid a user in the process of adding connectors and cli integration.

PR #1056, opened 2023-08-07
Chore[ingest]: adding parameter --partition-pdf-infer-table-structure

yuming-long:
### Summary

* adding --partition-pdf-infer-table-structure to unstructured-ingest

### Test

* check there are field `"text_as_html"` for extracted tables in output json file

### Note

all partition parameters are not passing into request correctly when use together with `partition-by-api`, so currently we have to stick with local partition (will be fixed soon).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.9.1-dev10
+## 0.9.1-dev11
 
 ### Enhancements
 
+* Adds --partition-pdf-infer-table-structure to unstructured-ingest.
 * Enable `partition_html` to skip headers and footers with the `skip_headers_and_footers` flag.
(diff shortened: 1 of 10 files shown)


jasonbot (commented):
Does this need to pass the `--partition-pdf-infer-table-structure` flag's value along to `partition_via_api` too? Did I miss the part in the code that does it?

yuming-long:
> Does this need to pass the --partition-pdf-infer-table-structure flag's value along to partition_via_api too? Did I miss the part in the code that does it?

yep we want to pass all partition kwargs to partition_via_api, but sadly its not working, and its using the local partition from /partition/auto.py
added a issue to fix the partition_via_api: https://github.com/Unstructured-IO/unstructured/issues/1064

jasonbot (approved):

PR #1058, opened 2023-08-07
feat: per-process ingest connections

ryannikolaidis:
Currently on ingest, connections to data sources are happening once per document (per IngestDoc). This will mean that likely with a lot of documents on a connector hitting an external service we’ll likely see new connections getting refused after N connections created (depending on the service limits). Instead, where possible (supported by the data source api), we should only create new connections per-process.

This defines a pattern for sharing a connection across all docs per process. Not every connector has the concept of a connection/session that can be kept open and shared across documents, so the session handle flow is opt-in by adding a ConnectorSessionHandleMixin with a method for creating the shared resources. Each connector will have bespoke data that needs to be shared, so the pattern allows the connector to define a SessionHandle subclass to define the data. Because this data in many cases will not be picklable it makes explicitly passing the shared data a bit tricky. This pattern uses an in memory variable to store and access the session handle.

Unit tests were added for unit specific functionality. The session handle flow was implemented for the Google Drive connector to validate end to end.


test_unstructured_ingest/unit/doc_processor/test_generalized.py
```diff
@@ -0,0 +1,44 @@
+# from dataclasses import dataclass
```
ryannikolaidis, line 1: debugging bizarre test failure..

unstructured/ingest/interfaces.py
```diff
@@ -18,6 +18,12 @@
 from unstructured.staging.base import convert_to_dict
 
 
+@dataclass
+class BaseSessionHandle(ABC):
+    """Abstract definition on which to define resources that are shared across each process/thread.
```
cragwolfe, line 23:
```suggestion
    """Abstract definition on which to define resources that local to an individual process.
```

unstructured/ingest/processor.py
```diff
@@ -74,15 +87,28 @@ def run(self):
...
                 return
 
+        # get a create_session_handle function if the connector supports it
+        create_session_handle_fn = (
+            partial(
+                cast(ConnectorSessionHandleMixin, self.doc_connector).create_session_handle,
+                cast(BaseConnector, self.doc_connector).config,
+            )
+            if isinstance(self.doc_connector, ConnectorSessionHandleMixin)
+            else None
```
cragwolfe, line 97: i don't think this handle should actually exist in the parent process. but i think i see the issue that the init process needs connection info, at the very least, following this approach.

i'm starting to think the cleanest way to do this is for the subprocess itself to create the SessionHandle lazily for the first IngestDoc it processes, since it will have the connector config at that time.

cragwolfe: or, it could pass the connector config of the first IngestDoc in `initargs` for the sub process's `initializer`, and then connector can choose to create its own SessionHandle if applicable.

unstructured/ingest/connector/google_drive.py
```diff
@@ -103,7 +114,9 @@ def get_file(self):
         from googleapiclient.errors import HttpError
         from googleapiclient.http import MediaIoBaseDownload
 
-        self.config.service = create_service_account_object(self.config.service_account_key)
+        if self.session_handle is None:
+            raise ValueError("Google Drive session handle was not set.")
+        self.session_handle = cast(GoogleDriveSessionHandle, self.session_handle)
```
rbiseck3, line 119: Since this is already set above: 
```python
session_handle: Optional[GoogleDriveSessionHandle] = None
```
I don't think the cast is necessary.

ryannikolaidis: iirce linting gets mad if you don't explicitly cast since it was optional

ryannikolaidis:
> I really wish there were a simpler implementation for this, it took me several readings to figure out what all was going on.
> 
> Just for my own edification: `create_session_handle` is correctly keying on config and if we connect to more than one Google Drive connection in the same process, they won't step on each other's toes, yeah?

yea, I agree. I think @cragwolfe's note will simplify this a bit, and have some additional thoughts that might help.

re: `more than one Google Drive connection in the same process`.
is that something we're trying to account for? this was actually undoing that since we're currently creating these per-document; instead lets re-use the same one per process. what is the use case of more than one connection per process?

unstructured/ingest/processor.py
ryannikolaidis: coming back to this fresh, I think I was over-engineering in case we ever needed to handle subprocess work async, but starting to feel like it's safe to just assume this is always serial within that subprocess. I like the idea of just lazily creating it, will head with that.

also, I'm now thinking the config should just own the definition for how to create the session handle. this would also be cleaner if we move that logic inside the subprocess since we don't pass the connector itself through (I don't believe).

cragwolfe (approved):
LGTM!
looking around the corner, we'll need a to serialize / de-serialize IngestDoc's for other use cases ("over the wire"), but i don't think this PR further complicates that issue.

PR #1072, opened 2023-08-09
Roman/notion tests

rbiseck3:
### Description
* Add ingest test for Notion docs
* Update default cache dir for connectors to include connector name. Makes debugging the cached content easier.


cragwolfe:
[Now realizing this is in draft mode, but comment still applies]

These are 2 distinct changes that ought to be different PR's.

On the the CI change is very significant and in of itself going from 4 jobs to 72 jobs (and counting with additional connectors). TBH, not sure where the actual limits of GH are, but have to image this is putting it to the test with many PR's. Also, it is burning a bunch of CI minutes cumulative on the cache setup for each test (68 * ~4mins ->  4 extra CI hours per run?).

IMO, this could be approached differently. If focused on debugging one connector, there could be an env var / config file set where each ingest scripts just exit if it is not the intended test.

CHANGELOG.md
```diff
@@ -81,15 +82,20 @@
...
 
 * Adds Sharepoint connector.
 
 ### Fixes
+<<<<<<< HEAD
 
 * Bump unstructured-inference==0.5.9:
   - ignores Tesseract errors where no text is extracted for tiles that indeed, have no text
+=======
+>>>>>>> 66c6cfc3 (update CHANGELOG)
```
ryannikolaidis, line 98: merge fix

unstructured/ingest/connector/discord.py
```diff
@@ -131,7 +131,7 @@ def __init__(
 
     def initialize(self):
         """Verify that can get metadata for an object, validates connections info."""
-        os.mkdir(self.standard_config.download_dir)
+        os.makedirs(self.standard_config.download_dir, exist_ok=True)
```
ryannikolaidis, line 134: funny caught this in another PR I'm working on, will have a fix for it. I think we actually don't want the folder creation to happen here at all and instead should happen on actually writing the file in the IngestDoc. reason being: if the IngestDoc were say ever processed somewhere (i.e. a different machine) other than where the the Connector was initialized, it could fail because it assumes that folder exists.

ryannikolaidis:
from PR description: `Update CI to run ingest tests as matrix, isolating each connector dependency and making logs easier to parse through` is no longer happening, right? probably good to scrub from the description for clarity?

unstructured/ingest/connector/discord.py
ryannikolaidis: does it? this looked specific to this discord connector. I haven't looked at all with a close eye, but it seemed that most were following [this convention](https://github.com/Unstructured-IO/unstructured/blob/8104fe274c935dbb824917206b2cacedb71fca99/unstructured/ingest/connector/confluence.py#L119) of making the directory when attempting to write the file (in IngestDoc). which is ideally just what we want to do here has well.

ryannikolaidis (approved):
lgtm!

Issue #1071, opened 2023-08-09
Optionally use UUIDs for the element ID


MthwRobinson:
Users who are interested in using the `element_id` as a primary key in a database need `element_id` to be unique. The goal of this issue is to allow users to optionally generate `element_id`s that are guaranteed to be unique instead of the deterministic SHA256 hash IDs that are used by default.

PR #1076, opened 2023-08-09
added UUID option for element_id arg in element constructor

ChrisPappalardo:
Addresses #1071

I followed how NoID() is used as a value for element_id by analogy.


MthwRobinson:
Thanks @ChrisPappalardo ! Could you add a changelog entry? Otherwise looks great

MthwRobinson (approved):
LGTM!

Issue #1057, opened 2023-08-07
bug/etree-parser-error


cacosandon:
**Describe the bug**
When reading a CSV that has emojis, it raises the error: `lxml.etree.ParserError: Document is empty`

**To Reproduce**
Run

```
>>> text = '<table><tbody><tr><td>👨\U+1F3FB🔧</td></tr></tbody></table>'
>>> text = lxml.html.document_fromstring(text)
```

**Expected behavior**
It should parse with the emojis.

**Screenshots**

<img width="1376" alt="Screenshot 2023-08-07 at 17 58 22" src="https://github.com/Unstructured-IO/unstructured/assets/30879716/969c8d4b-d2f3-4eca-9fec-a272c35957be">

v/s without emojis:

<img width="503" alt="Screenshot 2023-08-07 at 17 59 05" src="https://github.com/Unstructured-IO/unstructured/assets/30879716/2a9a9155-5620-499a-a5c0-facdae77cdf1">

**Additional context**
Definitely is because the characters of the emoji.

PR #1009, opened 2023-08-01
🐛 fixed emoji xlsx bug

hynky1999:
The xlsx loader is failing when the document contains emojis.
E. g. `!\n\nThanks 🙌🔥🤠`
[WIP] I will add test for this case
See https://github.com/Donohue/medium-to-jekyll/issues/3

PR #1077, opened 2023-08-09
Fix/1057 etree parser error

christinestraub:
### Summary
Applied the following approach to `partition_csv`.
- add functionality to check if a string contains any emoji characters
- add functionality to switch `html` text parser based on whether the `html` text contains emoji

**NOTE:** I expected this approach to apply not only to `partition_csv`, but also to other partitioning functions that use `lxml.html.document_fromstring` (`partition_xlsx`, `partition_tsv`). Inspired by PR [#1009](https://github.com/Unstructured-IO/unstructured/pull/1009).

### Testing
```
filename = "example-docs/stanley-cups-with-emoji.csv"
elements = partition_csv(filename)
```
---
Addresses [#1057](https://github.com/Unstructured-IO/unstructured/issues/1057) for `CSV`


MthwRobinson:
This approach looks good to me. Could we add this to `xlsx.py` and `tsv.py` too? They use a similar pattern.

MthwRobinson (approved):
LGTM!

PR #1085, opened 2023-08-10
feat: `unique_element_ids` kwarg for UUID elements

### LINKED ISSUE #1071: Optionally use UUIDs for the element ID

MthwRobinson:
### Summary

Closes #1071. Adds a `unique_element_ids` kwarg to partition functions. If `True`, the element IDs will be a UUID instead of a SHA-256 hash. This is helpful if you'd like to use the element IDs as the primary key in a database.

### Testing

```python

elements = partition_text(text="Here is some example text.", unique_element_ids=True)
# Should be a UUID like 256b25cc-9922-4673-8b49-d0afcf1a6271
elements[0].id

elements = partition_text(text="Here is some example text.")
# Should be 6b318ad9116de363fe8f131b7d103ff5
elements[0].id
```


test_unstructured/partition/test_text.py
```diff
@@ -464,3 +465,16 @@ def test_partition_text_from_text_with_custom_metadata_date(
     elements = partition_text(text=text, metadata_last_modified=expected_last_modification_date)
 
     assert elements[0].metadata.last_modified == expected_last_modification_date
+
+
+def test_partition_text_with_unique_ids():
+    elements = partition_text(text="hello there!")
+    assert elements[0].id == "c69509590d81db2f37f9d75480c8efed"
+    # Test that the element is JSON serializable. This shold run without an error
```
qued, line 473:
```suggestion
    # Test that the element is JSON serializable. This should run without an error
```

test_unstructured/partition/test_text.py
```diff
@@ -464,3 +465,16 @@ def test_partition_text_from_text_with_custom_metadata_date(
...
+    json.dumps(elements[0].to_dict())
+
+    elements = partition_text(text="hello there!", unique_element_ids=True)
+    assert len(elements[0].id) == 36
+    assert elements[0].id.count("-") == 4
+    # Test that the element is JSON serializable. This shold run without an error
```
qued, line 479:
```suggestion
    # Test that the element is JSON serializable. This should run without an error
```

qued (approved):
LGTM, a pair of small doc comments.

BalintKomjati:
Hi @MthwRobinson ! This makes element_id inconsistent with parent_id, at least in partition_docx which i am using right now. So parent_id is the very same as if unique_element_ids == False. Which makes parent_id rather useless as the id it points to does not exist. Is this the intended use? Would be pretty straightforward to me to use these unique ids in the parent_id field as well.

PR #1094, opened 2023-08-11
Fix/1057 etree parser error xlsx

christinestraub:
### Summary
Addresses [#1057](https://github.com/Unstructured-IO/unstructured/issues/1057) for `XLSX`. Related to [PR #1077](https://github.com/Unstructured-IO/unstructured/pull/1077).
### Testing
```
filename = "example-docs/emoji.xlsx"

elements = partition_xlsx(filename)
```


cragwolfe (approved):
LGTM!

cragwolfe (approved):

unstructured/partition/xlsx.py
```diff
@@ -11,17 +13,12 @@
...
+    contains_emoji,
     exactly_one,
     get_last_modified_date,
     get_last_modified_date_from_file,
     spooled_to_bytes_io_if_needed,
 )
-from unstructured.utils import dependency_exists
-
-if dependency_exists("bs4"):
-    from lxml.html.soupparser import fromstring as html_string_parser
```
cragwolfe, line 22: actually, i think the intent of this if statement was to allow "minimal" installs that don'thave bs4 installed to use the "lesser" html_string_parser.

to honor that convention, i think the better thing to do would be to strip emoji chars, then use document_fromstring if bs4 is not installed. wdyt?

cragwolfe (dismissed):
requesting changes in the interim per my earlier comment. (it's also mildly dismaying the test-ingest diff's look like a step backwards)

unstructured/partition/xlsx.py
```diff
@@ -65,6 +62,9 @@ def partition_xlsx(
     for sheet_name, table in sheets.items():
         page_number += 1
         html_text = table.to_html(index=False, header=False, na_rep="")
+        html_string_parser = (
+            soupparser_fromstring if contains_emoji(html_text) else document_fromstring
+        )
```
cragwolfe, line 67: given that the test results are worse with `document_fromstring` in test_unstructured/partition/test_auto.py right now, i think it is just better to always use soupparser_fromstring.

unstructured/partition/xlsx.py
christinestraub: Yes, it's reasonable to have `bs4` in a minimal unstructured install. That was already done in the previous PR for this issue - https://github.com/Unstructured-IO/unstructured/pull/1077/files.

unstructured/partition/xlsx.py
christinestraub: Sounds reasonable! After revisiting this, I also think it is better to always use `soupparser_fromstring`.

cragwolfe (approved):
LGTM!

PR #1097, opened 2023-08-11
Roman/compression support

rbiseck3:
### Description
Added support for extracting compressed zip and tar content. This applies to all Fsspec and Local connectors. It the file associated with the doc is detected as a compressed one, the original compressed file isn't processed and if the `--uncompress` flag is passed, the file is uncompressed and the resulting content is set as the children of the parent to be processed as well. 

Bug found using the LocalConnector to process the uncompressed data, if the same file name exists in the unarchived content but at different paths, this maps to the same output and only the final one written exists in the final result. Updated to create the output path similar to the Fsspec connector.


ryannikolaidis:
worth updating the test contents for one of the object store connectors? add a zip and tar and some files to unpack in each?

PR #1100, opened 2023-08-11
roman/notion module not found fix 

rbiseck3:
### Description
One of the added modules was missing an `__init__.py` file which made it undiscoverable in the path when running as a cli command via console script rather than the `PYTHONPATH=. python ...` approach.


ryannikolaidis (approved):
ahh, glad you found it!

cragwolfe:
worth a release, IMO

cragwolfe (approved):

PR #1102, opened 2023-08-11
build(deps): dependency cleanup

qued:
Attempt to clean up some pins that were prone to conflicts. `cryptography` was the main offender, and could have been cleaned up by moving the pin to `constraints.in`, but with this PR I'm floating the idea that the undocumented pins aren't necessary.

So if you have pinned something that's unpinned here for reasons other than to resolve a dependency conflict, please chime in, and let's get it pinned in the right place with a note.

#### Testing:
CI should pass, and no objections about the unpinning.


cragwolfe (approved):
LGTM.
let's make this a release commit since it is about time anyway.

PR #1107, opened 2023-08-13
chore: run make pip-compile on mac

cragwolfe:
so that `make install` works on mac (and will still work on linux). Hopefully,
this also resolves recent docker build issues.


christinestraub (approved):
LGTM!

PR #1109, opened 2023-08-13
Handle inline and lacking filename

mkly:
Handle Content-Disposition: inline and attachment without filename

* Add new email test example and test with Content-Disposition: inline.
* Move attachment_info above for loop so it is always defined
* Check if item is inline as well as attachment as these both lack an = character to split on
* Create filename if filename is not specified and write file.
* Update list_attachments with new filename

### DIFF
--- a/unstructured/partition/email.py
+++ b/unstructured/partition/email.py
@@ -5,7 +5,7 @@ import re
 import sys
 from email.message import Message
 from functools import partial
-from tempfile import SpooledTemporaryFile, TemporaryDirectory
+from tempfile import SpooledTemporaryFile, TemporaryDirectory, NamedTemporaryFile
 from typing import IO, Callable, Dict, List, Optional, Tuple, Union
 
 from unstructured.file_utils.encoding import (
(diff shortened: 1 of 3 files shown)


mkly:
Bit of an edge case email client [Claws Mail](https://www.claws-mail.org/). Noticed an issue with "Content-Disposition: inline" and included an example in the test case. Thanks for the very useful library and let me know if there's anything here I should adjust. :tropical_drink:

unstructured/partition/email.py
```diff
@@ -163,10 +163,10 @@ def extract_attachment_info(
             cdisp = part["content-disposition"].split(";")
             cdisp = [clean_extra_whitespace(item) for item in cdisp]
 
+            attachment_info = {}
             for item in cdisp:
-                attachment_info = {}
 
-                if item.lower() == "attachment":
+                if item.lower() == "attachment" or item.lower() == "inline":
```
cragwolfe, line 169:
```suggestion
                if item.lower() in ("attachment", "inline"):
```
tiny nit

cragwolfe (commented):
Awesome contribution! Please run `make tidy` for lint issues.

unstructured/partition/email.py
mkly: I considered that myself. Thanks for that feedback. Fixed.

mkly:
Apologies for missing the linting and changelog. I've gone ahead an updated this PR as well a #1110 Thank you for taking the time to review these and especially for the feedback.

cragwolfe:
For sure! Thanks for providing clean tests along with the fixes. Text book! :)

cragwolfe (approved):
LGTM!

cragwolfe:
@mkly , looks like this PR just needs some conflicts resolved. thanks again for the contribution!

PR #1110, opened 2023-08-13
Fix attachments with `=` in filename

mkly:
Fix attachments with `=` in filename

* Limit `split` to first match of `=` to prevent creating a list of more than two parts
* Add example email with attachment name and test for issue


cragwolfe (approved):
LGTM! 🎉

PR #1111, opened 2023-08-14
chore: bump pip version in published image

cragwolfe:
for consistency with the development environment, i.e. the Makefile.


shreyanid (commented):
Change LGTM. For testing, I did `make docker-build` then `make docker-test` and `make docker-smoke-test`. If this is to prod is there a different place I should download the image from and test?

shreyanid (approved):

PR #1119, opened 2023-08-14
fix pdf partition of list items being detected as titles in OCR only mode

### LINKED ISSUE #1010: List items are detected as titles in OCR only mode
In some cases list items are detected as titles when partitioning PDFs using "ocr_only" mode. The goal of this issue to take a look and see if we can correct this behavior.

### Testing
```

filename = "example-docs/list-item-example.pdf"
elements = partition_pdf(filename, strategy="ocr_only")
```

Coniferish:
### Summary
Closes #1010
adds `group_bullet_paragraph` func to handle grouping of bullet items that are split across multiple lines

### Testing
```
filename = "example-docs/list-item-example.pdf"
elements = partition_pdf(filename, strategy="ocr_only")
```

Note: This is not a complete fix for the testing document above as the second page of bullets is not interpreted as bullet points by `pytesseract`, so they cannot be interpreted as such downstream.


test_unstructured/cleaners/test_core.py
```diff
@@ -209,6 +209,30 @@ def test_group_broken_paragraphs_non_default_settings():
     )
 
 
+def test_group_broken_paragraphs_with_bullets():
+    text = """○The big red fox
```
benjats07, line 213: Do we really want to keep the ○ and · once we detected the element is inside a list?
If we want to clean the document probably deleting that characters could be better.

benjats07 (approved):
LGTM, just a comment about keeping characters · and ○

test_unstructured/cleaners/test_core.py
Coniferish: Yes, because they're needed later to define the elements as list elements. I added a comment in the function noting that. I can add one to the test as well if needed

Coniferish: My mistake! Looks like I removed the comment. I'll add it back noting as much

Coniferish:
@benjats07 Can you look at the test that's failing and advise what to do about it? I'm not getting that failure locally when I run `make test` and don't see the `rfc854.txt.json` locally that's failing.

benjats07:
Oh, two things:

1. Did you regenerate the expected outputs? Since you're changing the outputs the comparison could be failing if you didn't update those.
2. Did you test on x64 processor? The comparison should pass there not just in Mac. I had a similar problem testing on M1/2 because the output of the comparison is slightly different than the expected in CI.

Coniferish:
> Oh, two things:
> 
>     1. Did you regenerate the expected outputs? Since you're changing the outputs the comparison could be failing if you didn't update those.
> 
>     2. Did you test on x64 processor? The comparison should pass there not just in Mac. I had a similar problem testing on M1/2 because the output of the comparison is slightly different than the expected in CI.

1) No, haven't done that before I don't think. Is that just a matter of exporting `OVERWRITE_FIXTURES`?

2) I'm on a x64 windows (wsl). Is that fine or do I need to do something to double-check?

PR #1123, opened 2023-08-15
build(lint): don't check version in main against self

cragwolfe:
If on the main branch already, it does not make sense to check if the latest commit is the same non-dev version.

This fixes an annoyance where the CI Lint job would fail on release main commits, but besides that was not causing any other issues.

*Test Instructions*
```

git checkout main
# make sure latest unstructured/__version__.py is a non-dev release

# copy the scripts/version-sync.sh file from this PR locally
cd scripts; rm version-sync.sh; wget https://raw.githubusercontent.com/Unstructured-IO/unstructured/crag/version-sync-change/scripts/version-sync.sh; chmod ugo+x version-sync.sh; cd -
# commit to main
git add scripts/version-sync.sh
git commit -m "committing to local main for the sake of a test"
# this has no errors
make check-version 

# now git checkout local main into another branch and see the error
# since this branch has the same release version as main
git checkout -b tmp-test-check-version
make check-version 

# cleanup
git checkout main
git reset --hard HEAD~1
git branch -D  tmp-test-check-version
```


shreyanid (approved):
Changes look good and test worked!

PR #1132, opened 2023-08-16
Roman/delta table connector

rbiseck3:
### Description
Add delta table connector and test against a delta table generated via delta.io and uploaded to s3. Shows an example of how to use the connection options to leverage s3. 

I was able to get this to work with s3 if I pass in the access and secret keys as storage options. Even though the s3 bucket being used is public, would not work without those.


unstructured/ingest/connector/delta_table.py
```diff
@@ -0,0 +1,124 @@
...
+    def get_file(self):
+        self._create_full_tmp_dir_path()
+        self.config.get_logger().debug(f"Fetching {self} - PID: {os.getpid()}")
+
+        df: pandas.DataFrame = self.batch.to_pandas()
+        self.config.get_logger().info(f"writing {len(df)} rows to {self.filename}")
+        df.to_csv(self.filename)
+
+
+@requires_dependencies(["deltalake"])
```
ryannikolaidis, line 87:
```suggestion
@requires_dependencies(["deltalake"], extras="delta-table")
```

unstructured/ingest/connector/delta_table.py
```diff
@@ -0,0 +1,124 @@
...
+    def _output_filename(self):
+        """Create filename document id combined with a hash of the query to uniquely identify
+        the output file."""
+        return Path(self.standard_config.output_dir) / f"{self.id}.json"
+
+    def _create_full_tmp_dir_path(self):
+        self.filename.parent.mkdir(parents=True, exist_ok=True)
+        self._output_filename.parent.mkdir(parents=True, exist_ok=True)
+
+    @requires_dependencies(["elasticsearch"])
```
ryannikolaidis, line 76: don't need this, right?

unstructured/ingest/connector/delta_table.py
```diff
@@ -0,0 +1,124 @@
...
+@dataclass
+class SimpleDeltaTableConfig(BaseConnectorConfig):
+    verbose: bool
+    table_uri: Union[str, Path]
+    version: Optional[int] = None
+    storage_options: Optional[Dict[str, str]] = None
+    without_files: bool = False
+    columns: Optional[List[str]] = None
+    batch_size: Optional[int] = None
+    logger: Optional[logging.Logger] = None
```
ryannikolaidis, line 34: we should probably stick to only including primitives in the config in case this ever needs to be serialized

unstructured/ingest/connector/delta_table.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def get_logger(self) -> logging.Logger:
+        if self.logger:
+            return self.logger
+        return make_default_logger(logging.DEBUG if self.verbose else logging.INFO)
```
ryannikolaidis, line 39: maybe we should switch to just use the same default logger everywhere in ingest (rather than having this configurable in some)?

test_unstructured_ingest/test-ingest-delta-table.sh
```diff
@@ -0,0 +1,25 @@
...
+set -e
+
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+cd "$SCRIPT_DIR"/.. || exit 1
+OUTPUT_FOLDER_NAME=delta-table
+OUTPUT_DIR=$SCRIPT_DIR/structured-output/$OUTPUT_FOLDER_NAME
+DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
+
+if [ -z "$AWS_ACCESS_KEY_ID" ] && [ -z "$AWS_SECRET_ACCESS_KEY" ]; then
+   echo "aws credentials not found as env vars"
```
ryannikolaidis, line 12:
```suggestion
   echo "Skipping Delta Table ingest test because either AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY env var was not set."
```

ryannikolaidis: Also, it doesn't look like these are set in secrets nor passed through to the environment in ci.yml nor ingest-test-fixtures-update-pr.yml

unstructured/ingest/connector/delta_table.py
```diff
@@ -0,0 +1,124 @@
...
+from unstructured.ingest.interfaces import (
+    BaseConnector,
+    BaseConnectorConfig,
+    BaseIngestDoc,
+    ConnectorCleanupMixin,
+    IngestDocCleanupMixin,
+    StandardConnectorConfig,
+)
+from unstructured.ingest.logger import make_default_logger
+from unstructured.utils import requires_dependencies
```
ryannikolaidis, line 22: something is not quite right with order of imports. when I run the test prior to installing the extras I get an error specifically about importing pyarrow rather than the [require_dependencies warning message](https://github.com/Unstructured-IO/unstructured/blob/59962bc6d50b0ed6f8806f42d79767b5c4563e79/unstructured/utils.py#L40-L41) to install the appropriate extra., probably need to move imports and decorate with extra arg.

unstructured/ingest/connector/delta_table.py
rbiseck3: I originally wanted to add this to the `BaseConnectorConfig` class, but since we're supporting python3.8,  this doesn't have support for the `kw_only` param in `@dataclass(kw_only=True)`, so doing something like this: 
```python
@dataclass
class BaseConnectorConfig(ABC):
    """Abstract definition on which to define connector-specific attributes."""

    verbose: bool = False
    logger: Optional[logging.Logger] = None

    def get_logger(self) -> logging.Logger:
        if self.logger:
            return self.logger
        self.logger = make_default_logger(logging.DEBUG if self.verbose else logging.INFO)
        return self.logger
```
results in: `error: Attributes without a default cannot follow attributes with one`

ryannikolaidis: oh I worded poorly, I meant why do we need this configurable at all? can we just use the same logger everywhere?

ryannikolaidis:
not sure if they all make sense in the context of a db, but could you look at adding the various data source properties (mixed in here):

https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L111-L159

at a minimum we will definitely want record locator, but ideally all that make sense here.

unstructured/ingest/connector/delta_table.py
rbiseck3: When they used a shared logger, it ran into issues. I think because of the multiprocessing fork. It was dropping logs. This way each fork has a dedicated logger.

ryannikolaidis:
also reminder to bump [connector docs](https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/connectors.rst)

unstructured/ingest/connector/delta_table.py
ryannikolaidis: are all of the other connectors that don't have this logic currently dropping logs?

rbiseck3: I haven't thoroughly checked. They might be. I couldn't narrow exactly when logs were dropped but it seemed having this solution prevented it.

unstructured/ingest/connector/delta_table.py
ryannikolaidis: @rbiseck3 bump ^ probably need to move the top level imports inside relevant classes/functions, otherwise one just gets the error:

```
from deltalake import DeltaTable
ModuleNotFoundError: No module named 'deltalake'
Error: No module named 'deltalake'
```

PR #1134, opened 2023-08-16
bug: update partition_msg and partition_email so attachments also receive metadata_last_modified kwarg

### LINKED ISSUE #1027: Attached files in `partition_msg` and `partition_email` should use `metadata_last_modified` when passed
Per the test failure [here](https://github.com/Unstructured-IO/unstructured/actions/runs/5742956100/job/15566252574?pr=1009), it seems like when `metadata_last_modified` is passed to `partition_msg` (and potentially `partition_email`), the `last_modified` metadata attribute on the attachments does not reflect the date passed in through the kwarg. The goal of this issue is to verify that behavior and fix it if needed.

Coniferish:
### Summary
Closes #1027
The msg test in question was no longer failing after removing the quick-fix and comment explaining the issue. However, the test was not functioning as intended. Test was refactored to appropriately test `metadata_last_modified` of attachments. 
`partition_msg` was then updated to pass `metadata_last_modified` to `attachment_partitioner`.
The same was done for email partitioning.

### Testing
```

filename="example-docs/fake-email-attachment.msg"
elements = partition_msg(filename=filename, attachment_partitioner=partition_text, process_attachments=True, metadata_last_modified="0000-00-00")

# previously, these were different values because last_modified wasn't being updated in attachments
elements[1].metadata.last_modified 
elements[-1].text
elements[-1].metadata.last_modified

email_filename="example-docs/eml/fake-email-attachment.eml"
email_elements = partition_email(filename=email_filename, attachment_partitioner=partition_text, process_attachments=True, metadata_last_modified="0000-00-00")

email_elements[1].metadata.last_modified 
email_elements[-1].text
email_elements[-1].metadata.last_modified
```


unstructured/partition/email.py
```diff
@@ -416,7 +416,12 @@ def partition_email(
                     raise ValueError(
                         "Specify the attachment_partitioner kwarg to process attachments.",
                     )
-                attached_elements = attachment_partitioner(filename=attached_filename)
+                attached_elements = attachment_partitioner(
+                    filename=attached_filename,
+                    metadata_last_modified=metadata_last_modified,
+                    max_partition=max_partition,
+                    min_partition=min_partition,
```
cragwolfe, line 423: passing through min_partition/max_partition here is a good catch. should there be a test confirming this is working as expected?

Coniferish: Makes sense. I'll create one

test_unstructured/partition/test_msg.py
```diff
@@ -173,20 +169,46 @@ def test_partition_msg_can_process_attachments(
...
+        filename=filename,
+        attachment_partitioner=partition_text,
+        process_attachments=True,
+        min_partition=6,
+        max_partition=12,
+    )
+
+    assert elements[0].text.startswith("Hello!")
+    assert elements[-1].text == attachment_elements[-1].text
+    assert elements[-2].text == attachment_elements[-2].text
```
cragwolfe, line 209: it'd be nice to actually assert the elements length as well.

cragwolfe (approved):
minor comment, but LGTM!

test_unstructured/partition/test_email.py
```diff
@@ -489,11 +489,10 @@ def test_partition_email_can_process_min_max_with_attachments(
     assert elements[0].text.startswith("Hello!")
     assert elements[-1].text == attachment_elements[-1].text
     assert elements[-2].text == attachment_elements[-2].text
-    email_attachments = []
     for element in elements:
         if element.metadata.attached_to_filename is not None:
-            email_attachments.append(element)
-    assert len(email_attachments) == len(attachment_elements)
+            assert len(element.text) <= 13
```
cragwolfe, line 494: does this actually have to be 13 and not 12?

Coniferish: No, not sure why I did that in the moment. I do think that the min/pax partition logic can return a last element larger than the max param, though, if in the iteration the last item is smaller than the min partition (that is, if the last item is smaller than the min partition, it gets appended to the prior item, so the combined result might be larger than the max param). Doesn't seem to be the case here, though.

PR #1137, opened 2023-08-16
test: create CI pipelines for verifying base and extras pass respective tests

### LINKED ISSUE #747: feat: CI that tests minimal dependency installs
A version of the unstructured library may be released where runtime errors occur due to missing dependencies if no extras are installed.

In the near term, a smoke test in CI with minimal deps installed that runs the auto `partition` against various doc types in example-docs/.
In the longer term, CI that runs unittests which are marked appropriately for one python version (say 3.8) with only minimal deps installed. In other words, we probably don't need this test to run across the full matrix of supported python versions.

Got nothing.

newelh:
**Summary**
Closes #747
* Create CI Pipeline for running text, xml, email, and html doc tests against the library installed without extras
* Create CI Pipeline for running each library extra against their respective tests


Makefile
```diff
@@ -253,6 +257,21 @@ test:
...
+	PYTHONPATH=. CI=$(CI) pytest \
+		test_${PACKAGE_NAME}/partition/test_text.py \
+		test_${PACKAGE_NAME}/partition/test_email.py \
+		test_${PACKAGE_NAME}/partition/test_html_partition.py \
+		test_${PACKAGE_NAME}/partition/test_xml_partition.py 
+
+.PHONY: test-extra
+test-extra:
+	PYTHONPATH=. CI=$(CI) pytest \
+		test_${PACKAGE_NAME}/partition/${EXTRA_NAME}
```
cragwolfe, line 272: this isn't so friendly from the cli since EXTRA_NAME won't be defined.

maybe a make target for each EXTRA_NAME ?

cragwolfe (approved):
LGTM!

.github/workflows/ci.yml
```diff
@@ -152,6 +152,89 @@ jobs:
...
+    needs: [setup, lint]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/cache/restore@v3
+      id: virtualenv-cache
+      with:
+        path: |
+          .venv
+          nltk_data
+        key: unstructured-${{ runner.os }}-${{ matrix.python-version }}-${{ env.GHA_CACHE_KEY_VERSION }}-${{ hashFiles('requirements/*.txt') }}
```
cragwolfe, line 208: wait, should the cache key be different per extra?

cragwolfe (changes requested):
one question

.github/workflows/ci.yml
newelh: I think you're right, that's a good catch. Let me update it.

newelh: I think we can probably drop the .venv from the cache since it'll be different for every extra anyway

cragwolfe (commented):
Looks like the cache change surfaced a real dep issue!

Sorry I did not think of this earlier, but I think this change actually introduces too many caches on top of what is already too many caches in unstructured's github CI https://github.com/Unstructured-IO/unstructured/actions/caches .

Can we instead refactor this into one CI run where we create a venv per extra test?

cragwolfe (approved):

cragwolfe:
Looks good one conflicts are resolved. Also, please create github issue to consolidate these tests into one CI run and be more tactical with the inline OS dep and pip installs (i.e. the weaviate and argilla installs).

PR #1150, opened 2023-08-18
chore: skip consistently failing test in main

cragwolfe:
The reason this test is failing is the API is returning "fast" results when "hi_res" is requested, which is being tracked in this ticket: https://github.com/Unstructured-IO/unstructured-api/issues/188 .

This failure was only showing up on the `main` branch, per the commented out `pytest` skips.


ryannikolaidis (approved):
lgtm!

PR #1152, opened 2023-08-18
roman/ingest-custom-errors

rbiseck3:
### Description
Adds three custom errors to ingest:
* `SourceConnectionError`
* `DestinationConnectionError`
* `PartitionError`

Included is a base custom error class that adds a wrapper. This wrapper wraps any raised exception into the custom error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.5-dev2
+## 0.10.5-dev3
 
 ### Enhancements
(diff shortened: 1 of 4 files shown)


ryannikolaidis (approved):
like it! I think this covers a large swathe of cases for now but also leaves flexibility for individual connectors to more granularly specify errors. e.g. assuming case-by-case we'll want to wrap functions in various Connector class methods when connecting to list content.

PR #1154, opened 2023-08-18
chore: pass ocr_mode in partition_pdf_or_image

awalker4:
Set to individual_blocks for now to work around [this bug](https://github.com/Unstructured-IO/unstructured-inference/issues/179).

I verified by printing the current ocr_mode in inference. The `entire_page` default is overridden.


cragwolfe (approved):
LGTM! 

though, you'll need to update the test fixtures

PR #1160, opened 2023-08-19
fix: Don't call extractable_elements if strategy is ocr_only 

### LINKED ISSUE #1079: Refactor `partition_pdf_or_image` for when `strategy == "ocr_only"`
Currently `partition_pdf_or_image` partitions pdf documents twice if `strategy == "ocr_only"`.

The top snippet is run regardless of whether or not `ocr_strategy` is defined, so when a pdf is partitioned and `ocr_strategy` is defined as a kwarg, the document is partitioned and assigned to `extracted_elements`. However, `_partition_pdf_or_image_with_ocr` is returned later in the method (bottom snippet).

```
if not is_image:
        extracted_elements = extractable_elements( ...
```

```
elif strategy == "ocr_only":
        # NOTE(robinson): Catches file conversion warnings when running with PDFs
        with warnings.catch_warnings():
            return _partition_pdf_or_image_with_ocr( ...
```

cdpierse:
- fixes #1079 where partitioning is happening twice in the case of `strategy="ocr_only"`
- only calls `extractable_elements` if we can predetermine that `ocr_only` is not a possible strategy even if it was the intended strategy. 
- Adds additional assertion test that `_partition_pdf_or_image_with_ocr` is not called when falling back to `fast` from `ocr_only`


cdpierse:
Not sure exactly what your style preference is for changelog, let me know if I've gotten it wrong.

cragwolfe:
> Not sure exactly what your style preference is for changelog, let me know if I've gotten it wrong.

That's exactly correct, thanks!

cragwolfe (approved):
LGTM!

Eventually, we should be able to merge `pdf_text_extractable` with `ocr_only` output. But for now, certainly not using it.

Issue #1136, opened 2023-08-16
Improve elements ordering for pdf


christinestraub:
In reference to https://github.com/Sanster/xy-cut, I think this could solve our natural order woes for multi-column or mixed layout pages when we have elements with coordinates coming from `hi_res`. For what it's worth, this may also be useful for other modes (`fast`, `ocr_only` ), even if we do not have precise coordinates.

christinestraub:
Addressed by #1161 (`hi_res` and `fast` strategies) and #1811 (`ocr_only` strategy)

PR #1161, opened 2023-08-21
Feat/1136 elements ordering for pdf

christinestraub:
### Summary
Address [#1136](https://github.com/Unstructured-IO/unstructured/issues/1136) for `hi_res` and `fast` strategies. The `ocr_only` strategy does not include coordinates.
- add functionality to switch sort mode between the current `basic` sorting and the new `xy-cut` sorting for `hi_res` and `fast` strategies
- add the script to evaluate the `xy-cut` sorting approach
- add jupyter notebook to provide evaluation and visualization for the `xy-cut` sorting approach

### Evaluation
```
export PYTHONPATH=.:$PYTHONPATH && python examples/custom-layout-order/evaluate_xy_cut_sorting.py <file_path> <strategy>
```
Here, the file should be under the project root directory. For example,
```
export PYTHONPATH=.:$PYTHONPATH && python examples/custom-layout-order/evaluate_xy_cut_sorting.py example-docs/multi-column-2p.pdf fast
```


cragwolfe:
Generally, also as show in the notebook (soon to be colab), this sorting method generally groups content better in multi column.

For example, given [this PDF ](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7234218/pdf/sbaa031.073.pdf) as input  which is used to create test ingest fixture in main: https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/expected-structured-output/biomed-path/07/07/sbaa031.073.PMC7234218.pdf.json , the content starting with "S6. SLEEP ENDOPHENOTYPES " in the left column all the way down is properly grouped together with the sorting method.
image

However, it is not perfect. Chunks from the 2nd column still intermingle with the first. E.g. the top right discussion is one of the first few elements. Given the improvment, I think this is worth merging sooner. But, perhaps there is a quick tweak that may be made to this algorithm so that when a Y line is drawn, everything to the left is processed first?

christinestraub:
> Generally, also as show in the notebook (soon to be colab), this sorting method generally groups content better in multi column.
> 
> For example, given [this PDF ](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7234218/pdf/sbaa031.073.pdf) as input which is used to create test ingest fixture in main: https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/expected-structured-output/biomed-path/07/07/sbaa031.073.PMC7234218.pdf.json , the content starting with "S6. SLEEP ENDOPHENOTYPES " in the left column all the way down is properly grouped together with the sorting method.
> 
> The below screenshot uses jq to extract text from the main and current branch expected output .json's, e.g. with:
> 
> ```
> jq -rj '.[] | .text,"\n"' "$filename"
> ```
> 
> image
> 
> However, it is not perfect. Chunks from the 2nd column still intermingle with the first. E.g. the top right discussion is one of the first few elements. Given the improvment, I think this is worth merging sooner. But, perhaps there is a quick tweak that may be made to this algorithm so that when a Y line is drawn, everything to the left is processed first?

Sounds reasonable! I'll try to figure out a solution.
FYI, you can get visualizations for [the PDF](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7234218/pdf/sbaa031.073.pdf) using the Python script `evaluate_xy_cut_sorting.py` instead of the Jupyter Notebook.
I used the following command to get visualizations.
`PYTHONPATH=. python examples/custom-layout-order/evaluate_xy_cut_sorting.py test_unstructured_ingest/download/biomed-path/07/07/sbaa031.073.pdf hi_res`

cragwolfe (approved):
Massive improvement in ordering of elements! 🎉

PR #1167, opened 2023-08-22
enhancement: tell users to install missing extras

MthwRobinson:
### Summary

Updates `partition` to let users know to installs the appropriate extras if they're missing. Prior to this PR, users would get an exception stating `partition_pdf` (or whichever function that requires extras) does not exist.

### Testing

First `pip uninstall ebooklib`. Then run

```python

partition(filename="example-docs/winter-sports.epub")
```

The error should look like

```python
ImportError: partition_epub is not available. Install the epub dependencies with pip install "unstructured[epub]"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
-## 0.10.5-dev1
+## 0.10.5-dev2
 
 ### Enhancements
 * Create new CI Pipelines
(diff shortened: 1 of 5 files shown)


cragwolfe (approved):
Confirmed test instructions. LGTM!

PR #1168, opened 2023-08-22
Potter/salesforce2

potter-potter:
Adds Salesforce connector.

Able to partition Account, Case, Campaign, EmailMessage, Lead,


ahmetmeleq:
Adding documentation here might be good:
https://github.com/Unstructured-IO/unstructured/tree/potter/salesforce2/docs/source/upstream_connectors

unstructured/ingest/cli/cmds/salesforce.py
```diff
@@ -0,0 +1,62 @@
...
+)
+@click.option(
+    "--salesforce-consumer-key",
+    required=True,
+    help="password bla bla",
+)
+@click.option(
+    "--salesforce-private-key-path",
+    required=True,
+    help="token bla bla.",
```
ryannikolaidis, line 38: since this is visible in the cli with `unstructured-ingest salesfoce --help` lets fill in some better help text here :)

unstructured/ingest/connector/salesforce.py
```diff
@@ -0,0 +1,328 @@
...
+from dataclasses import dataclass
+from email.utils import formatdate
+from pathlib import Path
+from string import Template
+from textwrap import dedent
+from typing import Any, Dict, List, Type
+
+from dateutil import parser  # type: ignore
+from simple_salesforce import Salesforce
+from simple_salesforce.exceptions import SalesforceMalformedRequest
```
ryannikolaidis, line 20: have been changing elsewhere to make these import inline so that request_dependencies can properly catch and handle this. 

that said, @rbiseck3 is working on a change to wrap the runner instead. so if you want to wait until that merges then we can just leave this as is.

unstructured/ingest/connector/salesforce.py
```diff
@@ -0,0 +1,328 @@
...
+    @staticmethod
+    def parse_folders(folder_str: str) -> List[str]:
+        """Parses a comma separated string of Outlook folders into a list."""
+        return [x.strip() for x in folder_str.split(",")]
+
+
+@dataclass
+class SalesforceIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
+    record_type: str
+    record: Dict[Any, Any]
```
ryannikolaidis, line 138: is this record truly Any, Any? want to make sure this is serializable, so hopefully the types are more constrained?

unstructured/ingest/connector/salesforce.py
```diff
@@ -0,0 +1,328 @@
...
+            elif self.record_type == "Lead":
+                formatted_record = self.create_lead(self.record)
+            elif self.record_type == "Case":
+                formatted_record = self.create_case(self.record)
+            elif self.record_type == "Campaign":
+                formatted_record = self.create_campaign(self.record)
+            else:
+                raise MissingCategoryError(
+                    f"There are no categories with the name: {self.record_type}",
+                )
```
ryannikolaidis, line 266: can we filter out the missing categories back in the SalesforceConnector `get_ingest_docs` call?

ryannikolaidis (changes requested):
A couple questions/requests above, otherwise looks great

potter-potter:
> Adding documentation here might be good: https://github.com/Unstructured-IO/unstructured/tree/potter/salesforce2/docs/source/upstream_connectors

Nice. Thats where they moved to. Thank you.

unstructured/ingest/connector/salesforce.py
potter-potter: @ryannikolaidis you're talking about moving the imports to where there are used and not up top? Or are you talking about putting both imports in 1 line?

unstructured/ingest/cli/cmds/salesforce.py
potter-potter: Yep. Bla bla bla is on the vague side.

unstructured/ingest/connector/salesforce.py
potter-potter: Well key is always a string. But how would we type hint all the different value options? Bool, int, str etc.

That being said I'm going to rewrite it anyway to just be an id (str) and not a dict, and then we will GET all the records in `get_file`

We had discussed that this will create more api connections, but will be safer for especially email... :)

potter-potter:
Didn't mean to close pr.

going to rewrite the pr slightly to allow better serialization.

unstructured/ingest/connector/salesforce.py
potter-potter: Rewriting this whole part so that it is just storing Ids. Like some of the other conectors.

unstructured/ingest/connector/salesforce.py
potter-potter: I see from another PR that you want it where the imports are actually used.

unstructured/ingest/connector/salesforce.py
potter-potter: Yep. Good idea.

unstructured/ingest/connector/salesforce.py
ryannikolaidis: sorry, this got buried on my side. yes, exactly, import in the functions where they are used or where the exceptions are raised. then wrap with the request_dependencies decorator and then the user will get the `pip install unstructured[salesforce]` message rather than complaint about single dependency that is missing.

potter-potter: Cool. Makes sense. I think you meant `requires_dependencies` above instead of `request_dependencies`. That threw me a little bit.

I put the Exception import in the function.

ryannikolaidis: oh yep, sorry! awesome, thanks.

unstructured/ingest/cli/cmds/salesforce.py
```diff
@@ -0,0 +1,62 @@
...
+    map_to_standard_config,
+    run_init_checks,
+)
+from unstructured.ingest.logger import ingest_log_streaming_init, logger
+from unstructured.ingest.runner import salesforce as salesforce_fn
+
+
+@click.command()
+@click.option(
+    "--salesforce-categories",
```
rbiseck3, line 19: Given the cli command is `salesforce`, I usually find having `salesforce` as part of the params redundant.

ryannikolaidis (approved):
lgtm!

unstructured/ingest/connector/salesforce.py
```diff
@@ -0,0 +1,339 @@
...
+    def get_ingest_docs(self) -> List[SalesforceIngestDoc]:
+        """Get Salesforce Ids for the records.
+        Send them to next phase where each doc gets downloaded into the
+        appropriate format for partitioning.
+        """
+        from simple_salesforce.exceptions import SalesforceMalformedRequest
+
+        client = self.config._get_client()
+
+        record_ids = []
```
rbiseck3, line 317: The name is a bit misleading since it's a list of docs rather than IDs

unstructured/ingest/cli/cmds/salesforce.py
potter-potter: @rbiseck3  - I hear you. Back before the refactor, this wasn't the case. But I would definitely prefer to simplify. Thank you.

unstructured/ingest/connector/salesforce.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+class MissingCategoryError(Exception):
+    """There are no categories with that name."""
+
+
+ACCEPTED_CATEGORIES = ["Account", "Case", "Campaign", "EmailMessage", "Lead"]
+
+
+EMAIL_TEMPLATE = Template(
```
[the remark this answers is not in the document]
potter-potter: @rbiseck3 Could you expand on your question? Are you asking if they should/should not be in all caps? Or are you asking if we are going to end up changing them? 

They are the most important values for the particular categories of the connector in my opinion. Email has to be of a very specific format to be partitioned. They can be added to or modified if needed.

unstructured/ingest/connector/salesforce.py
potter-potter: Hmm. Interesting to think about. They are SalesforceIngestDoc . But the valuable piece we are sending through is the ids of the Salesforce records. I guess it is how you interpret the word docs. I do like a variable name to be the most accurate, so thank you.  @ryannikolaidis  thoughts?

unstructured/ingest/connector/salesforce.py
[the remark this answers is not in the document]
rbiseck3: i don't actually have enough context around how these templates were generated, I was just wondering if these were specific to the salesforce api and if there was a risk of them changing in the future?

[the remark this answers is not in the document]
potter-potter: Got ya. They are definitely specific to Salesforce. I considered putting them into another file, but that would create more mess I think.

I came up with them. As long as Salesforce records continue to have the basic key/values that are used in the templates, they should be fine. They are very basic keys that I assume will be around for a while. Of course nothing is guaranteed.

unstructured/ingest/cli/cmds/salesforce.py
ryannikolaidis: ah, yea, good catch here.

unstructured/ingest/connector/salesforce.py
ryannikolaidis: ohh I see. at instantiation they don't necessarily contain more than ids, but that's okay; they are instances of SalesforceIngestDoc which can use the id as the instruction set to fetch the raw document, so less "doc" in terms of raw document and rather in terms of IngestDoc. maybe we call this `ingest_docs` to remove any confusion?

ryannikolaidis:
just peeked at the ingest test error. you'll need to add a registry_name field for your IngestDoc class, as done [here](https://github.com/Unstructured-IO/unstructured/blob/a4ec43a85fc5afde9ffee679b65835f7a0dd6d09/unstructured/ingest/connector/confluence.py#L86) and you'll need to add the class to the lookup in [the registry](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/connector/registry.py).

PR #1170, opened 2023-08-22
Roman/ingest logging

rbiseck3:
### Description
The ingest code needs to be able to handle multiprocessing fork without issue. Currently the logger instance used was a singular shared one which is not safe for forked processes. This PR refactors the use of the logger that use a new instance for each one.


cragwolfe:
This may have been lost in a refactor, but for awhile this is why logging was initialized per sub-process in its initializer:

https://github.com/Unstructured-IO/unstructured/blob/70420b5c781e6eaeb/unstructured/ingest/main.py#L85-L86

though, this was shortly added later for mac + linux consistency:

https://github.com/Unstructured-IO/unstructured/blob/c51adb21e/unstructured/ingest/main.py#L26-L27

rbiseck3:
@cragwolfe, I don't believe these lines actually create a new logger instance for each pool to use: 
```python
            initializer=ingest_log_streaming_init,
            initargs=(logging.DEBUG if self.verbose else logging.INFO,),
```
Rather it simply configures the same logger. 

I'm not actually sure what issues had been run into to to require: 
```python
with suppress(RuntimeError):
    mp.set_start_method("spawn")
```

cragwolfe:
> Rather it simply configures the same logger.

i'm not sure about that. my understanding was it's the "same logger", but the logger object instance per process should be different and unique to the process it was instantiated in.

PR #1178, opened 2023-08-22
feat: serialize ingest docs as json

ryannikolaidis:
IngestDoc class instances (and by extension ConnectorConfig instances) are serialized to the child process pool via pickle. Pickle is not only an [insecure data format](https://docs.python.org/3/library/pickle.html#:~:text=The%20pickle%20module%20is%20not,could%20have%20been%20tampered%20with.), but it is opaque and does not include any mechanisms for making sure live objects can be rehydrated across versions. This could be particular problematic for users that want to pass these over-the-wire.

This PR updates all IngestDoc classes to be JSON serializable/de-serializable and updates the processing flow to leverage this for passing from the parent process into mp pool subprocesses.


unstructured/ingest/connector/outlook.py
```diff
@@ -37,6 +35,7 @@ class SimpleOutlookConfig(BaseConnectorConfig):
     authority_url: Optional[str] = field(repr=False)
     ms_outlook_folders: List[str]
     recursive: bool = False
+    registry_name: str = "outlook"
```
potter-potter, line 38: Could you give some explanation where/why registry name is eventually used.

unstructured/ingest/connector/box.py
```diff
@@ -27,36 +27,25 @@ class AccessTokenError(Exception):
 @dataclass
 class SimpleBoxConfig(SimpleFsspecConfig):
     @requires_dependencies(["boxfs"], extras="box")
-    def __post_init__(self):
+    def get_access_kwargs(self):
```
potter-potter, line 30: This is really nice.

unstructured/ingest/connector/outlook.py
ryannikolaidis: Yep, so registry_name is used so that we have a reference to get back to the original document class. By adding as a field it will end up in the serialized output of the IngestDoc. When we deserialize via the [registry](https://github.com/Unstructured-IO/unstructured/pull/1178/files#diff-d5c2c5d105fa4cf0da28254ec2fc514c1c5b6f9d5ce05167ae272611ecae6d36) `create_ingest_doc_from_json` method, we lookup the class using this name. I didn't want to just fetch/store the name of the class because this would that serialized artifacts would break if we ever refactored the name of a class (so lightly decoupling)

unstructured/ingest/connector/box.py
ryannikolaidis: Thanks! Yea, definitely needed to sit with this one a bit, but happy with where we landed.

rbiseck3:
How does the dataclasses-json library handle fields that are not defined on the class itself but are set as internal values via the methods? For example the IngestDoc gets `self.isd_elems_no_filename` set when the `process_file()` method is called but that's not one the the fields set on the class and therefor doesn't exist in the constructor. Would this not persist between serialization/de-serialization?

ryannikolaidis:
> How does the dataclasses-json library handle fields that are not defined on the class itself but are set as internal values via the methods? For example the IngestDoc gets `self.isd_elems_no_filename` set when the `process_file()` method is called but that's not one the the fields set on the class and therefor doesn't exist in the constructor. Would this not persist between serialization/de-serialization?

@rbiseck3 great question. so that particular case shouldn't matter, because we do not make any calls or assignments to the IngestDoc in between the instantiation by the connector and the point that we serialize it, and in principle we never should. that said, I believe dataclasses-json's `to_json` is operating specifically on the original defined class attributes. if fields aren't part of the original dataclass definition they will be ignored (you would have to define custom serialization/deserialization definitions for this).

just to be sure ran this experiment, which did show the `new_attribute` did not exist on the deserialized doc:

```
from unstructured.ingest.connector.s3 import S3Connector, SimpleS3Config

doc_connector = S3Connector(  # type: ignore
    standard_config=standard_config,
    config=SimpleS3Config(
        path="s3://utic-dev-tech-fixtures/small-pdf-set/",
        recursive=True,
        access_kwargs={"anon": True},
    ),
)

ingest_docs = doc_connector.get_ingest_docs()
ingest_doc = ingest_docs[0]
ingest_doc.new_attribute = 42
ingest_doc_json = ingest_doc.to_json()
ingest_doc = create_ingest_doc_from_json(ingest_doc_json)
print(hasattr(ingest_doc, 'new_attribute'))
```

unstructured/ingest/registry.py
```diff
@@ -0,0 +1,68 @@
+import json
```
rbiseck3, line 1: Does it make sense for this to live inside the connector package? Possibly in the `__init__` file? This is very specific to the connectors rather than the entire ingest code.

rbiseck3 (approved):

unstructured/ingest/registry.py
ryannikolaidis: yep, good point! moving it now.

unstructured/ingest/connector/registry.py
```diff
@@ -0,0 +1,68 @@
...
+    "s3": S3IngestDoc,
+    "sharepoint": SharepointIngestDoc,
+    "slack": SlackIngestDoc,
+    "wikipedia_html": WikipediaIngestHTMLDoc,
+    "wikipedia_text": WikipediaIngestTextDoc,
+    "wikipedia_summary": WikipediaIngestSummaryDoc,
+}
+
+
+def create_ingest_doc_from_json(ingest_doc_json: str) -> BaseIngestDoc:
```
cragwolfe, line 58: nice!!!

cragwolfe (approved):
🙌

unstructured/ingest/connector/s3.py
```diff
@@ -15,7 +15,11 @@ class SimpleS3Config(SimpleFsspecConfig):
     pass
 
 
+@dataclass
 class S3IngestDoc(FsspecIngestDoc):
+    remote_file_path: str
+    registry_name: str = "s3"
```
jasonbot, line 21: Since we can't count on some using using `~/.aws/config` to save key/secret in server environments, we should add those to the base class if they're added to `kwargs`. We can fix this forward, this PR is already huge.

unstructured/ingest/connector/registry.py
```diff
@@ -0,0 +1,74 @@
...
+    "notion_page": NotionPageIngestDoc,
+    "onedrive": OneDriveIngestDoc,
+    "outlook": OutlookIngestDoc,
+    "reddit": RedditIngestDoc,
+    "s3": S3IngestDoc,
+    "sharepoint": SharepointIngestDoc,
+    "slack": SlackIngestDoc,
+    "wikipedia_html": WikipediaIngestHTMLDoc,
+    "wikipedia_text": WikipediaIngestTextDoc,
+    "wikipedia_summary": WikipediaIngestSummaryDoc,
```
jasonbot, line 60: I don't know if it would be cleaner to have a `@register_ingest_doc` decorator that you could put on each data class and have it register with the `INGEST_DOC_NAME_TO_CLASS`.

This works too, and is more consistent with how we register fsspec classes. Disregard prior comment.

jasonbot (approved):
This is a perfectly serviceable solution, thank you for keeping it simple. Would be nice (but a monstrous ask) to add more tests; but the ones you have are enough to make me confident it'll work. Moving forward some IngestDocs (like S3) may need additional attributes but we can fix forward when we hit those issues.

unstructured/ingest/connector/registry.py
ryannikolaidis: yea, I also went down the path of trying to introspect as well, but that got a little gnarly. landed here for consistency, exactly as you said.

I'm not wild on needing to maintain these lists, but if we do resolve should probably solve consistently everywhere.

unstructured/ingest/connector/s3.py
ryannikolaidis: yep, we can add as optional flags in the cli and then just pass through to the access_kwargs dictionary, which does get serialized. that should be a quick, pretty straightforward change.

PR #1180, opened 2023-08-23
Added plain-text comparison for tests

benjats07:
This PR adds a comparison during ingest test for the content of the files in plain text (i.e.: without JSON format)


cragwolfe:
I tried modifying 
https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/expected-structured-output/local-single-file/english-and-korean.png.json
and running

    ./test_unstructured_ingest/test-ingest-local-single-file.sh

and got this output:

\ No newline at end of file
xargs: jq '.[].text'<{}>/Users/cragwolfe/r/unstructured/test_unstructured_ingest/expected-text-output/local-single-file: No such file or directory
xargs: jq '.[].text'<{}>/Users/cragwolfe/r/unstructured/test_unstructured_ingest/text-output/local-single-file: No such file or directory

Which makes sense because a file is a valid test path in addition to a directory. Besides fixing that, additional comments:

* please word wrap long lines to multiple lines (say 80 characters per line) for easier diff viewing.
* `store-just-text.sh` could have a more descriptive name. something like `json-to-clean-text-file.sh` ?
* please add a line or two to .gitignore so the generated text files are ignored by git

benjats07:
@cragwolfe I addressed those issues

test_unstructured_ingest/json-to-clean-text-folder.sh
```diff
@@ -0,0 +1,16 @@
...
+#   $2 folder to place the text field for all entries, for all files at $1
+ 
+set +e
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+INPUT_FOLDER_NAME=$1
+OUTPUT_DIR_TEXT=$2
+echo "INPUT FOLDER: $INPUT_FOLDER_NAME"
+echo "OUTPUT FOLDER: $OUTPUT_DIR_TEXT"
+mkdir -p "$OUTPUT_DIR_TEXT"
+find "$INPUT_FOLDER_NAME" -type f -print0| xargs -IX --null -n1 "$SCRIPT_DIR"/json-to-text.sh X "$OUTPUT_DIR_TEXT"
```
cragwolfe, line 16: newline nit per GH UI.

test_unstructured_ingest/json-to-text.sh
```diff
@@ -0,0 +1,13 @@
...
+# text elements. The resulting file will be stored at the $2 folder with the same
+# name as the original file appending .txt as suffix.
+# Arguments:
+# - $1 path to the file to clean
+# - $2 path to folder to store the result
+# 
+
+BASE=$(basename "$1")
+DEST=$2/$BASE.txt
+jq '.[].text'<"$1"|fold -w 80 -s > "$DEST"
```
cragwolfe, line 13: newline nit per GH UI.

test_unstructured_ingest/.gitignore
```diff
@@ -1,2 +1,3 @@
 structured-output
 download
+*text-output
```
cragwolfe, line 3: let's make this a bit more specific, i think the syntax is:
```
test_unstructured_ingest/expected-text-output/**
test_unstructured_ingest/text-output/**
```

cragwolfe, line 3: also can add an extra newline to avoid the GH UI complaint.

test_unstructured_ingest/.gitignore
```diff
@@ -1,2 +1,4 @@
 structured-output
 download
+test_unstructured_ingest/expected-text-output/**
```
qued, line 3: From the folder name it sounds like this is where the text output fixtures would go. Don't we want to track changes in git for these fixtures?

cragwolfe: there are not text fixtures so far, only json fixtures. this PR derives the clean text from json's at runtime.

cragwolfe (approved):
Definitely a big improvement to view partitioned text output changes! 👍

PR #1181, opened 2023-08-23
ci: fix documentation build flow

ryannikolaidis:
* pins sphinx-X library versions to workaround lack of availability in CI
* pins furo because our docs `conf.py` config is not compatible with the latest version


qued (approved):
LGTM. I'm a little worried about the `safetensors` bump, but CI isn't complaining.

PR #1192, opened 2023-08-23
Data source properties

rvztz:
Adds data source properties to documents:
- data_created
- date_modified
- version
- record_locator

Adds logic to check if file exists in the remote source. For connectors that don't directly support it, adds exception handling to check any issues while retrieving the file.


ryannikolaidis:
I haven't gone through this yet in too much detail, but one larger observation to revisit all of these with: in several of places (like `airtable` for example) you are using a pattern where you initialize variables to None on `post_init` and then rely on a call to `get_file` to set the internal values. this is problematic because it means if you call to say `exists` before calling `get_file` it will not properly test the value (it actually might error because with `airtable` you check `>= 1`). if were relying on `exists` to determine whether or not to call `get_file` you could see why this would be an issue. largely, we should never assume some other method was called prior, and the state of these shouldn't change after other methods are called. also, ideally these don't rely on internal variables at all.

ryannikolaidis (changes requested):
see note above

rbiseck3:
To touch on @ryannikolaidis point about internal values being used, take a look at how I handled the check for existence in the notion connector: [connector/notion/connector.py#L145](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/connector/notion/connector.py#L145). I'm all for using internal variables to cache the value from any other methods that may provide insight into it (i.e. `exists`), but you need to keep track of if _anything_ has checked and if not, explicitly do it if possible in the method that's meant to return that value.

ryannikolaidis:
> To touch on @ryannikolaidis point about internal values being used, take a look at how I handled the check for existence in the notion connector: [connector/notion/connector.py#L145](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/connector/notion/connector.py#L145). I'm all for using internal variables to cache the value from any other methods that may provide insight into it (i.e. `exists`), but you need to keep track of if _anything_ has checked and if not, explicitly do it if possible in the method that's meant to return that value.

ahh yes, good point! definitely meant "aside from caching".

rbiseck3:
Hey @rvztz, a couple general comments:
1. Can you include a PR description just make it easy to quickly understand what the PR is tackling and any relevant information that makes it easy for a reviewer  (i.e. how it was tested, choices made that might warrant some conversation, etc)?
2. I think there are some merge issues at the moment that are making it hard to parse through the relevant changes being introduced in this PR. For example it's showing a bunch of changes to the partition code in the PR but I'm pretty sure those aren't changes you introduced. 

Thanks!

ryannikolaidis:
> Hey @rvztz, a couple general comments:
> 
> 1. Can you include a PR description just make it easy to quickly understand what the PR is tackling and any relevant information that makes it easy for a reviewer  (i.e. how it was tested, choices made that might warrant some conversation, etc)?
> 2. I think there are some merge issues at the moment that are making it hard to parse through the relevant changes being introduced in this PR. For example it's showing a bunch of changes to the partition code in the PR but I'm pretty sure those aren't changes you introduced.
> 
> Thanks!

@rvztz just want to make sure you saw these notes ^.

also if you have to rework this a bit anyways, may be worth chunking these into smaller PRs. doesn't have to be all at once and will probably be a bit easier for a reviewer.

ryannikolaidis:
closing since these are now broken down into smaller prs

PR #1194, opened 2023-08-24
Move all dependency requirement check to cli code rather than connectors

rbiseck3:
### Description
This refactor is happening for a couple reasons:
* The formatted error for a missing response makes sense in the context of a user running the cli (`s3fs` for example):
```
Following dependencies are missing: s3fs. Please install them using `pip install "unstructured[s3]"`
``` 
If a user wanted to import the connector code itself to use, this might not apply so having functions wrapped in that codebase doesn't make sense
* The code is meant to only wrap a function but was being used to also wrap classes which can break the inheritance of classes already wrapped by `@dataclass`.
* This also helps consolidate what needs to be wrapped with the `@requires_dependencies` decorator to a single method, being the CLI entrypoint for each connector.

### DIFF
--- a/unstructured/ingest/cli/cmds/airtable.py
+++ b/unstructured/ingest/cli/cmds/airtable.py
@@ -12,6 +12,7 @@ from unstructured.ingest.cli.common import (
 )
 from unstructured.ingest.logger import ingest_log_streaming_init, logger
 from unstructured.ingest.runner import airtable as airtable_fn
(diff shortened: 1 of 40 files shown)


ryannikolaidis:
we connected async on this, but dropping line here for context. assuming we probably need to rethink with the addition of the registry, since we don't want top level imports to trigger when we reference classes for which the user might not have the dependencies installed.

PR #1197, opened 2023-08-24
fix: respect `<pre>` tag order in `partition_html`

### LINKED ISSUE #1184: html elements are in a strange order
**Describe the bug**
if using partition_html the elements is not always in the same order as in the html document. Especially if there are "pre" tags this will be added at the end of the elements. 

**To Reproduce**
```python
elements = partition_html(filename=html_file)
all_text = []
for el in elements:
    el.apply(group_broken_paragraphs)
    el.apply(clean_extra_whitespace)
    print(el.to_dict()['text'])
```

**Expected behavior**
I would expect that the itered elements have the same order as in the html document.

**Environment Info**
unstructured==0.10.5

Is there a way I am not aware of and have the same order?

MthwRobinson:
### Summary

Closes #1184. Updates `partition_html` to respect the ordering of `<pre>` tags in HTML documents.

### Testing

The elements in the following example should be in the correct order.

```python

    html_text = """
    <pre>The Big Brown Bear</pre>
    <div>The big brown bear is growling.</div>
    <pre>The big brown bear is sleeping.</pre>
    <div>The Big Blue Bear</div>
    """
```


qued (commented):
I don't see a difference on the results of running the test code here vs. `main`.

MthwRobinson:
@qued - Fixed the test example, it was picking up on the inner tags. You should see the `<pre>` tags at the end on `main` now and the correct ordering on the feature branch. Updated the unit test as well.

qued (approved):
LGTM!

PR #1223, opened 2023-08-28
Add s3 writer

rbiseck3:
### Description
Convert s3 cli code to also support writing to s3. Writers are added as optional subcommands to the parent command with their own arguments. Custom `click.Group` introduced to add some custom formatting and text in help messages. 

To limit the scope of this PR, most existing files were not touched but instead new files were added for the new flow. This allowed _only_ the s3 connector to be updated without breaking any other ones.

### DIFF
--- a/examples/ingest/s3-small-batch/ingest.sh
+++ b/examples/ingest/s3-small-batch/ingest.sh
@@ -12,5 +12,9 @@ PYTHONPATH=. ./unstructured/ingest/main.py \
         s3 \
          --remote-url s3://utic-dev-tech-fixtures/small-pdf-set/ \
          --anonymous \
-         --structured-output-dir s3-small-batch-output \
-         --num-processes 2
+         --output-dir s3-small-batch-output \
+         --num-processes 2 \
+         --verbose \
+         s3 \
+         --anonymous \
+         --remote-url s3://utic-dev-tech-fixtures/small-pdf-set-output
--- a/unstructured/ingest/cli/cmds/utils.py
+++ b/unstructured/ingest/cli/cmds/utils.py
@@ -5,6 +5,11 @@ import click
 
 class Group(click.Group):
     def parse_args(self, ctx, args):
+        """
+        This allows for subcommands to be called with the --help flag without breaking
+        if parent command is missing any of its required parameters
+        """
+
         try:
             return super().parse_args(ctx, args)
         except click.MissingParameter:
@@ -17,8 +22,9 @@ class Group(click.Group):
             return super().parse_args(ctx, args)
 
     def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
-        """Extra format methods for multi methods that adds all the commands
-        after the options.
+        """
+        Copy of the original click.Group format_commands() method but replacing
+        'Commands' -> 'Destinations'
         """
         commands = []
         for subcommand in self.list_commands(ctx):
--- a/unstructured/ingest/cli/cmds/utils.py
+++ b/unstructured/ingest/cli/cmds/utils.py
@@ -1,0 +1,48 @@
+from gettext import gettext as _
+
+import click
+
+
+class Group(click.Group):
+    def parse_args(self, ctx, args):
+        try:
+            return super().parse_args(ctx, args)
+        except click.MissingParameter:
+            if "--help" not in args:
+                raise
+
+            # remove the required params so that help can display
+            for param in self.params:
+                param.required = False
+            return super().parse_args(ctx, args)
+
+    def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
+        """Extra format methods for multi methods that adds all the commands
+        after the options.
+        """
+        commands = []
+        for subcommand in self.list_commands(ctx):
+            cmd = self.get_command(ctx, subcommand)
+            # What is this, the tool lied about a command.  Ignore it
+            if cmd is None:
+                continue
+            if cmd.hidden:
+                continue
+
+            commands.append((subcommand, cmd))
+
+        # allow for 3 times the default spacing
+        if len(commands):
+            if formatter.width:
+                limit = formatter.width - 6 - max(len(cmd[0]) for cmd in commands)
+            else:
+                limit = -6 - max(len(cmd[0]) for cmd in commands)
+
+            rows = []
+            for subcommand, cmd in commands:
+                help = cmd.get_short_help_str(limit)
+                rows.append((subcommand, help))
+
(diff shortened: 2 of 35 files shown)


examples/ingest/s3-small-batch/ingest.sh
```diff
@@ -12,5 +12,9 @@ PYTHONPATH=. ./unstructured/ingest/main.py \
...
          --remote-url s3://utic-dev-tech-fixtures/small-pdf-set/ \
          --anonymous \
-         --structured-output-dir s3-small-batch-output \
-         --num-processes 2
+         --output-dir s3-small-batch-output \
+         --num-processes 2 \
+         --verbose \
+         s3 \
+         --anonymous \
+         --remote-url s3://utic-dev-tech-fixtures/small-pdf-set-output
```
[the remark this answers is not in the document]
rbiseck3: These are no different than the connector configs being used for the upstream connection, but used for the downstream one. It might look exceptionally weird here because we're reading and writing to s3 with the same configs but this won't always be the case.

unstructured/ingest/cli/cmds/utils.py
```diff
@@ -0,0 +1,48 @@
...
+            return super().parse_args(ctx, args)
+
+    def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
+        """Extra format methods for multi methods that adds all the commands
+        after the options.
+        """
+        commands = []
+        for subcommand in self.list_commands(ctx):
+            cmd = self.get_command(ctx, subcommand)
+            # What is this, the tool lied about a command.  Ignore it
```
ryannikolaidis, line 26: is this due to some weird handling by click?

unstructured/ingest/cli/cmds/utils.py
```diff
@@ -0,0 +1,48 @@
...
+                limit = -6 - max(len(cmd[0]) for cmd in commands)
+
+            rows = []
+            for subcommand, cmd in commands:
+                help = cmd.get_short_help_str(limit)
+                rows.append((subcommand, help))
+
+            if rows:
+                with formatter.section(_("Destinations")):
+                    formatter.write_dl(rows)
```
ryannikolaidis, line 48: what's going on here?

rbiseck3: This is a copy of the existing `format_commands` method on the `Group` class in click but replaces `Commands` -> `Destinations` to make is more obvious that the subcommands are related to downstream connectors.

examples/ingest/s3-small-batch/ingest.sh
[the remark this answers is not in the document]
ryannikolaidis: oh no, meant more broadly that the general form of the subcommand of s3 to specify output. appreciate the re-use of the configs here. i.e. just generally calling 

```
./unstructured/ingest/main.py s3 --some --options s3 --some --other -options
```

and tracking that this is input and output. though maybe making it clear in the docs and help is enough, have definitely worked through far more confusing conventions in clis.

unstructured/ingest/cli/cmds/utils.py
ryannikolaidis: got it. is this potentially brittle to updates in click?

rbiseck3: It's still using the same logic so only changes to this _exact_ method wouldn't be reflected, but this it really just used for formatting so I don't seen this being high risk.

examples/ingest/s3-small-batch/ingest.sh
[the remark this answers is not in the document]
ryannikolaidis: oh yea, totally appreciate the complexity and don't have any immediate thoughts.

unstructured/ingest/doc_processor/generalized.py
```diff
@@ -59,9 +60,11 @@ def process_document(ingest_doc_json: str, **partition_kwargs) -> Optional[List[
         # the results. Instead, the Processor (caller) may work with the aggregate
         # results across all docs in memory.
         doc.write_result()
-    except Exception:
+    except Exception as e:
         # TODO(crag) save the exception instead of print?
         logger.error(f"Failed to process {doc}", exc_info=True)
+        raise e
```
ryannikolaidis: if we raise here, does this kill other subprocesses from completing?

rbiseck3: That’s outdated. Had it in for debugging. Should be gone now.

unstructured/ingest/cli/cmds/utils.py
```diff
@@ -0,0 +1,48 @@
+from gettext import gettext as _
+
+import click
+
+
+class Group(click.Group):
+    def parse_args(self, ctx, args):
```
ryannikolaidis, line 7: maybe a docstring here. esp on why we needed / problems we're solving..etc.

unstructured/ingest/interfaces2.py
```diff
@@ -0,0 +1,446 @@
...
+        default_factory=lambda: ["element_id", "text", "type", "metadata"],
+    )
+    flatten_metadata: bool = False
+    metadata_exclude: t.List[str] = field(default_factory=list)
+    metadata_include: t.List[str] = field(default_factory=list)
+    partition_endpoint: t.Optional[str] = None
+    api_key: t.Optional[str] = None
+
+    @staticmethod
+    def get_schema() -> dict:
```
ryannikolaidis, line 54: is the idea that we would want to validate against this in the future? allow users to inspect what's needed?

ryannikolaidis: needed for adding the s3 writer? or just adding as something we should do across the board?

rbiseck3: Removed this from the base config class, forgot to remove the implementation. This is now gone. I had cherry picked this from the other PR I had out but don't need if for this specific one.

ryannikolaidis:
overall looks good. appreciate that you have kept it straightforward. a couple comments/questions. 

not this pr, but once we add s3 write support for all connectors, reminder we'll want to bump some docs: README (to include steps for output connectors?), [s3 example ](https://github.com/Unstructured-IO/unstructured/blob/main/examples/ingest/s3-small-batch/ingest.sh)to include output?, output connectors[ documentation doc](https://github.com/Unstructured-IO/unstructured/tree/main/docs/source/downstream_connectors).

unstructured/ingest/interfaces2.py
ryannikolaidis: cool cool, thanks

Issue #1232, opened 2023-08-29
bug: cv2 should not be top level import


cragwolfe:
**Describe the bug**
cv2 is used for debugging/visualizing the algorithm in unstructured/partition/utils/xycut.py , however it should not be imported as a runtime partition dependency since the underlying OS dep libGL may not be available (and cause a stack trace at runtime).

**Additional context**
The cleanest "fix" may be to move the cv2 functions to another module under unstructured/partition/utils, but lazy imports would also work.

MthwRobinson:
Closed in #1234

PR #1234, opened 2023-08-29
fix: make cv2 dependency optional

qued:
Makes CV2 dependency optional. It's not used in any of the functional code.

Note: None of this module has tests. We need tests ASAP.


cragwolfe (changes requested):
Can we move it to another module that isn't imported on the hot (partition) path? They are useful functions, as used in:

https://colab.research.google.com/drive/1q-wjJdFjGhhei5C9m-hzXszYANrXuWbT
and
https://colab.research.google.com/drive/1HgBvHNPnY-dXO043DftvvMeynlLPgQ_p

cragwolfe:
Related issue: https://github.com/Unstructured-IO/unstructured/issues/1232

cragwolfe (approved):

cragwolfe:
worth a release?

PR #1242, opened 2023-08-30
[issue 1237] fix empty coordinates break sorting bug

### LINKED ISSUE #1237: bug/`sort_page_elements` with empty coordinates (`chipper` output for `PDF`s)
**Describe the bug**
`AttributeError: 'NoneType' object has no attribute 'points'` when using the partition auto method with `chipper` model with a `.pdf` file as input. I see this happening because `chipper` returns empty bounding boxes information, and now the default method for sorting `SORT_MODE_XY_CUT` counts on having those. It breaks the [example notebook](https://github.com/Unstructured-IO/unstructured/blob/main/examples/chipper-local-inference/chipper-local-inference.ipynb) for chipper local inference.

**To Reproduce**
```python
filename = 'layout-parser-paper-fast.pdf'
model_name = "chipper"
elements = partition(filename=filename, strategy='hi_res', model_name=model_name, pdf_image_dpi=300)

# to reproduce it with an image input (restart kernel first)
import os
os.environ['UNSTRUCTURED_HI_RES_MODEL_NAME'] = model_name
filename = 'layout-parser-paper-fast.jpg'
model_name = "chipper"
elements = partition(filename=filename, strategy='hi_res')
```

**Expected behavior**
Unstructured elements output.

**Screenshots**
image

**Environment Info**
I run this on a colab. I installed the libraries as:
```
!pip install unstructured --quiet
!pip install unstructured-inference --quiet
!apt-get update -y && apt-get install -y poppler-utils --quiet
!apt-get -qq install tesseract-ocr
!pip install -q --user --upgrade pillow
```

The full code is the [example notebook](https://github.com/Unstructured-IO/unstructured/blob/main/examples/chipper-local-inference/chipper-local-inference.ipynb) to run local inference with `chipper` which was working without bugs at the time of its creation. 

**Additional Context**
At first glance I see its related to last introduced changes to solve the double column sorting of elements in [sorting.py](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/utils/sorting.py). 
```python
sort_page_elements(
    page_elements: List[Element],
    sort_mode: str = SORT_MODE_XY_CUT,
```

badGarnet:
This PR resolves #1237 by checking if any coordinates are `None`; if yes do not attempt to sort with xy cut method and return the list as is.


unstructured/partition/utils/sorting.py
```diff
@@ -41,10 +42,13 @@ def sort_page_elements(
 
     if sort_mode == SORT_MODE_XY_CUT:
         coordinates_list = [el.metadata.coordinates for el in page_elements]
-        boxes = [coordinates_to_bbox(coords) for coords in coordinates_list]
-        res: List[int] = []
-        recursive_xy_cut(np.asarray(boxes).astype(int), np.arange(len(boxes)), res)
-        sorted_page_elements = [page_elements[i] for i in res]
+        if any(coords is None for coords in coordinates_list):
+            sorted_page_elements = page_elements
```
cragwolfe, line 46: probably worth a warning, because i don't think should get called in the first place under normal assumptions

cragwolfe (approved):

PR #1249, opened 2023-08-30
feat: `partition_xml` infers element type on each leaf node

### LINKED ISSUE #1229: enhancement: XML partitioning logic shouldn't be identical to plain text
The current `partition_xml` logic is to extract the text from leaf nodes in the XML, concatenate it via newlines, then delegate to `partition_text`.

While it makes sense to classify the XML text using the same logic we use for plain text, it may not make sense to delegate the splitting. We should consider leaving the elements split up as they are in the leaf nodes, and only calling `element_from_text` to get the element type.

MthwRobinson:
### Summary

Closes #1229. Updates `partition_xml` so that the element type is inferred on each leaf node when `xml_keep_tags=False` instead of delegating splitting and partitioning to `partition_xml`. If `xml_keep_tags=True`, the file is treated like a text file still and partitioning is still delegated to `partition_text`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
-## 0.10.10-dev0
+## 0.10.10-dev1
 
 ### Enhancements
+
+* If `xml_keep_tags=False`, `partition_xml` infers the element type for each leaf element
+  separately without delegating to `partition_text`. Note, `max_partition` and `min_partition`
+  no longer apply when `xml_keep_tags=True` since `partition_xml` no longer delegates
+  to `partition_text` in that case.
 * Bump `unstructured-inference==0.5.18`, change non-default detectron2 classification threshold
 
 ### Features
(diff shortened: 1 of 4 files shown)


qued (approved):
Overall LGTM. I confirmed the behavior, but we probably should have a test with a fixture that has a single leaf node with content separated by newlines, and the test makes sure that the output matches the results of processing the same content as text when `xml_keep_tags` is `True` and doesn't match when it's `False`.

MthwRobinson:
Added a test confirming the updated behavior.

MthwRobinson:
Lemme actually try one more tweak to improve the output for `xml_keep_tags=True` though, that still looks weird.

MthwRobinson:
@qued - Updated `xml_keep_tags` so all of the raw text gets treated as a single `Text` elements, similar to what's done with Excel and CSV now. That gets rid of the bad splits on returns for that as well.

PR #1251, opened 2023-08-30
dev: add py-spy profiling

badGarnet:
This PR adds a new developer tool for profiling performance: `py-spy`. Additionally it adds a new make command to start a docker with your local `unstructured` repo mounted for quick testing code in a Rocky Linux environment (see usage below for intent).

### py-spy

It is a sampling profiler https://github.com/benfred/py-spy and in practice usually provides more readily usable information than commonly used `cProfiler`. It also supports output to `speedscope` format, [which](https://github.com/jlfwong/speedscope#usage) provides a rich view of the profiling result.

### usage

The new tool is added to the existing `profile.sh` script and is readily discoverable in the interactive interface. When select to view the new speedscope format profile it would show up in your local browser if you followed the readme to install speedscope locally via `npm install -g speedscope`.

On macOS the profiling tool needs superuser privilege. If you are not comfortable with that feel free to run the profiling inside a Linux container if your local dev env is macOS.


badGarnet:
please advise on if this needs changelog update

scripts/performance/run_partition.py
```diff
@@ -12,6 +13,11 @@
 
     file_path = sys.argv[1]
     strategy = sys.argv[2]
-    result = partition(file_path, strategy=strategy)
+    model_name = None
+    if len(sys.argv) > 3:
+        model_name = sys.argv[3]
+    else:
+        model_name = os.environ.get("PARTITION_MODEL_NAME")
+    result = partition(file_path, strategy=strategy, model_name=model_name)
```
ryannikolaidis, line 21: 🎉

ryannikolaidis (approved):
thanks for the quick change on with the sudo command. lgtm!

PR #1252, opened 2023-08-30
bug/empty-elements

mallorih:
- This PR adds a function to check if a piece of text only contains a bullet (no text) to prevent creating an empty element. 
- Also fixed a test that had a typo.

**Testing**
```Python
elements = partition_text('example-docs/fake-email.txt')

for ele in elements:
    print(ele)

```


unstructured/partition/text.py
```diff
@@ -247,7 +247,7 @@ def partition_text(
     for ctext in file_content:
         ctext = ctext.strip()
 
-        if ctext:
+        if ctext and clean_bullets(ctext):
```
badGarnet, line 250: can we then just pass in `clean_bullets(ctext)` to `element_from_text` and skip clean_bullets in the next function? Here we are calling the same regex pattern multiple times.

mallorih: So, I've moved the logic to `elements_from_text` but we can't remove the `clean_bullets` logic or maybe I'm misunderstood your suggestion.

unstructured/partition/text.py
```diff
@@ -261,11 +261,13 @@ def element_from_text(
...
     if is_bulleted_text(text):
-        return ListItem(
-            text=clean_bullets(text),
-            coordinates=coordinates,
-            coordinate_system=coordinate_system,
-        )
+        clean_text = clean_bullets(text)
+        if clean_text:
+            return ListItem(
+                text=clean_bullets(clean_text),
```
cragwolfe, line 267:
```suggestion
                text=clean_text,
```
don't need to `clean_text` again?

mallorih: Yep that's what I meant! Thanks

cragwolfe (commented):
Just needs a test? otherwise LGTM

unstructured/partition/text.py
```diff
@@ -261,11 +261,13 @@ def element_from_text(
...
-            coordinates=coordinates,
-            coordinate_system=coordinate_system,
-        )
+        clean_text = clean_bullets(text)
+        if clean_text:
+            return ListItem(
+                text=clean_text,
+                coordinates=coordinates,
+                coordinate_system=coordinate_system,
+            )
```
qued, line 270: If the `if` statement fails, there's no fallback, so this function returns `None`.

qued (changes requested):
I added a test that exposed an issue with the new logic.

CHANGELOG.md
```diff
@@ -46,6 +47,7 @@
 * **Fix a bug in Table partitioning** Previously the `skip_infer_table_types` variable used in `partition` was not being passed down to specific file partitioners. Now you can utilize the `skip_infer_table_types` list variable when calling `partition` to specify the filetypes for which you want to skip table extraction, or the `infer_table_structure` boolean variable on the file specific partitioning function.
 * **Fix partition docx without sections** Some docx files, like those from teams output, do not contain sections and it would produce no results because the code assumes all components are in sections. Now if no sections is detected from a document we iterate through the paragraphs and return contents found in the paragraphs.
 
+>>>>>>> 135aa65906109c06540cffc5c8e0815f5e9f2a1c
```
qued, line 50: Remove this please.

CHANGELOG.md
```diff
@@ -370,6 +372,7 @@ allowing the document to be loaded. Fix: Change parent class for Formula to Text
 ### Fixes
 
 * Fix a bug where `xy-cut` sorting attemps to sort elements without valid coordinates; now xy cut sorting only works when **all** elements have valid coordinates
+>>>>>>> a2af72bb79d94941f867d93dfeaed9c4a5d510f7
```
qued, line 375: Remove this please.

unstructured/partition/text.py
```diff
@@ -157,6 +157,10 @@ def combine_paragraphs_less_than_min(
     return combined_paras
 
 
+def has_text(text: str):
```
qued, line 160: Could we call this something slightly more descriptive like maybe `is_not_empty_bullet`? Also needs a short docstring to clarify the purpose, and a type hint for the output.

qued (commented):
A few minor changes needed, but the logic works for me. Ping me to approve once the changes are made.

unstructured/partition/text.py
qued: My reasoning for the name change is something like this is counterintuitive:
image

qued (approved):
LGTM!

PR #1255, opened 2023-08-31
revert pdf changes and add new pdf for empty page testing

badGarnet:
- revert the layout parser fast pdf file to original with just two pages
- add a new file that has one empty page and one page says "this page is intentionally left blank" for tests


qued (approved):
LGTM!

cragwolfe (changes requested):
since most of these tests are mocks, can we go back with the old `example-docs/layout-parser-paper-fast.pdf` in the fixtures?
(only change any test that needs to be changed)

badGarnet:
> since most of these tests are mocks, can we go back with the old `example-docs/layout-parser-paper-fast.pdf` in the fixtures? (only change any test that needs to be changed)

I would argue all the partition tests should be testing against a doc with potential blank page. The existing tests are testing different parameters for partitioning. Alternatively we should probably give a variety of documents (the empty page is just one edge case) to test for each partition setup as another parametrized fixture.

cragwolfe:
ok, i wasn't sure all of these tests were doing real partitioning vs. testing the flow of the code.
assuming it is doing real partitioning, the -empty- files seems like it is more real (ignoring empty) pages than it needs to be which will add significant time to CI vs the existing -fast.pdf file (which isn't processed super "fast," to be fair).

badGarnet:
> ok, i wasn't sure all of these tests were doing real partitioning vs. testing the flow of the code. assuming it is doing real partitioning, the -empty- files seems like it is more real (ignoring empty) pages than it needs to be which will add significant time to CI vs the existing -fast.pdf file (which isn't processed super "fast," to be fair).

is it adding a lot though? Looking at successful action runtime it seem to have barely changed compared to existing ones. Also this change adds one blank page and one page with one line of text. If this do actually add a lot of time we do have something to worry about.
With that said, yes, running tesseract on a single page is adding a few seconds on linux so for those tests we are adding a few seconds to each test. 
The new setup of one real blank and one with one line (that also allows us to test difference between fast and other strategies) is justifiable since it adds two most basic edge case testing scenarios without adding more. Now I do think we could do better with the testing by saving the results from `tersseract` calls for later use on the same page (which probably can be a feature even...)

cragwolfe:
ok, i was thrown off by:
image
looking much larger. i thought you were taking `layout-parser-paper.pdf` and adding pages which would have been much slower.
still not crazy about extra seconds when one test should be sufficient to show we are handling blank pages correctly. 🤷 these tests show we are not crashing (which is fine), but since we have the extra test file, we might as well assert the `metatdata.page_number`s we are getting back are what's expected.

badGarnet:
> ok, i was thrown off by: image looking much larger. i thought you were taking `layout-parser-paper.pdf` and adding pages which would have been much slower. still not crazy about extra seconds when one test should be sufficient to show we are handling blank pages correctly. 🤷 these tests show we are not crashing (which is fine), but since we have the extra test file, we might as well assert the `metatdata.page_number`s we are getting back are what's expected.

that's fair; let me just use the empty page one for that one test for now

cragwolfe (approved):
🙏 🙏

PR #1259, opened 2023-08-31
Fix bbox coordinates for ocr_only strategy

### LINKED ISSUE #1227: bug/bbox for pdf gives out nan
When I tried to extract the bbox for a pdf file `layout-parser-paper-fast.pdf` with `ocr_only` strategy, it gives out `nan` values.

```
elements[20].metadata.coordinates
CoordinatesMetadata(points=((nan, nan), (nan, 2200.0), (0.0, 2200.0), (0.0, nan)), 
    system=<unstructured.documents.coordinates.PixelSpace object at 0x2c7b94dc0>)
```

For additional info, the text value of that element is `'2 Z. Shen et al.'`

Coniferish:
### Summary
Closes #1227, which found that `nan` values were present in the coordinates being generated for some elements.
This breaks logic out from `add_pytesseract_bbox_to_elements` to new functions `_get_element_box` and `convert_multiple_coordinates_to_new_system`. It also updates the logic to check that the current bounding box matches the first character of the element's text (as to avoid the `~` characters that `pytesseract.image_to_boxes` includes, but are not present in `pytesseract.image_to_string`.

### Testing
```
from PIL import Image, ImageDraw

filename="example-docs/layout-parser-paper-with-table.jpg"
elements = partition_image(filename=filename, strategy="ocr_only")
image = Image.open(filename)
draw = ImageDraw.Draw(image)
for i, element in enumerate(elements):
    print(i, element.metadata.coordinates)
    if element.metadata.coordinates:
        draw.polygon(element.metadata.coordinates.points, outline="red", width=2)
output = "example-docs/box-layout-parser-paper-with-table.jpg"
image.save(output)
image.close()
```


unstructured/partition/pdf.py
```diff
@@ -464,7 +464,52 @@ def convert_pdf_to_images(
...
+        if char not in element.text:
+            char_count += 1
+            continue
+        box_text = box_text + char
+
+        x1, y1, x2, y2 = map(int, [_x1, _y1, _x2, _y2])
+        min_x = min(min_x, x1)
+        min_y = min(min_y, y1)
+        max_x = max(max_x, x2)
+        max_y = max(max_y, y2)
```
badGarnet, line 500: would this be better if we get word level bounding boxes so it can save us operations? `pytesseract.image_to_data` returns a table with bounding boxes for each "word" (spans bounded by white space). What's more the return can be a pd dataframe so we can do vector match with the table instead of running this in a loop with string operations, type casing, and comparison operations in python (which is more costly than in c++ where tesseract runs)

Coniferish: That sounds like a great idea. Didn't realize pytesseract would return words

unstructured/documents/coordinates.py
```diff
@@ -68,6 +68,19 @@ def convert_coordinates_to_new_system(
         rel_x, rel_y = self.convert_to_relative(x, y)
         return new_system.convert_from_relative(rel_x, rel_y)
 
+    def convert_multiple_coordinates_to_new_system(
```
badGarnet, line 71: nit (aka as a followup ticket): this multiple conversion is just calling single conversion in serial. Looking at the code it seems we might be able to vectorize the operation to improve performance (memory and speed actually; you'd be surprised how much more memory python list takes vs. numpy array).

Coniferish: I'll create a ticket for this then.

unstructured/partition/pdf.py
Coniferish: Should this also be a follow-up issue or should I do this before merging?

unstructured/partition/pdf.py
```diff
@@ -464,7 +464,52 @@ def convert_pdf_to_images(
...
+    max_y = 0
+    box_text = ""
+    for box in boxes[box_idx : box_idx + char_count]:  # noqa
+        char, _x1, _y1, _x2, _y2, _ = box.split()
+
+        # pytesseract cleans some characters from text that still appear in the bounding boxes,
+        # so those bounding boxes should be skipped
+        if char not in element.text:
+            char_count += 1
+            continue
```
badGarnet, line 493: so `char` here might still be not in the `element`. e.g., if the actual text should have been "apple vs. oranges" and `element.text` is `aple vs. oranges`, it is missing the second `p`. Then this logic will still pass because there is still a `p` in the text. More realistically think about a long page it likely has all the most common characters so even if a whole word gets cleaned out the individual characters can still be in `element.txt`. And actually even whole word may still fall into this kind of trap.

unstructured/partition/pdf.py
badGarnet: a follow up for sure

unstructured/partition/pdf.py
Coniferish: Yah, I realize it's not the best check. It was from when I was initially trying to figure out what to do with the random '~' characters that were showing up. I think this might actually be handled by the `while` loop I later added to compare the first char of element.text and the bounding box char.

badGarnet: so maybe we can remove this logic then?

unstructured/partition/pdf.py
```diff
@@ -582,7 +620,12 @@ def _partition_pdf_or_image_with_ocr(
...
-            add_pytesseract_bbox_to_elements(_elements, _bboxes, width, height)
+            add_pytesseract_bbox_to_elements(
+                elements=_elements,
+                bboxes=_bboxes,
+                width=width,
+                height=height,
+            )
 
             elements.extend(_elements)
             if include_page_breaks:
```
Coniferish, line 631: @badGarnet Can your # FIXME comment be removed in `_partition_pdf_or_image_with_ocr` with your PR that changed the var names to `_element` and such? Still seeing it in the code and wondering about removing it with this PR

badGarnet: ah; that is regarding to the identical metadata being copied (albeit reference I think) into every element. I think we can remove it for now. I should try to understand its purpose better and propose an alternative if still feel it should be fixed.

unstructured/partition/pdf.py
```diff
@@ -464,7 +464,41 @@ def convert_pdf_to_images(
...
-def add_pytesseract_bbox_to_elements(elements, bboxes, width, height):
+def _get_element_box(
+    boxes: List[str],
+    box_idx: int,
+    char_count: int,
+):
+    """Helper function to get the bounding box of an element.
+
+    Args:
+        element
```
badGarnet, line 475:
```suggestion
```

unstructured/partition/pdf.py
```diff
@@ -464,7 +464,41 @@ def convert_pdf_to_images(
...
+        boxes (List[str], optional): Defaults to [].
+        box_idx (int, optional): Defaults to 0.
+        char_count (int, optional): Defaults to 0.
+    """
+    min_x = float("inf")
+    min_y = float("inf")
+    max_x = 0
+    max_y = 0
+    for box in boxes[box_idx : box_idx + char_count]:  # noqa
```
badGarnet, line 484: why can't we just pass in `boxes[box_idx: box_ids+char_count]`? So this function is simplified to iterate through input boxes.

badGarnet, line 484: why can't we just pass in `boxes[box_idx: box_ids+char_count]`? So this function is simplified to iterate through input boxes.

unstructured/partition/pdf.py
Coniferish: Okay. I'll remove it. It doesn't look to me like there's duplicate metadata being generated, but maybe I'm not looking closely enough.

badGarnet (approved):

unstructured/partition/pdf.py
```diff
@@ -464,21 +464,50 @@ def convert_pdf_to_images(
...
+        x1, y1, x2, y2 = map(int, [_x1, _y1, _x2, _y2])
+        min_x = min(min_x, x1)
+        min_y = min(min_y, y1)
+        max_x = max(max_x, x2)
+        max_y = max(max_y, y2)
+
+    return ((min_x, min_y), (min_x, max_y), (max_x, max_y), (max_x, min_y)), char_count
+
+
+def add_pytesseract_bbox_to_elements(
```
qued, line 493: Should it be `add_pytesseract_bboxes_to_elements`? On the other hand the name `bboxes` is confusing to me for a string, although I'm not familiar with the output of `pytesseract.image_to_boxes`.

qued (changes requested):
Here's some preliminary nits, I'm still trying to understand the changes.

unstructured/partition/pdf.py
Coniferish: Will change the name. "bbox" stands for "bounding box", which I saw used as the var name somewhere while working on this

qued: I understand the abbreviation, but I would expect a variable called `bboxes` to contain a sequence of geometric objects representing bounding boxes... not a string.

Coniferish: Ahh.. okay. Pytesseract returns a giant string from `image_to_boxes`. Open to recommendations for another name, though.

unstructured/partition/pdf.py
```diff
@@ -464,21 +464,50 @@ def convert_pdf_to_images(
             yield image
 
 
-def add_pytesseract_bbox_to_elements(elements, bboxes, width, height):
+def _get_element_box(
+    boxes: List[str],
+    char_count: int,
+) -> Tuple[Tuple[int, int], int]:
```
qued, line 470:
```suggestion
) -> Tuple[Tuple[Tuple[float, float], Tuple[float, int], Tuple[int, int], Tuple[int, float]], int]:
```

unstructured/partition/pdf.py
qued: Ok, how about we just change `bboxes -> bboxes_string`.

unstructured/partition/pdf.py
```diff
@@ -464,21 +464,50 @@ def convert_pdf_to_images(
...
+    bboxes: str,
+    width: int,
+    height: int,
+) -> List[Text]:
     """
     Get the bounding box of each element and add it to element.metadata.coordinates
 
     Args:
         elements: elements containing text detected by pytesseract.image_to_string.
         bboxes (str): The return value of pytesseract.image_to_boxes.
```
qued, line 504:
```suggestion
        bboxes_string (str): The return value of pytesseract.image_to_boxes.
```

qued (approved):
LGTM pending CI passing.

PR #1262, opened 2023-09-01
fix: wrapped error handling for connectors

ryannikolaidis:
The CustomError that we use to wrap custom ingest errors inherits from BaseException rather than Exception (as we should, per specification [here](https://docs.python.org/3/library/exceptions.html#BaseException)). This resulted in exceptions not properly raising as expected. This PR changes the inheritance which resolves the known issue.

Additionally, our base definition for get_file on IngestDoc was wrapped with SourceConnectionError, however this must be explicitly decorating each subclass definition in order to function. This PR does that.

## Testing
Some unit test coverage was added for the error wrapping class, however this wasn't properly recreating the issue we are seeing when running ingest tests.

To recreate that issue one can intentionally raise an exception in the [partition_file](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L214C9-L214C23) definition and then run any ingest test. Prior to this change: the code and logs suggest that everything ran without exception, but the partitioned output was not generated (as a result the test will fail without any clues as to what went wrong). With this update, the expected custom partition error, error message, and stack trace will be visible.


ahmetmeleq:
Is it possible or practical to wrap `get_file` in `BaseIngestDoc` to avoid modifying individual connector files? (except the connectors with unique method names)

ryannikolaidis:
> Is it possible or practical to wrap `get_file` in `BaseIngestDoc` to avoid modifying individual connector files? (except the connectors with unique method names)

that's actually what was originally done, but decorators aren't inherited like that, so actually none of these were getting wrapped. you can test this by adding a `raise` in a connector's `get_file` and see that before this PR it does not get wrapped, but after it does.

ahmetmeleq (approved):
I understand now. Also checked and couldn't find a way to inherit a decorator to an overridden method.

- Reviewed all the code 
- Reproduced obtaining error messages from partition in local succesfully
- I had reproduced not being able to see the error message from partition in the past
- Reproduced generating SourceConnectionError message succesfully

LGTM!

PR #1266, opened 2023-09-01
fix: version-sync

ryannikolaidis:
The check to verify version changes has recently permitted commits with the same release version to merge; the intention is that this should be prevented.

The issue is that the version-sync script's call to get the version on `main` attempts to look at the local main branch. When this doesn't exist the variable MAIN_VERSION is assigned to an empty value rather than a semantic version string. This means that later comparisons are meaningless. [Here](https://github.com/ryannikolaidis/unstructured/actions/runs/6045554705/job/16405814607#step:6:77) is an example of the script "succeeding" with this condition.

Instead, this PR adds logic to fetch origin main and compare against the version value there. It also adds `set -eux` at the top of the script so we can debug issues if they come up in the future and also force this to fail and exit if values are unset (which would have prevented the bug from resulting in success).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
-## 0.10.12-dev0
+## 0.10.12-dev1
 
 ### Enhancements
 
 ### Features
 
 ### Fixes
+
+* Update version-sync to prevent duplicate release versions
 
 ## 0.10.11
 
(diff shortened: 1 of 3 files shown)


cragwolfe (approved):
🙏 🙏 🙏

PR #1268, opened 2023-09-01
Feat: Create a naive hierarchy for elements

newelh:
## **Summary**
By adding hierarchy to unstructured elements, users will have more information for implementing vector db/LLM chunking strategies. For example, text elements could be queried by their preceding title element. The hierarchy is implemented by a parent_id tag in the element's metadata.

### Features
- Introduces a parent_id to ElementMetadata (The id of the parent element, not a pointer)
- Creates a rule set for assigning hierarchies. Sensible default is assigned, with an optional override parameter
- Sets element parent ids if there isn't an existing parent id or matches the ruleset

### How it works

Hierarchies are assigned via a parent id field in element metadata. Elements are read sequentially and evaluated against a ruleset. For example take the following elements:

1. Title, "This is the Title"
2. Text, "this is the text"

And the ruleset: `{"title": ["text"]}`. When evaluated, the parent_id of 2 will be the id of 1.  The algorithm for determining this is more complex and resolves several edge cases, so please read the code for further details.

### Schema Changes

```
@dataclass
class ElementMetadata:
      coordinates: Optional[CoordinatesMetadata] = None
      data_source: Optional[DataSourceMetadata] = None
      filename: Optional[str] = None
      file_directory: Optional[str] = None
      last_modified: Optional[str] = None
      filetype: Optional[str] = None
      attached_to_filename: Optional[str] = None
+     parent_id: Optional[Union[str, uuid.UUID, NoID, UUID]] = None
+     category_depth: Optional[int] = None

...
```

### Testing
```
from typing import List

elements = partition(filename="./unstructured/example-docs/fake-html.html", strategy="auto")

for element in elements:
    print(
        f"Category:  {getattr(element, 'category', '')}\n"\
        f"Text:      {getattr(element, 'text', '')}\n"
        f"ID:        {element.id}\n" \
        f"Parent ID: {element.metadata.parent_id}\n"\
        f"Depth:     {element.metadata.category_depth}\n" \
    )
```

### Additional Notes
Implementing this feature revealed a possibly undesired side-effect in how element metadata are processed. In `unstructured/partition/common.py` the `_add_element_metadata` is invoked as part of the `add_metadata_with_filetype` decorator for filetype partitioning. This method is intended to add additional information to the metadata generated with the element including filename and filetype, however the existing metadata is merged into a newly created metadata object rather than the other way around. Because of the way it's structured, new metadata fields can easily be forgotten and pose debugging challenges to developers. This likely warrants a new issue.

I'm guessing that the implementation is done this way to avoid issues with deserializing elements, but could be wrong.


mallorih:
Will this Rule Set ensure List Items of the same list have the same parent id but a different parent id from other lists in the document? Are we able to use Subheadlines as parent ids? I think I would like to test this on a few sample documents to see the outcome.

newelh:
> Will this Rule Set ensure List Items of the same list have the same parent id but a different parent id from other lists in the document? Are we able to use Subheadlines as parent ids? I think I would like to test this on a few sample documents to see the outcome.

Good questions. I'll write some test cases so we can verify

qued:
The structure I get back from the test code is as follows:
image

This isn't correct, but maybe this is what we get for now with the naive approach?

newelh:
> The structure I get back from the test code is as follows: image
> 
> This isn't correct, but maybe this is what we get for now with the naive approach?

You're right, this is unexpected behavior, let me look into this

benjats07 (approved):
LGTM

PR #1274, opened 2023-09-01
build(deps): Remove pillow pin

qued:
Removed pin for `PIL` as `detectron2` repo has been updated, and so has `unstructured-inference` (see [154](https://github.com/Unstructured-IO/unstructured-inference/pull/154)).

#### Tests:

Any problems should have shown up as a unit test failure in `unstructured-inference` (that tests with `PIL==10.0.0`) so we should be good to go.


benjats07 (approved):
LGTM

PR #1275, opened 2023-09-01
fix: split on break elements in `partition_html`

### LINKED ISSUE #1230: bug/html partition returns all paragraphs as one element
`ideas-page.html` returned text as one element.

ideas-page-view

ideas-page-element-returned

Definition of Done
* ElementType returned each paragraph separately

MthwRobinson:
### Summary

Closes #1230. Updates `partition_html` to split sections separated by break elements such as `<br>` into separate elements.

### Testing

The following splits into three elements on the feature branch and is a single element on `main`.

```python

html_text = "<html><p>I'm a <strong>parrot</strong>.<br>Parrots like to squawk.<br>Parrots like to sing.<p></html>"
```


test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/ideas-page.json
```diff
@@ -22,6 +22,256 @@
...
+        null
+      ],
+      "emphasized_text_contents": [
+        "(Someone fed my essays into GPT to make something that could answer\nquestions based on them, then asked it where good ideas come from.  The\nanswer was ok, but not what I would have said. This is what I would have said.)"
+      ],
+      "emphasized_text_tags": [
+        "i"
+      ]
+    },
+    "text": "questions based on them, then asked it where good ideas come from. The"
```
qued, line 50: This fragmented text seems worse to me. Also the emphasized text contents seems to be derived from the text before it was split.

qued (changes requested):
This seems to break up the text by the placement of the newlines in the raw HTML. Looking at the raw and rendered HTML, I think it would be better to break by the placement of the `<br>` (and other similar) tags.

test_unstructured_ingest/expected-structured-output/box/nested-1/ideas-page.html.json
Klaijan, line 1: These results look suspicious to me. It seems like it breaks when applying `split_by_paragraph` on top of the extracted html text which was not in the intended paragraph format. I also agree that splitting with `<br>` could result in a prettier output :)

test_unstructured_ingest/expected-structured-output/box/nested-1/ideas-page.html.json
```diff
@@ -21,6 +21,246 @@
...
+      "link_urls": [
+        "index.html",
+        "https://twitter.com/stef/status/1617222428727586816"
+      ],
+      "link_texts": [
+        null,
+        null
+      ],
+      "emphasized_text_contents": [
+        "(Someone fed my essays into GPT to make something that could answer\nquestions based on them, then asked it where good ideas come from.  The\nanswer was ok, but not what I would have said. This is what I would have said.)"
```
Klaijan, line 42: These results look suspicious to me. It seems like it breaks when applying `split_by_paragraph` on top of the extracted html text which was not in the intended paragraph format. I also agree that splitting with `<br>` could result in a prettier output :)

test_unstructured_ingest/expected-structured-output/box/nested-1/ideas-page.html.json
```diff
@@ -21,6 +21,246 @@
...
+      ],
+      "emphasized_text_tags": [
+        "i"
+      ]
+    },
+    "text": "questions based on them, then asked it where good ideas come from. The"
```
Klaijan, line 48: These results look suspicious to me. It seems like it breaks when applying `split_by_paragraph` on top of the extracted html text which was not in the intended paragraph format. I also agree that splitting with `<br>` could result in a prettier output :)

qued:
I'm not sure if this is ready for re-review yet, but I tried it out on `ideas-page.html` just now, and elements 1-3 looked like what we expect, but element 0 was all the text repeated a few times.

MthwRobinson:
@qued - Closing in favor of #1314. Found a simpler way to do this.

PR #1279, opened 2023-09-01
Adds data source properties to fsspec-based connectors

rvztz:
aAdds data source properties to fsspec interface for fsspec-based connectors:
-  data_created
-  date_modified
-  version
-  record_locator

Separates the logic between fetching the file from source and `get_file`. Retrieves file metadata when any of the properties are called.


potter-potter:
Nice. This works well for the happy path. I need to check how it works if the document is not at the source. Will check it today. Also I will check what happens if the file is updated.

The above might be easier to do with unit tests and mocking.

unstructured/ingest/connector/fsspec.py
```diff
@@ -111,13 +124,77 @@ def get_file(self):
...
+        self.file_exists = fs.exists(self.remote_file_path)
+        date_created = self._catch_not_implemented(fs.created)
+        date_modified = self._catch_not_implemented(fs.modified)
+        version = str(fs.checksum(self.remote_file_path))
+
+        self.file_metadata = FsspecFileMeta(
+            date_created,
+            date_modified,
+            version,
+        )
```
ryannikolaidis, line 160: instead of this pattern where we check if a thing exists and otherwise call get_file_metadata, what about instead just make `file_metadata` [a cached property](https://docs.python.org/3/library/functools.html#functools.cached_property)?

unstructured/ingest/connector/fsspec.py
```diff
@@ -111,13 +124,77 @@ def get_file(self):
...
         )
         logger.debug(f"Fetching {self} - PID: {os.getpid()}")
-        fs.get(rpath=self.remote_file_path, lpath=self._tmp_download_file().as_posix())
+        try:
+            fs.get(rpath=self.remote_file_path, lpath=self._tmp_download_file().as_posix())
+            self.file_exists = True
+        except Exception as e:
+            if isinstance(e, FileNotFoundError):
+                self.file_exists = False
+            raise
```
ryannikolaidis, line 133: I don't think we really need to be updating file_exists in get_file. by the time the process is calling get_file the parent process has already already checked if file_exists

ryannikolaidis:
let's also define the [source_url](https://github.com/Unstructured-IO/unstructured/blob/95b62953070c9c8bee1e4028ed830f7047ec9f69/unstructured/ingest/interfaces.py#L157C1-L161C1) data source property while we're updating these

unstructured/ingest/connector/fsspec.py
```diff
@@ -111,13 +127,76 @@ def get_file(self):
...
         )
         logger.debug(f"Fetching {self} - PID: {os.getpid()}")
-        fs.get(rpath=self.remote_file_path, lpath=self._tmp_download_file().as_posix())
+        try:
+            fs.get(rpath=self.remote_file_path, lpath=self._tmp_download_file().as_posix())
+        except FileNotFoundError as e:
+            logger.error("Failed to find file in remote source.")
+            logger.error(e)
+
+    def _catch_not_implemented(self, method, **kwargs) -> Optional[str]:
```
ryannikolaidis, line 136: this name is confusing. really what we're doing here is attempting to fetch a date for a file path using an arbitrary method. it happens to account for NotImplementedError. call this something like `_get_formatted_date`?

unstructured/ingest/connector/fsspec.py
```diff
@@ -111,13 +127,76 @@ def get_file(self):
...
         """The filename of the file after downloading from cloud"""
         return self._tmp_download_file()
 
+    @property
+    def date_created(self) -> Optional[str]:
+        return self.file_metadata.date_created  # type: ignore
+
+    @property
+    def date_modified(self) -> Optional[str]:
+        return self.file_metadata.date_created  # type: ignore
```
ryannikolaidis, line 179: why is this date_created? typo?

unstructured/ingest/connector/fsspec.py
```diff
@@ -111,13 +127,76 @@ def get_file(self):
...
+        fs: AbstractFileSystem = get_filesystem_class(self.config.protocol)(
+            **self.config.get_access_kwargs(),
+        )
+        date_created = self._catch_not_implemented(fs.created)
+        date_modified = self._catch_not_implemented(fs.modified)
+        version = str(fs.checksum(self.remote_file_path))
+        source_url = self._catch_not_implemented(
+            fs.sign(self.remote_file_path),
+            expiration=SIGNED_URL_EXPIRATION,
+        )
```
ryannikolaidis, line 158: I'm really confused here. _catch_not_implemented is assuming we're getting a date, right?

unstructured/ingest/connector/fsspec.py
```diff
@@ -111,13 +127,76 @@ def get_file(self):
             **self.config.get_access_kwargs(),
         )
         logger.debug(f"Fetching {self} - PID: {os.getpid()}")
-        fs.get(rpath=self.remote_file_path, lpath=self._tmp_download_file().as_posix())
+        try:
+            fs.get(rpath=self.remote_file_path, lpath=self._tmp_download_file().as_posix())
+        except FileNotFoundError as e:
+            logger.error("Failed to find file in remote source.")
+            logger.error(e)
```
ryannikolaidis, line 134: hmm, what is the value of catching here? would we not want this to raise? if we can't get the file we definitely don't want to move to the next step of trying to partition the downloaded file.

unstructured/ingest/connector/fsspec.py
ryannikolaidis: The NotImplementedError catching is secondary to that imo.

ryannikolaidis: the only method that doesn't return a date that we're passing through this method (incorrectly) is the fs.sign. I think in that case we should just put the try/catch inline rather than going through another function

unstructured/ingest/connector/fsspec.py
ryannikolaidis: when you pull in from latest you'll find that we added a wrapper this so that errors raised in get_file will be wrapped as a [SourceConnectionError](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/error.py#L28) with the wrapping decorator like[ this](https://github.com/Unstructured-IO/unstructured/blob/92692ad8d7d5001601dd88fef869a29660f492cb/unstructured/ingest/connector/fsspec.py#L105).

ryannikolaidis: This will mean that all errors raised in this method will get a wrapped with a log `Error in connecting to upstream data source: `

ryannikolaidis: which is why I think we can do nothing here (i.e. don't try/catch just let it raise the exception)

ryannikolaidis: which is why I think we can do nothing here (i.e. don't try/catch just let it raise the exception)

unstructured/ingest/connector/fsspec.py
```diff
@@ -113,11 +129,69 @@ def get_file(self):
...
+
+        date_created = None
+        with suppress(NotImplementedError):
+            date_created = fs.created(self.remote_file_path)
+            date_created = date_created.isoformat()
+
+        date_modified = None
+        with suppress(NotImplementedError):
+            date_modified = fs.modified(self.remote_file_path)
+            date_modified = date_modified.isoformat()
```
ryannikolaidis, line 150:
```suggestion
            date_modified = fs.modified(self.remote_file_path).date_modified.isoformat()
```

unstructured/ingest/connector/fsspec.py
```diff
@@ -113,11 +129,69 @@ def get_file(self):
...
+        from fsspec import AbstractFileSystem, get_filesystem_class
+
+        fs: AbstractFileSystem = get_filesystem_class(self.config.protocol)(
+            **self.config.get_access_kwargs(),
+        )
+
+        date_created = None
+        with suppress(NotImplementedError):
+            date_created = fs.created(self.remote_file_path)
+            date_created = date_created.isoformat()
```
ryannikolaidis, line 145:
```suggestion
            date_created = fs.created(self.remote_file_path).date_created.isoformat()
```

unstructured/ingest/connector/fsspec.py
```diff
@@ -25,6 +30,8 @@
     "dropbox",
 ]
 
+SIGNED_URL_EXPIRATION = 300
```
ryannikolaidis, line 33: still needed?

ryannikolaidis (approved):
LGTM!

ryannikolaidis (changes requested):
let's make sure we have steps (even if manual) to validate that access of the documents doesn't affect the checksum --> version.

ryannikolaidis (approved):

PR #1283, opened 2023-09-01
Adds data source properties to airtable, confluence and discord

rvztz:
Addresses #1316.

Adds data source properties to airtable, confluence and discord:
-  data_created
-  date_modified
-  version
-  record_locator

Separates the logic between fetching the file from source and `get_file`. Retrieves file metadata when any of the properties are called.

Adds logic to check if file exists in the remote source. For connectors that don't directly support it, adds exception handling to check any issues while retrieving the file.


unstructured/ingest/connector/confluence.py
```diff
@@ -100,22 +105,87 @@ def _output_filename(self):
...
         # have a separate connection object for each process
-        confluence = Confluence(
-            self.config.url,
-            username=self.config.user_email,
-            password=self.config.api_token,
-        )
 
-        result = confluence.get_page_by_id(page_id=self.file_meta.document_id, expand="body.view")
+        result = self._get_page()
+        self.get_file_metadata(result)
```
ryannikolaidis, line 188: why is `get_file_metadata` called here?

unstructured/ingest/connector/airtable.py
```diff
@@ -49,39 +60,117 @@ class AirtableIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
...
+            return None, None
+
+        if len(rows) == 0:
+            logger.info("Empty document, retrieved table but it has no rows.")
+        return rows, table_url
+
+    @cached_property
+    def file_metadata(self):
+        """Gets file metadata from the current table."""
+        rows, table_url = self._get_table_rows()
```
rbiseck3, line 127: Not sure how `@cached_property` or if the underlying variable used to cache it available outside this method, but since you're using `self._get_table_rows()` to derive the values, would be nice for the cached value to be set in that method if it's called before this one so that we don't have to make that call twice. 

Also general question, values such as `date_created` shouldn't be changing, but `date_modified` might, so I'm not sure if we would want to avoid caching that value. @ryannikolaidis, not how the intention of this property might impact how we get this value.

ryannikolaidis: @rbiseck3 [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) only caches the property for the lifetime of that IngestDoc instance. I don't think we need to be concerned about changes that occur during the duration of processing a document, with respect to these properties.

ryannikolaidis: ideally we don't want separate get calls for each piece of data here since in most cases a single `get` is going to give us most or all of this and it's probably unlikely that they'll change over the span of processing a document and should be able to query the value at any time, but to your point (@rbiseck3 ) it seems like the most technically accurate way to do this would be to explicitly set (or overwrite) the mutable values _at the time_ of fetching the doc to insure that we write the correct metadata of the raw document when fetched.

ryannikolaidis: any thoughts?

unstructured/ingest/connector/confluence.py
```diff
@@ -92,37 +94,88 @@ def filename(self):
...
-        result = confluence.get_page_by_id(page_id=self.file_meta.document_id, expand="body.view")
+        result = self._get_page()
+        if result is None:
+            raise ValueError(f"Failed to retrieve page with ID {self.document_meta.document_id}")
+
         self.document = result["body"]["view"]["value"]
         self.filename.parent.mkdir(parents=True, exist_ok=True)
         with open(self.filename, "w", encoding="utf8") as f:
             f.write(self.document)
+        self.set_source_metadata(page=result)
```
ryannikolaidis, line 178: maybe just move this up directly after we get the result (since this should be as close as possible to the fetch of the doc/content itself)

unstructured/ingest/connector/confluence.py
```diff
@@ -92,37 +94,88 @@ def filename(self):
...
         """Create output file path based on output directory, space id and document id."""
-        output_file = f"{self.file_meta.document_id}.json"
-        return Path(self.partition_config.output_dir) / self.file_meta.space_id / output_file
+        output_file = f"{self.document_meta.document_id}.json"
+        return Path(self.partition_config.output_dir) / self.document_meta.space_id / output_file
+
+    @property
+    def record_locator(self) -> t.Optional[t.Dict[str, t.Any]]:
+        return {
+            "page_id": self.document_meta.document_id,
```
ryannikolaidis, line 110: record_locator should have everything needed to programmatically get back to that source document. assuming we need more than just page_id here. some root domain to make sure we point to the right space?

unstructured/ingest/connector/discord.py
```diff
@@ -92,28 +87,68 @@ async def on_ready():
...
+        self._create_full_tmp_dir_path()
+        if self.connector_config.verbose:
+            logger.debug(f"fetching {self} - PID: {os.getpid()}")
 
+        messages, jump_url = self._get_messages()
+        if messages == []:
+            raise ValueError(f"Failed to retrieve messages from Discord channel {self.channel}")
         self._tmp_download_file().parent.mkdir(parents=True, exist_ok=True)
+        # self._output_filename.parent.mkdir(parents=True, exist_ok=True)
+
```
ryannikolaidis, line 131: okay to remove?

unstructured/ingest/connector/discord.py
```diff
@@ -92,28 +87,68 @@ async def on_ready():
...
         with open(self._tmp_download_file(), "w") as f:
             for m in messages:
                 f.write(m.content + "\n")
+        self.update_source_metadata(messages=messages, jump_url=jump_url)
```
ryannikolaidis, line 135: move this after the _get_messages (and error check)

unstructured/ingest/connector/confluence.py
```diff
@@ -92,37 +94,89 @@ def filename(self):
...
-        result = confluence.get_page_by_id(page_id=self.file_meta.document_id, expand="body.view")
+        result = self._get_page()
+        if result is None:
+            raise ValueError(f"Failed to retrieve page with ID {self.document_meta.document_id}")
+
         self.document = result["body"]["view"]["value"]
         self.filename.parent.mkdir(parents=True, exist_ok=True)
         with open(self.filename, "w", encoding="utf8") as f:
             f.write(self.document)
+        self.update_source_metadata(page=result)
```
ryannikolaidis, line 179: move to  just after _get_page() (and the error)

unstructured/ingest/interfaces.py
```diff
@@ -160,6 +180,10 @@ def wrapper(self, *args, **kwargs):
 
         return wrapper
 
+    def update_source_metadata(self, **kwargs) -> None:
+        """Sets the SourceMetadata and the  properties for the doc"""
+        pass
```
ryannikolaidis, line 185: double checking: this is still in progress?

ryannikolaidis: ohh woops, this was my bad, I misread this as being a subclass implementation of it. yep, agreed. maybe TODO: make this raise after we implement in existing connectors?

unstructured/ingest/connector/airtable.py
```diff
@@ -48,39 +50,101 @@ class AirtableIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
...
+        try:
+            table = api.table(self.table_meta.base_id, self.table_meta.table_id)
+            table_url = table.url
+            rows = table.all(
+                view=self.table_meta.view_id,
+            )
+        except Exception as e:
+            # TODO: more specific error handling?
+            logger.error(e)
+            return None, None
```
ryannikolaidis, line 96: errrm why are we swallowing errors here? seems like if we can't pull the table this _should_ fail, right?

ryannikolaidis: per conversation, let's narrow this to only catch the error if it doesn't exist on remote (i.e. not if something like a connection issue occurs)

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
...
 * **Better debug output related to sentence counting internals**. Clarify message when sentence is not counted toward sentence count because there aren't enough words, relevant for developers focused on `unstructured`s NLP internals.
 * **Faster ocr_only speed for partitioning PDF and images.** Use `unstructured_pytesseract.run_and_get_multiple_output` function to reduce the number of calls to `tesseract` by half when partitioning pdf or image with `tesseract`
+* **Adds SourceMetadata to base classes and `update_source_metadata` to` base interface.**
+* **Adds data source properties (date_created, date_modified, version, exists, source_url, record_locator) to Airtable, Confluence and Discord connectors.** Implements `update_source_metadata` function to fetch properties from files/objects on Airtable, Confluence and Discord connectors.
 * **Adds data source properties to fsspec connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive.
```
ryannikolaidis, line 16:
```suggestion
* **Adds data source properties to connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive.
```

ryannikolaidis (approved):

ryannikolaidis (approved):
lgtm!

ryannikolaidis (changes requested):
include Airtable, Confluence, and Discord in changelog reference to data property additions

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 0.10.16-dev1
```
ryannikolaidis, line 1:
```suggestion
## 0.10.16-dev0
```

CHANGELOG.md
```diff
@@ -21,7 +32,7 @@
...
 * **Faster ocr_only speed for partitioning PDF and images.** Use `unstructured_pytesseract.run_and_get_multiple_output` function to reduce the number of calls to `tesseract` by half when partitioning pdf or image with `tesseract`
-* **Adds data source properties to fsspec connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
+* **Adds data source properties to fsspec connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc. Connectors implementing these changes: Airtable, Confluence, Discord
```
ryannikolaidis, line 35: no, don't update the changelog for for previous versions

ryannikolaidis (approved):

unstructured/ingest/connector/confluence.py
```diff
@@ -92,33 +94,85 @@ def filename(self):
...
+            self.source_metadata = SourceMetadata(
+                exists=False,
+            )
+            return
+        document_history = page["history"]
+        date_created = datetime.strptime(
+            document_history["createdDate"],
+            "%Y-%m-%dT%H:%M:%S.%fZ",
+        ).isoformat()
+        if last_updated := document_history.get("lastUpdated", "").get("when", ""):
```
ahmetmeleq, line 147: This seems like it'll throw an `AttributeError: 'str' object has no attribute 'get'` when there's no `lastUpdated` key in `document_history` (since the default value for the first get is an empty string)

ahmetmeleq:
```suggestion
        if last_updated := document_history.get("lastUpdated", {}).get("when", ""):
```

ahmetmeleq:
Just one comment, everything else looks great!

PR #1293, opened 2023-09-03
build(release): bump to unstructured-inference==0.5.21

cragwolfe:


badGarnet (approved):

christinestraub (changes requested):
The `requirements/constraints.in` doesn't seem to have changed.
- https://github.com/Unstructured-IO/unstructured/blob/crag/bump-uns-inf-0.5.21/requirements/constraints.in#L26
- https://github.com/Unstructured-IO/unstructured/blob/crag/bump-uns-inf-0.5.21/requirements/extra-pdf-image.txt#L210

badGarnet:
> The `requirements/constraints.in` doesn't seem to have changed.
> 
> * https://github.com/Unstructured-IO/unstructured/blob/crag/bump-uns-inf-0.5.21/requirements/constraints.in#L26
> * https://github.com/Unstructured-IO/unstructured/blob/crag/bump-uns-inf-0.5.21/requirements/extra-pdf-image.txt#L210

good catch; created an issue to maybe create an automated way to ensure dependencies are correct

cragwolfe:
> The requirements/constraints.in doesn't seem to have changed.

🤦 . thanks for catching that!

> good catch; created an issue https://github.com/Unstructured-IO/unstructured/issues/1294 to maybe create an automated way to ensure dependencies are correct

there is `make check-deps` which is in CI "to ensure dependencies are correct." in this case, i forgot to update a constraints.in file the first place. that's just pure human error vs. what i claimed in the PR. don't think we needed automated way to prevent that.

PR #1295, opened 2023-09-04
add makefile to requirements

### LINKED ISSUE #1294: feat/pip-compile-make-command
Currently when bumping dependencies it is up to the developer to ensure all `.txt` requirement files are updated as needed and this can be error prune.

An automated `make pip-compile` make command that:
- sets the dependencies between different `.in` files
- provides command to run to compile affected `.txt` files (via the dependencies)
- also set up a CI step that runs pip-compile when changes in `requirements` folder is detected

Currently way is the alternative: human checking and it is error prune.

**Additional context**

example of makefile

```
PIP = pip-compile -q -o $@ $<

extra-csv.txt: extra-csv.in constraints.in base.txt
	$(PIP)

dev.txt: dev.in constraints.in base.txt test.txt
	$(PIP)

test.txt: test.in base.txt
	$(PIP)

base.txt: base.in constraints.in
	$(PIP)
```

badGarnet:
This PR resolves #1294 by adding a Makefile to compile requirements. This makefile respects the dependencies between file and will compile them in order. E.g., extra-*.txt will be compiled __after__ base.txt is updated.

Test locally by simply running `make pip-compile` or `cd requirements && make clean && make all`


cragwolfe:
why this is better than the existing solution? in the off chance that the existing `make pip-compile` results in inconsistent dependencies, this would get caught by https://github.com/Unstructured-IO/unstructured/blob/b710baf/.github/workflows/ci.yml#L70 .

but as long as pins are being kept in `constraints.in`, this really shouldn't happen in the first place.

badGarnet:
> why this is better than the existing solution? in the off chance that the existing `make pip-compile` results in inconsistent dependencies, this would get caught by https://github.com/Unstructured-IO/unstructured/blob/b710baf/.github/workflows/ci.yml#L70 .
> 
> but as long as pins are being kept in `constraints.in`, this really shouldn't happen in the first place.

The current method relies on `base.txt` is compiled first because of alphabetical order. The other extras that depend on `base.txt` are able to get updated `base.txt` because of naming. Additionally if we have more layered dependencies like `dev.txt` we may not be able to rely on the file names.

Makefile is designed to handle compilation dependency so I would argue that is the safest way to ensure pip-compile runs correctly. Now I do concede that with current requirements this may seem a bit over engineering.

badGarnet:
probably not needed for now

rbiseck3:
Got this:
```shell
❯ make pip-compile
scripts/pip-compile.sh: line 11: MAKE[1]:: command not found
cp: requirements/build.txt: No such file or directory
make: *** [pip-compile] Error 1
```

scripts/pip-compile.sh
```diff
@@ -8,5 +8,5 @@ if ! python -c "import sys; assert sys.version_info.major == $major and sys.vers
   exit 1
 fi
 
-cd ./requirements && $(MAKE) *.txt
+cd ./requirements && make *.txt && cd ..
```
rbiseck3: Can use `pushd`/`popd` to navigate to/from the requirements dir. Has some pleasantries: [Navigating with popd](https://opensource.com/article/19/8/navigating-bash-shell-pushd-popd#navigating-with-popd)

qued:
Also there used to be logic to copy the `build.txt` file in `requirements/` into the `docs/` folder overwriting `docs/requirements.txt`. In other words, the file `build.txt` is really the requirements file for `sphinx` builds. I think that logic is missing from what we are currently doing as well.

badGarnet:
> Also there used to be logic to copy the `build.txt` file in `requirements/` into the `docs/` folder overwriting `docs/requirements.txt`. In other words, the file `build.txt` is really the requirements file for `sphinx` builds. I think that logic is missing from what we are currently doing as well.

updated makefile to work on subdirectories and added back the copy line

qued (approved):
Overall this LGTM!

Two other non-blocking nice-to-have things I'd flag are:
1. It looks like the formatting of the output files is slightly different (see attached image) so maybe we want to do a full re-compile of everything to put it in the new format? image
2. Speaking of doing a full re-compile, I'm not totally sure how it's being determined when files need to be updated, but sometimes they're skipped (see image). I'm not sure it can make the determination that nothing needs to be done when new versions could have been released to pypi. But even assuming the logic is fine, it would be nice to have a way to force a compile on everything. 
image

badGarnet:
> Overall this LGTM!
> 
> Two other non-blocking nice-to-have things I'd flag are:
> 
> 1. It looks like the formatting of the output files is slightly different (see attached image) so maybe we want to do a full re-compile of everything to put it in the new format? image
> 2. Speaking of doing a full re-compile, I'm not totally sure how it's being determined when files need to be updated, but sometimes they're skipped (see image). I'm not sure it can make the determination that nothing needs to be done when new versions could have been released to pypi. But even assuming the logic is fine, it would be nice to have a way to force a compile on everything.
> 
> image

1. this is due to where the pip compile command is executed; using the make file it is executed inside `requirements` folder
2. I can add a `make clean` to remove `.txt` then in script we can do `make clean && make all` to force recompile everything.
cc @rbiseck3 @ryannikolaidis

Issue #1185, opened 2023-08-23
feat/min and max partition made available to all partitions


ron-unstructured:
**Is your feature request related to a problem? Please describe.**
Tried to use partition_html to download and chunk long webpage. However, min_partition and max_partition isn't an option.

**Describe the solution you'd like**
Right now, `min_partition` and `max_partition` are only available for XML, Text, PDF, Msg, Email partitions. These args can be made available for all `partition_` calls.

**Describe alternatives you've considered**
(1) Chunk the long, extracted texts from `partition_html` by element type, and (2) pass the texts from `partition_html` to partition_text to chunk the texts with min and max partitions. 

**Additional context**
Use decorator in `partition_` calls

Coniferish:
@qued and @cragwolfe, for this issue it's looking like handling coordinates properly will be a challenge. Would it be acceptable to have two elements with the same coordinates if they originate from a single element that's broken into two because of `max_partition`? I'm not as familiar with all the requirements around coordinates.

Coniferish:
Great. I haven't touched the unstructured-api repo yet, but do you want me to create an issue there? I can work on adding the `chuncking_strategy` parameter to the partition functions here, though

Coniferish:
@cragwolfe Great. I haven't touched the unstructured-api repo yet, but do you want me to create an issue there? I can work on adding the `chuncking_strategy` parameter to the partition functions here, though

orlandounstructured:
Closing due to #1222 ; reopen if still relevant

scanny:
Just a note on this, in case anyone reopens later. In my view, an element (partition) is a _naturally bounded_ object, generally those bounds being the extent of the paragraph it came from.

Any _synthetic_ sizing should be performed by _chunking_ elements, which can divide large elements using text-splitting and combine small elements into larger ones if desired.

Maybe that's the conclusion the conversation around this ticket came to, just wanted to argue against adding size arguments to partitioners.

PR #1304, opened 2023-09-05
chunk_by_title decorator

Coniferish:
### Summary
Partial solution to #1185.
Related to #1222.
Creates decorator from `chunk_by_title` cleaning brick.
Breaks a document into sections based on the presence of Title elements. Also starts a new section under the following conditions:

- If metadata changes, indicating a change in section or page or a switch to processing attachments. If `multipage_sections=True`, sections can span pages. `multipage_sections` defaults to True.
- If the length of the section exceeds `new_after_n_chars` characters. The default is 1500. The **chunking function does not split individual elements**, so it's possible for a section to exceed that threshold if an individual element if over `new_after_n_chars characters`, which could occur with a long NarrativeText element.

Combines sections under these conditions
- Sections under `combine_under_n_chars` characters are combined. The default is 500.

### Testing
```

url = "https://understandingwar.org/backgrounder/russian-offensive-campaign-assessment-august-27-2023-0"
chunks = partition_html(url=url, chunking_strategy="by_title")

for chunk in chunks:
    print(chunk)
    print("\n\n" + "-"*80)
    input()
```


Coniferish:
@MthwRobinson and/or @cragwolfe
I still need to add tests, but wanted to check in to make sure this is looking good so far.

cragwolfe:
> @MthwRobinson and/or @cragwolfe
I still need to add tests, but wanted to check in to make sure this is looking good so far.

the pattern generally looks good to me. i'll observe, though, that we have a lot of metadata fields floating around now (somehave more doc types that others, e.g. "emphasized_text" and "href's"), and merging will be non-trivial.
what should happen when `page_number` is multiple pages?

Coniferish:
> emphasized_text

@cragwolfe  Yah, I've noticed that as well and that's actually part of the reason why I put the docstring info for this in the decorator. 

I'm not clear on your question about `page_number`. Are you asking what if multiple sections have the same `page_number` or when a section spans multiple pages? I admittedly didn't think about that since I was just relying on Matt's function. I can investigate, though

cragwolfe:
@Coniferish , please add a test for `partition()` from `unstructured.partition.auto` as well.

unstructured/chunking/title.py
```diff
@@ -33,6 +35,17 @@ def chunk_by_title(
...
+    if (
+        combine_under_n_chars is not None
+        and new_after_n_chars is not None
+        and (
+            combine_under_n_chars > new_after_n_chars
+            or combine_under_n_chars < 0
+            or new_after_n_chars < 0
+        )
+    ):
+        raise ValueError("Invalid values for combine_under_n_chars and/or new_after_n_chars.")
```
cragwolfe, line 47: please add a unittest for this

cragwolfe:
Please add unittests for all the partition_ functions for the sake of testing completeness (even though logically, "should be fine").

cragwolfe:
> > the pattern generally looks good to me. i'll observe, though, that we have a lot of metadata fields floating around now (somehave more doc types that others, e.g. "emphasized_text" and "href's"), and merging will be non-trivial. what should happen when `page_number` is multiple pages?
> 
> @cragwolfe Yah, I've noticed that as well and that's actually part of the reason why I put the docstring info for this in the decorator. What do you mean by "merging will be non-trivial" though?
> 
> I'm not clear on your question about `page_number`. Are you asking what if multiple sections have the same `page_number` or when a section spans multiple pages? I admittedly didn't think about that since I was just relying on Matt's function. I can investigate, though

we can kick this off to a separate issue -- comprehensive review of how to handle metadata merges when chunking.

cragwolfe (approved):

PR #1308, opened 2023-09-06
feat: improved chipper elements mapping and new category_depth metadata

LaverdeS:
Two changes: 
1. Improved mapping of `chipper` element types `Headline` (to `Title`), `Subheadline`(to `Title`)  and `Abstract`( to `NarrativeText`.
2. New element metadata `category_depth`: `None` unless is `Headline` (`category_depth=1`), or `Subheadline` (`category_depth=2`). The update of `category_depth` happens during  the transform `normalize_layout_element`.

Test have been added to confirm `Headline`s and `Subheadline`s as `Title`s with correspondent `category_depth` metadata value.


unstructured/documents/elements.py
```diff
@@ -167,6 +167,9 @@ class ElementMetadata:
     # Metadata extracted via regex
     regex_metadata: Optional[Dict[str, List[RegexMetadata]]] = None
 
+    # Title depth to map chipper elements Headline and Subheadline
+    title_depth: Optional[int] = None
```
badGarnet, line 171: design wise it seems we are throwing everything into the `ElementMetaData` and results in a lot of `Optional` fields. This can be cumbersome down the road when this list becomes unwieldily long. Maybe we should consider refactor this into a base class and mixins/inherited classes for things like `TitleMetaData`. The optional also creates uncertainty in downstream process when we are not guaranteed an input `ElementMetaData` will have a certain attribute for processing and we'd have to check if a field is `None` potentially a lot. E.g., a downstream process function on metadata would always need an accompanying parent `Element`'s category and knowledge from dev what category has which attributes.

unstructured/partition/common.py
```diff
@@ -104,11 +104,17 @@ def normalize_layout_element(
...
+        _element_class = _element_class(
             text=text,
             coordinates=coordinates,
             coordinate_system=coordinate_system,
         )
+        if element_type in ["Title", "Headline", "Subheadline"]:
+            title_depth = 0
+            title_depth = 1 if element_type == "Headline" else title_depth
+            title_depth = 2 if element_type == "Subheadline" else title_depth
+            _element_class.metadata.title_depth = title_depth
```
badGarnet, line 116:
```suggestion
        if element_type == "Title":
            _element_class.metadata.title_depth = 0
        elif element_type == "Headline":
            _element_class.metadata.title_depth = 1
        elif element_type == "Subheadline":
            _element_class.metadata.title_depth = 2
```

unstructured/documents/elements.py
LaverdeS: makes sense, in that case we would have to refactor several `Optional` metadata fields in `ElementMetadata` and consider a similar logic + grouping. Totally in favour of having a shorter list of Optional fields if any.

badGarnet (commented):
maybe a test please? 😬

benjats07 (approved):
LGTM

benjats07:
> maybe a test please? 😬

The ingest-test result could be useful, I created a [this PR](https://github.com/Unstructured-IO/unstructured-inference/pull/195) to execute it with this PR and chipper model as default (Just for getting the results, Chipper still in beta)

unstructured/documents/elements.py
newelh: Other elements may benefit from having a depth such as bullet/list items. Could we rename this parameter to simply `depth` to enable it to be used across element types? Also, I think a sensible default would be 0, as in no depth

newelh: After discussion, `category_depth` will better represent the intention of the field -- the depth relative to the element's category (e.g. Title, Bullet, or List) not the depth relative to all elements

qued (commented):
Echoing other discussions, I'm fine with the changes pending:
1. A unit test to verify the new structure
2. Changing the name of `title_depth` per the discussion from the stand up

qued (changes requested):
See above.

qued (approved):
LGTM!

PR #1309, opened 2023-09-06
Klaijan/improve is_possible_narrative_text and is_possible_title

Klaijan:
Improve `ElementType` detection by adding more rules to the `NarrativeText` and `Title` determining functions.

**`is_possible_narrative_text()`**

Original calculation:
```
    ratio = capitalized / len(tokens)
```

New calculation:

- Ignore a word in text when the first word of the detected line is already capitalized
- Use nltk's `ne_chunk` to help ignore words with name-entity where it must be capitalized (e.g. person, organization, place, etc)

```
    unrequisite_capitalized = sum(
        np.logical_and(
            is_alpha,
            np.logical_and(np.logical_xor(is_capitalized, is_ne), np.logical_not(is_ne)),
        ),
    )

    ratio = unrequisite_capitalized / alpha_token
```

**`is_possible_title()`**

- Add condition where title should not ends with period, but can ends in 3 dots

```
    if text.endswith(".") and not text.endswith("..."):
        return False
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
 
 ### Enhancements
 
+* Updated `is_possible_title` detection rule by adding condition that title does not end with period, but can end with 3 dots
+* Updated `exceeds_cap_ratio` calculation by ignoring name-entity and first-word capitalized
 * Updated documentation: Added back support doc types for partitioning, more Python codes in the API page,  RAG definition, and use case.
 * Updated Hi-Res Metadata: PDFs and Images using Hi-Res strategy now have layout model class probabilities added ot metadata.
 * Updated the `_detect_filetype_from_octet_stream()` function to use libmagic to infer the content type of file when it is not a zip file.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Enhancements
 
+* **Update `is_possible_title` detection rule by adding condition that title does not end with period, but can end with 3 dots.** The function previously captures short sentence, line, or text chunk as title in `.txt` file as `Title`. Adding a rule can avoid a faulty element extraction.
+* **Update `exceeds_cap_ratio` calculation by ignoring name-entity and first-word capitalized.** Capitalized words do not always imply a title text. This added functionality leverages the NER knowledge to help justify a legitimate capitalized words. Currently supperts English language only.
 * Clarify message when sentence is not counted toward sentence count b/c there aren't enough words
 * Adds numbered ListItem grouping when pdfminer broke down by line-by-line using coordinates
 * Use text-based classification hen elements come back uncategorized from PDF/Image partitioning
@@ -40,8 +42,6 @@
 
 ### Enhancements
 
-* Updated `is_possible_title` detection rule by adding condition that title does not end with period, but can end with 3 dots
-* Updated `exceeds_cap_ratio` calculation by ignoring name-entity and first-word capitalized
 * Updated documentation: Added back support doc types for partitioning, more Python codes in the API page,  RAG definition, and use case.
 * Updated Hi-Res Metadata: PDFs and Images using Hi-Res strategy now have layout model class probabilities added ot metadata.
 * Updated the `_detect_filetype_from_octet_stream()` function to use libmagic to infer the content type of file when it is not a zip file.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,27 +4,24 @@
 
 * **Update `is_possible_title` detection rule by adding condition that title does not end with period, but can end with 3 dots.** The function previously captures short sentence, line, or text chunk as title in `.txt` file as `Title`. Adding a rule can avoid a faulty element extraction.
 * **Update `exceeds_cap_ratio` calculation by ignoring name-entity and first-word capitalized.** Capitalized words do not always imply a title text. This added functionality leverages the NER knowledge to help justify a legitimate capitalized words. Currently supperts English language only.
-* Clarify message when sentence is not counted toward sentence count b/c there aren't enough words
-* Adds numbered ListItem grouping when pdfminer broke down by line-by-line using coordinates
-* Use text-based classification hen elements come back uncategorized from PDF/Image partitioning
-* Updated HTML Partitioning to extract tables
-* Create and add `add_chunking_strategy` decorator to partition functions
-* Adds `languages` as an input parameter and marks `ocr_languages` kwarg for deprecation in pdf partitioning functions
-* Adds `xlsx` and `xls` to `skip_infer_table_types` default list in `partition`
-* Adds `languages` as an input parameter and marks `ocr_languages` kwarg for deprecation in image partitioning functions
-* Adds `languages` as an input parameter and marks `ocr_languages` kwarg for deprecation in auto partition
-* Replaces `language` with `languages` as an input parameter to unstructured-partition-text_type functions
-* Removes `UNSTRUCTURED_LANGUAGE` env var. To skip English specific checks, set the `languages` parameter to non-English language(s).
+* **Better ListItem grouping for PDF's (fast strategy).** The `partition_pdf` with `fast` strategy previously broke down some numbered list item lines as separate elements. This enhancement leverages the x,y coordinates and bbox sizes to help decide whether the following chunk of text is a continuation of the immediate previous detected ListItem element or not, and not detect it as its own non-ListItem element.
+* **Fall back to text-based classification for uncategorized Layout elements for Images and PDF's**. Improves element classification by running existing text-based rules on previously UncategorizedText elements
+* **Adds table partitioning for Partitioning for many doc types including: .html, .epub., .md, .rst, .odt, and .msg.** At the core of this change is the .html partition functionality, which is leveraged by the other effected doc types. This impacts many scenarios where `Table` Elements are now propery extracted.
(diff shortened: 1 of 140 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ### Enhancements
 
 * Updated documentation: Added back support doc types for partitioning, more Python codes in the API page,  RAG definition, and use case.
+* Updated `is_possible_title` detection rule
```
badGarnet, line 6: maybe worth writing a bit more on what changes are introduced?

unstructured/partition/text_type.py
```diff
@@ -291,7 +294,17 @@ def exceeds_cap_ratio(text: str, threshold: float = 0.5) -> bool:
...
-    ratio = capitalized / len(tokens)
+    # NOTE(klaijan) - detect the first word in the sentence, whether it is alpha or not
+    is_first_word_capitalized = (text.split()[0]).istitle()
+    ne_chunks = ne_chunk(text)
+    count_ne = sum([hasattr(chunk, "label") * len(chunk) for chunk in ne_chunks])
+
+    if len(tokens) > is_first_word_capitalized + count_ne:
+        ratio = (capitalized - is_first_word_capitalized - count_ne) / (
+            len(tokens) - is_first_word_capitalized - count_ne
+        )
```
badGarnet, line 305: - `is_first_word_capitalized + count_ne` used multiple times so might as well compute once and store in one variable (also can consolidate lines 298-300)
- can we have a test for this case please? The existing tests checks if first word is capitalized but doesn't check the entity capitalized case. 
- careful with edge case where every token has `"label"` -> division by 0

Klaijan: > * `is_first_word_capitalized + count_ne` used multiple times so might as well compute once and store in one variable (also can consolidate lines 298-300)
I thought having them separate is easier to follow the logic or make and adjustments moving forward as they are two different cases. I can consolidate line 299-300 together and combine `ne_chunk` variables into the `count_ne`

> * can we have a test for this case please? The existing tests checks if first word is capitalized but doesn't check the entity capitalized case.
Will do!

> * careful with edge case where every token has `"label"` -> division by 0
The edge case was captured with if-else. If the overall token lengths exceed the subtracted part, then we apply the new calculation, else it uses the same calculation. Allow me to think through this a little bit more whether the original calculation is always the best case to fall to or not.

unstructured/partition/pptx.py
```diff
@@ -125,10 +126,11 @@ def partition_pptx(
                     elements.append(ListItem(text=text, metadata=metadata))
                 elif is_email_address(text):
                     elements.append(EmailAddress(text=text))
+                elif is_possible_title(text) and not page_title_detected:
```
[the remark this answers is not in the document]
Klaijan: Could you please point me to the nltk running in `is_possible_title` please? I might have missed it from my quick scroll. However, `is_possible_narrative_text` calls `exceeds_cap_ratio` that also uses nltk models for name entity and therefore should be equally, if not more, expensive. Additionally, with our current semi-universal `text_type` rule, the order of the rule matters a lot as many of the texts are `True` for both cases. How about we walk this through in more detail?

CHANGELOG.md
Klaijan: On it!

unstructured/partition/text_type.py
Klaijan: Also I noticed that I might have missed the case where NE is also the first word. I'll add on that.

cragwolfe:
Despite the other github CI issues, it looks like there is a real error in the Dockerfile CI test

```
   Resource maxent_ne_chunker not found.
E         Please use the NLTK Downloader to obtain the resource:
E       
E         >>> import nltk
E         >>> nltk.download('maxent_ne_chunker')
E         
E         For more information see: https://www.nltk.org/data.html
```

seems like this should be added to https://github.com/Unstructured-IO/unstructured/blob/09cc4bf/Dockerfile#L60 and
https://github.com/Unstructured-IO/unstructured/blob/09cc4bf/Makefile#L42 ?
btw, what is the size of the added download?

Klaijan:
> Despite the other github CI issues, it looks like there is a real error in the Dockerfile CI test
> 
> ```
>    Resource maxent_ne_chunker not found.
> E         Please use the NLTK Downloader to obtain the resource:
> E       
> E         >>> import nltk
> E         >>> nltk.download('maxent_ne_chunker')
> E         
> E         For more information see: https://www.nltk.org/data.html
> ```
> 
> seems like this should be added to https://github.com/Unstructured-IO/unstructured/blob/09cc4bf/Dockerfile#L60 and https://github.com/Unstructured-IO/unstructured/blob/09cc4bf/Makefile#L42 ? btw, what is the size of the added download?

It is 14 MB

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ### Enhancements
 
 * Updated documentation: Added back support doc types for partitioning, more Python codes in the API page,  RAG definition, and use case.
-* Updated `is_possible_title` detection rule
+* Updated `is_possible_title` detection rule by adding condition that title would not ends with period, but can ends with 3 dots
```
qued, line 6:
```suggestion
* Updated `is_possible_title` detection rule by adding condition that title does not end with period, but can end with 3 dots
```

test_unstructured/partition/test_text_type.py
```diff
@@ -211,12 +211,14 @@ def test_contains_english_word(text, expected, monkeypatch):
...
     [
-        ("Intellectual Property in the United States", True),
+        ("Intellectual Property in the United States", False),
         ("Intellectual property helps incentivize innovation.", False),
         ("THIS IS ALL CAPS. BUT IT IS TWO SENTENCES.", False),
         ("LOOK AT THIS IT IS CAPS BUT NOT A TITLE.", True),
         ("This Has All Caps. It's Weird But Two Sentences", False),
         ("The Business Report is expected within 6 hours of closing", False),
+        ("United States' Intellectual Property", True),
+        ("United States' intellectual property", False),
```
badGarnet, line 221: 🎆 but do have to say this took me a second to think through "yes, this makes sense" 😅

unstructured/partition/text_type.py
```diff
@@ -283,16 +291,33 @@ def exceeds_cap_ratio(text: str, threshold: float = 0.5) -> bool:
...
+
+    ignored_capitalize = sum(np.logical_and(is_alpha, np.logical_or(is_capitalized, is_ne)))
+
+    if sum(is_alpha) > ignored_capitalize:
+        ratio = (sum(np.logical_and(is_alpha, is_capitalized)) - ignored_capitalize) / (
+            sum(is_alpha) - ignored_capitalize
+        )
+    else:
+        ratio = sum(np.logical_and(is_alpha, is_capitalized)) / sum(is_alpha)
+    return bool(ratio > threshold)
```
badGarnet, line 320: devil's advocate (for maybe a followup ticket): since `threshold` is used only in the end as gatecheck maybe we want to refactor this so there is a function that just returns the ratio just in case in the future we may want to use the ratio for other purposes. This also makes the function logically more modular (just computes the ratio of capitalized letters).

unstructured/partition/text_type.py
```diff
@@ -91,6 +93,9 @@ def is_possible_narrative_text(
     return True
 
 
+# add test case for entity capitalized case
+
+
```
badGarnet, line 98: remove?

Klaijan: Missed that one. Thank you!

test_unstructured_ingest/expected-structured-output/pdf-fast-reprocess/s3/small-pdf-set/Silent-Giant-(1).pdf.json
```diff
@@ -680,7 +680,7 @@
     "text": "Despite impressive recent growth, the stark reality is that renewables alone will not be able to resolve our dependence on fossil fuels. Clearly, the sun does not always shine, and the wind does not always blow, and this is compounded by the fact that many times these periods coincide with when electricity demand is at its highest, but renewables can be complementary to nuclear energy. Storage solutions, such as batteries, will not be able to power our societies for days or weeks when the weather is not favourable. Natural gas is currently the most used solution for the intermittency problem, which only serves to reinforce our economy’s dependence of fossil fuels, and severely undermines the apparently ‘green credentials’ of many renewables."
   },
   {
-    "type": "NarrativeText",
+    "type": "Title",
```
benjats07, line 683: Probably add a character limit for avoiding case like this one.

unstructured/partition/text_type.py
```diff
@@ -283,16 +288,33 @@ def exceeds_cap_ratio(text: str, threshold: float = 0.5) -> bool:
...
+    # tokens = [tk for tk in word_tokenize(text) if tk.isalpha()]
+    tokens = word_tokenize(text)
+    is_alpha = np.array([token.isalpha() for token in tokens])
 
-    # NOTE(jay-ylee) - If word_tokenize(text) is empty, return must be True to
-    # avoid being misclassified as Narrative Text.
-    if len(tokens) == 0:
+    # NOTE(klaijan) - If all alphas in word_tokenize(text) is empty
+    # return must be True to avoid being misclassified as Narrative Text.
+    if sum(is_alpha) == 0:
```
cragwolfe, line 297: nit: `sum(is_alpha)` this is used in 4 places, make it a variable?

cragwolfe:
I haven't looked at about ~25 ingest test diff's and they seem slightly worse to me (11 worse, 6 better, 10 meh). 

Some of the not so good ones:
image
image
image
image

Klaijan:
> I haven't looked at about ~25 ingest test diff's and they seem slightly worse to me (11 worse, 6 better, 10 meh).
> 
> Some of the not so good ones: image image image image

There was a bug in the logic which is resolved. The result is now improving.

Klaijan:
Temporarily closed. Needs to add more rules on top to reduce both false negative (before) and false positive (after). Current state: htmls and pdfs Title give too high FP.

PR #1314, opened 2023-09-06
fix: separate elements by `<br>` tag in `partition_html`

### LINKED ISSUE #1230: bug/html partition returns all paragraphs as one element

MthwRobinson:
### Summary

Closes #1230. Updates `partition_html` to split on `<br>` tags that appear within text elements.

### Testing

The following is code previously produced one giant element on `main`.

```python

filename = "example-docs/ideas-page.html"
elements = partition_html(filename=filename)

len(elements) # Should be 4
print("\n\n".join([str(el) for el in elements)])
```

The output should be:

```python
January 2023

(Someone fed my essays into GPT to make something that could answer
questions based on them, then asked it where good ideas come from.  The
answer was ok, but not what I would have said. This is what I would have said.)

The way to get new ideas is to notice anomalies: what seems strange,
or missing, or broken? You can see anomalies in everyday life (much
of standup comedy is based on this), but the best place to look for
them is at the frontiers of knowledge.

Knowledge grows fractally.
From a distance its edges look smooth, but when you learn enough
to get close to one, you'll notice it's full of gaps. These gaps
will seem obvious; it will seem inexplicable that no one has tried
x or wondered about y. In the best case, exploring such gaps yields
whole new fractal buds.
```


qued (approved):
LGTM! CI is being a jerk today though.

PR #1322, opened 2023-09-06
enhancement: backup text categorization

qued:
Currently there are some cases when `partition_pdf` is run using the `hi_res` strategy, in which elements can come back with category `UncategorizedText`. This happens when the detection model fails to detect an element, but we're able to find it anyway either because it was embedded in the PDF, or we found it using OCR.

This PR is to allow for attempting to categorize these uncategorized elements using our text-based classification function, `element_from_text`.

#### Testing:

```python

elements = partition_pdf("../example-docs/layout-parser-paper.pdf", strategy="hi_res")

# Should be 122 on main, 29 on this branch
print(len([el for el in elements if el.category == "UncategorizedText"]))

```


pravin-unstructured (approved):
Makes sense to me.

benjats07 (approved):
LGTM

cragwolfe (approved):
LGTM, there is a clear overall improvement looking at ingest-test fixture diff's.

However, there are some clear misses converting an element to Title. E.g. this is one of a few of similar pattern:

    "type": "Title",
    "element_id": "2e26dc2c4d8d6e4e53865d5697d3a983",
    "text": "n u J"
    
I'd say that's worth a follow up ticket.

PR #1323, opened 2023-09-06
Add Model Probabilities to Hi-Res strategy MetaData for Images + PDFs.

pravin-unstructured:
If a layout model is used from unstructured-inference, you get back class probabilities in the element metadata from partition. extra-pdf-image-in in requirements already has the newest version of unstructured-inference in there without a pinned version. Is there any place else that the unstructured-inference version needs to be updated to the required release version, 0.5.22?


cragwolfe:
> Is there any place else that the unstructured-inference version needs to be updated to the required release version, 0.5.22?

yes, update https://github.com/Unstructured-IO/unstructured/blob/main/requirements/constraints.in , run `make pip-compile` with python3.8 on mac, and push the updates.

qued (approved):
LGTM pending updating the tests.

PR #1324, opened 2023-09-07
Adding table extraction to partition_html

amanda103:
Adding table extraction to HTML partitioning.

This PR utilizes 'table' HTML elements to extract and parse HTML tables and return them in partitioning.

Still working on updating/adding to unit tests, but to manually test:
```
# checkout this branch, go into ipython shell
In [1]: from unstructured.partition.html import partition_html
In [2]: path_to_html = "{html sample file with table}"
In [3]: elements = partition_html(path_to_html)
```
you should see the table in the elements list!


test_unstructured/partition/test_html_partition.py
```diff
@@ -267,27 +267,27 @@ def test_partition_html_raises_with_too_many_specified():
...
+        "The\nanswer was ok, but not what I would have said. This is what"
+        " I would have said.) </td><td>The way to get new ideas is to notice"
+        " anomalies: what seems strange,\nor missing, or broken? You can see"
+        " anomalies in everyday life (much\nof standup comedy is based on this), "
+        "but the best place to look for\nthem is at the frontiers of knowledge. "
+        "</td><td>Knowledge grows fractally.\nFrom a distance its edges look smooth,"
+        " but when you learn enough\nto get close to one, you&#x27;ll notice "
+        "it&#x27;s full of gaps. These gaps\nwill seem obvious; it will seem "
+        "inexplicable that no one has tried\nx or wondered about y. In the best case,"
+        " exploring such gaps yields\nwhole new fractal buds. </td></tr>\n</tbody>\n</table>",
```
cragwolfe, line 285: the convention for `Table` elements is for `.text` to just be concatentated text, but `metadata.text_as_html` to have the html string, as above, so it looks like the html parser should be updated to reflect that convention.

amanda103: Ahhh makes sense 👍🏻

test_unstructured/documents/test_html.py
```diff
@@ -77,14 +78,7 @@ def test_parses_tags_correctly():
...
     el = doc.elements[0]
-    assert el.ancestortags + (el.tag,) == (
-        "html",
-        "body",
-        "table",
-        "tbody",
-        "tr",
-        "td",
-    )
+    assert el.ancestortags + (el.tag,) == ("table",)
```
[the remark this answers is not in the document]
amanda103: yes that looks to be the case 😆 will fix!

badGarnet:
test example's html output is:

Screenshot 2023-09-08 at 4 59 17 PM

I think the original html do have some elements that is non text but maybe those literal `\n`s should be actual line breaks?

badGarnet (approved):

unstructured/partition/html.py
```diff
@@ -63,7 +63,7 @@ def partition_html(
     headers
         The headers to be used in conjunction with the HTTP request if URL is set.
     ssl_verify
-        If the URL parameter is set, determines whether or not partition uses SSL verification
+        If the URL parameter is set, determines weather or not partition uses SSL verification
```
shreyanid, line 66:
```suggestion
        If the URL parameter is set, determines whether or not partition uses SSL verification
```
was correct initially :)

unstructured/documents/html.py
```diff
@@ -441,6 +456,36 @@ def _is_text_tag(tag_elem: etree.Element, max_predecessor_len: int = 5) -> bool:
...
+                table_text = ""
+                html_table = ""
+            return (
+                HTMLTable(
+                    text=table_text,
+                    text_as_html=html_table.replace("\n", "<br>"),
+                    tag=tag_elem.tag,
+                    ancestortags=tuple(el.tag for el in tag_elem.iterancestors())[::-1],
+                ),
+                tag_elem,
```
shreyanid, line 483: why does `next_element` need to be returned? In the case of `_process_list_item` the docstring mentions "Also returns the next html elements so that we can skip processing if bullets are found in a div element.", but I don't see a similar use case here. Can remove returning `next_element` if not needed or used (which it doesn't look like it is where it's being called in the case in `class HTMLDocument`), or add a comment on why it's necessary to keep track of.

PR #1325, opened 2023-09-07
Fix bbox coordinates for ocr_only strategy

### LINKED ISSUE #1227: bug/bbox for pdf gives out nan

Coniferish:
### Summary
Duplicate PR of #1259 because of issues with checks
Closes #1227, which found that `nan` values were present in the coordinates being generated for some elements.
This breaks logic out from `add_pytesseract_bbox_to_elements` to new functions `_get_element_box` and `convert_multiple_coordinates_to_new_system`. It also updates the logic to check that the current bounding box matches the first character of the element's text (as to avoid the `~` characters that `pytesseract.image_to_boxes` includes, but are not present in `pytesseract.image_to_string`.

### Testing
```
from PIL import Image, ImageDraw

filename="example-docs/layout-parser-paper-with-table.jpg"
elements = partition_image(filename=filename, strategy="ocr_only")
image = Image.open(filename)
draw = ImageDraw.Draw(image)
for i, element in enumerate(elements):
    print(i, element.metadata.coordinates)
    if element.metadata.coordinates:
        draw.polygon(element.metadata.coordinates.points, outline="red", width=2)
output = "example-docs/box-layout-parser-paper-with-table.jpg"
image.save(output)
image.close()
```


qued (approved):
LGTM pending resolving whatever this docker test problem is.

unstructured/partition/pdf.py
```diff
@@ -572,7 +572,7 @@ def _add_pytesseract_bboxes_to_elements(
     boxes = bboxes_string.strip().split("\n")
     box_idx = 0
     for element in elements:
-        if not element.text:
```
qued: What logic are we looking for here, because it looks like anything truthy in `boxes[box_idx][0]` will cause it to skip this element, and that doesn't sound right to me.

unstructured/partition/pdf.py
```diff
@@ -548,40 +577,32 @@ def add_pytesseract_bbox_to_elements(elements, bboxes, width, height):
         height=height,
     )
 
-    boxes = bboxes.strip().split("\n")
-    i = 0
+    boxes = bboxes_string.strip().split("\n")
+    box_idx = 0
     for element in elements:
+        if not element.text or boxes[box_idx][0]:
```
qued, line 583: Previous comment was supposed to be attached to this line.

Coniferish: Honesty didn't check to see what object was making this fail (guessing it is None), but I'll go through and see and come up with a better check. I was just getting it to not fail any more.

Coniferish: The index error was just coming from the while loop (presumably if it was at the end of the element text). I put it in a try/except to catch this, but let me know if there's a different preferred method.

PR #1326, opened 2023-09-07
Add clean_ligatures to core cleaners

walsha2:
# Background

[Ligatures](https://en.wikipedia.org/wiki/Ligature_(writing)#Ligatures_in_Unicode_(Latin_alphabets)) can sometimes show up during the text extraction process when they should not. Very common examples of this are with the Latin `f` related ligatures which can be **very subtle** to spot by eye (see example below), but can wreak havoc later.

```python
"ﬀ": "ff",
"ﬁ": "fi",
"ﬂ": "fl",
"ﬃ": "ffi",
"ﬄ": "ffl",
```

Several libraries already do something like this. Most recently, `pdfplumber` added this sort of capability as part of the text extraction process, see https://github.com/jsvine/pdfplumber/issues/598

Instead of incorporating any sort of breaking change to the PDF text processing in `unstructured`, it is best to add this as another cleaner and allow users to opt in. In turn, the `clean_ligatures` method has been added in this PR - with accompanying tests.

# Example

Here is an example PDF that causes the issue. For example: `Beneﬁts`, which should be `Benefits`.

[example.pdf](https://github.com/Unstructured-IO/unstructured/files/12544344/example.pdf)

```bash
curl -X 'POST' \
    'https://api.unstructured.io/general/v0/general' \
    -H 'unstructured-api-key: ${UNSTRUCTURED_API_KEY}' \
    -F 'files=@example.pdf' \
    -s | jq -C .
```

# Notes

An initial list of mappings was added with the most common ligatures. There is some subjectivity to this, but this should be a relatively safe starting set. Can always be expanded as needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Features
 
 * Add Jira Connector to be able to pull issues from a Jira organization
+* Add `clean_ligatures` function to expand ligatures in text
 
 ### Fixes
 
--- a/test_unstructured/cleaners/test_core.py
+++ b/test_unstructured/cleaners/test_core.py
@@ -65,6 +65,32 @@ def test_clean_ordered_bullets(text, expected):
 @pytest.mark.parametrize(
     ("text", "expected"),
     [
+        ("The æther is a classic element.", "The aether is a classic element."),
+        ("In old texts, Æsop's fables are", "In old texts, AEsop's fables are"),
+        ("The buﬀer zone is there.", "The buffer zone is there."),
+        ("The ﬁle was found in the system.", "The file was found in the system."),
+        ("She had a ﬂower in her hair.", "She had a flower in her hair."),
+        ("The coﬃn was placed in the grave.", "The coffin was placed in the grave."),
+        ("The buﬄe zone was clearly marked.", "The buffle zone was clearly marked."),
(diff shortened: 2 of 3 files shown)


cragwolfe:
@walsha2 , please add a bullet near the top of CHANGELOG.md. this is a great cleaning addition, thanks!
i'm tempted for this to be included by default in many partition cases, tbh, but that can follow in a later PR.

walsha2:
> add a bullet near the top of CHANGELOG.md

Done.

> i'm tempted for this to be included by default in many partition cases

I thought about it as well, but I just started messing with `unstructured` this week. Will defer that decision about partition incorporation to someone more in the know! 😄 

------

First PR to get feet wet with repo. Awesome package. Looking forward to making more contributions in the future!

cragwolfe (approved):

cragwolfe:
> @cragwolfe I cant really make sense of the CI actions. Not sure why the `CI / lint (3.8)` is failing. Does not look to be PR related at all.
> 
> ```
> make: black: No such file or directory
> ```
> 
> https://github.com/Unstructured-IO/unstructured/actions/runs/6105531406/job/16569692294?pr=1326

yes, github CI was having a rough time for some unknown reason -- frustrating! it seems to be fine now.
i just merged in changes to main, let's see how it goes 🤞

PR #1328, opened 2023-09-07
feat: improved salesforce partitioning

ryannikolaidis:
Currently email bodies are created with only the body text. Instead we should leverage the html body such that partitioning will better structure the content.

Additionally, the Salesforce connector currently partitions all non-email records, which are (sometimes nested) dictionaries of structured data, by writing them to a templated, raw text. The result is that the key value pairs are limited to only keys that we specify in the template and that the partitioning may sometimes incorrectly group content. Instead by using xml as the intermediary format we can guarantee the element content will adhere to the key-value pairs.

* leverage html content for email files
* use xml as intermediary format for well better structured key-values
* bonus: wraps get_file with @SourceConnectionError.wrap

### DIFF
--- a/unstructured/ingest/connector/salesforce.py
+++ b/unstructured/ingest/connector/salesforce.py
@@ -8,13 +8,13 @@ https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_de
 https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_auth_connected_app.htm
 """
 import os
+from collections import OrderedDict
 from dataclasses import dataclass
 from email.utils import formatdate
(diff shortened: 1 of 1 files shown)


ryannikolaidis:
Still in draft. For some reason emails written were empty (related to my change). Though this is definitely working against demo account I set up last night.

potter-potter:
There was a reason I was doing Text Body instead of HtmlBody. I can't quite remember, but I was definitely looking at HtmlBody first. And then concluded the Text Body was best. Though I didn't go down the rabbit hole too far.

potter-potter:
Html seems to work.

XML is super cool. We're just grabbing everything that it gives us and save to an xml file. And I thought XML was not as cool as json.

potter-potter:
Oh. Strange. Second time I did email. It did not work....

PR #1334, opened 2023-09-07
refactor languages parameter for pdf partition functions

shreyanid:
### Summary

In order to support language functionality other than Tesseract OCR, we want to represent languages provided for either partitioning accuracy or OCR as a standard list of langcodes as strings.

### Details

Adds `languages` (a list of strings) as a parameter to pdf partitioning functions. Marks `ocr_languages` for deprecation. Adds a new file `lang.py` for language-related helper functions.

Coming up: langcode standardization, language detection

### Test

Call `partition_pdf` or `partition_pdf_or_image` with a variety of strategies, languages, or `ocr_languages`. 
- inclusion of `ocr_languages` as a parameter should display a deprecation warning
- the other valid call outputs should be no different from the current outputs.

ex:
```

elements = partition_pdf(filename="example-docs/DA-1p.pdf", strategy="hi_res", languages=["eng", "spa"])
```


unstructured/partition/pdf.py
```diff
@@ -94,13 +97,28 @@ def partition_pdf(
         The last modified date for the document.
     """
     exactly_one(filename=filename, file=file)
+
+    if languages and ocr_languages:
```
amanda103, line 101: this seems strange to raise, since we have a default value for languages. might be confusing to users if they didn't actually pass in the `languages` param

shreyanid: was following this pattern https://github.com/Unstructured-IO/unstructured/blob/e4e25c9febe8e46874cbb91f0201cfaf5330a672/unstructured/partition/auto.py#L185

amanda103: Yes I think this makes sense to raise when both params are passed in by the user as in the example you linked. But in this case where we're defining a default we will return this error message if they only pass in `ocr_languages` param which is confusing.

unstructured/partition/lang.py
```diff
@@ -0,0 +1,31 @@
+from typing import List
+
+
+# convert language param (list of strings) into tesseract ocr language format (639-2 variant joined by + signs)
+# NOTE(Shreya): does this arg need a default value?
+def prepare_languages_for_tesseract(languages: List[str] = ["eng"]):
+    # NOTE(Shreya): assumes language codes are in tesseract format (will be updated later)
+    tesseract_languages = ""
+    for lang in languages:
+        tesseract_languages = tesseract_languages + lang + "+"
```
newelh, line 10: List comprehension would be great here `"".join([f"{lang}+" for lang in languages).strip("+")`

unstructured/partition/lang.py
```diff
@@ -0,0 +1,31 @@
...
+        if plus_sign_index != -1:
+            lang = ocr_languages[:plus_sign_index]
+            ocr_languages = ocr_languages[plus_sign_index + 1 :]
+        else:  # no plus sign found
+            lang = ocr_languages  # entire string is one language
+            end = True
+
+        languages.append(lang)
+
+    return languages
```
newelh, line 31: Could this function not just be `ocr_languages.split("+")`?

unstructured/partition/pdf.py
```diff
@@ -94,13 +100,34 @@ def partition_pdf(
         The last modified date for the document.
     """
     exactly_one(filename=filename, file=file)
+
+    if not isinstance(languages, list):
+        raise TypeError("The language parameter must be a list of language codes as strings.")
+
+    if ocr_languages is not None:
```
amanda103, line 107: this is much clearer what's happening 🥇

unstructured/partition/pdf.py
```diff
@@ -94,13 +100,34 @@ def partition_pdf(
...
+            )
+
     return partition_pdf_or_image(
         filename=filename,
         file=file,
         include_page_breaks=include_page_breaks,
         strategy=strategy,
         infer_table_structure=infer_table_structure,
-        ocr_languages=ocr_languages,
+        # ocr_languages=ocr_languages,
```
amanda103, line 129: is there a reason we're commenting out as opposed to removing?

amanda103 (approved):
lgtm! I think the only q I have is around the commenting vs. removing the params. If there's a reason to keep the comments I can be convinced, otherwise probably makes sense to remove :)

unstructured/partition/pdf.py
shreyanid: nope! was just for visibility of change during draft, will remove now

PR #1341, opened 2023-09-08
build(deps): PDF images, unstructured-inference==0.5.23

cragwolfe:
Bumps unstructured-inference==05.23 to pull in @christinestraub's fix:
https://github.com/Unstructured-IO/unstructured-inference/pull/198 , so embedded Images
in PDF's are now included in partition results ("hi_res").

From the perspective of elements with clean text, this is not a big win as a lot of the images have OCR garbage. However, it is important to preserve image elements for other downstream use cases, so overall this is a step forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.13-dev4
+## 0.10.13-dev5
 
 ### Enhancements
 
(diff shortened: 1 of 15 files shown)


benjats07 (approved):
LGTM

Issue #494, opened 2023-04-18
Libmagic detection for "application/octet-stream"


cragwolfe:
In `unstructured.file_utils.fileytpe::detect_filetype()`, if a content-type is "application/octet-stream",  and the file processed is determined to be a zip file in
[_detect_filetype_from_octet_stream()](https://github.com/Unstructured-IO/unstructured/blob/7ec8527/unstructured/file_utils/filetype.py#L275), it checks the zip archive content to try to determine the type.

However, if it is not a zip file, the function gives up and returns a [FileType.UNK](https://github.com/Unstructured-IO/unstructured/blob/7ec8527/unstructured/file_utils/filetype.py#L316). Instead, the content type could be inferred with libmagic as is already the case [here](https://github.com/Unstructured-IO/unstructured/blob/7ec8527/unstructured/file_utils/filetype.py#L194) . This logic could also be applied to unzipp'ed content where the filename does not match a known filetype.

**Definition of Done**

Test(s) are added where files previously identified as `FileType.UNK` due to above issue are now identified by libmagic.

PR #1347, opened 2023-09-08
Chore: Libmagic detection for "application/octet-stream" when  it is not a zip file.

srjchauhan:
Addressed the issue #494 .
Updated the `_detect_filetype_from_octet_stream()` function to use libmagic to infer the content type of file when it is not a zip file.

### DIFF
--- a/unstructured/file_utils/filetype.py
+++ b/unstructured/file_utils/filetype.py
@@ -397,6 +397,10 @@ def _detect_filetype_from_octet_stream(file: IO) -> FileType:
         elif all(f in archive_filenames for f in EXPECTED_PPTX_FILES):
             return FileType.PPTX
 
+    if LIBMAGIC_AVAILABLE:
+        # Infer mime type using magic if octet-stream is not zip file
+        mime_type = magic.from_buffer(file.read(4096), mime=True)
+        return STR_TO_FILETYPE.get(mime_type, FileType.UNK)
     logger.warning(
         "Could not detect the filetype from application/octet-stream MIME type.",
     )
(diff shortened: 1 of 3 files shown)


unstructured/file_utils/filetype.py
```diff
@@ -397,6 +397,10 @@ def _detect_filetype_from_octet_stream(file: IO) -> FileType:
         elif all(f in archive_filenames for f in EXPECTED_PPTX_FILES):
             return FileType.PPTX
 
+    if LIBMAGIC_AVAILABLE:
```
cragwolfe, line 400: `elif`?

cragwolfe (commented):
one comment/question.
please add a test to verify functionality.
otherwise looks good to me, thanks!

srjchauhan:
I have already added a test case, If you want me to handle any specific case please do mention it I will add that case
Thanks

cragwolfe:
> I have already added a test case, If you want me to handle any specific case please do mention it I will add that case
> Thanks

my bad, one year later and i'm still not used to seeing tests above source codes in diff (thanks to alphabetization and unstructured ). what about the `elif`?

srjchauhan:
Elif is not required as if lib magic is not available it will  return filetype.unk from last return statement, it's already there you might have missed it because  it is not showing in changes

cragwolfe (approved):

unstructured/file_utils/filetype.py
srjchauhan: It will return filetype.unk from line 407 if libmayic not available so,elif not required

cragwolfe:
Thanks for the contribution, @srjchauhan !

srjchauhan:
Thank you @cragwolfe

Issue #1296, opened 2023-09-04
bug: partition() fails on a certain PDF


cragwolfe:
**Describe the bug**
Partitioning https://utic-dev-tech-fixtures.s3.us-east-2.amazonaws.com/pastebin/ravak-profi-2022_en.pdf
results in an error.

**To Reproduce**

Snippet and stack trace in https://gist.github.com/cragwolfe/8014dcc0ec1c13c2f4b00dfcd110cb5f

**Expected behavior**
No stack trace, partitioned elements are returned.

**Additional context**

One can work around the issue with `sort_mode="basic"` e.g.:

    partition(filename=filename, strategy=strategy, sort_mode="basic")

cragwolfe:
CC @christinestraub

awalker4:
Another reproducer from the community slack: https://publicplansdata.org/reports/CA_ORANGECOUNTY-OCERS_CAFR_2021_142.pdf

awalker4:
[Forward Partners Financials.pdf](https://github.com/Unstructured-IO/unstructured/files/12540961/Forward.Partners.Financials.pdf)

The attached pdf gives a different error in `sort_page_elements`. I can make a separate bug if requires a different fix.

```
  File "/home/notebook-user/prepline_general/api/general.py", line 367, in pipeline_api
    elements = partition(
  File "/home/notebook-user/.local/lib/python3.10/site-packages/unstructured/partition/auto.py", line 293, in partition
    elements = _partition_pdf(
  File "/home/notebook-user/.local/lib/python3.10/site-packages/unstructured/documents/elements.py", line 237, in wrapper
    elements = func(*args, **kwargs)
  File "/home/notebook-user/.local/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 528, in wrapper
    elements = func(*args, **kwargs)
  File "/home/notebook-user/.local/lib/python3.10/site-packages/unstructured/partition/pdf.py", line 97, in partition_pdf
    return partition_pdf_or_image(
  File "/home/notebook-user/.local/lib/python3.10/site-packages/unstructured/partition/pdf.py", line 174, in partition_pdf_or_image
    extracted_elements = extractable_elements(
  File "/home/notebook-user/.local/lib/python3.10/site-packages/unstructured/partition/pdf.py", line 118, in extractable_elements
    return _partition_pdf_with_pdfminer(
[...]

Issue #671, opened 2023-06-01
ingest with  fast strategy on PDFs have different output on different python versions


yuming-long:
The ingest is having different output on different python versions (python 3.8, 3.9, 3.10 on ci) I think this happens on tables specifically. See the [ci error](https://github.com/Unstructured-IO/unstructured/actions/runs/5136058229/jobs/9242350596) on [draft PR](https://github.com/Unstructured-IO/unstructured/pull/614).

**To Reproduce**
Run `scripts/ingest-test-fixtures-update.sh ` on different python versions, and the output maybe different.

**Screenshots**
Filename: `s3/small-pdf-set/2023-Jan-economic-outlook.pdf`
The table: 
![image](https://github.com/Unstructured-IO/unstructured/assets/63475068/5b0f1707-7042-4872-8ebe-e0ee87035173)
Part of the output from the table with fast strategy on python 3.8.15:
![image](https://github.com/Unstructured-IO/unstructured/assets/63475068/b04fc1a4-529b-4791-8063-50e682970817)
ci error on python 3.9 (same with fast) strategy:
![image](https://github.com/Unstructured-IO/unstructured/assets/63475068/62d4b49a-cc95-4d93-8aa3-4203a6f7322f)

yuming-long:
close as fixed in https://github.com/Unstructured-IO/unstructured/pull/614

PR #1361, opened 2023-09-09
fix: avoid PDF sorting error on negative coords

cragwolfe:
The default sorting algorithm for PDF's, "xycut," would cause an error when partitioning a document if Y coordinate points were negative. This change checks for that condition (or more broadly, any negative coordinates) and falls back to the "basic" sort if that is the case.

This PR does not address the underlying issue of "bad points" which still should be investigated. However, the sorting code should be less brittle to unexpected bounding boxes in the first case.

Resolves: https://github.com/Unstructured-IO/unstructured/issues/1296

~~**Bonus:** previously, 2 sorts were always called per PDF page for non-basic sorts. Really, the intention was to fall back to basic if the non-basic sort (which right now is just xycut) exited out due to a stricter validation check. Now, only one sort is actually ever applied. This change led to some subtle diff's in ingest-test outputs.~~
Always run the basic sort first now for determinism in element outputs, for now.

**Test instructions**

Run the following snippet using https://utic-dev-tech-fixtures.s3.us-east-2.amazonaws.com/pastebin/fpf-p89.pdf as a test file.

    import logging
    import sys

    logging.getLogger("unstructured.trace").setLevel(logging.DEBUG)
    logging.getLogger("unstructured.trace").addHandler(logging.StreamHandler())

    print(f"unstructured version: {__version__}")
    elements = partition(filename=sys.argv[1], strategy="fast")
    print(f"number of elements is {len(elements)}")

E.g.

    python test-snippet.py fpf-p89.pdf

When running with unstructured `main` or `unstructured==0.10.12`, witness the stack trace. When running from this branch, no stack trace.


christinestraub:
For the `fast` strategy, I needed to keep the basic sort(SORT_MODE_BASIC) when applying the xycut sort because the ingest with `fast` strategy on PDFs failed on different Python versions without the basic sort(SORT_MODE_BASIC) - https://github.com/Unstructured-IO/unstructured/pull/1161/commits/9da3efd5818d4d7ea5bc2c2f922aeb4457ae8045
The basic sort was implemented to fix [this issue](https://github.com/Unstructured-IO/unstructured/issues/671).

yuming-long (approved):
LGTM!

PR #1366, opened 2023-09-11
ingest test file cleanup

rbiseck3:
### Description
* `check-diff-expected-output.sh`: Because the text content is being generated at runtime, this adds a final clean up step to delete those folders by default. An environment variable can be set to skip clean up if desired. When running tests locally, not cleaning up caused some issues. This isn't a problem when running in CI since a new environment is used each time and the generated content isn't persisted in the repo. 
* All base ingest tests got a trap added to delete the generated `$OUTPUT_DIR` to avoid memory issues in CI since all tests are run on a single runner. 

Closes out https://github.com/Unstructured-IO/unstructured/issues/1432


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 * **Faster ocr_only speed for partitioning PDF and images.** Use `unstructured_pytesseract.run_and_get_multiple_output` function to reduce the number of calls to `tesseract` by half when partitioning pdf or image with `tesseract`
 * **Adds data source properties to fsspec connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive.
 * **Non-HTML text files now return unstructured-elements as opposed to HTML-elements.** Previously the text based files that went through `partition_html` would return HTML-elements but now we preserve the format from the input using `source_format` argument in the partition call.
-
+* **Ingest tests cleanup** Add cleanup to all ingest tests to delete created content by default. This remedies memory issues of running all the tests on a single runner in the CI.
```
ryannikolaidis, line 17: we should probably keep changelog notes specific to code changes / can omit notes that relate to tests

test_unstructured_ingest/test-ingest-azure.sh
```diff
@@ -8,6 +8,19 @@ OUTPUT_FOLDER_NAME=azure
...
+  if [ -d "$OUTPUT_DIR" ]; then
+    echo "cleaning up tmp directory: $OUTPUT_DIR"
+    rm -rf "$OUTPUT_DIR"
+  fi
+
+  echo "--- Cleanup done ---"
+}
+
+trap cleanup EXIT
+
```
ryannikolaidis, line 23: can we make this a common script that just gets called rather than so much duplication?

test_unstructured_ingest/check-diff-expected-output.sh
```diff
@@ -14,12 +14,30 @@ set +e
...
+      echo "cleaning up tmp directory: $OUTPUT_DIR_TEXT"
+      rm -rf "$OUTPUT_DIR_TEXT"
+    fi
+  else
+    echo "skipping tmp directory cleanup"
+  fi
+}
+
+trap cleanup EXIT
+
```
ryannikolaidis, line 40: if we're calling cleanup in individual scripts, do we need to do it here?
cleanup here feels like an unexpected consequence when calling check-diff

rbiseck3: this cleanup is specific to the text files generated within this script.

ryannikolaidis (approved):
lgtm!

ryannikolaidis:
note: admin merging, we've seen this succeed regularly across enough attempts but have hit some unrelated intermittent connectivity issues causing ingest to fail.

Issue #1319, opened 2023-09-06
feat: document embeddings


ahmetmeleq:
**1- Related Problem**

Many NLP applications require the ability to represent parts of documents in a semantic way. Ability to embed documents, or parts of documents will enable users to make use of these semantic representations.

**2- Aimed Solution / Definition of Done**

Create a module with functions that receive chunks of text, embeds them, and saves the embedding representations to the disk.

PR #1368, opened 2023-09-11
feat: document embeddings

ahmetmeleq:
Closes https://github.com/Unstructured-IO/unstructured/issues/1319, closes https://github.com/Unstructured-IO/unstructured/issues/1372

This module:

- implements EmbeddingEncoder classes which track embedding related data
- implements embed_documents method which receives a list of Elements, obtains embeddings for the text within Elements, updates the Elements with an attribute named embeddings , and returns the updated Elements
- the module uses langchain to obtain the embeddings
-----
- The PR additionally fixes a JSON de-serialization issue on the metadata fields.

To test the changes, run `examples/embed/example.py`


unstructured/embed/embedder/example.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+from unstructured.embed.embedder.open_ai import OpenAIEmbedder, OpenAIEmbedderConfig
+
+config = OpenAIEmbedderConfig(
+    api_key=os.environ["OPENAI_API_KEY"],
+    list_of_elements_json_paths="""/Users/ahmet/elements_results/1.docx.json
+                                   /Users/ahmet/elements_results/2.pdf.json""",
```
rbiseck3, line 8: Don't want to include paths specific to the user.

unstructured/embed/embedder/open_ai.py
```diff
@@ -0,0 +1,155 @@
...
+    Args:
+        api_key: API Key, generated for the user
+        model_name: Model name for OpenAI
+
+    Returns:
+        OpenAI Embeddings object
+    """
+    from langchain.embeddings.openai import OpenAIEmbeddings
+
+    embedder = OpenAIEmbeddings(
```
badGarnet, line 45: usually don't want to comment on names at this stage but... I'd propose using `encoder` instead of `embedder`. `encoder` is more commonly used.

badGarnet, line 45: usually don't want to comment on names at this stage but... I'd propose using `encoder` instead of `embedder`. `encoder` is more commonly used.

badGarnet: I see we have `encoder` elsewhere... welp

unstructured/embed/embedder/open_ai.py
```diff
@@ -0,0 +1,155 @@
...
+    def get_elements_from_json(self) -> List[Element]:
+        """Reads elements"""
+        return partition_json(filename=self.filename)
+
+    @requires_dependencies(["langchain"], extras="jira")
+    def embed_and_write_result(self):
+        logger.debug(f"Reading: {self} - PID: {os.getpid()}")
+
+        elements = self.get_elements_from_json()
+        embeddings = self.session_handle.service.embed_documents([str(e) for e in elements])
```
badGarnet, line 110: this can be a long running process; is this function call gonna be async?

unstructured/embed/embedder/open_ai.py
```diff
@@ -0,0 +1,155 @@
...
+        for i in range(len(elements)):
+            elements[i].embedding = embeddings[i]
+
+        return elements
+
+    def write_result(self):
+        """Write the structured json result for this doc. result must be json serializable."""
+        self._output_filename.parent.mkdir(parents=True, exist_ok=True)
+        with open(self._output_filename, "w", encoding="utf8") as output_f:
+            json.dump(self.elements, output_f, ensure_ascii=False, indent=2, cls=ElementEncoder)
```
badGarnet, line 127: we are probably just prototyping here so writing to json is fine but depending on the embedding model this can blow up the json file size very quickly. e.g., the most commonly used ones have 748 dimension, so that is 748 floating point numbers per embedding.
speaking of which we should probably record encoder metadata somewhere in the process/output out well, like the vector dimension and if it is unit vector (two attributes relevant to information retrieval tasks -> computing distance)

unstructured/embed/embedder/example.py
ahmetmeleq: Updated with https://github.com/Unstructured-IO/unstructured/pull/1368/commits/d7a8689463f69a5591cc5bb6fbd2cabd84e60f96

unstructured/embed/embedder/open_ai.py
ahmetmeleq: resolving due to duplicate

ahmetmeleq: > I see we have `encoder` elsewhere... welp

I hadn't thought about that while naming it, but I agree that it might be confused with `ElementEncoder` if we name it as encoder

This object manages the requests to OpenAI API - I wanted to name it as `embedder` so users with less ML experience can also make sense of it (due to it being used as a trendy word in wider circles)

unstructured/embed/embedder/open_ai.py
ahmetmeleq: Per document, yes, each OpenAIEmbeddingsDoc instance will be handled separately when we integrate this into ingest. Each instance corresponds to a file.

However, in this implementation, sub-document parallelisation would not be supported. I agree that we might want to support it, considering that we might have long documents

unstructured/embed/embedder/open_ai.py
ahmetmeleq: On saving the embeddings to the disk, what are some good alternatives? 

What I think is:
- If we'd like to isolate the task of encoding from the task of writing to a DB, and want an intermediate place to store the embeddings:
  - we can save embeddings to a separate file, again in a basic format. these files can be processed later on to be written into a database
  - we can use a message queue so the embeddings later get consumed by a job and be written into the database
- Or, we could directly write the embeddings into a database

unstructured/embed/embedder/open_ai.py
```diff
@@ -0,0 +1,155 @@
...
+        """Create output file path."""
+        output_file = f"{self.filename}_with_embeddings.json"
+
+        return (Path(self.config.output_dir) / output_file).resolve()
+
+    def get_elements_from_json(self) -> List[Element]:
+        """Reads elements"""
+        return partition_json(filename=self.filename)
+
+    @requires_dependencies(["langchain"], extras="jira")
```
ryannikolaidis, line 105: extras="jira"? typo?

unstructured/embed/embedder/open_ai.py
ryannikolaidis: how about we just call this `embeddings`, per the class name and that seems to be the name they give to the instance in [their documentation](https://api.python.langchain.com/en/latest/embeddings/langchain.embeddings.openai.OpenAIEmbeddings.html).

unstructured/embed/embedder/open_ai.py
ryannikolaidis: do we actually need to write this to file? or can we just rely on return the in-memory elements with embeddings?

unstructured/embed/embedder/open_ai.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def get_embed_docs(self):
+        """Reads all result files to embed them."""
+        return [
+            OpenAIEmbeddingsDoc(
+                self.config,
+                filename,
+            )
+            for filename in self.config.list_of_elements_json_paths.split()
+        ]
```
ryannikolaidis, line 155: probably for v0 this just takes a single result (list of Elements) or json file and returns the list of Elements with embeddings? i.e. not sure we need to support iterating a list of docs (at least for v0)

unstructured/embed/embedder/open_ai.py
ahmetmeleq: Depends on how the downstream jobs want to obtain and use these embeddings. If we'd like embed large corpora, we'd like to index the embeddings in a DB. To do that, the job that'll write the embeddings into the DB can read the embeddings from memory, from a message queue, or from the disk

If adequate, we can return the embeddings in-memory for now; and open another ticket to bring in persistence

ahmetmeleq:
todo: add metadata properties for the embedding model

unstructured/embed/embedder/open_ai.py
ryannikolaidis: > Depends on how the downstream jobs want to obtain and use these embeddings
yes, definitely. lets start with the requirements for ingest and what makes most sense there and how this interfaces.  CC @rbiseck3 

aside: let's avoid a message queue

unstructured/embed/embedder/open_ai.py
ryannikolaidis: though maybe we need to look at how we want this to slot in for [ingest here](https://github.com/Unstructured-IO/unstructured/blob/2b571eb9a370a0ac10e3d78815c626015ffb7691/unstructured/ingest/processor.py#L96-L98) for a requirements perspective.

unstructured/embed/interfaces.py
```diff
@@ -0,0 +1,26 @@
...
+
+class BaseEmbeddingEncoder(ABC):
+    @abstractmethod
+    def initialize():
+        """Initializes the embedding encoder class. Should also validate the instance
+        is properly configured: e.g., embed a single a element"""
+        pass
+
+    @property
+    def num_of_dimensions(self) -> int:
```
rbiseck3, line 15: Should this method be an abstract method and require the implementation of this base class to define it? Otherwise the return type should be `Optional[int]` since the default behavior is to return `None`.

unstructured/embed/interfaces.py
```diff
@@ -0,0 +1,26 @@
...
+        """Number of dimensions for the embedding vector."""
+        return None
+
+    @property
```
rbiseck3, line 19: Same comment as above.

unstructured/embed/openai.py
```diff
@@ -0,0 +1,46 @@
...
+class OpenAIEmbeddingEncoder(BaseEmbeddingEncoder):
+    def __init__(self, api_key: str, model_name: Optional[str] = "text-embedding-ada-002"):
+        self.api_key = api_key
+        self.model_name = model_name
+        self.initialize()
+
+    def initialize(self):
+        self.openai_client = self.get_openai_client()
+
+    def embed(self, elements: Optional[List[Element]]) -> List[Element]:
```
jasonbot, line 20: As long as we have this method with the `Optional[List[Element]]) -> List[Element]` signature to work with, we should be fine in Enterprise to consume this.

unstructured/embed/embedder/open_ai.py
ahmetmeleq: Changed the namings with https://github.com/Unstructured-IO/unstructured/pull/1368/commits/499795ae9364593a4954a30e0e568598f3d15f33, please let me know if we'd like further changes on the names

unstructured/embed/embedder/open_ai.py
ahmetmeleq: Update: For the current ticket, we've decided to make this work fully in memory. Inputs are read from memory, results are passed on via memory. https://github.com/Unstructured-IO/unstructured/pull/1368/commits/499795ae9364593a4954a30e0e568598f3d15f33

unstructured/embed/embedder/open_ai.py
ahmetmeleq: That's right, missed this one. Addressed with https://github.com/Unstructured-IO/unstructured/pull/1368/commits/499795ae9364593a4954a30e0e568598f3d15f33

unstructured/embed/embedder/open_ai.py
ahmetmeleq: Re-implemented as taking a single list of elements and returning the updated list with embeddings in https://github.com/Unstructured-IO/unstructured/pull/1368/commits/499795ae9364593a4954a30e0e568598f3d15f33

ahmetmeleq: For fitting into ingest, and a single doc, I think instantiating the class and putting it as `self.embedding_encoder.embed(doc)` should be possible within those lines:
https://github.com/Unstructured-IO/unstructured/blob/2b571eb9a370a0ac10e3d78815c626015ffb7691/unstructured/ingest/processor.py#L96-L98

However, to process multiple docs, we need an async / parallelised solution. Would pool.map be possible like with partition, I wonder:
https://github.com/Unstructured-IO/unstructured/blob/2b571eb9a370a0ac10e3d78815c626015ffb7691/unstructured/ingest/processor.py#L71-L87

A note, since these are API calls and not actual compute intensive operations, our users will ultimately be bound by the model provider (in the current case OpenAI) API. Paid users get good [rate limits](https://platform.openai.com/docs/guides/rate-limits/what-are-the-rate-limits-for-our-api), therefore if OpenAI doesn't have hidden limits behind the scenes (ie if they allocate more resources to the user when the load from that user is high), all types of parallelisation should benefit our users' runtime.

unstructured/embed/interfaces.py
ahmetmeleq: Agreed, addressed with https://github.com/Unstructured-IO/unstructured/pull/1368/commits/04468f0e5f6e19546dfa4ef17ef8d4a2a7064f71

unstructured/embed/interfaces.py
ahmetmeleq: Agreed, and addressed with https://github.com/Unstructured-IO/unstructured/pull/1368/commits/04468f0e5f6e19546dfa4ef17ef8d4a2a7064f71

unstructured/embed/interfaces.py
ahmetmeleq: Please also note https://github.com/Unstructured-IO/unstructured/pull/1368/commits/8c51d8ade6423cda7085cb2a3f6f0927dd2d1878

unstructured/embed/interfaces.py
ahmetmeleq: Please also note https://github.com/Unstructured-IO/unstructured/pull/1368/commits/8c51d8ade6423cda7085cb2a3f6f0927dd2d1878

ahmetmeleq:
Ready for review assuming parallelisation with this design is doable/easy in ingest, just like it is in enterprise

unstructured/embed/openai.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, api_key: str, model_name: Optional[str] = "text-embedding-ada-002"):
+        self.api_key = api_key
+        self.model_name = model_name
+        self.initialize()
+
+    def initialize(self):
+        self.openai_client = self.get_openai_client()
+
+    def num_of_dimensions(self):
+        return np.shape(self.examplary_embedding)
```
rbiseck3, line 23: pretty sure every method that calls `self.examplary_embedding` would raise an error is called before that's being set. Should add some logic around that.

ahmetmeleq: self.examplary_embedding is set in the __init__ method, via this route:

__init__ -> initialize -> get_openai_client 

would that work?

rbiseck3: Perfect.

unstructured/embed/openai.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def is_unit_vector(self):
+        return np.isclose(np.linalg.norm(self.examplary_embedding), 1.0)
+
+    def embed_query(self, query):
+        return self.openai_client.embed_documents(str(query))
+
+    def embed_documents(self, elements: List[Element]) -> List[Element]:
+        embeddings = self.openai_client.embed_documents([str(e) for e in elements])
```
rbiseck3, line 32: Just curious, does the ordering of the elements affect the embeddings? Might want to sort them somehow if so.

rbiseck3: Oh wait elements are already sorted, ignore.

rbiseck3 (approved):

CHANGELOG.md
```diff
@@ -67,7 +74,10 @@
...
+* **Fixes a chunking issue via dropping the field "coordinates".**
+  * Problem: chunk_by_title function was chunking each element to its own individual chunk while it needed to group elements into a fewer number of chunks. We've discovered that this happens due to a metadata matching logic in chunk_by_title function, and discovered that elements with different metadata can't be put into the same chunk. At the same time, any element with "coordinates" essentially had different metadata than other elements, due each element locating in different places and having different coordinates.
+  * Fix: That is why we have included the key "coordinates" inside a list of excluded metadata keys, while doing this "metadata_matches" comparision.
+  * Importance: This change is crucial to be able to chunk by title for documents which include "coordinates" metadata in their elements.
```
ryannikolaidis, line 80: intentionally updating this older note? ...guessing this will be resolved when you update to latest, but look out for it.

CHANGELOG.md
```diff
@@ -8,8 +7,18 @@
 
 ### Features
 
+* **Adds the embedding module to be able to embed Elements**
+  * Problem: Many NLP applications require the ability to represent parts of documents in a semantic way. Until now, Unstructured did not have text embedding ability within the core library.
+  * Feature: This embedding module is able to track embeddings related data with a class, embed a list of elements, and return an updated list of Elements with the *embeddings* property. The module is also able to embed query strings.
+  * Importance: Ability to embed documents, or parts of documents will enable users to make use of these semantic representations in different NLP applications, such as search, retrieval, and retrieval augmented generation.
+
```
ryannikolaidis, line 14: appreciate adding the importance here. probably should follow the single paragraph format of all of the other recent changelog items

CHANGELOG.md
ahmetmeleq: Yes, this was to modify the changelog item into the multiple paragraph format 

Now I've re-modified all changelog items into single paragraph format so this change no longer exists 👍

PR #1383, opened 2023-09-12
Roman/expand fsspec downstream connectors

rbiseck3:
### Description
Because all of the fsspec connectors have their interactions with the various underlying file system libraries abstracted away by the parent `fsspec` library, adding the rest on top of the existing s3 downstream connector is fairly simple. This adds the following downstream connectors: 
 * azure
 * box
 * dropbox
 * gcs

Also updated docs/ingest tests for s3 since that falls in the fsspec category.


ryannikolaidis:
overall looks good. add examples, tests, docs?

PR #1384, opened 2023-09-12
Fix bbox coordinates for ocr_only strategy <- Ingest test fixtures update

ryannikolaidis:
This pull request includes updated ingest test fixtures.
Please review and merge if appropriate.


Coniferish:
This is actually using the wrong document for testing, so this should not be merged. Closing

PR #1388, opened 2023-09-12
Table processing test for RTF

amanda103:
This PR does two things:
1. Adds test case (and alters sample doc) for rtf and epub files with table
2. Adds `xls/x` file extension to `skip_infer_table_types` default list

### DIFF
--- a/unstructured/partition/auto.py
+++ b/unstructured/partition/auto.py
@@ -126,7 +126,7 @@ def partition(
     encoding: Optional[str] = None,
     paragraph_grouper: Optional[Callable[[str], str]] = None,
     headers: Dict[str, str] = {},
-    skip_infer_table_types: List[str] = ["pdf", "jpg", "png"],
+    skip_infer_table_types: List[str] = ["pdf", "jpg", "png", "xls", "xlsx"],
     ssl_verify: bool = True,
     ocr_languages: str = "eng",
     pdf_infer_table_structure: bool = False,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.15-dev5
+## 0.10.15-dev6
 
 ### Enhancements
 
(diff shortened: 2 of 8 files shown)


shreyanid:
Could you share the context on why xls/x is being added to the to skip_infer_table_types default list? If table extraction is skipped for an Excel file what else is there to get processed 😅

amanda103:
> Could you share the context on why xls/x is being added to the to skip_infer_table_types default list? If table extraction is skipped for an Excel file what else is there to get processed 😅

Yup! My understanding is that by default we can't assume these docs _only_ contain tables, as opposed to something like a CSV that makes more sense to process as a table by default. A user can still process an xls/x file as a table but would have to set the skip_infer_table_types to not contain the file type extension :)

shreyanid:
I see, so unlike pdf/image which may just contain some table elements, processing an excel file as a table assumes the entire document is a table, which is possible not to be the case, so that's why we don't want to extract tables from it by default. 

It would be nice if there were a middle ground for Excel instead, of extracting the tables within an Excel file by default but not processing the entire Excel file as a table, but this skipping table extracting by default solution makes sense as a next best 👍

unstructured/partition/auto.py
```diff
@@ -126,7 +126,7 @@ def partition(
     encoding: Optional[str] = None,
     paragraph_grouper: Optional[Callable[[str], str]] = None,
     headers: Dict[str, str] = {},
-    skip_infer_table_types: List[str] = ["pdf", "jpg", "png"],
+    skip_infer_table_types: List[str] = ["pdf", "jpg", "png", "xls", "xlsx"],
```
shreyanid, line 129: Could you add in the README that Excel tables are now also not extracted by default?

amanda103: I don't see an existing reference to this in the README. Do you mean in the docs? https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/api.rst#table-extraction-for-other-filetypes

shreyanid (approved):
LGTM!

unstructured/partition/auto.py
amanda103: ah yes good find! that'll have to be a separate PR for the api readme 👍🏻

PR #1397, opened 2023-09-13
chore: add languages parameter for html partition functions

shreyanid:
### Summary
In order to support language functionality other than Tesseract OCR, we want to represent languages provided for either partitioning accuracy or OCR as a standard list of langcodes as strings. In the case of non-image functions where ocr_languages is not a parameter, we can pass in language for later usage in element type detection.

### Details
Adds languages (a list of strings) as a parameter to `partition_html` and `convert_and_partition_html`. 

### Test
Only adds a parameter as currently unused, does not make any changes to partitioning yet.

`make test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.15-dev7
+## 0.10.15-dev8
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


shreyanid:
Closing since the parameter isn't being used yet.

PR #1399, opened 2023-09-13
chore: refactor languages parameter for text_type functions

shreyanid:
### Summary
In order to support language functionality other than Tesseract OCR, we want to represent languages provided for either partitioning accuracy or OCR as a standard list of langcodes as strings. To identify element types such as NarrativeText and Title, continue the refactor into functions that use language checks to determine those potential classifications.

### Details
Replaces `language` with `languages` (a list of strings) as a parameter to `is_possible_narrative_text` and `is_possible_title`. Removes the `UNSTRUCTURED_LANGUAGE` environment variable and its usages.

### Test
Call `is_possible_narrative_text` and `is_possible_title` with text in a variety of languages and different inputs for `languages`. The resulting element classifications should be no different from the current outputs.

ex: see `test_text_type_handles_multi_language_examples` in `test_unstructured/partition/test_text_type.py`.

### DIFF
--- a/unstructured/partition/text_type.py
+++ b/unstructured/partition/text_type.py
@@ -48,7 +48,7 @@ def is_possible_narrative_text(
         The minimum proportion of alpha characters the text needs to be considered
         narrative text
     languages
-        The list of languages present in the document. Defaults to "eng" for English
+        The list of languages present in the document. Defaults to ["eng"] for English
     language_checks
         If True, conducts checks that are specific to the chosen language. Turn on for more
         accurate partitioning and off for faster processing.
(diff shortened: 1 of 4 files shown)


shreyanid:
question for reviewer: how do I ensure the UNSTRUCTURED_LANGUAGE environment variable has been removed, other than removing all its usages?

unstructured/partition/text_type.py
```diff
@@ -47,8 +47,8 @@ def is_possible_narrative_text(
     non_alpha_threshold
         The minimum proportion of alpha characters the text needs to be considered
         narrative text
-    language
-        The two letter language code for the text. defaults to "en" for English
+    languages
+        The list of languages present in the document. Defaults to "eng" for English
```
amanda103, line 51: nit. `Defaults to ["eng"]` since it's a list

amanda103 (approved):
tiny nit, otherwise lgtm! just remember to update the changelog/version for whichever you merge second :)

amanda103:
> UNSTRUCTURED_LANGUAGE

Sorry missed this - let me do some digging to recall where this is being set. I believe it'll be a separate PR though.

shreyanid:
update for posterity: we could not find where the UNSTRUCTURED_LANGUAGE env var was being set/used, so just added a note in CHANGELOG that it is being removed.

PR #1408, opened 2023-09-13
update constraints to pin weaviate during ci

badGarnet:
This PR ensures the version for `weaviate` is consistent in CI testing. Latest (3.24.1) is not compatible with our test needs and last version that run successfully in CI is 3.23.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,10 +145,6 @@ jobs:
         sudo add-apt-repository -y ppa:alex-p/tesseract-ocr5
         sudo apt-get install -y tesseract-ocr tesseract-ocr-kor
         tesseract --version
-        # NOTE(robinson) - Installing weaviate-client separately here because the requests
-        # version conflicts with label_studio_sdk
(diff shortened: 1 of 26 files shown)


cragwolfe (approved):
🙏 🙏

qued:
I'm not sure pinning is the solution here. I get the impression they are permanently removing client-side schema validation, so it's no good hanging on to a test on our side that depends on that.

badGarnet:
> I'm not sure pinning is the solution here. I get the impression they are permanently removing client-side schema validation, so it's no good hanging on to a test on our side that depends on that.

wait, does the validation actually sends data to some server somewhere?

qued:
I'll reiterate I think it's better to skip/remove the test than pin to a version of Weaviate that can potentially fall behind. I don't think it does us any good, for example, to validate a schema against a validation check that's out of date. (And if it's not out of date, then our schema is still valid whether we are testing it or not.)

That being said, this needs to go through so other PRs aren't blocked, so I'm not going to hold things up.

PR #1413, opened 2023-09-14
ci: add new release version alert

ryannikolaidis:
GHA workflow that detects if a commit in a PR branch bumps the version to a release (non-dev) version. Since we only want this to alert once, it only runs when the version file changes in a given commit (i.e. subsequent changes unrelated to the version will not trigger the alert). 

Example from workflow running with version bump in commit: https://github.com/Unstructured-IO/unstructured/actions/runs/6193995455/job/16816334805


cragwolfe (approved):
🙌

qued (approved):
LGTM, very cool!

PR #1422, opened 2023-09-14
docx refactor

scanny:
Reviewers: I recommend reviewing commit-by-commit or just looking at the final version of `partition/docx.py` as View File.

This refactor solves a few problems but mostly lays the groundwork to allow us to refine further aspects such as page-break detection, list-item detection, and moving python-docx internals upstream to that library so our work doesn't depend on that domain-knowledge.


unstructured/partition/docx.py
```diff
@@ -134,305 +151,625 @@ def partition_docx(
...
-            emphasized_texts = _get_emphasized_texts_from_paragraph(paragraph)
-            emphasized_text_contents, emphasized_text_tags = _extract_contents_and_tags(
-                emphasized_texts,
+
+class _DocxPartitioner:
+    """Provides `.partition()` for MS-Word 2007+ (.docx) files."""
+
+    # TODO: I think we can do better on metadata.filename. Should that only be populated when a
+    #       `metadata_filename` argument was provided to `partition_docx()`? What about when not but
+    #       we do get a `filename` arg or a `file` arg that has a `.name` attribute?
```
Klaijan, line 178: I think it is ok to have it the way it is (e.g. `None` when no argument was provided). One way to think about this is maybe we have another field to record the filename if provided, and fall back to `.name` when not found. But I'll take more opinions on this too!

scanny: @Klaijan Present behavior for `partition_docx(filename="some_file.docx")` is to record `None` for `metadata.filename`. I'm thinking `metadata_filename` should take precedence, but isn't `filename` better than `None` for the case where they're just partitioning a file on a filesystem?

unstructured/partition/docx.py
```diff
@@ -41,89 +54,93 @@
...
+    return elements
 
 
 @process_metadata()
 @add_metadata_with_filetype(FileType.DOCX)
 @add_chunking_strategy()
 def partition_docx(
     filename: Optional[str] = None,
-    file: Optional[Union[IO[bytes], SpooledTemporaryFile]] = None,
+    file: Optional[Union[BinaryIO, SpooledTemporaryFile[bytes]]] = None,
```
Klaijan, line 137: I tried to run this locally and encountered an error. Or is it just me?

```
file: Optional[Union[BinaryIO, SpooledTemporaryFile[bytes]]] = None,
TypeError: 'type' object is not subscriptable
```

scanny: @Klaijan Hmm, good catch. What Python version are you running? I think this problem only arises in 3.7 and 3.8 which would explain why I didn't see it myself. Fix is easy though, just adding:
```python
```
should do the trick as long as we're support 3.7 and 3.8.

scanny: Ok, yeah, there are a few problems like this when running under 3.7. I'll get those fixed up. Thanks @Klaijan :)

Klaijan:
All looks good to me from the walk through and local tests. I still do see some test failures for the formatting, but that could be easily fix.

unstructured/partition/docx.py
```diff
@@ -134,305 +170,625 @@ def partition_docx(
...
+    def _sectPrs(self) -> Sequence[CT_SectPr]:
+        """All w:sectPr elements in document, in document-order."""
+        return self._sectPr.xpath(
+            "/w:document/w:body/w:p/w:pPr/w:sectPr | /w:document/w:body/w:sectPr"
+        )
 
-        headers_and_footers.append((headers, footers))
 
-    return headers_and_footers
+# == monkey-patch docx.text.Paragraph.runs ===========================================
```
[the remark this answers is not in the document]
scanny: Hi Klaijan, that "divider" comment is meant to visually segregate the `_get_paragraph_runs()` function and the `Paragraph.runs = property(...` statement below it because those actually modify the `python-docx` library code and need to move upstream. So this comment line and the one on the last line of the file are a "matched pair" indicating the boudaries of that block.

I placed it at the bottom for two reasons, one because it is not referenced by the docx-partitioning code in the module and two because it will be going away soon when we move it upstream.

Klaijan (approved):
LGTM

qued (approved):
LGTM, just one suggestion related to consistency with how we've been notifying users when they are missing a dependency.

unstructured/partition/docx.py
```diff
@@ -103,35 +103,44 @@ def convert_and_partition_docx(
         Determines whether or not metadata is included in the metadata attribute on the elements in
         the output.
     """
-    if filename is None:
-        filename = ""
+    if "pypandoc" not in globals():
```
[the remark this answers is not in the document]
scanny: Ah, good to know, thanks @qued I'll check that out and replace this :)

PR #1425, opened 2023-09-14
add concurency setting to cancel previous runs if new one is triggerered

tabossert:
from the same ref


cragwolfe:
We can do this, but is not a no-brainer. Sometimes there is value in letting a test from a previous commit that is going for ~8 minutes run to completion.

It is, of course, bad when people push a bunch of successive commits, and this helps with that.

tabossert:
closed due to upgrade

PR #1426, opened 2023-09-14
fix: Pass partition_image kwargs downstream

qued:
`partition_pdf` allows for passing a `model_name` parameter. Given the similarity between the image and PDF pipelines, the expected behavior is that `partition_image` should support the same parameter, but `partition_image` was unintentionally not passing along its `kwargs`. This was corrected by adding the kwargs to the downstream call.

#### Testing:

```python

output1 = partition_image("example-docs/layout-parser-paper-fast.jpg", model_name="detectron2_onnx")
output2 = partition_image("example-docs/layout-parser-paper-fast.jpg", model_name="yolox")

# These shouldn't be the same, since they were produced using different models.
assert output1 != output2

```
The assertion should fail on `main`, but pass on this branch.


unstructured/partition/image.py
```diff
@@ -81,4 +81,5 @@ def partition_image(
         languages=languages,
         strategy=strategy,
         metadata_last_modified=metadata_last_modified,
+        **kwargs,
```
badGarnet, line 84: probably worth adding to docstring where `kwargs` go? (so use can anticipate what can go in?)

badGarnet (approved):

PR #1433, opened 2023-09-15
FEAT: Add hierarchy for indented text

mallorih:
This PR uses the first coordinate point to determine if the text is in the right or left column (for 2 column documents) based on the median of the coordinates. Then if first coordinate point of (previous element/current element) >= 0.79, the `parent_id` is of the current element is modified to the `id` of the previous element (the text above it). This code skips elements that are indented and have an element type of `Title`. 

**Code Gotchas**
- The code skips elements with `Title` category, so if a piece of text is incorrectly labeled the code will not modify the `parent_id`.
- Get confused about text in tables and will incorrectly modify `parent_id` on these elements.

**Tested**
- Tested on `YOLOX: Exceeding YOLO Series in 2021` and `Dense Passage Retrieval for Open-Domain Question Answering` papers which are pdfs.


ryannikolaidis:
worth adding a unit test to cover this?

unstructured/partition/common.py
```diff
@@ -183,6 +186,69 @@ def normalize_layout_element(
...
+            right.append((ele, "right"))
+
+    # Assign parent id to indented text
+    prev_ele = []
+    i = 0
+    for ele, col in left:
+        if prev_ele:
+            if (
+                ele.metadata.coordinates.points[0][0] - prev_ele.metadata.coordinates.points[0][0]
+            ) > 0.79 and isinstance(ele, Title):
```
qued, line 216: What's the significance of 0.79 here? Is that like a minimum distance that if the left edges of two elements are more than that far apart we consider it "indented"?

unstructured/partition/common.py
```diff
@@ -183,6 +186,69 @@ def normalize_layout_element(
...
+    # Split text from multicolumn page in left and right lists
+    # Will not work for more than 2 columns
+    for ele in elements:
+        if ele.metadata.coordinates.points[0][0] <= median:
+            left.append((ele, "left"))
+        else:
+            right.append((ele, "right"))
+
+    # Assign parent id to indented text
+    prev_ele = []
```
qued, line 210: This looks like a way of assigning a null type to `prev_ele`, and then `prev_ele` gets updated after the first run through the code. I think `None` would be a better null type to assign rather than an empty list.

unstructured/partition/common.py
mallorih: 0.79 is the threshold for if something is on the left or right side of the page.

PR #1441, opened 2023-09-17
fix: fix ingest paddle hanging issue

yuming-long:
## Summary

Ingest tests are having paddle OOM issue which cause the tests to hang forever. The fix here is to remove paddle from ci and set both OCR env `TABLE_OCR` and `ENTIRE_PAGE_OCR` to `tesseract`. (will have follow up PR to investigate why this is failing)

## Test
please check ingest tests in CI


tabossert (approved):
LGTM!

ryannikolaidis (approved):
🎉 thank you!

Issue #1317, opened 2023-09-06
feat: add unstructured-ingest support for compressed files


ryannikolaidis:
Handle at least zip and tar for now. Every decompressed file of supported file type should be partitioned. We’ll need to do this for all object stores, so if any of the flow is reusable/generalizable it’s worth trying to make it so. 

**Definition of Done**

Connectors handle zip and tar files, all of the contents are decompressed and every supported file type is processed.

PR #1443, opened 2023-09-18
Roman/fsspec compression support

rbiseck3:
### Description
As part of the `get_ingest_docs` method of the `FsspecSourceConnector`, any file that ends in an extension that is recognized as either a zip or tar file will be first pulled down, decompressed and those files will be added to the list of docs as `LocalIngestDoc` to be processed alongside all of the original `FsspecIngestDoc`. This adds a bit of overhead to the creation of the Fsspec connector but then all docs can be parallelized using multiprocessing Pool. 

Closes out https://github.com/Unstructured-IO/unstructured/issues/1317

### DIFF
--- a/unstructured/ingest/connector/fsspec.py
+++ b/unstructured/ingest/connector/fsspec.py
@@ -1,10 +1,17 @@
+import copy
 import os
 import re
+import tarfile
 import typing as t
+import zipfile
 from contextlib import suppress
 from dataclasses import dataclass, field
 from pathlib import Path
 
+from unstructured.ingest.connector.local import (
+    LocalSourceConnector,
+    SimpleLocalConfig,
+)
 from unstructured.ingest.error import SourceConnectionError
 from unstructured.ingest.interfaces import (
     BaseConnectorConfig,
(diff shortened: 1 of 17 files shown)


unstructured/ingest/connector/fsspec.py
```diff
@@ -209,16 +220,104 @@ def _list_files(self):
...
+            with tarfile.open(filename, "r:gz") as tfile:
+                tfile.extractall(path=path)
+            return path
+        else:
+            raise ValueError(
+                "filename {} not a recognized compressed extension: {}".format(
+                    filename,
+                    ", ".join(zip_file_extensions + tar_file_extensions),
+                ),
+            )
```
ryannikolaidis, line 252: make this re-usable/common/accessible to other connectors that might support compressed files in the future?

unstructured/ingest/connector/fsspec.py
```diff
@@ -209,16 +220,104 @@ def _list_files(self):
...
+        local_connector = LocalSourceConnector(
+            connector_config=SimpleLocalConfig(
+                input_path=path,
+                recursive=True,
+            ),
+            read_config=new_read_configs,
+            partition_config=new_partition_configs,
+        )
+        local_connector.initialize()
+        return local_connector.get_ingest_docs()
```
ryannikolaidis, line 277: does this need to be FsspecIngestDoc specific? could this be a mixin so other connectors can reuse?

unstructured/ingest/connector/fsspec.py
```diff
@@ -209,16 +220,104 @@ def _list_files(self):
...
+
     def get_ingest_docs(self):
```
ryannikolaidis, line 279: any of this logic reusable outside of fsspec?

ryannikolaidis (commented):
changes look good. obviously primary question here is: how much of this can we make common such that it will be a much lighter touch for any addition non-fsspec connectors to support this

unstructured/ingest/connector/fsspec.py
potter-potter: Can you write a unit test for this? Just to make sure it does what we all expect.

PR #1453, opened 2023-09-19
feat: introduce language detection function for text partitioning function

shreyanid:
### Summary
Uses `langdetect` to detect all languages present in the input document.

### Details
- Converts all language codes (whether user inputted or detected using `langdetect`) to a standard ISO 639-3 code.
- Adds `languages` field to the metadata
- Will revisit how to nonstandardly represent simplified vs traditional Chinese scripts internally (separate PR).
- Update ingest test results to add `languages` field to documents. Some other side effects are changes in order of some elements and changes in element categorization

### Test
You can test the detect_languages function individually by importing the function and inputting a text sample and optionally a language:
```
text = "My lubimy mleko i chleb."
doc_langs = detect_languages(text)
print(doc_langs)
```
-> ['ces', 'pol', 'slk']


test_unstructured/partition/test_lang.py
```diff
@@ -45,3 +45,24 @@ def test_prepare_languages_for_tesseract_warns_non_tesseract_language(caplog):
...
+    languages = ["en"]
+    assert lang.detect_languages(text, languages) == ["eng"]
+
+
+def test_detect_languages_korean_auto():
+    text = "안녕하세요"
+    assert lang.detect_languages(text) == ["kor"]
+
+
+def test_detect_languages_gets_multiple_languages():
```
amanda103, line 66: 👌🏻 👌🏻 👌🏻 👌🏻 👌🏻

unstructured/partition/lang.py
```diff
@@ -201,3 +202,56 @@ def _get_all_tesseract_langcodes_with_prefix(prefix: str):
...
+    languages: List[str] = ["auto"],
+) -> List[str]:
+    """
+    Detects the list of languages present in the text (in the default "auto" mode),
+    or formats and passes through the user inputted document languages if provided.
+    """
+    if text.strip() == "":
+        return ["eng"]  # english as default
+
+    DetectorFactory.seed = 0
```
amanda103, line 227: potentially a comment around this - if I were looking at this the first time it's not clear what this is for

unstructured/partition/lang.py
```diff
@@ -201,3 +202,56 @@ def _get_all_tesseract_langcodes_with_prefix(prefix: str):
...
+
+    # user inputted language
+    if languages and "auto" not in languages:
```
amanda103, line 230: is there a case where users pass in auto and other languages? if so this check would pass and we would continue, that seems odd that would happen but curious if the rest of the flow is ok if so? and if not, let's catch it before here :)

amanda103: psyche - missed the 'not' in languages 🤪 disregard^

unstructured/partition/text.py
```diff
@@ -166,6 +167,7 @@ def partition_text(
     paragraph_grouper: Optional[Callable[[str], str]] = None,
     metadata_filename: Optional[str] = None,
     include_metadata: bool = True,
+    languages: List[str] = ["auto"],
```
amanda103, line 170: should this typing be Optional[List[str]]?

unstructured/partition/text.py
```diff
@@ -201,6 +205,9 @@ def partition_text(
     if text is not None and text.strip() == "" and not file and not filename:
         return []
 
+    if not isinstance(languages, list):
+        raise TypeError("The language parameter must be a list of language codes as strings.")
```
amanda103, line 209: this is getting returned to the user yeah? if so, maybe an example is useful :)

unstructured/partition/text.py
shreyanid: I don't think so - we are using a default "auto" value, so any other non-default value passed in by a user should also not be None. Let me know if that makes sense!

amanda103 (approved):
lgtm!

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
...
 * **Add functionality to save embedded images in PDF's separately as images** This allows users to save embedded images in PDF's separately as images, given some directory path. The saved image path is written to the metadata for the Image element. Downstream applications may benefit by providing users with image links from relevant "hits."
 * **Azure Cognite Search destination connector** New Azure Cognitive Search destination connector added to ingest CLI.  Users may now use `unstructured-ingest` to write partitioned data from over 20 data sources (so far) to an Azure Cognitive Search index.
 * **Improves salesforce partitioning** Partitions Salesforce data as xlm instead of text for improved detail and flexibility. Partitions htmlbody instead of textbody for Salesforce emails. Importance: Allows all Salesforce fields to be ingested and gives Salesforce emails more detailed partitioning.
+* **Add document level language detection functionality.** Introduces the "auto" default for the languages param, which then detects the languages present in the document using the `langdetect` package. Implemented only for the partition_text function to start.
```
amanda103, line 9: Might also want to add the metadata addition here, and potentially link to the standard lang codes that will be returned?

shreyanid: Added the metadata field addition note, and that the info is stored internally as ISO 639-3 codes. This may change in the future to handle special cases for either of the OCR agents though

PR #1459, opened 2023-09-19
feat: adds Azure Cognitive Search (full text) destination connector

rbiseck3:
### Description
New [Azure Cognitive Search](https://azure.microsoft.com/en-us/products/ai-services/cognitive-search) destination connector added. Writes each json element from the created json files via partition and writes that content to an index. 

**Bonus bug fix:** Due to a recent change where the default version of python used in the repo was bumped to `3.10` from `3.8`, this means running `pip-compile` now runs it against that version rather than the lowest we support which is still `3.8`. This breaks the setup for those lower versions because some of the versions pulled in by `pip-compile` exist for `3.10` but not `3.8`. `pip-compile` was updates to run as a script that checks the version of python being used first, which helps guarantee that all dependencies meet the minimum python version requirement.

Closes out https://github.com/Unstructured-IO/unstructured/issues/1466


ahmetmeleq:
Looks great 👍

unstructured/ingest/connector/azure_cognitive_search.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Create a client
+        credential = AzureKeyCredential(self.connector_config.key)
+        self.client = SearchClient(
+            endpoint=self.connector_config.endpoint,
+            index_name=self.write_config.index,
+            credential=credential,
+        )
+
+    def conform_dict(self, data: dict) -> None:
```
ryannikolaidis, line 47: should this be a common function that can be reused by other connectors that need to do this?

ryannikolaidis:
reminder: still needs an end to end ingest test against a live endpoint.

ryannikolaidis:
not for this pr, but in a follow on, let's expose an option to generate embeddings to support vector search https://learn.microsoft.com/en-us/azure/search/vector-search-overview

test_unstructured_ingest/test-ingest-azure-cognitive-search.sh
```diff
@@ -0,0 +1,85 @@
...
+
+echo "sleeping 5 seconds to let index finish catching up after writes"
+sleep 5
+
+# Check the contents of the index
+docs=$(curl "https://utic-test-ingest-fixtures.search.windows.net/indexes/$DESTINATION_INDEX/docs/\$count?api-version=$API_VERSION" \
+  --header "api-key: $AZURE_SEARCH_API_KEY" \
+  --header 'content-type: application/json' | jq)
+
+expected_docs=0
```
ryannikolaidis, line 76: nit: `expected_docs_count`?

test_unstructured_ingest/test-ingest-azure-cognitive-search.sh
```diff
@@ -0,0 +1,85 @@
...
+  --endpoint "$AZURE_SEARCH_ENDPOINT" \
+  --index "$DESTINATION_INDEX"
+
+echo "sleeping 5 seconds to let index finish catching up after writes"
+sleep 5
+
+# Check the contents of the index
+docs=$(curl "https://utic-test-ingest-fixtures.search.windows.net/indexes/$DESTINATION_INDEX/docs/\$count?api-version=$API_VERSION" \
+  --header "api-key: $AZURE_SEARCH_API_KEY" \
+  --header 'content-type: application/json' | jq)
```
ryannikolaidis, line 74: nit: `docs_count`? or `num_docs`?

unstructured/ingest/cli/cmds/azure_cognitive_search.py
```diff
@@ -0,0 +1,92 @@
...
+        raise click.ClickException("parent command missing info name")
+    source_cmd = ctx.parent.info_name.replace("-", "_")
+    runner_fn = runner_map[source_cmd]
+    parent_options: dict = ctx.parent.params if ctx.parent else {}
+    conform_click_options(options)
+    conform_click_options(parent_options)
+    verbose = parent_options.get("verbose", False)
+    ingest_log_streaming_init(logging.DEBUG if verbose else logging.INFO)
+    log_options(parent_options, verbose=verbose)
+    log_options(options, verbose=verbose)
```
ryannikolaidis, line 71: not for this pr, but at some point maybe we should consider if there's anything we can do to reduce the duplicated steps across all of these?

unstructured/ingest/connector/azure_cognitive_search.py
ryannikolaidis: also maybe quick docstring just explaining what/why we're doing this

unstructured/ingest/connector/azure_cognitive_search.py
```diff
@@ -0,0 +1,118 @@
...
+                json_list.extend(json_content)
+        logger.info(
+            f"writing {len(json_list)} documents to destination "
+            f"index at {self.write_config.index}",
+        )
+        try:
+            results = self.client.upload_documents(documents=json_list)
+
+        except azure.core.exceptions.HttpResponseError as http_error:
+            raise WriteError(f"http error: {http_error}") from http_error
```
ryannikolaidis, line 101: do we know the difference between connectivity and unrecoverable issue writing? ...i.e. should a caller wrap WriteError in retry? mostly with custom errors want to distinguish based on potentially actionable issues. if it's connectivity, retrying might help. if it's an issue with writing because something is invalid, retrying won't help

unstructured/ingest/runner/__init__.py
```diff
@@ -24,6 +26,35 @@
 from .slack import slack
 from .wikipedia import wikipedia
 
+runner_map: t.Dict[str, t.Callable] = {
```
ryannikolaidis, line 29: nothing actionable here, but realizing we've got a lot of these lists that we now have to maintain. also not sure that we've got better alternatives, just thinking outloud.

unstructured/ingest/connector/azure_cognitive_search.py
rbiseck3: At least right now this is specific to the schema used by Azure Cognitive Search index, but I could see something similar being used later for other json-based doc stores. Will add doc string but leave this in here for now.

unstructured/ingest/connector/azure_cognitive_search.py
rbiseck3: Can incorporate with the new retry strategy introduced in the other PR

scripts/pip-compile.sh
```diff
@@ -0,0 +1,18 @@
...
+fi
+
+for file in requirements/*.in; do
+  if [[ "$file" =~ "constraints" ]]; then
+    continue;
+  fi;
+  echo "running: pip-compile --upgrade $file"
+  pip-compile --upgrade "$file"
+done
+cp requirements/build.txt docs/requirements.txt
```
ryannikolaidis, line 18: was this needed specifically for this PR?

ryannikolaidis (commented):
qq above, otherwise this looks good

scripts/pip-compile.sh
ryannikolaidis: got it, probably worth mentioning in the pr description as a bonus change for anyone digging back through

rbiseck3: Added

ryannikolaidis (approved):
lgtm! appreciate if you can update the description to call out that the pip-compile change

Issue #1460, opened 2023-09-19
bug/ Invalid coordinates since upgrading to 0.10.15


cdpierse:
**Describe the bug**
I just upgraded to `0.10.15` and have noticed that PDF's that previously worked for me are no longer partitioning successfully and
result in a `ValueError: Invalid coordinates.` being thrown. 

**To Reproduce**
A common PDF I use to test new versions below.
```python
from unstructured.partition.auto import partition

elements = partition(
    url="https://www.apple.com/environment/pdf/Apple_Environmental_Progress_Report_2022.pdf",
    strategy="fast",
)
narrative_texts = [e.text for e in elements if e.category == "NarrativeText"]
```
**Expected behavior**
To return a list of elements. 
**Screenshots**

**Environment Info**
Please run `python scripts/collect_env.py` and paste the output here. 
This will help us understand more about the environment in which the bug occurred.
```
OS version:  macOS-13.4-arm64-arm-64bit
Python version:  3.11.4
unstructured version:  0.10.15
unstructured-inference version:  0.5.28
pytesseract version:  0.3.10
Torch version:  2.0.1
Detectron2 is not installed

[notice] A new release of pip is available: 23.1.2 -> 23.2.1
[notice] To update, run: pip install --upgrade pip

[notice] A new release of pip is available: 23.1.2 -> 23.2.1
[notice] To update, run: pip install --upgrade pip
PaddleOCR is not installed
Libmagic version: file-5.41
magic file from /usr/share/file/magic
Traceback (most recent call last):
  File "/Users/charlespierse/Documents/tactic/genie/../../unstructured/scripts/collect_env.py", line 251, in <module>
    main()
  File "/Users/charlespierse/Documents/tactic/genie/../../unstructured/scripts/collect_env.py", line 243, in main
    libreoffice_version = get_libreoffice_version()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/charlespierse/Documents/tactic/genie/../../unstructured/scripts/collect_env.py", line 171, in get_libreoffice_version
    result = subprocess.run(
             ^^^^^^^^^^^^^^^
  File "/Users/charlespierse/.pyenv/versions/3.11.4/lib/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1462, opened 2023-09-19
fix: coordinates bug on pdf parsing

amanda103:
Addresses: https://github.com/Unstructured-IO/unstructured/issues/1460

We were raising an error with invalid coordinates, which prevented us from continuing to return the element and continue parsing the pdf. Now instead of raising the error we'll return early.

to test:
```

elements = partition(url='https://www.apple.com/environment/pdf/Apple_Environmental_Progress_Report_2022.pdf', strategy="fast")
```


unstructured/partition/pdf.py
```diff
@@ -761,7 +761,7 @@ def check_coords_within_boundary(
         a float ranges from [0,1] to scale the horizontal (x-axis) boundary
     """
     if not coord_has_valid_points(coordinates) and not coord_has_valid_points(boundary):
-        raise ValueError("Invalid coordinates.")
+        return False
```
cragwolfe, line 764: worth adding a trace detail message i think, like: https://github.com/Unstructured-IO/unstructured/blob/9a3e24fcbb4efef30d5f046952e794666345b7ea/unstructured/partition/utils/sorting.py#L70
.
is there a single page this fails on that could be added to a test? one could extract the problematic page with something like:

    qpdf --empty --pages orignal.pdf 89 -- ~/tmp/docs/single-page-p89.pdf

finally, does not need to be addressed in this PR, but what is the why for the invalid coordinate? is it negative values? (which i think are actually legit in PDF's in certain cases)

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 ### Fixes
 
+* **GH Issue 1460** Fixes a bug that prevented certain pdf's from bypassing new coordinate logic
```
cragwolfe, line 12:
```suggestion
* ***Fixes an issue that caused a partition error for some PDF's.** Fixes GH Issue 1460 by bypassing a coordinate check if an element has invalid coordinates.
```

unstructured/partition/pdf.py
amanda103: Sounds good - I'll add trace detail and a test!

And yes - the values that it was erroring on were negative.

amanda103: here's where it was failing: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/utils/sorting.py#L30

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+* ***Fixes an issue that caused a partition error for some PDF's.** Fixes GH Issue 1460 by bypassing a coordinate check if an element has invalid coordinates.
```
cragwolfe, line 13:
```suggestion
* ***Fixes an issue that caused a partition error for some PDF's.** Fixes GH Issue 1460 by bypassing a coordinate check if an element has invalid coordinates.

```
newline

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.10.16-dev2
+## 0.10.16
```
cragwolfe, line 2:
```suggestion
```
rm line

cragwolfe (approved):
🎉

Issue #1285, opened 2023-09-01
tests: more robust element json serialization/deserialization tests


cragwolfe:
**Is your feature request related to a problem? Please describe.**
Tests exists to ensure that, in pseudo-code:

    output.json = elements_to_json(partition(raw_doc.<extension>))
    partition(output.json) == partition(raw_doc.<extension>)

However, it is testing individual element objects, which themselves do not compare all metadata fields in the [overloaded \_\_eq__ operator](https://github.com/Unstructured-IO/unstructured/blob/f6a745a/unstructured/documents/elements.py#L417-L423).

**Describe the solution you'd like**

Rather than comparing the element objects, compare the serialized json strings of the LHS and RHS. Pretty much all the tests that need to be updated are in: https://github.com/Unstructured-IO/unstructured/pull/1187

scanny:
Fixed in https://github.com/Unstructured-IO/unstructured/pull/1681

scanny:
Merged on Oct 12, 2023.

PR #1468, opened 2023-09-20
roman/define json schema

rbiseck3:
### Description
Add schema that can be used for validation of the json output produced via partition.

### Testing
* Unit test added that validate the element schema across each of the element types we support. 

### Open conversation:
* When, if ever, should this schema be enforced to make sure any breaking changes to it are caught? 
* Should this be published in our docs/README? And if so, have a reference to it rather than a copy-paste of the python.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## 0.10.17-dev1
+## 0.10.17-dev2
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


newelh:
> When, if ever, should this schema be enforced to make sure any breaking changes to it are caught?

There should probably be a test that validates the schema against a mocked element with all the metadata

cragwolfe:
FWIW, the schema itself is less important than proving (and enforcing through tests) that de-serializing a json document is the same as partitioning the original document, per https://github.com/Unstructured-IO/unstructured/issues/1285 . Different strategies and doc types will also have metadata differences, so it's also important not to be too strict.

rbiseck3:
I would argue that both are important. By providing a concrete schema, it allows users to understand how to map the content to any downstream datastore. It also provides some guarantee of consistency as we modify the element content in processes like the ingest cli (outside the partition method) where metadata is added.

PR #1470, opened 2023-09-20
Add languages to metadata

newelh:
## Summary
Adds languages to element metadata. 

```
@dataclass
class ElementMetadata:
      coordinates: Optional[CoordinatesMetadata] = None
      data_source: Optional[DataSourceMetadata] = None
      filename: Optional[str] = None
      file_directory: Optional[str] = None
      last_modified: Optional[str] = None
      filetype: Optional[str] = None
      attached_to_filename: Optional[str] = None
      parent_id: Optional[Union[str, uuid.UUID, NoID, UUID]] = None
      category_depth: Optional[int] = None
+   languages: Optional[List[str]] = None
...
```

## Testing
```
elements = partition(filename=filename, strategy="ocr_only", languages=["eng"])
assert elements[0].metadata.languages == ["eng"]
```


newelh:
@shreyanid, can you verify, does your branch currently support language auto-detection?

unstructured/documents/elements.py
```diff
@@ -137,6 +137,9 @@ class ElementMetadata:
     parent_id: Optional[Union[str, uuid.UUID, NoID, UUID]] = None
     category_depth: Optional[int] = None
 
+    # Languages in element. TODO(newelh) - More strongly type languages
+    languages: Optional[List[str]] = None
```
shreyanid, line 141: To answer your question, yes we do currently support "auto" as an an input value to indicate auto language detection. I've been enforcing the type of `languages` to be non-Optional (`List[str] = ["auto"]`), let me know if you have cases where a None value makes sense

newelh: Not using optional will break metadata functionality. By more strongly type I perhaps meant an enum for each language so we could enforce validation for functions that take a "language" and have stronger typing throughout.

newelh:
This will likely not past ingest tests when moving to merge into main. I suspect that this is because it will include languages in the outputs for many documents which currently do not have the language field in the json it's being tested against

unstructured/chunking/title.py
```diff
@@ -84,7 +84,7 @@ def chunk_by_title(
                                 item["start"] += start_char
                                 item["end"] += start_char
 
-                        if i > 0:
+                        if i > 0 and _value != value:
```
badGarnet, line 87: what is this extra logic solving? adding specified data into other objects' metadata? asking since this doesn't seem to have anything to do with language so likely found a bug along the way?

test_unstructured/partition/test_auto.py
```diff
@@ -385,6 +385,12 @@ def test_auto_partition_formats_languages_for_tesseract():
         )
 
 
+def test_auto_partition_element_metadata_user_provided_languages():
+    filename = "example-docs/chevron-page.pdf"
+    elements = partition(filename=filename, strategy="ocr_only", languages=["eng"])
+    assert elements[0].metadata.languages == ["eng"]
```
badGarnet, line 391: maybe also add a test for no input and assert the `languages` attribute is `None`?

unstructured/chunking/title.py
newelh: Yes, I can go add some more detailed comments as to why we need to add this but essentially if the list is `['eng']` and there's 15 or so elements, for each run of the loop, the number of elements in the list would double `['eng', 'eng']`. `['eng', 'eng', 'eng', 'eng']` etc. This would eventually cause a memory error if since the number of items in the list is 2 ^ n where n is the number of elements in the document. Perhaps this isn't the right fix, but this check prevents the list from extending itself with itself if the value is already there.

badGarnet: then wouldn't it be better to not use `extend` below but use set operations to ensure we don't have duplicates? From the interface the value here for language can be a list of languages so we can have `['eng', 'ita']` as `_value` and `['eng']` as `value`. In this case they won't be equal so we'd extend one with the other which results in double `'eng'`.

newelh: Yeah that's a great idea. I'll implement that. One of my concerns is that I'm not aware how else this function is used -- what other metadata list items are supposed to be run through this and is there expected behavior for those list items that I'm missing? I'm assuming there's proper test coverage for this but can never be certain. I think using set ops is a better alternative though and should maintain the desired functionality

shreyanid, line 87: can you remove this change? it's not present on main so I don't want to introduce it

newelh: https://github.com/Unstructured-IO/unstructured/pull/1470#discussion_r1333392891

unstructured/documents/elements.py
shreyanid: update per discussion: all metadata types are optional so this will follow that pattern, but that doesn't affect the typing fot the actual `languages` param

unstructured/chunking/title.py
shreyanid: ah just saw it thanks

test_unstructured/partition/test_auto.py
shreyanid: if the language parameter is not passed to partition_text (the only function that implements language detection so far), the default value will be auto -> the resulting language will be detected, not `None`. still good to test for no input but the result will be an actual language list. the change has not yet been extended to other partition functions and auto

unstructured/chunking/title.py
newelh: After further evaluation, because value could be any type, using a set will raise an exception because lists and dictionaries can't be hashed in a set operation. Instead we can iterate through the elements and append them to _value if they're not in the list.

test_unstructured/partition/test_auto.py
badGarnet: wait, @newelh I think @shreyanid is saying we should add a test without input for `languages` and test the output. And I think that is important to test with and without input for the kwarg. The rest of this PR looks good.

newelh: Yep I added a test here: https://github.com/Unstructured-IO/unstructured/commit/d3bfa02d143fe6412a994bfeb606eb815da2c67f

newelh: But I can also add one with the missing kwarg for auto

newelh:
@shreyanid Before we merge this could you confirm this mapping of what the expected results are for each combination of strategy and present language parameter? I want to ensure our test coverage properly represents what we're expecting.

### Expected language metadata for an english document
| Strategy                                    | Auto      | Fast      | OCR Only | Hi Res |
|------------------------------|---------|--------|-----------|-------|
|Language Param not Provided | ["eng"] if Fast or OCR Only  | ["eng"] | ["eng"] (default) |  X  |
|Language Param = ["eng"].      | ["eng"] | ["eng"] | ["eng"]    | ["eng"]|

badGarnet (approved):
changes look good; let's wait for @shreyanid to confirm the behavior; thanks for checking on that!

shreyanid:
The row where English is provided looks correct. When the language param is not provided, I believe the results should also all be English (including hi_res).

When the language param is not supplied, the default value is "auto" language, which calls the language detection package to determine that the document is in fact english. This flow is strategy independent, but if you have seen otherwise let me know.

test_unstructured/partition/test_auto.py
shreyanid: auto partition does not yet have language detection, so this partition_text test looks good for now! thank you

newelh:
Is it dependent on the document type at all? Or should this be the case for all partitioners of all strategies. You mentioned earlier that auto-detection is only available for partition_text. I'm trying to prevent undefined behavior where a user either does or doesn't provide a language and the language that comes in the metadata matches or not. Under what situations is the language being changed, assumed or processed based on what was provided -- how can we make that obvious to users and test each case? For example, if I partition a word document and specify `languages=["fra"]` under the current implementation what's the flow for this being processed? 

1. Under what cases is auto-detection - OR - a default language being assigned? What strategies and what partitioners?
2. Does a user provided parameter always override auto-detection or default languages?
3. Are there cases where a user provides a languages list where it isn't used? What strategies and what partitioners?
4. Should metadata always match what a user provided, whether the user's input was used or not?

shreyanid:
good questions!

No, it is mostly not dependent on document type. The language detection function only takes in text, which is just from whatever text is generated by partitioning a document. Since we need text available for language detection, this change will initially only be applied to non-image based documents (everything but pdf and image). My current PR that you are merging this one into only made changes to partition_text, but a follow on PR will be made to continue this pattern to all other non-image-based document partition functions. So yes, the auto language flow I described should be the case for all strategies, and all partitioning functions except pdf and image (not due to the strategy involved, but due to the lack of text available before OCR to do detection).

For your example, partition_docx does not currently implement language detection, but following the same pattern established with partition_text, the flow would simply take the user's input (since it does not contain "auto" in languages), standardize it to the 3-letter ISO 639-3 code if not already standardized, and return it. No language detection will occur in the flow where a user specifies language input.

So the language detection only occurs when the user does not specify a list of languages (or in some weird case where the user specifies "auto" in the list along with other languages, at which point the other inputted languages will be ignored).

shreyanid:
1. auto detection occurs when "auto" is present somewhere in the parameter. This is the default behavior unless overridden by the user. ex. ["auto"], ["eng", "auto"] or no provided input (which defaults to ["auto"]) all lead to auto language detection. This occurs for all strategies, and all documents that have text available (not pdf or image)

2 and 3. A user provided parameter always overrides auto-detection. The ONLY case where it does not is if the user provided parameter contains the value "auto" (which would be weird to include along with other languages, but it is a possibility). 
ex. 
user provides ["en"] or ["rus", "spa"] -> no language detection
user provides ["auto"] or ["auto", "eng"] -> language detection is triggered by the "auto" value

4. Metadata will match the output of the `detect_languages` function, which follows the flow described above. This output should be standardized, and should NOT contain "auto" since a language will have been detected instead. So no, it may not always exactly match the user input, depending on standardization or the presence of "auto".

newelh:
Ok, I think I'm ok with us merging this as-is with the caveat that in your branch there will be temporary logging that let's a user know that they've supplied a language parameter that's not being used by the partitioner. Also, any additional tickets cut to add the language detection functionality include updating the language metadata field for those partitioners.

shreyanid:
Will reflect the logger warning before merging my change, and will include updating the metadata when new partitioners implement the language detection flow!

PR #1475, opened 2023-09-20
Potter/improved salesforce partitioning

potter-potter:
Based on Ryan's improved salesforce partitioning.

htmlbody doesn't partition well so we used a hack until html partitioning improves.


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.10.17-dev3
+
+### Enhancements
+
+* **Improves salesforce partitioning** Partitions Salesforce data as xlm instead of text for improved detail and flexibility. Partitions htmlbody instead of textbody for Salesforce emails. Importance: Allows all Salesforce fields to be ingested and gives Salesforce emails more detailed partitioning.
+
+### Features
+
+### Fixes
```
ryannikolaidis, line 9: description here is good, but just need to bump the 0.10.17-dev2 below to 0.10.17-dev3 and include this enhancement note in that existing section

unstructured/ingest/connector/salesforce.py
```diff
@@ -242,9 +130,15 @@ def create_eml(self, email_json: t.Dict[str, t.Any]) -> str:
             from_email=email_json.get("FromAddress"),
             to_email=email_json.get("ToAddress"),
             textbody=email_json.get("TextBody"),
+            # TODO: This is a hack to get emails to process correctly.
+            # The HTML partitioner seems to have issues with <br> and text without tags like <p>
+            htmlbody=email_json.get("HtmlBody", "")  # "" because you can't .replace None
+            .replace("<br />", "<p>")
+            .replace("<body", "<body><p"),
```
ryannikolaidis, line 137: is the tag `<p` meant to be left open like that? i.e. not `<p>`?

ryannikolaidis: should we also be replacing the `</body>` with `</p></body>`

potter-potter: Yeah. This is not pretty for sure. The open `<p` is completed by the style tags and other stuff that I can't reliably predict (unless I do a regex which may be better.

```
<body style="height: auto; min-height: auto;">
```
turns to 
```
<body><p style="height: auto; min-height: auto;">
```

The partitioning doesn't seem to care if there are </p>

ryannikolaidis: ohhh, got it, I missed the fact that the `<body` tag was also dangling. yea, not pretty, but seems to do the trick and hopefully just temporary. thanks for clarifying!

CHANGELOG.md
ryannikolaidis: @potter-potter bump ^ (also probably need to merge up now)

ryannikolaidis (approved):
lgtm!

PR #1476, opened 2023-09-21
pptx refactor

scanny:
Reviewers: I recommend reviewing commit-by-commit or just looking at the final version of  partition/pptx.py using as View File.

This refactor solves a problem or two, the big one being recursing into group-shapes to get all shapes on the slide, but mostly lays the groundwork to allow us to refine further aspects such as list-item detection, off-slide shape detection, and image-capture going forward.


unstructured/file_utils/filetype.py
```diff
@@ -1,14 +1,17 @@
 from __future__ import annotations
 
+import enum
+import functools
 import inspect
 import json
 import os
 import re
 import zipfile
-from enum import Enum
```
newelh, line 8: Any particular reason for switching to `import enum` from `enum import Enum`? I believe the latter is what's suggested in the python documentation and how it's used throughout unstructured

unstructured/partition/pdf.py
```diff
@@ -262,7 +265,7 @@ def partition_pdf_or_image(
             layout_elements = []
             for el in _layout_elements:
                 if hasattr(el, "category") and el.category == "UncategorizedText":
-                    new_el = element_from_text(el.text)
+                    new_el = element_from_text(cast(Text, el).text)
```
newelh, line 268: I'm assuming here that we can safely assume that the element is Text and only a cast is needed for type hinting? Is there any safety issues with the way we're handling these objects?

unstructured/partition/pptx.py
```diff
@@ -58,97 +67,235 @@ def partition_pptx(
...
-    elif file is not None:
-        last_modification_date = get_last_modified_date_from_file(file)
-        presentation = pptx.Presentation(
-            spooled_to_bytes_io_if_needed(
-                cast(Union[BinaryIO, SpooledTemporaryFile], file),
-            ),
+
+    if isinstance(file, SpooledTemporaryFile):
+        file.seek(0)
+        file = io.BytesIO(file.read())
```
newelh, line 83: If the spooled temporary file is being loaded into memory here, doesn't that defeat the point of using spooled temporary files? Also, if spooled temporary file typing has been removed from the file kwargs, why would we need to check the instance of file? are we expecting users to inject a spooled temporary file anyway?

unstructured/partition/pptx.py
```diff
@@ -58,97 +67,235 @@ def partition_pptx(
...
+            include_slide_notes,
+            metadata_filename,
+            metadata_last_modified,
+        )._iter_presentation_elements()
+
+    def _iter_presentation_elements(self) -> Iterator[Element]:
+        """Generate each document-element in presentation in document order."""
+
+        for slide in self._presentation.slides:
+            yield from self._increment_page_number()
```
newelh, line 142: The yield pattern works well for creating specific element types and folding them into a generator. This may be less obvious to people new to the library or with less python experience. Could you add a bit more in-line documentation for how this pattern works?

unstructured/partition/pdf.py
scanny: Yes. In this case the CI lint mypy error was something like: `Element has no attribute 'text'.` So either all `_layout_elements` items that satisfy the `if` statement on L267 are sub-types of `Text` or this code will raise `AttributeError` at runtime. `_layout_elements` comes from `_partition_pdf_or_image_local()` which is `() -> List[Element]`.

It's possible that `isinstance(e, Text)` can be inferred from the `if hasattr(el, "category") and el.category == "UncategorizedText"` line above, I'm not sure. It's really a question for the "owner" of that code to answer, I didn't want to change the logic and it seemed like it was working before, so my interest was primarily to fix the `mypy` error from CI with the smallest and most localized change possible.

Another alternative, maybe better now that I think about it, would be to insert `assert isinstance(e, Text)` on the line above instead of casting. That might be a fair inference from the presence of `e.category == "UncategorizedText"` for the author.

Either way though, an exception will be raised if `e` is not `Type[Text]` (`Text` or a subtype of it). So the runtime behavior is unchanged. `cast()` calls an empty function at runtime.

unstructured/partition/pptx.py
scanny: I sure thought so! :)

But in speaking with Crag, apparently `SpooledTemporaryFile` has exploded for us in the past when the file is huge (GBs size) and the Python version is below some threshold. So apparently the conversion is legitimately required.

This is essentially what `spooled_to_bytes_io_if_needed()` does, but that function needs some work and is used in a lot of other places so I thought for the cost of a couple lines I'd just pull out the needful here with the idea we might extract a "file-loader" object at some point that would encapsulate this and many of the surrounding lines here.

unstructured/partition/pptx.py
scanny: Sure, good idea :)

unstructured/partition/pptx.py
newelh: Makes sense, could you add an inline todo/note that summarizes that?

unstructured/file_utils/filetype.py
newelh: Looks like we're currently 50/50 on which way we import enums through the lib. So, we could keep this but we should create a follow-up issue to standardize imports

unstructured/partition/pptx.py
scanny: k, turns out this was put in for a particular downstream package (maybe PDF-miner or something) and just got rubber-stamped around it looks like. I'm cutting it out as `python-docx` and `python-pptx` like `SpooledTemporaryFile` just fine and do exactly what the `spooled_to_bytes_io_if_needed()` function does which is read the whole thing once from the top. I am going to leave in the `seek(0)` though (when `file` has `.seek()`), just in case.

scanny: Hmm, okay, I think I discovered the reason and it's that `python-docx` and `python-pptx` both use ZipFile (because the format is based on a zip-archive), and `ZipFile` checks for a `seekable` attribute that `SpooledTemporaryFile` doesn't have. So it turns out to be a real problem after all. And it makes the comment I need to add clear, so that's a plus :)

scanny:
k, Newel, the fix is in. Tests probably take a half-hour to work through, but the only real changes are comments. I'll leave it with you :)

unstructured/partition/pptx.py
newelh: Looks fantastic. Thanks

newelh:
Pulled the tests and checked them out. Looks great. I'll approve

newelh (approved):

PR #1479, opened 2023-09-21
ci: more reliable release version alerts

ryannikolaidis:
This addresses a few issues with the release version alert workflow:

* Missed alerts when the commit with the version change isn't processed by the workflow -- Now we process all changes across the entire branch regardless of which file change in the last commit. In order to prevent re-triggering the Slack alert on each push, it creates a small cache item to validate whether it has already alerted Slack and skips in that case.
* Edge cases where the pr doesn't exist or isn't discovered. -- even though this only runs on PR->main, I have seen this happen in the logs. Now if the PR is not found, we still alert with a link to the corresponding branch.
* Force pushes result in error because the "last commit" is not found. -- The existing implementation relies on only looking at changes in the last commit in order to only trigger the alert on the single commit that changes the version file to a release version. This errors when trying to find the last commit when the user force pushes. This implementation evaluates against all changes in the branch on every commit, so is tolerant to this condition.

## Testing
(I ran tests with the slack alert only echoing for test purposes)

[Example push success](https://github.com/Unstructured-IO/unstructured/actions/runs/6256390493/job/16987144593): logged on a commit with previous version change in a previous commit (after fixing bug with version check itself). 

[Example skip](https://github.com/Unstructured-IO/unstructured/actions/runs/6256425177/job/16987233894)
after running an empty commit: `git commit --allow-empty -m "empty commit" && git push`, immediately following the preceding success. Because it found a duplicate alert had already been "sent" by checking the message cache, it skips.

[Example skip](https://github.com/Unstructured-IO/unstructured/actions/runs/6256442942/job/16987279477) after running:
```
git commit --allow-empty -m "empty commit" && git push
git reset --hard HEAD~1
git push -f origin
```
This demonstrates that it doesn't error when handling the force push case. Note from logs that the version alert would have fired if we hadn't previously alerted (found in cache)


scripts/check-new-release-version.sh
```diff
@@ -12,7 +12,7 @@ function is_non_dev_version {
 function get_main_branch_version {
     local VERSION
     git fetch origin main
-    VERSION=$(git show origin/main:unstructured/__version__.py | grep -o -m 1 -E "(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)")
+    VERSION=$(git show origin/main:unstructured/__version__.py | grep -o -m 1 -E "(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9.-]+)?")
```
ryannikolaidis, line 15: now actually includes -dev from the version on main 🤦

qued (approved):
LGTM as far as examination of the results posted, and nothing about the logic sticks out at me. I'm a little nervous trusting anything to do with github caches, but if it works it works.

ryannikolaidis:
> LGTM as far as examination of the results posted, and nothing about the logic sticks out at me. I'm a little nervous trusting anything to do with github caches, but if it works it works.

thanks! re: github caches, I totally agree. couldn't find a better straightforward path to maintain statefulness. in this case, I think the worst that happens if something goes wrong with the cache is we just get duplicate notifications which doesn't seem too bad, would probably rather caution on extra notifications than none at all.

PR #1480, opened 2023-09-21
Discord fix when a channel is not found.

rvztz:
- **issue**:  Discord connector raises issues when trying to fetch `jump_url`
- **cause**:  The `get_channel` function from the discord client returns None when the given channel couldn't be fetched or it doesn't exists. 
- **fix**: Call `bot.get_channel` only once within the callback function used by the discord bot. If the fetching of the channel fails, we note it earlier.


unstructured/ingest/connector/discord.py
```diff
@@ -88,12 +89,19 @@ async def on_ready():
                 if self.days:
                     after_date = dt.datetime.utcnow() - dt.timedelta(days=self.days)
                 channel = bot.get_channel(int(self.channel))
+                if channel is None:
+                    raise FileNotFoundError("Channel not found.")
```
ryannikolaidis, line 93: do we know for a fact if this is None it's because the channel wasn't found versus a connectivity issue? asking because we hit the error earlier and it seemed to be an ephemeral issue, which would suggest the latter, right? I want to make sure we don't mislead the caller here.

ryannikolaidis: Got it, but then why were we seeing that in CI intermittently? ...because obviously that channel didn't disappear.

ryannikolaidis: also probably good to drop a line to that documentation as comment in the code here, but probably helpful for us to understand why this is the case intermittently in CI. seems like some _other_ case is also resulting in this being None (if None is what triggered the original error we were seeing)

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 * **Fixes a metadata source serialization bug** Problem: In unstructured elements, when loading an elements json file from the disk, the data_source attribute is assumed to be an instance of DataSourceMetadata and the code acts based on that. However the loader did not satisfy the assumption, and loaded it as a dict instead, causing an error. Fix: Added necessary code block to initialize a DataSourceMetadata object, also refactored DataSourceMetadata.from_dict() method to remove redundant code. Importance: Crucial to be able to load elements (which have data_source fields) from json files.
 
+* **Fixes issue when channel raises `None`** Retrieves the channel only once, moves `jump_url` property setting within the message retrieval context.
+
```
ryannikolaidis, line 16: follow structure of the log above this one:
title should clearly state that this fixes issue where channel isn't found by the Discord source connector.
then state the Problem: describe what is currently broken and any relevant background. Then state the Fix: what changed in order to fix this. Finally include the Importance: how this impacts the user.

ryannikolaidis:
looks like we have no actual changes at this point? I know this is still draft, assuming something is in progress?

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 * **Fixes a metadata source serialization bug** Problem: In unstructured elements, when loading an elements json file from the disk, the data_source attribute is assumed to be an instance of DataSourceMetadata and the code acts based on that. However the loader did not satisfy the assumption, and loaded it as a dict instead, causing an error. Fix: Added necessary code block to initialize a DataSourceMetadata object, also refactored DataSourceMetadata.from_dict() method to remove redundant code. Importance: Crucial to be able to load elements (which have data_source fields) from json files.
 
+* **Fixes issue when channel raises `None`** Retrieves the channel only once, moves `jump_url` property setting within the message retrieval context.
```
ryannikolaidis, line 15: should specify that this is wrt Discord connector. should clearly state the Problem, Fix, and Importance (see the line above this)

ryannikolaidis: (see my earlier comment on this)

CHANGELOG.md
```diff
@@ -25,7 +25,7 @@
 should be generated, however the Formula class inherits from Element instead of Text. After this change the element is correctly created with the correct class 
 allowing the document to be loaded. Fix: Change parent class for Formula to Text. Importance: Crucial to be able to load documents that contain formulas.
 * **Fixes Sphinx errors.** Fixes errors when running Sphinx `make html` and installs library to suppress warnings.
-
+* **Fixes issue when channel returns `None`** Problem: Getting the `jump_url` out of the discord `channel` when it's non-existent fails. Fix: property `jump_url`is retrieved within the same context as the messages from the channel. Importance: Avoid possible issues when the connector fails to fetch information about the discord channel.
```
ryannikolaidis, line 28:
```suggestion
* **Fixes issue with Discord connector when a channel returns `None`** Problem: Getting the `jump_url` from a nonexistent Discord `channel` fails. Fix: property `jump_url` is now retrieved within the same context as the messages from the channel. Importance: Avoids cascading issues when the connector fails to fetch information about a Discord channel.
```

ryannikolaidis (approved):

PR #1481, opened 2023-09-21
fix/bad formed formula

benjats07:
@ron-unstructured reported that loading files with:

```

elements_yolox = partition_pdf(filename="1706.03762.pdf", strategy='hi_res', model_name="yolox")
print(elements_yolox)
```

Throws an error. After debugging the execution I found that the issue is that an object of class Formula is being created, however, this class doesn't contain an __init__ method. This PR solves the issue of adding a constructor method with an empty string for the element.

The file can be found at:
https://drive.google.com/drive/folders/1hDumyps0hA4_d-GZxs3Hij15Cpa5fjWY?usp=sharing

After this PR is merged this file is correctly processed


cragwolfe:
@benjats07 , please include a unittest in your PR that would have failed without this fix.

qued (commented):
I actually think the code that was there before will work, but `Formula` should subclass `Text` instead of `Element`, unless I'm missing something.

benjats07:
> I actually think the code that was there before will work, but `Formula` should subclass `Text` instead of `Element`, unless I'm missing something.

Yeah, that is better, just need to reorder the file 😄

test_unstructured/documents/test_base.py
```diff
@@ -25,6 +37,12 @@ def test_get_narrative():
     document.print_narrative()
 
 
+def test_get_formula():
+    document = MockDocumentWithFormula()
+    formula = [e for e in document.elements if isinstance(e, Formula)]
+    assert isinstance(formula[0], Formula)
```
qued, line 43: Small nit, non-blocking, but asserting `isinstance` here doesn't seem necessary as it would have to be to be in this `formula` list. We could also just assert `formula` isn't empty.

qued (approved):
LGTM, just one tiny nit in the test.

test_unstructured/documents/test_base.py
benjats07: Sure, I already made the changes 😄

PR #1492, opened 2023-09-21
feat: add Notion connector retry strategy

rbiseck3:
### Description
In order to add a retry strategy to the notion http calls, leveraging a generic backoff library with some tweaks to pass in values from the CLI.


test_unstructured_ingest/test-ingest-notion.sh
```diff
@@ -22,7 +22,8 @@ PYTHONPATH=. ./unstructured/ingest/main.py \
     --database-ids "122b2c22996b435b9de2ee0e9d2b04bc" \
     --num-processes 2 \
     --recursive \
-    --verbose
+    --verbose \
+    --max-retries 10
```
ryannikolaidis, line 26: with backoff how long is the 10th retry? curious what's the max time this takes to fail (or succeed)

rbiseck3: Set this to a concrete 60 seconds for now.

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
-## 0.10.17-dev3
+## 0.10.17-dev4
 
 ### Enhancements
 
 * **Adds data source properties to SharePoint, Outlook, Onedrive, Reddit, and Slack connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
 * **Add functionality to save embedded images in PDF's separately as images** This allows users to save embedded images in PDF's separately as images, given some directory path. The saved image path is written to the metadata for the Image element. Downstream applications may benefit by providing users with image links from relevant "hits."
+* **Add retry strategy for source connectors** Dynamic retry strategy added with exponential backoff added to notion source connector.
```
ryannikolaidis, line 7: add a sentence on "why this matters"

requirements/base.txt
```diff
@@ -1,9 +1,11 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.10
```
ryannikolaidis, line 2: didn't you call out an issue on compiling with 3.10? assuming you'll hit that when you merge up?

ryannikolaidis (commented):
this looks good. maybe worth a few separate runs to prove this resolves the current, intermittent notion connectivity issues? (since presumably we'll undo the skip test on main when this goes in, don't want to re-break things)

ryannikolaidis:
> this looks good. maybe worth a few separate runs to prove this resolves the current, intermittent notion connectivity issues? (since presumably we'll undo the skip test on main when this goes in, don't want to re-break things)

bump ^ were you able to do this?

ryannikolaidis:
I guess alternatively we could check this in and make sure we leave the test disabled. biggest concern is that I don't want to re-enable the notion test until we're convinced that this resolves the issue

rbiseck3:
Re-enabled notion checks, will let ingest tests run a couple times to validate.

ryannikolaidis:
requires pip install of tiktoken:

```
  File "/Users/ryannikolaidis/Development/unstructured/unstructured/unstructured/ingest/error.py", line 22, in wrapper
    raise cls(cls.error_string.format(str(error))) from error
unstructured.ingest.error.EmbeddingEncoderConnectionError: Error in connecting to the embedding model provider: Could not import tiktoken python package. This is needed in order to for OpenAIEmbeddings. Please install it with `pip install tiktoken`.
```

we should probably extend as an extra included with  langchain and openai which are needed for embedding[ here](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/embed/openai.py#L42-L57)

potter-potter:
Could you add a comment on why you forked the backoff library? I couldn't find the exact reason.

rbiseck3:
@potter-potter, this was to support passing in values via the CLI. The library itself expects setting all values on the decorator, but you can't pass in instance values to a decorator on a method that uses `self`,  so the refactor expects wrapping an instance method and having the values it needs on the `self` object. All the files added here are ones that exist in the original library which the code slightly tweaked.

test_unstructured_ingest/test-ingest.sh
```diff
@@ -57,17 +57,8 @@ trap print_last_run EXIT
...
-    echo "Running ./test_unstructured_ingest/$script"
-    ./test_unstructured_ingest/"$script"
-    set -e
-    echo "--------- FINISHED SCRIPT $script ---------"
-  else
-    echo "--------- RUNNING SCRIPT $script ---------"
-    echo "Running ./test_unstructured_ingest/$script"
-    ./test_unstructured_ingest/"$script"
-    echo "--------- FINISHED SCRIPT $script ---------"
-  fi
```
ryannikolaidis, line 72: should we leave this in in some capacity so that we can quickly disable in the future?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * **Improve hierarchy detection in pptx documents** List items, and other slide text are properly nested under the slide title. This will enable better chunking of pptx documents.
 * **Refactor of the ingest cli workflow** The refactored approach uses a dynamically set pipeline with a snapshot along each step to save progress and accommodate continuation from a snapshot if an error occurs. This also allows the pipeline to dynamically assign any number of steps to modify the partitioned content before it gets written to a destination.
 * **Applies `max_characters=<n>` argument to all element types in `add_chunking_strategy` decorator** Previously this argument was only utilized in chunking Table elements and now applies to all partitioned elements if `add_chunking_strategy` decorator is utilized, further preparing the elements for downstream processing.
+* **Add retry strategy for source connectors** Dynamic retry strategy added with exponential backoff added to notion source connector.
```
ryannikolaidis, line 13: right now this only adds to Notion, correct? headline here might be misleading?

test_unstructured_ingest/test-ingest.sh
rbiseck3: We should probably open up another issue to find a more dynamic approach to omit tests at run time (i.e. CLI variable)

unstructured/ingest/connector/notion/connector.py
```diff
@@ -24,6 +27,13 @@ class SimpleNotionConfig(BaseConnectorConfig):
     database_ids: t.List[str]
     recursive: bool
     api_key: str
+    max_retries: t.Optional[int] = None
+    max_time: t.Optional[float] = None
+
+    def get_retry_strategy(self):
+        if self.max_time or self.max_retries:
+            return RetryStrategy(max_time=self.max_time, max_tries=self.max_retries)
+        return None
```
ryannikolaidis, line 36: are other connectors going to be defining _different_ retry strategies? or mostly re-defining this every time?

unstructured/ingest/connector/notion/client.py
```diff
@@ -16,14 +17,57 @@
...
+        "{kwargs}".format(**details),
+    )
+
+
+def giveup_handler(details):
+    logger.error(
+        "Gave up after {elapsed:0.1f} seconds/{tries} tries "
+        "calling function {target} with args {args} and kwargs "
+        "{kwargs}".format(**details),
+    )
```
ryannikolaidis, line 39: are connectors each going to need to redefine these?

test_unstructured_ingest/test-ingest.sh
ryannikolaidis: right, was thinking more in the meantime

unstructured/ingest/connector/notion/connector.py
rbiseck3: Figured it made sense to start off with this until we look into how the other connectors might use this, and refactor then.

unstructured/ingest/connector/notion/client.py
rbiseck3: Same comment as above, can move out once we know more about how other connectors are going to be using the retry strategy.

CHANGELOG.md
rbiseck3: The enhancement was the addition of the retry strategy itself. The comment associated with the bullet points states it was only applied to the notion connector.

unstructured/ingest/connector/notion/connector.py
ryannikolaidis: cool, yep, makes sense!

CHANGELOG.md
ryannikolaidis, line 13:
```suggestion
* **Add common retry strategy utilities for unstructured-ingest** Dynamic retry strategy with exponential backoff added to notion source connector.
```

ryannikolaidis: right, added suggestion as to how to clarify

ryannikolaidis (approved):
lgtm!

PR #1503, opened 2023-09-22
Add notification that we will be utilizing scarf for docker and python downloads

tabossert:
We've created a custom domain, downloads.unstructured.io that redirects to quay.io
(using https://scarf.sh/). This custom domain allows us to swap the
underlying container registry without impacting users. It also provides
us with important metrics about container and package usage, without surfacing PII
like IP addresses.

Python package follows the same pattern at packages.unstructured.io


cragwolfe (approved):
Rendered markdown LGTM!

PR #1505, opened 2023-09-22
Feat: Native hierarchies for docx element types

newelh:
# Summary
Leveraging native hierarchy in word documents allows us to build more accurate hierarchies. Especially for list/bullet items and headings/titles. 

# TODO:
Improve summary
Testing
Before/after
Note about detecting by style ilvl instead of paragraph ilvl


test_unstructured/partition/docx/test_docx.py
```diff
@@ -416,3 +428,97 @@ def test_add_chunking_strategy_on_partition_docx(filename="example-docs/handbook
...
+        "example-docs/category-level.docx",
+        None,
+        None,
+        False,
+        None,
+    )
+
+    # Remember that category depths are 0-indexed and relative to the category type
+    # Title, list item, bullet, narrative text, etc.
+    paragraph = partitioner._document.paragraphs[0]
```
scanny, line 444: You might extracting `paragraphs = partitioner._document.paragraphs` to a line at the top here just to reduce the word-count below. The repeated dereferencings don't buy you anything as far as I can see. They're also expensive, but of course that only matters at scale; still, as a matter of style you wouldn't want repeated identical dereferencings (say more than two; each dot is a dict lookup) in your production code so consistent style would be to apply the same to test code.

You could also consider a more compact structure like:

```python
    test_cases = [
        (0, "Some text before any heading"),
        (0, "A Heading 1"),
        # etc. ...
    ]

    partitioner = _DocxPartitioner("example-docs/category-level.docx", None, None, False, None)
    paragraphs = partitioner._document.paragraphs

    for idx, (depth, text) in enumerate(test_cases):
        paragraph = paragraphs[idx]
        actual_depth = partitioner._parse_category_depth_by_style(paragraph)
        assert (
            actual_depth == depth
        ), f"expected paragraph[{idx}] to have depth=={depth}, got {actual_depth}"
        assert text in paragraph.text, f"paragraph[{idx}].text does not contain {text}"
```

Then you could probably see the whole test without scrolling. This also has the benefit being able to produce nice test-failure messages without repeating yourself a lot.

Your call though, just an idea :)

test_unstructured/partition/docx/test_docx.py
```diff
@@ -416,3 +428,97 @@ def test_add_chunking_strategy_on_partition_docx(filename="example-docs/handbook
...
+    assert partitioner._parse_category_depth_by_style(paragraph) == 0
+
+
+def test_parse_category_depth_by_style_name():
+    partitioner = _DocxPartitioner(
+        None,
+        None,
+        None,
+        False,
+        None,
```
scanny, line 499: This all fits on one line, doesn't it? I think if you leave off the trailing comma Black doesn't spread it out for you.

Probably a matter of personal taste, but I value the vertical real-estate more than a scannable column when it fits.

unstructured/partition/common.py
```diff
@@ -460,7 +460,8 @@ def convert_to_bytes(
 
 
 def convert_ms_office_table_to_text(
-    table: Union["docxtable", "pptxtable"], as_html: bool = True
+    table: Union["docxtable", "pptxtable"],
+    as_html: bool = True,
```
scanny, line 464: Is Black doing this? Or just a project convention? If I'm running the wrong Black version or config I'd like to fix it :)

scanny (approved):
LGTM Newel :)

A couple of remarks to consider.

unstructured/partition/common.py
newelh: My black configuration was misconfigured and I think set some of these. I can fix.

test_unstructured/partition/docx/test_docx.py
newelh: Much more elegant. Thank you!

unstructured/partition/common.py
newelh: Looks like these adjustments are in line with the results of `make tidy` and `make check`

test_unstructured/partition/docx/test_docx.py
newelh: Trailing comma is required by repository formatting config

PR #1508, opened 2023-09-23
Adding scarf tgz file into reqs with note

tabossert:
This allows anonymous tracking of downloads

Related to: https://github.com/Unstructured-IO/unstructured#chart_with_upwards_trend-analytics

Testing: 
`pip install -r requirements/base.in`

Result:
all packages should install as normal

### DIFF
--- a/requirements/extra-pdf-image.txt
+++ b/requirements/extra-pdf-image.txt
@@ -219,7 +219,7 @@ typing-extensions==4.8.0
     #   torch
 tzdata==2023.3
     # via pandas
-unstructured-inference==0.5.28
+unstructured-inference==0.5.31
     # via -r requirements/extra-pdf-image.in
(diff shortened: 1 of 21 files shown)


requirements/extra-pdf-image.txt
```diff
@@ -219,7 +217,7 @@ typing-extensions==4.8.0
     #   torch
 tzdata==2023.3
     # via pandas
-unstructured-inference==0.5.31
+unstructured-inference==0.5.28
```
ryannikolaidis, line 220: hmm, this downgrades unstructured-inference? feel like I remember someone mentioning changes to constraints recently, guessing this isn't what we want here..

tabossert: Thanks for catching, I have changed it back to latest version

PR #1516, opened 2023-09-25
Adds data source properties to the Jira connector

rvztz:
Defines update_source_metadata for the delta_table connector.


CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
-## 0.10.17-dev7
+## 0.10.17-dev8
 
 ### Enhancements
 
-* **Adds data source properties to SharePoint, Outlook, Onedrive, Reddit, and Slack connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
+* **Adds data source properties to SharePoint, Outlook, Onedrive, Reddit, Slack and Jira connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
```
ryannikolaidis, line 5:
```suggestion
* **Adds data source properties to SharePoint, Outlook, Onedrive, Reddit, Slack, and Jira connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
```
oxford comma

unstructured/ingest/connector/jira.py
```diff
@@ -323,18 +296,42 @@ def _output_filename(self):
...
+        self.source_metadata = SourceMetadata(
+            date_created=datetime.strptime(
+                self.parsed_fields["created"],
+                "%Y-%m-%dT%H:%M:%S.%f%z",
+            ).isoformat(),
+            date_modified=datetime.strptime(
+                self.parsed_fields["updated"],
+                "%Y-%m-%dT%H:%M:%S.%f%z",
+            ).isoformat(),
+            version=version,
```
ryannikolaidis, line 318: I don't see a version in any of our expected outputs. how are we validating this?

ryannikolaidis: got it, so in theory this could exist, but doesn't in our test content.

unstructured/ingest/connector/jira.py
```diff
@@ -323,18 +296,42 @@ def _output_filename(self):
...
 
+    def update_source_metadata(self, **kwargs) -> None:
+        exists = bool(self.issue)
+        if not exists:
+            self.source_metadata = SourceMetadata(
+                exists=exists,
+            )
+            return
+        version = None
+        if len(self.parsed_fields["versions"]) > 1:
```
ahmetmeleq, line 307: Otherwise, we won't be making use of the version info when there is only one version element in the version list

If we do not assign version info before there are at least two versions of a doc, please ignore this suggestion

```suggestion
        if len(self.parsed_fields["versions"]) >= 1:
```

unstructured/ingest/connector/jira.py
ahmetmeleq: Is this field a "(~release) version" for a project, or a version for the issue content (say, before and after an edit to the issue content)?

ahmetmeleq: If the field is a version for the issue content itself, testing this might need updating documents at the source to introduce a second version to some issues

I think it'll be worth it to do those updates in our test jira org and test if it's working as intended (mainly since we're introducing the version field with this PR)

At best, it'll only require a change in a document in our test jira org and rerunning the CI. At worst, it will save us some debugging and a patch later on

CHANGELOG.md
```diff
@@ -27,7 +27,7 @@
...
+* **Adds data source properties to SharePoint, Outlook, Onedrive, Reddit, Slack, DeltaTable and Jira connectors** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
 * **Add functionality to save embedded images in PDF's separately as images** This allows users to save embedded images in PDF's separately as images, given some directory path. The saved image path is written to the metadata for the Image element. Downstream applications may benefit by providing users with image links from relevant "hits."
```
ryannikolaidis, line 31: no, we shouldn't change this one. these were previous version changes

ryannikolaidis:
@rvztz see comment on changelog. ping me when fixed so we can merge

PR #1517, opened 2023-09-25
fix: Better logic for setting `category_depth` metadata for `Title` elements

LaverdeS:


test_unstructured/partition/test_common.py
```diff
@@ -405,3 +417,12 @@ def test_set_element_hierarchy_custom_rule_set():
...
         elements[5].metadata.parent_id == elements[4].id
     ), "FigureCaption should be child of Title 2"
+
+
+def test_document_to_element_list_sets_category_depth_titles():
+    layout_with_hierarchies = MockDocumentLayout()
+    elements = document_to_element_list(layout_with_hierarchies)
+    assert elements[0].metadata.category_depth == 1
+    assert elements[1].metadata.category_depth == 2
+    assert elements[2].metadata.category_depth is None
```
newelh, line 427: What's the reasoning for this category_depth being None vs being set to 0?

unstructured/partition/common.py
```diff
@@ -569,6 +569,14 @@ def document_to_element_list(
                 element.metadata.text_as_html = (
                     layout_element.text_as_html if hasattr(layout_element, "text_as_html") else None
                 )
+                try:
+                    if (isinstance(element, Title) and not element.metadata.category_depth) and any(
+                        el.type in ["Headline", "Subheadline"] for el in page.elements
+                    ):
+                        element.metadata.category_depth = 0
+                except AttributeError:
+                    # HTML instances
```
newelh, line 578: Could we add a logging statement here probably at the info level? Worried that passing this exception may hide a side effect that may be difficult to debug

unstructured/partition/common.py
```diff
@@ -569,6 +569,14 @@ def document_to_element_list(
                 element.metadata.text_as_html = (
                     layout_element.text_as_html if hasattr(layout_element, "text_as_html") else None
                 )
+                try:
+                    if (isinstance(element, Title) and not element.metadata.category_depth) and any(
```
newelh, line 573: and not element.metadata.category_depth will return false for 0 or None, is this intentional or should this be `and element.metadata.category_depth is None`

newelh (changes requested):

test_unstructured/partition/test_common.py
LaverdeS: `elements[2]` is a `Text` type. The default value for the `category_depth` metadata for all elements is None. The method `to_dict` of `ElementMetadata` returns all metadata that is not None ([here](https://github.com/Unstructured-IO/unstructured/blob/ab29de8dbdd3e75cbf0d61b5515efec322547a9b/unstructured/documents/elements.py#L194)). Many tests count on this behaviour and expect empty metadata so the Element method `to_dict` returns empty metadata and the json contains only the Elements types + text instead of a list of elements all with 0 value in `category_depth`. I did consider this before so I get you but after discussion with @qued, we thought the None value was more convenient.

unstructured/partition/common.py
LaverdeS: done

unstructured/partition/common.py
LaverdeS: fixed

newelh (approved):

newelh:
LGTM!

CHANGELOG.md
```diff
@@ -15,9 +15,9 @@
...
 * **Fixes issue where unstructured-inference was not getting updated** Problem: unstructured-inference was not getting upgraded to the version to match unstructured release when doing a pip install.  Solution: using `pip install unstructured[all-docs]` it will now upgrade both unstructured and unstructured-inference. Importance: This will ensure that the inference library is always in sync with the unstructured library, otherwise users will be using outdated libraries which will likely lead to unintended behavior.
-* **Fixes SharePoint connector failures if any document has an unsupported filetype** Problem: Currently the entire connector ingest run fails if a single IngestDoc has an unsupported filetype. This is because a ValueError is raised in the IngestDoc's `__post_init__`. Fix: Adds a try/catch when the IngestConnector runs get_ingest_docs such that the error is logged but all processable documents->IngestDocs are still instantiated and returned. Importance: Allows users to ingest SharePoint content even when some files with unsupported filetypes exist there.
```
cragwolfe, line 20: intentional?

CHANGELOG.md
```diff
@@ -15,9 +15,9 @@
 
 ### Fixes
 
+* **Fixes category_depth None value for Title elements** Problem: The mapping of `Headline` and `Subheadline` element types from `chipper` to `Title`, assigns metadata `category_depth` 1 and 2 respectively. During the same mapping, the `Title` elements from `chipper` are left unchanged, with metadata `category_depth` = None. Fix: Whenever `Headline` or `Subheadline` type layout elements are present in a page, then all `Title` elements with `category_depth` = None should be set to have a depth of 0 instead.
```
cragwolfe, line 18: does this make sense semantically?
remind me what  the `category_depth` for an `H1` element is again? if it is 0, does this imply all other Titles from Chipper are like html `H1`'s?

Please add "Why does it matter? for the 2nd sentence.

CHANGELOG.md
LaverdeS: non-intentional. I just added the missing bullet for SharePoint in Fixes

LaverdeS: non-intentional. I just added the missing bullet for SharePoint in Fixes

CHANGELOG.md
LaverdeS: correct, `H1` is `category_depth` 0. For, `chipper`: `Title` elements are like `H1`, `Headline` like `H2`, and `Subheadline` like `H3`. I am changing the description of the fix for clarification.

PR #1527, opened 2023-09-26
rfctr: base _Partitioner class

qued:


unstructured/partition/base.py
```diff
@@ -0,0 +1,106 @@
...
+            filename,
+            file,
+            metadata_filename,
+            metadata_last_modified,
+        )._iter_document_elements()
+
+    @abstractmethod
+    def _iter_document_elements(self) -> Iterator[Element]:
+        """Generate each document-element in document order."""
+        pass
```
scanny, line 60: There are two roles that can be fulfilled by a class like this and in my experience they are usually better kept separate:

1. **Abstract interface definition.** Well handled by a `Protocol` and the `AbstractPartitioner` protocol definition above is about all you can say about these classes because they only have the one "public" method. This defines what an object needs to implement to be considered one of these things (this is _structural typing_).
2. **Base class.** A base class provides default implementations and possibly abstract-methods, although I think those can easily become heavy-handed. I'll say more about that below, but in this particular case, do we really care that folks use a method named `._iter_document_elements()` to implement the main functionality of the class? I would consider that more of a convention/template type of concern and not rigidly require that, unless you were going to inherit `iter_document_elements()` in the sub-classes, which I don't think you can (because they have different option-types).

unstructured/partition/base.py
```diff
@@ -0,0 +1,106 @@
...
+        filename, file = self._filename, self._file
+
+        if filename is not None:
+            return self._load_document_from_filename(filename)
+
+        assert file is not None
+        if isinstance(file, SpooledTemporaryFile):
+            file.seek(0)
+            file = io.BytesIO(file.read())
+        return self._load_document_from_file_object(file)
```
scanny, line 87: I'm not sure I'd try to generalize this off the bat, at least not this far.

I _do_ however think it may be worthwhile to have a utility function like:
```python
def open_file(file: str | IO[bytes]) -> IO[bytes]:
    """Encapsulate details of getting from a path-or-file to a usable file-like object."""
    if isinstance(file, str):
        return ... # -- opened file or whatever, but who closes it when we're done? --
    if isinstance(file, SpooledTemporaryFile):
        file.seek(0)
        return io.BytesIO(file.read())
    return file
```

Not sure of the details and this would probably require some looking around at the different partitioners to find the common denominator, if any.

Overall I'd say this flexible-source-file handling question needs more study and I'd be inclined to leave the details to the particular patitioner for now. Main thing is not trying to combine getting a source-file in a suitable form and extracting any required metadata from it, which are two distinct functions and need not be coupled.

unstructured/partition/base.py
```diff
@@ -0,0 +1,106 @@
...
+        file_path, file = self._filename, self._file
+
+        # -- if the file is on the filesystem, get its date from there --
+        if file_path is not None:
+            return None if file_path.startswith("/tmp") else get_last_modified_date(file_path)
+
+        # -- otherwise try getting it from the file-like object (unlikely since BytesIO and its
+        # -- brethren have no such metadata).
+        assert file is not None
+        return get_last_modified_date_from_file(file)
```
scanny, line 106: This one does strike me as a good one to extract to a base class assuming we have a uniform policy for computing this which I expect we do. Everything we need to do this job (`file: str | IO[bytes], opts: PartitionerOpts)` is specified in the abstract interface so the implementation would work for all document types.

One thing to note here that I haven't ticketed yet is that a `file_path` starting with "/tmp" is unreliable for detecting a temporary file. It might work on Linux but definitely not on Mac and I expect not on Windows either.

scanny (commented):
A few ideas in the review comments :)

MthwRobinson:
@scanny @qued - Is there anything in this PR that you'd still like to do?

scanny:
I'd be inclined to return this one to the void (i.e. close it) and think freshly about it if we want to revisit it some time in future.

This was from like the day after I started, so I'm sure my thoughts have changed quite a bit as I've become more broadly familiar with the code base.

Also I'm inclined now toward a more incremental approach. I'll close for now and we might look back at it later for ideas or something. But if we wanted to delete it that would be fine with me too.

PR #1529, opened 2023-09-26
Allow request timeouts be passed for urls in partition

cdpierse:
In python the default behaviour of `requests.get` without a `timeout` being set is to hang indefinitely. We have a production use case where the desired behaviour would be to raise a timeout error rather than have the application just hang. 

This PR adds a new optional keyword parameter `request_timeout` to `partition` which is passed to `file_and_type_from_url` in the case where we are fetching from a URL. This is then passed to `requests.get`

### DIFF
--- a/unstructured/partition/auto.py
+++ b/unstructured/partition/auto.py
@@ -137,6 +137,7 @@ def partition(
     xml_keep_tags: bool = False,
     data_source_metadata: Optional[DataSourceMetadata] = None,
     metadata_filename: Optional[str] = None,
+    request_timeout: Optional[int] = None,
     **kwargs,
 ):
     """Partitions a document into its constituent elements. Will use libmagic to determine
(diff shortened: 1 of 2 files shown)


unstructured/partition/auto.py
```diff
@@ -407,8 +412,9 @@ def file_and_type_from_url(
     content_type: Optional[str] = None,
     headers: Dict[str, str] = {},
     ssl_verify: bool = True,
+    request_timeout: Optional[int] = None,
 ) -> Tuple[io.BytesIO, Optional[FileType]]:
-    response = requests.get(url, headers=headers, verify=ssl_verify)
+    response = requests.get(url, headers=headers, verify=ssl_verify, timeout=request_timeout)
```
badGarnet, line 417: can we add a test (maybe using mocker spy) to confirm the timeout is pass down as intended. This is more for preventing future code changes from accidentally break this behavior.

awalker4: @Coniferish do you mind taking a look at a small unit test here? Not a huge priority

qued:
Closing in favor of #2013 which adds a unit test.

unstructured/partition/auto.py
qued: Already handled in #2013.

PR #1538, opened 2023-09-26
feat/add sources from unstructured inference

benjats07:
This PR adds support for `source` property from `unstructured_inference`, allowing the user to be able to see the origin of the data under `detection_origin`field environment variable UNSTRUCTURED_INCLUDE_DEBUG_METADATA=true

In order to try this feature you can use this code:
```

yolox_elements = partition_pdf_or_image(filename='example-docs/loremipsum-flat.pdf', strategy='hi_res', model_name='yolox')

sources = [e.detection_origin for e in yolox_elements]
print(sources)
```
And will print 'yolox' as source for all the elements


cragwolfe (dismissed):
some quick feedback. this is mainly for debugging purposes, right?

* let's avoid overloading the word "source" which is has a meaning for data connectors.
* assuming this "breadcrumb" exists for debugging purposes, it should not exist in any element json output by default

benjats07:
> some quick feedback. this is mainly for debugging purposes, right?
> 
>     * let's avoid overloading the word "source" which is has a meaning for data connectors.
> 
>     * assuming this "breadcrumb" exists for debugging purposes, it should not exist in any element json output by default

* "source" isn't used anymore in this PR
* Restored original output

qued (commented):
`data_origin` sounds to me like a data source. I think the things we are identifying the origin of are the category classification, the bounding boxes, and maybe the classification probability. Those are the outputs of an object detection model, so my suggestion is `detection_origin`, but am open to other ideas.

benjats07:
> `data_origin` sounds to me like a data source. I think the things we are identifying the origin of are the category classification, the bounding boxes, and maybe the classification probability. Those are the outputs of an object detection model, so my suggestion is `detection_origin`, but am open to other ideas.
Sounds good to me.
🤔 At the main library we have other "sources" like OCR and all other libraries, in that case, would be correct to call it "detection" too?

qued (approved):
LGTM!

test_unstructured/partition/pdf-image/test_pdf.py
```diff
@@ -131,6 +133,8 @@ def _test(result):
         assert len(result) > 10
         # check that the pdf has multiple different page numbers
         assert {element.metadata.page_number for element in result} == expected
+        if UNSTRUCTURED_INCLUDE_DEBUG_METADATA:
+            assert {element.metadata.detection_origin for element in result} == {origin}
```
badGarnet, line 137: maybe use mocker to set the constant to true so we are certain this is tested; I understand that in the ci we set the env variable so this is tested in CI but this if statement here can cause confusion for local testing vs. ci; and potentially have code silently fail if ci for some reason dropped the env

unstructured/partition/common.py
```diff
@@ -158,40 +167,48 @@ def normalize_layout_element(
...
+            detection_origin=origin,
         )
         if element_type == "Headline":
             _element_class.metadata.category_depth = 1
         elif element_type == "Subheadline":
             _element_class.metadata.category_depth = 2
         return _element_class
     elif element_type == "Checked":
-        return CheckBox(
+        checkbox = CheckBox(
```
badGarnet, line 178: why can't we keep the original pattern of just return instead of using a variable to capture the return then return?

unstructured/partition/docx.py
```diff
@@ -305,7 +305,7 @@ def _increment_page_number(self) -> Iterator[PageBreak]:
         """Increment page-number by 1 and generate a PageBreak element if enabled."""
         self._page_counter += 1
         if self._include_page_breaks:
-            yield PageBreak("")
+            yield PageBreak("", detection_origin="docx")
```
badGarnet, line 308: lets put the string `"docx"` as a constant for this sub module (e.g., after import define a `DETECTION_ORIGIN="docx"`) and reuse below

unstructured/partition/common.py
benjats07: You're right, I revert to use that pattern

unstructured/partition/docx.py
benjats07: Also refactored other origins 😄

test_unstructured_ingest/expected-structured-output/notion/c47a4566-4c7a-488b-ac2a-1292ee507fcb.json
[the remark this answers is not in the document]
benjats07: I'll re-run ingestion tests updates to be sure.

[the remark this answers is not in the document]
benjats07: Nice catch, file should be back

unstructured/documents/elements.py
```diff
@@ -192,6 +193,15 @@ class ElementMetadata:
     # Detection Model Class Probabilities from Unstructured-Inference Hi-Res
     detection_class_prob: Optional[float] = None
 
+    if UNSTRUCTURED_INCLUDE_DEBUG_METADATA:
+        detection_origin: Optional[str] = None
+
+    def __setattr__(self, key, value):
+        if not UNSTRUCTURED_INCLUDE_DEBUG_METADATA and key == "detection_origin":
```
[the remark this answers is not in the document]
benjats07: Probably, but at this point, I prefer not doing that as this is the only parameter with that behavior, I can add it if we begin to add variables just during debugging

PR #1539, opened 2023-09-26
Klaijan/pdf get embedded url index

Klaijan:
**Executive Summary**

Adds PDF functionality to capture hyperlink (external or internal) for pdf fast strategy along with associate text.

**Technical Details**

- `pdfminer` associates `annotation` (links and uris) with bounding box rather than text. Therefore, the link and text matching is not a perfect pair but rather a logic-based and calculation matching from bounding box overlapping.
- There is no word-level bounding box. Only character-level (access using `LTChar`). Thus in order to get to word-level, there is a window slicing through the text. The words are captured in alphanumeric and non-alphanumeric separately, meaning it will split the word if contains both, on the first encounter of non-alphanumeric.)
- The bounding box calculation is calculated using start and stop coordinates for the corresponding word calculated from above. The calculation is simply using distance between two dots.

The result now contains `links` in `metadata` as shown below:

```
            "links": [
                {
                    "text": "link",
                    "url": "https://github.com/Unstructured-IO/unstructured",
                    "start_index": 12
                },
                {
                    "text": "email",
                    "url": "mailto:unstructuredai@earlygrowth.com",
                    "start_index": 30
                },
                {
                    "text": "phone number",
                    "url": "tel:6505124019",
                    "start_index": 49
                }
            ]
```


unstructured/partition/pdf.py
```diff
@@ -51,6 +67,8 @@
 )
 from unstructured.utils import requires_dependencies
 
+# from lxml import etree
```
awalker4, line 70: Guessing this can be removed

awalker4 (approved):
LGTM! Verified that the new sample pdf has links in the metadata.

PR #1540, opened 2023-09-27
chore: Table chunking

amanda103:
This change is adding to our `add_chunking_strategy` logic so that we are able to chunk Table elements' `text` and `text_as_html` params. In order to keep the functionality under the same `by_title` chunking strategy we have renamed the `combine_under_n_chars` to `max_characters`. It functions the same way for the combining elements under Title's, as well as specifying a chunk size (in chars) for TableChunk elements.

*renaming the variable to `max_characters` will also reflect the 'hard max' we will implement for large elements in followup PRs

Additionally -> some lint changes snuck in when I ran `make tidy` hence the minor changes in unrelated files :) 

TODO:
✅ add unit tests
--> note: added where I could to unit tests! Some unit tests I just clarified that the chunking strategy was now 'by_title' because we don't have a file example that has Table elements to test the 'by_num_characters' chunking strategy
✅  update changelog

To manually test:
```
In [1]: filename="example-docs/example-10k.html"

In [2]: from unstructured.chunking.title import chunk_table_element

In [3]: from unstructured.partition.auto import partition

In [4]: elements = partition(filename)

# element at -2 happens to be a Table, and we'll get chunks of char size 4 here
In [5]: chunks = chunk_table_element(elements[-2], 4)

# examine text and text_as_html params
ln [6]: for c in chunks:
                    print(c.text)
                    print(c.metadata.text_as_html)
```


unstructured/chunking/title.py
```diff
@@ -9,11 +10,80 @@
...
+        text_chunk = text[:num_characters]
+        table_chunk = TableChunk(
+            text=text_chunk,
+            metadata=copy.copy(metadata),
+        )
+        if text_as_html:
+            text_as_html_chunk = text_as_html[:num_characters]
+            table_chunk.metadata.text_as_html = text_as_html_chunk
+            # Remove the processed chunk from text_as_html
+            text_as_html = text_as_html[num_characters:]
```
badGarnet, line 71: so there is really no correlation between the text chunk and the `text_as_html` chunk since the html will have html formatting characters so it will be different from text
additionally what does this mean to this chunk? This chunk would still contain `2 * num_character` of text if we got both html and text. And this means the total number of characters in this chunk is twice that of other chunks without html. What would this mean for RAG downstream (main usage for chunks?)

amanda103: That's correct - this chunk will still contain text and text_as_html params, both potentially at the given char # input.

My understanding is that the downstream workflow understands tables better in their HTML form, so I'm guessing if someone is chunking tables they're sending the text_as_html elements and not the text, or separately? But it's definitely important to provide the chunking for the text_as_html param if the element has it after partitioning.

unstructured/chunking/title.py
```diff
@@ -9,11 +10,80 @@
...
+    chunked_elements: List[Element] = []
+
+    for element in elements:
+        if isinstance(element, Table):
+            char_len = len(element.text)
+            html_table = element.to_dict().get("text_as_html")
+            if html_table:
+                char_len = len(html_table)
+            if char_len > num_characters:  # type: ignore
+                chunked_elements.extend(chunk_table_element(element, num_characters))
```
badGarnet, line 41: here we are potentially mixing `Element` objects with `Chunk` objects; are those two types of objects compatible? Are there attributes that exist in one but not another?

badGarnet: nevermind, I see element is essentially the parent

unstructured/chunking/title.py
```diff
@@ -9,11 +10,80 @@
...
+        A list of unstructured elements. Usually the output of a partition functions.
+    num_characters
+        Splits elements text field so that a section matches the desired length."""
+
+    chunked_elements: List[Element] = []
+
+    for element in elements:
+        if isinstance(element, Table):
+            char_len = len(element.text)
+            html_table = element.to_dict().get("text_as_html")
```
badGarnet, line 37:
```suggestion
            html_table = getattr(element, "text_as_html")
```

amanda103: shoot - I remember why I didn't do this:

```
>               html_table = getattr(element, "text_as_html")
E               AttributeError: 'Table' object has no attribute 'text_as_html'
```

I think I need to do the same thing as I did below in the actual chunking function:
`    text_as_html = element.text_as_html if hasattr(element, "text_as_html") else None`

unstructured/chunking/title.py
```diff
@@ -9,105 +10,153 @@
...
-        A list of unstructured elements. Usually the ouput of a partition functions.
+        A list of unstructured elements. Usually the output of a partition functions.
     multipage_sections
         If True, sections can span multiple pages. Defaults to True.
-    combine_under_n_chars
+    combine_text_under_n_chars
         Combines elements (for example a series of titles) until a section reaches
         a length of n characters.
-    new_after_n_chars
+    max_characters
```
cragwolfe, line 83: `new_after_n_chars` was a "soft limit".
theoretically, soft limits are nice because one is never going to chop an element in half, for instance.

i can see wanting both the soft limit and hard limit (`max_characters`).

for example, if my accumulated chunk is over 800 chars (a soft limit), start a new clean chunk. but don't let the chunk exceed 1000 chars (a hard limit).

imo, ^^ is worth a follow on github issue.

unstructured/chunking/title.py
```diff
@@ -9,105 +10,153 @@
     Element,
     ElementMetadata,
     Table,
+    TableChunk,
     Text,
     Title,
 )
 
 
+def chunk_table_element(
```
newelh, line 19: I think we could simplify the branching in this function to be something like:
```
def chunk_table_element(element: 'Table', max_chars: int = 1500) -> List[Union['Table', 'TableChunk']]:
    text = element.text
    html = getattr(element, 'text_as_html', None)

    if len(text) <= max_chars and (html is None or len(html) <= max_chars):
        return [element]

    chunks = []
    metadata = copy.copy(element.metadata)
    is_continuation = False

    while text or html:
        text_chunk, text = text[:max_chars], text[max_chars:]
        table_chunk = TableChunk(text=text_chunk, metadata=copy.copy(metadata))
        
        if html:
            html_chunk, html = html[:max_chars], html[max_chars:]
            table_chunk.metadata.text_as_html = html_chunk

        if is_continuation:
            table_chunk.metadata.is_continuation = True
        
        chunks.append(table_chunk)
        is_continuation = True

    return chunks
```

unstructured/chunking/title.py
```diff
@@ -9,105 +10,153 @@
...
     sections = _split_elements_by_title_and_table(
         elements,
         multipage_sections=multipage_sections,
-        combine_under_n_chars=combine_under_n_chars,
-        new_after_n_chars=new_after_n_chars,
+        combine_text_under_n_chars=combine_text_under_n_chars,
+        max_characters=max_characters,
     )
-
     for section in sections:
```
newelh, line 106: This may be a good opportunity to clean up this loop a bit as it's getting unwieldy
```
    for section in sections:
        if not section:
            continue

        first_element = section[0]
        
        if not isinstance(first_element, Text):
            chunked_elements.extend(section)
            continue
        
        if isinstance(first_element, Table):
            chunked_elements.extend(chunk_table_element(first_element, max_characters))
            continue

        text = ""
        metadata = first_element.metadata
        start_char = 0

        for element in section:
            if isinstance(element, Text):
                text += "\n\n" if text else ""
                start_char = len(text)
                text += element.text

            for attr, value in vars(element.metadata).items():
                if isinstance(value, list):
                    _value = getattr(metadata, attr, []) or []
                    
                    if attr == "regex_metadata":
                        for item in value:
                            item["start"] += start_char
                            item["end"] += start_char

                    _value.extend(item for item in value if item not in _value)
                    setattr(metadata, attr, _value)

        chunked_elements.append(CompositeElement(text=text, metadata=metadata))

    return chunked_elements
```

newelh (changes requested):

unstructured/chunking/title.py
badGarnet: oof, agree; simple step might even be taking the element loop out into own function first so that:
- can do unit test
- reduce nesting
now more to this might involve data structure; we store data as intuitive objects right now but may not be most efficient for compute; one alternative is to store data as block (e.g., numpy array) and attributes are just reference to the data underneath. This makes expression still natural but computation can be better optimized. But that's really for another ticket on its own; for now lets at least refactor the inner loop out

newelh: +1

unstructured/chunking/title.py
cragwolfe: the problem of giant elements could exist anywhere, i.e., we should chunk a giant NarrativeText block the same way we would a Table very soon if not in this PR.

> So basically we'll have a soft and hard max but table is only using hard max and title is only using soft max until the next ticket/pr :)

i think that works, as long as we are prioritizing having a soft max and hard max that is consistent across element types soon afterwards.

amanda103: re: soon - Yes definitely! Will add this context to the tickets in the next sprint

unstructured/chunking/title.py
amanda103: Sounds good! Can definitely simplify the existing loop, and I'll make note of this convo for the chunking work for other element types that is in the next sprint :)

newelh (approved):

test_unstructured/partition/docx/test_docx.py
```diff
@@ -489,3 +481,29 @@ def test_parse_category_depth_by_style_name():
...
+    filename="example-docs/fake-doc-emphasized-text.docx",
+):
+    chunk_elements = partition_docx(
+        filename,
+        chunking_strategy="by_title",
+        max_characters=9,
+        combine_text_under_n_chars=5,
+    )
+    elements = partition_docx(filename)
+    chunks = chunk_by_title(elements, max_characters=9, combine_text_under_n_chars=5)
```
badGarnet, line 507: can we also have assert to check for this example that
- without combining text under n chars set to 5 there are tiny chunks
- with the setting we have text with minimum 5 chars

test_unstructured/chunking/test_title.py
badGarnet, line 1: can we add a parametrized unittest on the parameters:
```
    combine_text_under_n_chars: int = 500,
    new_after_n_chars: int = 1000,
    max_characters: int = 1500,
```
to test edge cases? We seem to have encountered quite some edge case related bugs recently so want to do more preventions on that by testing the inputs with edge cases like:
- 0 (there are tests for that so thats good)
- negative values
- values larger than length of text
- combinations of parameters that shouldn't make sense (e.g., one parameters asks maximum length being 100 while another sets minimum size to be 110)

unstructured/chunking/title.py
```diff
@@ -218,8 +250,9 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> List[Element]:
                 elements = chunk_by_title(
                     elements,
                     multipage_sections=params.get("multipage_sections", True),
-                    combine_under_n_chars=params.get("combine_under_n_chars", 500),
-                    new_after_n_chars=params.get("new_after_n_chars", 1500),
+                    combine_text_under_n_chars=params.get("combine_text_under_n_chars", 500),
+                    new_after_n_chars=params.get("new_after_n_chars", 1000),
+                    max_characters=params.get("max_characters", 1500),
```
badGarnet, line 255: maybe set the default values like 500 as constants? Here unlike in the function API it is not exposed to the caller so the default value is hardwired in and hidden, which can make it difficult to maintain. Also we have those constants being reused in multiple places so it is better if we use a constant variable so that it is easier to maintain the code as well.

amanda103: Yes definitely! will do :)

test_unstructured/chunking/test_title.py
amanda103: Great idea! I've added to the existing test that was checking for ValueError raised, as well as added some conditions of when to raise :)

PR #1547, opened 2023-09-27
enhancement: memory efficient xml partitioning

### LINKED ISSUE #1236: bug/Out of memory in xml parsing
**Describe the bug**
Trying to parse XML (large 1.1GB) and getting "out of memory" exception
**To Reproduce**
To parse large XML

**Environment Info**
OS version:  macOS-13.5-arm64-arm-64bit
Python version:  3.11.3
unstructured version:  0.9.0
unstructured-inference version:  0.5.7
pytesseract version:  0.3.10
Torch version:  2.1.0.dev20230811
Detectron2 is not installed
PaddleOCR is not installed
Libmagic version: file-5.41

**Additional context**
```
Failed to ingest /Users/slava/Documents/Development/private/AI/cloud-20230805/tickets.xml due to Traceback (most recent call last):
  File "/Users/slava/Documents/Development/private/AI/h2ogpt/src/gpt_langchain.py", line 2005, in path_to_doc1
    res = file_to_doc(file, base_path=None, verbose=verbose, fail_any_exception=fail_any_exception,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/Documents/Development/private/AI/h2ogpt/src/gpt_langchain.py", line 1894, in file_to_doc
    doc1 = loader.load()
           ^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/site-packages/langchain/document_loaders/unstructured.py", line 86, in load
    elements = self._get_elements()
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/site-packages/langchain/document_loaders/xml.py", line 42, in _get_elements
    return partition_xml(filename=self.file_path, **self.unstructured_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/site-packages/unstructured/documents/elements.py", line 222, in wrapper
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/site-packages/unstructured/file_utils/filetype.py", line 628, in wrapper
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/site-packages/unstructured/partition/xml.py", line 105, in partition_xml
    raw_text = get_leaf_elements(filename=filename, file=file, xml_path=xml_path)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/site-packages/unstructured/partition/xml.py", line 40, in get_leaf_elements
    root = ET.fromstring(raw_text)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/slava/.pyenv/versions/3.11.3/lib/python3.11/xml/etree/ElementTree.py", line 1338, in XML
    parser.feed(text)
xml.etree.ElementTree.ParseError: out of memory: line 1, column 0
```

qued:
Closes #1236. Partitions XML documents iteratively in most cases*, never loading the entire tree into memory. This ends up being much faster.

(* The exception is when the argument `xml_path` is passed to filter elements. I was not able to find a way in Python to compare XPaths while streaming the elements, aside from writing a custom XPath parser. So the shortest way forward was to bite the bullet and load the whole tree in memory when filtering by XPath.)

Memory usage is about 20% of usage on `main` when processing a 470MB XML file. Time to process is 10s vs 900s.

Output is slightly different, but appears to be an improvement, adding lines of text that are skipped in current partitioning. No text is lost.

#### Testing:
Test file I used is available compressed [here](https://ftp.ncbi.nlm.nih.gov/pubmed/baseline/pubmed23n1106.xml.gz).

Test code (requires installation of [memory-profiler](https://pypi.org/project/memory-profiler/)):
```python
import time

from memory_profiler import memory_usage


def main():
    from unstructured.partition.xml import get_leaf_elements

    return get_leaf_elements(filename="pubmed23n1106.xml")


t0 = time.time()
mem = memory_usage(main)
t1 = time.time()

total = t1 - t0
print(f"Max memory usage: {max(mem)}")
print(f"Time: {total}")

```
Compare results on `main` vs this branch (warning, took about 15 minutes on my machine on `main`).


unstructured/partition/xml.py
```diff
@@ -20,41 +22,58 @@
...
+        b = BytesIO(bytes(cast(str, text), encoding="utf-8"))
+        return _get_leaf_elements(b, xml_path=xml_path)
 
-    for elem in root.findall(xml_path):
-        for subelem in elem.iter():
-            if is_leaf(subelem) and is_string(subelem.text):
-                leaf_elements.append(subelem.text)
 
-    return leaf_elements
+def _get_leaf_elements(
```
badGarnet, line 48: is the return ever used as a list or just used as iterators? If it is only used as iterator later might be even better to return a generator instead of list (so that even more efficient for memory)

qued: That's a good point, I'll check the downstream uses. My only concern would be if we actually use it lazily then I assume the file will stay open, but in practice that probably won't matter much.

qued: So the immediate thing is that having `get_leaf_elements` and `_get_leaf_elements` return a generator doesn't have much affect since `partition_xml` immediately wraps it in a list. If we go all the way and have `partition_xml` return a generator, then a number of tests fail.

Tests can be updated, but since `partition_pdf` is at the boundary of our interface, I feel like we should hold off on such a change unless we want to make that change consistent across all `partition` functions.

I'll make `get_leaf_elements` and `_get_leaf_elements` return generators though.

badGarnet (approved):

PR #1551, opened 2023-09-27
refactor: unstructured ingest as a pipeline

rbiseck3:
### Description
As we add more and more steps to the pipeline (i.e. chunking, embedding, table manipulation), it would help seperate the responsibility of each of these into their own processes, running each in parallel using json files to share data across. This will also help guarantee data is serializable if this code was used in an actual pipeline. Following is a flow diagram of the proposed changes. As part of this change:
* A parent pipeline class will be responsible for running each `node`, which can optionally be run via multiprocessing if it supports it, or not. Possible nodes at this moment:
  * Doc factory: creates all the ingest docs via the source connector
  * Source: reads/downloads all of the content to process to the local filesystem to the location set by the `download_dir` parameter.
  * Partition: runs partition on all of the downloaded content in json format. 
  * Any number of reformat nodes that modify the partitioned content. This can include chunking, embedding, etc. 
  * Write: push the final json into the destination via the destination connector
* This pipeline relies on the information of the ingest docs to be available via their serialization. An optimization was introduced with the `IngestDocJsonMixin` which adds in all the `@property` fields to the serialized json already being created via the `DataClassJsonMixin`
* For all intermediate steps (partitioning, reformatting), the content is saved to a dedicated location on the local filesystem. Right now it's set to `$HOME/.cache/unstructured/ingest/pipeline/STEP_NAME/`.
* Minor changes: made sense to move some of the config parameters between the read and partition configs when I explicitly divided the responsibility to download vs partition the content in the pipeline. 
* The pipeline class only makes the doc factory, source and partition nodes required, keeping with the logic that has been supported so far. All reformatting nodes and write node are optional. 
* Long term, there should also be some changes to the base configs supported by the CLI to support pipeline specific configs, but for now what exists was used to minimize changes in this PR. 
* Final step to copy the final output to the location designated by the `_output_filename` value of the ingest doc.
* Hashing occurs at each step by hashing the parameters of that step (i.e. partition configs) along with the previous step via the filename used. This allows each step to be the same _if_ all the parameters for it have not changed and the content so far is the same. 
* The only data that is shared and has writes to across processes is the dictionary of ingest json data. This dict is created using the `multiprocessing.manager.DictProxy` to make sure any interaction with it is behind a lock. 

### Minor refactors included:
* Utility methods added to extract configs from the click options
* Utility method to add common options to click commands.
* All writers moved to using the class approach which extracts a lot of the common code so there's less copy-paste when new runners are added. 
* Use `@property` for source metadata on base ingest doc to add logic to call `update_source_metadata` if it's still `None` at the time it's fetched. 

### Additional bug fixes included
* Fsspec connectors were not serializable due to the `ingest_doc_cls`. This was removed from the fields captured by the `@dataclass` decorator and added in a `__post_init__` method. 
* Various reddit connector params were missing. This doesn't have an explicit ingest test at the moment so was never caught. 
* Fsspec connector had the parent `update_source_metadata` misnamed as `update_source_metadata_metadata` so it was never being called.  

### Flow Diagram

ingest_pipeline


badGarnet (commented):
some general comments:
- can we switch to use concurrent executor pool from `concurrent.futures` instead of multiprocessing pool? The executor is better at avoiding thread locks than pool. Given we potentially run models with their own mp backend (e.g., huggingface tokenizer uses rust backend) we can get into trouble with threadlock easily
- the rust backend from huggingface is also an example for careful setting up the pipeline: there are nodes in this design that inherently does multi-threading/procesing on its own: While the attribute `support_multiprocessing` can help id them it doesn't help when some are preferred to not be called AFTER or BEFORE python's multi processing: e.g., https://github.com/huggingface/tokenizers/issues/187
- how does this pipeline optimize a workflow where part of the process is on GPU and we can utilize the CPU for other tasks at the same time (e.g., one of the reformat node uses a GPU model and another runs a randomforest model on CPU and they don't depend on each other)? Maybe we want to think about aync tasks? Which would favor likely a pool of concurrent executors as well

rbiseck3:
@badGarnet valid points, I'm not sure if some of them should be scoped to their own effort? To limit how much is changed in this PR, the pool approach is what's already being used, this only splits it up into discrete steps run by a pipeline class.

ryannikolaidis:
> -output-dir...Should we still support this parameter? If so, should a step be added to copy the final json files to this location?

yes, I think this is a pretty reasonable expectation for the simple source connector only flow, which folks likely will still hit. I think adding that final copy step makes sense to me.

ryannikolaidis:
> Does it make sense to hash ingest docs on download location? If a new pipeline was ran with the same files but not the same parameters for the pipeline steps, this might not make sense. But we want to be able to pick up from a node in the pipeline to avoid reprocessing if we don't have to.

I know we hash the download location by url, but what do you mean by hash ingest docs ? where do IngestDocs themselves get hashed? or is this in the new flow..is this a hash used for storing data from the intermediate nodes/stages? if it’s the latter, then I agree, at a minimum the hash should include any parameters at that node/stage and all of the parameters from upstream nodes/stages.

ryannikolaidis:
> How to force reprocessing of a particular step? Or would this apply to the whole pipeline if the --reprocess flag passed into the CLI invocation without further granularity?

I think we’ll still want a way to just allow a user to reprocess the whole pipeline. maybe --reprocess still does that and add a new flag to specify reprocessing a specific stage? presumably you can only specify one specific one since anything downstream of the earliest reprocessed node would also have to get reprocessed?

rbiseck3:
@ryannikolaidis, regarding the hash, the pipeline context that gets passed around across nodes contains a
```python
ingest_docs_map: dict
``` 
Which gets populated via 
```python
ingest_docs_map[get_ingest_doc_hash(doc)] = doc
```
Where `doc` is `ingest_doc.to_json()`. And the hash is derived from 
```python
hashed = hashlib.sha256(json_as_dict.get("filename").encode()).hexdigest()[:32]
```

This allows every stage be able to map the content they're producing back to the original ingest doc class.

ryannikolaidis:
> @ryannikolaidis, regarding the hash, the pipeline context that gets passed around across nodes contains a
> 
> ```python
> ingest_docs_map: dict
> ```
> 
> Which gets populated via
> 
> ```python
> ingest_docs_map[get_ingest_doc_hash(doc)] = doc
> ```
> 
> Where `doc` is `ingest_doc.to_json()`. And the hash is derived from
> 
> ```python
> hashed = hashlib.sha256(json_as_dict.get("filename").encode()).hexdigest()[:32]
> ```
> 
> This allows every stage be able to map the content they're producing back to the original ingest doc class.

got it. yea, then I agree, you definitely can't just rely on download location for this hash.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * **Adds data source properties to the Jira connector** These properties (date_created, date_modified, version, source_url, record_locator) are written to element metadata during ingest, mapping elements to information about the document source from which they derive. This functionality enables downstream applications to reveal source document applications, e.g. a link to a GDrive doc, Salesforce record, etc.
 * **Improve title detection in pptx documents** The default title textboxes on a pptx slide are now categorized as titles.
 * **Improve hierarchy detection in pptx documents** List items, and other slide text are properly nested under the slide title. This will enable better chunking of pptx documents.
+* **Refactor of ingest cli workflow** The new approach uses a dynamically set pipeline with a snapshot along each step to save progress and allow to continue from there if an error occurred. Also allows to dynamically set any number of steps to modify the partition content before it gets written to a destination.
```
ryannikolaidis, line 9:
```suggestion
* **Refactor of the ingest cli workflow** The refactored approach uses a dynamically set pipeline with a snapshot along each step to save progress and accomadate continuation from a snapshot if an error occurs. This also allows the pipeline to dynamically assign any number of steps to modify the partitioned content before it gets written to a destination.
```

ryannikolaidis:
given how much this changes,  probably worth a refresh to the [README](https://github.com/Unstructured-IO/unstructured/tree/main/unstructured/ingest#readme) and supporting diagrams to describe the entire flow in the context of these changes.

test_unstructured_ingest/test-ingest-salesforce.sh
```diff
@@ -17,12 +18,18 @@ CI=${CI:-"false"}
...
     cleanup_dir "$DOWNLOAD_DIR"
   fi
 }
 trap cleanup EXIT
 
+if [ -z "$SALESFORCE_USERNAME" ] || [ -z "$SALESFORCE_CONSUMER_KEY" ]; then
+  echo "Skipping Salesforce ingest test because SALESFORCE_USERNAME and SALESFORCE_CONSUMER_KEY env vars not set"
+  exit 0
+fi
+
```
ryannikolaidis, line 32: good catch

test_unstructured_ingest/unit/test_paths.py
```diff
@@ -23,15 +23,25 @@ class FakeConfigFolder:
     download_dir = "/fakeuser/fake_download"
 
 
+class DropboxIngestDocTest(DropboxIngestDoc):
+    def __post_init__(self):
+        return None
+
+
+class FsspecIngestDocTest(FsspecIngestDoc):
```
ryannikolaidis, line 31: nit: can we call these like `DropboxIngestTestDoc` or `TestDropboxIngestDoc`? rather than `DropboxIngestDocTest`. otherwise I'm expecting this class itself to be a test.

unstructured/ingest/cli/cmds/azure.py
```diff
@@ -60,25 +58,17 @@ def azure_source(ctx: click.Context, **options):
...
 def get_source_cmd() -> click.Group:
     cmd = azure_source
-    AzureCliConfig.add_cli_options(cmd)
-    CliRemoteUrlConfig.add_cli_options(cmd)
-    CliRecursiveConfig.add_cli_options(cmd)
-
-    # Common CLI configs
-    CliReadConfig.add_cli_options(cmd)
-    CliPartitionConfig.add_cli_options(cmd)
-    cmd.params.append(click.Option(["-v", "--verbose"], is_flag=True, default=False))
```
ryannikolaidis, line 83: not having to do this is nice

unstructured/ingest/cli/cmds/reddit.py
```diff
@@ -57,6 +56,17 @@ def add_cli_options(cmd: click.Command) -> None:
                 type=str,
                 help="If set, return posts using this query. Otherwise, use hot posts.",
             ),
+            click.Option(
+                ["--num-posts"],
+                required=True,
+                type=click.IntRange(0),
+                help="If set, return posts using this query. Otherwise, use hot posts.",
```
ryannikolaidis, line 63: errr, I think you copied and pasted this from above?

ryannikolaidis: better help here?

unstructured/ingest/cli/cmds/reddit.py
```diff
@@ -57,6 +56,17 @@ def add_cli_options(cmd: click.Command) -> None:
...
+            ),
+            click.Option(
+                ["--user-agent"],
```
ryannikolaidis, line 66: add a help and describe what this is

unstructured/ingest/cli/interfaces.py
```diff
@@ -1,25 +1,99 @@
...
+                    (Path.home() / ".cache" / "unstructured" / "ingest" / "pipeline").resolve(),
+                ),
+                show_default=True,
+                help="Where to place working files when processing each step",
+            ),
+            click.Option(
+                ["--num-processes"],
+                default=2,
+                show_default=True,
+                help="Number of parallel processes to process docs in.",
```
ryannikolaidis, line 90:
```suggestion
                help="Number of parallel processes with which to process docs",
```

unstructured/ingest/cli/interfaces.py
```diff
@@ -51,6 +125,12 @@ def add_cli_options(cmd: click.Command) -> None:
                 "is not specified and "
                 "skip processing them through unstructured.",
             ),
+            click.Option(
+                ["--max-docs"],
+                default=None,
+                type=int,
+                help="If specified, process at most specified number of documents.",
```
ryannikolaidis, line 132:
```suggestion
                help="If specified, process at most the specified number of documents.",
```

unstructured/ingest/cli/utils.py
```diff
@@ -12,39 +23,35 @@ def conform_click_options(options: dict):
...
-                        "{value!r} is not one of {choices}.",
-                        len(self.choices),
-                    ).format(value=s, choice=choices_str, choices=choices_str),
-                    param,
-                    ctx,
-                )
-        return split
+def extract_configs(
+    data: dict,
+    validate: t.Optional[t.List[t.Type[BaseConfig]]] = None,
```
ryannikolaidis, line 28: okay so validation is "can I instantiate from data dictionary" and the `validate` argument is actually a list of Config subclasses to validate? I'd probably name something clearer. maybe validation_config_classes? other ideas?
also probably worth a one-liner on what we're doing here.

unstructured/ingest/connector/fsspec.py
```diff
@@ -90,6 +90,9 @@ class FsspecIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
     connector_config: SimpleFsspecConfig
     remote_file_path: str
 
+    def __post_init__(self):
+        self.update_source_metadata_metadata()
```
ryannikolaidis, line 94: this looks like a typo..in the original file assuming we should have called this `update_source_metadata` right?

unstructured/ingest/connector/salesforce.py
```diff
@@ -87,6 +87,16 @@ class SalesforceIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
     record_type: str
     record_id: str
     registry_name: str = "salesforce"
+    _record: OrderedDict = field(default_factory=lambda: OrderedDict())
+
+    def __post_init__(self):
+        self.update_source_metadata()
```
ryannikolaidis, line 93: why calling this in post_init rather than opportunistically? seems like this defeats the point of lazy fetching right? also this guarantees were doing this both when the connector creates the ingest_doc and when we rehydrate in processing, right?

unstructured/ingest/ingest_doc_json_mixin.py
```diff
@@ -0,0 +1,54 @@
+import json
+import typing as t
+from pathlib import Path
+
+from dataclasses_json import DataClassJsonMixin
+from dataclasses_json.core import _ExtendedEncoder
+
+
+class IngestDocJsonMixin(DataClassJsonMixin):
```
ryannikolaidis, line 9: docstring explaining what/why

i.e. adds in all the @property fields to the serialized json already being created via the DataClassJsonMixin...etc

unstructured/ingest/pipeline/interfaces.py
```diff
@@ -0,0 +1,184 @@
...
+        if path := self.get_path():
+            logger.info(f"Creating {path}")
+            path.mkdir(parents=True, exist_ok=True)
+
+    def get_path(self) -> t.Optional[Path]:
+        return None
+
+
+@dataclass
+class DocFactoryNode(PipelineNode):
```
ryannikolaidis, line 81: docstring

unstructured/ingest/pipeline/interfaces.py
```diff
@@ -0,0 +1,184 @@
...
+    BaseDestinationConnector,
+    BaseSourceConnector,
+    PartitionConfig,
+    ProcessorConfig,
+)
+from unstructured.ingest.logger import ingest_log_streaming_init, logger
+
+
+@dataclass
+class PipelineContext(ProcessorConfig):
```
ryannikolaidis, line 23: docstring

unstructured/ingest/runner/airtable.py
```diff
@@ -2,59 +2,45 @@
...
-
-    process_documents(
-        source_doc_connector=source_doc_connector,
-        partition_config=partition_config,
-        verbose=verbose,
-        dest_doc_connector=dest_doc_connector,
-    )
+
+
+class Airtable(Runner):
```
ryannikolaidis, line 10: can we call each of these XRunner. So this would be AirtableRunner? otherwise it's a bit confusing to see a class broadly named Airtable that actually is doing this very specific task.

unstructured/ingest/connector/salesforce.py
rbiseck3: I wanted to move it out of the `get_file` method since that can be skipped if the file exists and then all this metadata wouldn't exist. This was the simplest approach but I see you point. I can further refactor to get the data.

rbiseck3:
@ryannikolaidis, I added a new issue to update all documentation given this new approach. If I was to do that in the same PR, it would a ton more changed files to the already large list of changes.

ryannikolaidis:
> @ryannikolaidis, I added a new issue to update all documentation given this new approach. If I was to do that in the same PR, it would a ton more changed files to the already large list of changes.

yep, all good. thanks!

ryannikolaidis (approved):
looks good! thanks for all of the changes.

PR #1553, opened 2023-09-27
Bump Dockerfile to use latest base image

tabossert:
New base image includes security fixes. This is an ongoing process to remediate security issues as they are identified.


yuming-long (approved):
LGTM!

PR #1556, opened 2023-09-27
Feat: Detect all text in HTML Heading tags as titles

newelh:
## Summary
This will increase the accuracy of hierarchies in HTML documents and provide more accurate element categorization. If text is in an HTML heading tag and is not a list item, address categorize it as a title. 

## Testing
```
elements = partition_html(url="https://www.eda.gov/grants/2015")
```
Before, the date headers at the given url would not be correctly parsed as titles, after this change they are now correctly identified.

A unit test to verify the functionality has been added: `test_html_partition::test_html_heading_title_detection` that includes values that were previously detected as narrative text and uncategorized text


badGarnet (approved):
confirm the test would fail with `<h1>` content being id'ed as text with the version on main. 
the test website from PR note returns 504 😢 so can't verify that case
code does make sense
grabbed a few random websites but do not see difference between this PR and main (could be missing those naked `<h1>` tags)

newelh:
> confirm the test would fail with `<h1>` content being id'ed as text with the version on main. the test website from PR note returns 504 😢 so can't verify that case code does make sense grabbed a few random websites but do not see difference between this PR and main (could be missing those naked `<h1>` tags)

Yes, confirmed that the test fails on main branch. (e.g. something that would be categorized by main as NarrativeText or Uncategorized text is now properly categorized as a title)

PR #1557, opened 2023-09-27
build: ignore failing delta lake test ingest for now

badGarnet:


test_unstructured_ingest/test-ingest.sh
```diff
@@ -56,7 +56,7 @@ trap print_last_run EXIT
 
 for script in "${scripts[@]}"; do
   CURRENT_SCRIPT=$script
-  if [[ "$CURRENT_SCRIPT" == "test-ingest-notion.sh" ]]; then
+  if [[ "$CURRENT_SCRIPT" == "test-ingest-notion.sh" ]] || [[ "$CURRENT_SCRIPT" == "test-ingest-notion.sh" ]]; then
```
qued, line 59: Am I missing something? It looks like this just changed from `if A` to `if A or A` which seems like a no-op to me. I'm guessing the second thing should be whatever script handles delta tables?

badGarnet: good catch; fixed

ryannikolaidis (approved):

PR #1586, opened 2023-09-29
Klaijan/fix: isalnum reference before assignment

Klaijan:
**Executive Summary**
Fix bug on the `get_word_bounding_box_from_element` function that prevent `partition_pdf` to run.

**Technical Details**
- The function originally first define `isalnum` on the first index. Now switched to conditional on flag value.

### DIFF
--- a/unstructured/partition/pdf.py
+++ b/unstructured/partition/pdf.py
@@ -868,6 +868,23 @@ def get_uris(
     coordinate_system: Union[PixelSpace, PointSpace],
     page_number: int,
 ) -> List[dict]:
+    """
+    Extracts URI annotations from a single or a list of PDF object references on a specific page.
+    The type of annots (list or not) depends on the pdf formatting. The function detectes the type
+    of annots and then pass on to get_uris_from_annots function as a List.
+
+    Args:
+        annots (Union[PDFObjRef, List[PDFObjRef]]): A single or a list of PDF object references
+            representing annotations on the page.
+        height (float): The height of the page in the specified coordinate system.
+        coordinate_system (Union[PixelSpace, PointSpace]): The coordinate system used to represent
+            the annotations' coordinates.
+        page_number (int): The page number from which to extract annotations.
+
+    Returns:
+        List[dict]: A list of dictionaries, each containing information about a URI annotation,
+        including its coordinates, bounding box, type, URI link, and page number.
+    """
     if isinstance(annots, List):
         return get_uris_from_annots(annots, height, coordinate_system, page_number)
     return get_uris_from_annots(annots.resolve(), height, coordinate_system, page_number)
@@ -879,6 +896,21 @@ def get_uris_from_annots(
     coordinate_system: Union[PixelSpace, PointSpace],
     page_number: int,
 ) -> List[dict]:
+    """
+    Extracts URI annotations from a list of PDF object references.
+
+    Args:
+        annots (List[PDFObjRef]): A list of PDF object references representing annotations on
+            a page.
+        height (Union[int, float]): The height of the page in the specified coordinate system.
+        coordinate_system (Union[PixelSpace, PointSpace]): The coordinate system used to represent
+            the annotations' coordinates.
+        page_number (int): The page number from which to extract annotations.
+
+    Returns:
+        List[dict]: A list of dictionaries, each containing information about a URI annotation,
+        including its coordinates, bounding box, type, URI link, and page number.
+    """
     annotation_list = []
     for annotation in annots:
         annotation_dict = try_resolve(annotation)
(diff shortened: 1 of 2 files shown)


unstructured/partition/pdf.py
```diff
@@ -1002,8 +1106,9 @@ def get_word_bounding_box_from_element(
 
                 # TODO(klaijan) - isalnum() only works with A-Z, a-z and 0-9
                 # will need to switch to some pattern matching once we support more languages
-                if index == 0:
+                if not set_alnum:
                     isalnum = char.isalnum()
+                    set_alnum = True
 
                 if char.isalnum() != isalnum:
```
badGarnet, line 1113: maybe think about this problem from a different angle? What does `isalnum` 's intention? It seems that if a `char`'s check is not the same as the last check we add a word? So on first pass we always expect the next condition to pass at line 1113 and word would be not empty after that. 
So in other words if word is empty we know we don't need to check at line 1113 so we can replace the highlighted lines with

``` 
if not word:
    isalnum = char.isalnum()
if word and char.isalnum() != isalnum:
```

badGarnet (commented):
commented on the code; also can we add a test that would fail without the change please (that triggers the bug)

Klaijan:
> commented on the code; also can we add a test that would fail without the change please (that triggers the bug)

@badGarnet will require us to find a document with url and first element of the object is not `LTChar`. I will ask for the document sample from community Slack. But otherwise, I can come back to find the document later late night today or tomorrow.

badGarnet:
> > commented on the code; also can we add a test that would fail without the change please (that triggers the bug)
> 
> @badGarnet will require us to find a document with url and first element of the object is not `LTChar`. I will ask for the document sample from community Slack. But otherwise, I can come back to find the document later late night today or tomorrow.

we can mock a doc to trigger this condition

unstructured/partition/pdf.py
```diff
@@ -1002,10 +1105,9 @@ def get_word_bounding_box_from_element(
 
                 # TODO(klaijan) - isalnum() only works with A-Z, a-z and 0-9
                 # will need to switch to some pattern matching once we support more languages
-                if index == 0:
+                if not word:
                     isalnum = char.isalnum()
-
-                if char.isalnum() != isalnum:
+                if word and char.isalnum() != isalnum:
```
cragwolfe, line 1110: if i'm following this logic correctly, can this be simplified to:

    isalnum = char.isalnum()
    if word and isalnum == False:
        ....

Klaijan: This condition here is to use `isalnum` flag to help us split the `word` by alphanumeric and non-alphanumeric property in the text sentence. For example, `Lorem ipsum dolor sit amet, consectetur "adipiscing" elit.` would be split as 

```
Lorem 
ipsum 
dolor 
sit 
amet
, 
consectetur
"
adipiscing
" 
elit
.
```

nikunjgoel95: @Klaijan Did you got a test PDF to test the function? if not https://documentation.nokia.com/sr/23-7-2/pdf/Interface_Configuration_Guide_23.7.R1.pdf

nikunjgoel95 (approved):

cragwolfe (approved):

cragwolfe:
> @Klaijan Did you got a test PDF to test the function? if not https://documentation.nokia.com/sr/23-7-2/pdf/Interface_Configuration_Guide_23.7.R1.pdf

LGTM assuming problematic .pdf passes. 

Also, it would be worth extracting the problematic page and adding it is a test. The page could be extracted with a command line:

    qpdf --empty --pages orignal.pdf 89 -- ~/tmp/docs/single-page-p89.pdf

Rmaram07:
Hi, I'm facing the same issue. When will this be merged to main branch? I see updated code in your repo but not in the main branch.

vprzybylo:
I am still experiencing this issue. Are any updates or reinstallations needed on our end?  Unfortunately, I can't share the document due to corporate policy.

File "C:\Users\przybv\Anaconda3\envs\sentiment\Lib\site-packages\unstructured\partition\pdf.py", line 1004, in get_word_bounding_box_from_element
    if char.isalnum() != isalnum:

Klaijan:
Hi @vprzybylo please try to pull the latest version from main and rerun your document again. Do not hesitate to reach out if the problem still persist for you!

anjanikant06:
Hi @Klaijan , I am using AWS Glue to run our code, hence need to do pip install unstructured. but still it shows the same issue for me.  local variable 'isalnum' referenced before assignment. can you please help what to do ? Thanks

Klaijan:
@anjanikant06 Then please stay tuned for our new pypi release to update the package which will include the bug fix for `isalnum` issue. Thank you!

Klaijan:
@anjanikant06 actually we just released new version 0.10.19. Please try to upgrade the package :)

PR #1587, opened 2023-09-30
fix: Fix various cases of HTML text missing after partition

### LINKED ISSUE #1543: bug/HTML partitioner doesn't partition plain text or break tags correctly
**Describe the bug**
The html partitioner doesn't partition correctly for completely plain html body text (without tags). The text doesn' t show up in the json result.

Also break tags, `<br>, <br /> cause the text before the break to not show up in the json result.

**To Reproduce**
Try to partition this simple html file:
```
<html>
<body>
body test
<p>p test</p>
<p>p test 2 <br>
p test 2 after break</p>
</body>
</html>
```

This results in the following json:
```
[
  {
    "type": "Title",
    "element_id": "720717b9e461ce3a899d3b814f675685",
    "metadata": {
      "data_source": {
        "date_processed": "2023-09-27T04:09:51.958194"
      },
      "filename": "simplest-html.html",
      "file_directory": "html-test",
      "last_modified": "2023-09-24T21:36:09",
      "filetype": "text/html",
      "category_depth": 0,
      "page_number": 1
    },
    "text": "p test"
  },
  {
    "type": "Title",
    "element_id": "601b2acb792f7bedc5544f24d8518a78",
    "metadata": {
      "data_source": {
        "date_processed": "2023-09-27T04:09:51.958194"
      },
      "filename": "simplest-html.html",
      "file_directory": "html-test",
      "last_modified": "2023-09-24T21:36:09",
      "filetype": "text/html",
      "category_depth": 0,
      "page_number": 1
    },
    "text": "p test 2 after break"
  }
]
```
Note the lines that are missing.

**Expected behavior**
I would expect `body test` and `p test 2` to be in the json also. But they are not.

**Environment Info**
OS version:  macOS-13.5.2-arm64-arm-64bit
Python version:  3.8.15
unstructured version:  0.10.11
unstructured-inference version:  0.5.1
pytesseract version:  0.3.10
Torch version:  2.0.1
Detectron2 is not installed

**Additional context**
Break tags and plain text may not be common in html today. But the html parser is not just used for modern web pages. The html partitioner is also used in .eml files and probably other internal files.

unifyh:
Fix 4 cases of text missing after partition:
1. Text immediately after `<body>`
```html
<body>
  missing1
  <div>hello</div>
</body>
```

2. Text inside container and immediately after `<br/>`
```html
<div>hello<br/>missing2</div>
```

3. Text immediately after a text opening tag, if said tag contains `<br/>`
```html
<p>missing3<br/>hello</p>
```

4. Text inside `<body>` if it is the only content (different cause from case 1)
```html
<body>missing4</body>
```

Also fix problem causing `test_unstructured/documents/test_html.py::test_exclude_tag_types` to not work as intended.

This will close #1543

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 
 ### Fixes
 
+* **Fix various cases of HTML text missing after partition**
+  Problem: Under certain circumstances, text immediately after some HTML tags will be misssing from partition result.
+  Fix: Updated code to deal with these cases.
+  Importance: This will ensure the correctness when partitioning HTML and Markdown documents.
+
 
 ## 0.10.18
 
--- a/test_unstructured/documents/test_html.py
+++ b/test_unstructured/documents/test_html.py
@@ -17,6 +17,7 @@ from unstructured.documents.elements import (
 from unstructured.documents.html import (
     HEADING_TAGS,
     LIST_ITEM_TAGS,
+    SECTION_TAGS,
     TABLE_TAGS,
     TEXT_TAGS,
     HTMLDocument,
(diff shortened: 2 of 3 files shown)


cragwolfe:
Thanks for the contribution, @unifyh ! Please add a unittest confirming the fix. 🙏

unifyh:
Unit tests added. ☺️

cragwolfe (approved):
LGTM!

Issue #1588, opened 2023-09-30
chore: connectors should leverage the `languages` argument rather than `ocr_languages`


ryannikolaidis:
**Is your feature request related to a problem? Please describe.**
Warnings about the soon to be deprecated ocr_languages argument appear in the output when running partition via Unstructured connectors.

```
The ocr_languages kwarg will be deprecated in a future version of unstructured. Please use languages instead.
```

**Describe the solution you'd like**
Unstructured connectors are updated to leverage the `languages` argument rather than `ocr_languages`

**Describe alternatives you've considered**
NA

**Additional context**
NA

PR #1605, opened 2023-10-02
roman/update language kwarg partition ingest 

rbiseck3:
### Description
Change partition kwarg from ocr_languages -> languages

Closes out https://github.com/Unstructured-IO/unstructured/issues/1588

### DIFF
--- a/unstructured/ingest/processor.py
+++ b/unstructured/ingest/processor.py
@@ -107,10 +107,11 @@ def process_documents(
     verbose: bool,
     dest_doc_connector: t.Optional[BaseDestinationConnector] = None,
 ) -> None:
+    languages = partition_config.ocr_languages.split("+") if partition_config.ocr_languages else []
     process_document_with_partition_args = partial(
         process_document,
         strategy=partition_config.strategy,
-        languages=partition_config.ocr_languages,
+        languages=languages,
         encoding=partition_config.encoding,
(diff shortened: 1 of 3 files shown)


unstructured/ingest/processor.py
```diff
@@ -107,10 +107,11 @@ def process_documents(
     verbose: bool,
     dest_doc_connector: t.Optional[BaseDestinationConnector] = None,
 ) -> None:
+    languages = partition_config.ocr_languages.split("+") if partition_config.ocr_languages else []
```
badGarnet, line 110: how about we define the joiner character/pattern somewhere in the repo and document it? 
also since the definition for `ocr_languages` is str, not optional str we can safely just write 
```
languages = partition_config.ocr_languages.split("+")
```

rbiseck3: The joiner character right now is documented in the CLI help string: [unstructured/ingest/cli/interfaces.py#L101](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/cli/interfaces.py#L101):
```python
            click.Option(
                ["--ocr-languages"],
                default="eng",
                help="A list of language packs to specify which languages to use for OCR, "
                "separated by '+' e.g. 'eng+deu' to use the English and German language packs. "
                "The appropriate Tesseract "
                "language pack needs to be installed."
                "Default: eng",
            ),
```

PR #1607, opened 2023-10-02
feat: change default hi_res model to yolox quantized

badGarnet:
- refactor the way to set default model for hi_res mode for image and pdf partition into a function that is callable and returns either an env varaible or a default model
- this keeps the current pattern for setting the default hi_res model
- change the default model name from `detectron2_onnx` to `yolox_quantized`
- the new default mode has better recall for tables and richer categories for partitioned elements than detectron2


badGarnet:
please review comments here for the ingest diff eval: https://github.com/Unstructured-IO/unstructured/pull/1618

benjats07 (approved):
LGTM, however, I would like to count with the metrics to make this change (not necessarily a blocker). I was able to manually found a lot of the text of the previous model in the new ingest tests.

badGarnet:
> LGTM, however, I would like to count with the metrics to make this change (not necessarily a blocker). I was able to manually found a lot of the text of the previous model in the new ingest tests.

actually do we have gold label for those ingest test documents? if not maybe that is something we should have even before generating metrics

unstructured/partition/pdf.py
```diff
@@ -70,6 +70,13 @@
 RE_MULTISPACE_INCLUDING_NEWLINES = re.compile(pattern=r"\s+", flags=re.DOTALL)
 
 
+def default_hi_res_model() -> str:
+    # a light config for the hi res model; this is not defined as a constant so that no setting of
+    # the default hi res model name is done on importing of this submodule; this allows (if user
+    # perfers) for setting env after importing the sub module and change the default model name
```
qued, line 76:
```suggestion
    # prefers) for setting env after importing the sub module and changing the default model name
```

qued (approved):
LGTM other than a documentation nit.

PR #1614, opened 2023-10-02
roman/drop downloads in ingest tests

rbiseck3:
### Description
In an effort to mitigate resource consumption when running CI tests, cleanup download dir for ingest tests after each one.


ryannikolaidis (approved):

cragwolfe (changes requested):
can we only cleanup if the CI env var is defined? having downloaded files tends to be more useful for devs than not by default, i think

rbiseck3:
I usually just comment that line out when running locally if I want to review the downloaded content.

cragwolfe (approved):
approving so as not to block, but i think the default for dev's really ought to be downloads exist locally (more useful debugging by default), with a super obvious switch to turn it off for those that are not used to running ingest all the time.

PR #1616, opened 2023-10-02
Feat: Native hierarchies for elements from pptx documents

newelh:
## Summary
**Improve title detection in pptx documents** The default title textboxes on a pptx slide are now categorized as titles.
**Improve hierarchy detection in pptx documents** List items, and other slide text are properly nested under the slide title. This will enable better chunking of pptx documents.

Hierarchy detection is improved by determining category depth via the following:
- Check if the paragraph item has a level parameter via the python pptx paragraph. If so, use the paragraph level as the category_depth level.
- If the shape being checked is a title shape and the item is not a bullet or email, the element will be set as a Title with a depth corresponding to the enumerated paragraph increment (e.g. 1st line of title shape is depth 0, second is depth 1 etc.). 
- If the shape is not a title shape but the paragraph is a title, the increment will match the level + 1, so that all paragraph titles are at least 1 to set them below the slide title element

## Testing
This PR adds an additional pptx file to example-docs: presentation-sample.pptx. This document illustrates slide titles, and bullet hierarchies in a pptx file.

### Try it yourself:
```
document = partition_pptx("example-docs/sample-presentation.pptx")
for element in document:
    print(
        f"Depth : {element.metadata.category_depth}\n"
        f"Parent: {element.metadata.parent_id}\n" \
        f"ID    : {element.id}\n" \
        f"Text  : {element.text}\n" \
        f"Type  : {element.category}\n\n" \
    )
```

If you have `graphviz` and `pygraphviz` installed, you can visualize the hierarchy with:
```
G = pgv.AGraph(strict=False, directed=True, dpi=300, rankdir='TB')

def wrap_text(text, width=50):
    return textwrap.fill(text, width=width)

# Adding nodes and edges
for element in document:
    wrapped_text = wrap_text(element.text)
    G.add_node(element.id, label=f"{element.category}\n{wrapped_text}", shape='box')
    if element.metadata.parent_id is not None:
        G.add_edge(element.metadata.parent_id, element.id)

# Render the graph
G.layout(prog='dot')
G.draw('hierarchy.png')
```

### Unit Tests
1. `test_partition_pptx_title_shape_detection`: Tests title element detection for text in title shapes. It includes a check for additional paragraphs within the title shape text frame and interprets them as Titles with incremental depth. 
2. `test_partition_pptx_level_detection`: Tests if the indentation level pulled from the pptx xml via python_pptx is correctly being set as a category depth.
3. `test_partition_pptx_hierarchy_sample_document`: Tests if the hierarchy in the sample file is being correctly parsed 

## Example
Sample Slides
Screenshot 2023-10-02 at 3 00 44 PM

### Before
The sample powerpoint yields no hierarchy relationships

###  After
document_structure_hierarchy


unstructured/partition/pptx.py
```diff
@@ -156,13 +156,18 @@ def _iter_presentation_elements(self) -> Iterator[Element]:
...
                 if shape.has_table:
                     assert isinstance(shape, GraphicFrame)
                     yield from self._iter_table_element(shape)
                 elif shape.has_text_frame:
                     assert isinstance(shape, Shape)
-                    yield from self._iter_paragraph_elements(shape)
+                    yield from self._iter_paragraph_elements(
+                        shape,
+                        is_title_shape=bool(shape == title_shape),
+                    )
```
scanny, line 170: I'm inclined to think we would want to unconditionally process the title shape first, when it exists, because it provides hierarchical context for all the other shapes on the slide. Isn't that right? Or am I misunderstanding hierarchy or do we want to except shapes that appear above the title somehow?

If we want to process it separately, then `._order_shapes()` would still get this new signature, possibly `-> (title_shape: Shape | None, body_shapes: List[Shape])` giving title-shape pride of place, and then process title shape outside the loop when present.

Also in that case, I'd be inclined to make a new `._iter_title_shape_elements()` rather than add a flag argument to `._iter_paragraph_elements()`. That's Uncle Bob's Clean Code Tip #12 Avoid Boolean Flag Arguments although he's not the only one and perhaps not even the first to mention that in the literature:

https://martinfowler.com/bliki/FlagArgument.html
https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

By the way, with the benefit of a few nights sleep and looking back on this, I'm thinking `._iter_paragraph_elements()` should really be `._iter_shape_elements()`, named for what it's decomposing rather than what it decomposes the shape into.

unstructured/partition/pptx.py
```diff
@@ -284,7 +304,7 @@ def iter_shapes(shapes: _BaseGroupShapes) -> Iterator[BaseShape]:
         def sort_key(shape: BaseShape) -> Tuple[int, int]:
             return shape.top or 0, shape.left or 0
 
-        return sorted(iter_shapes(slide.shapes), key=sort_key)
+        return sorted(iter_shapes(slide.shapes), key=sort_key), getattr(slide.shapes, "title", None)
```
scanny, line 307: `Shapes.title` is always present, isn't it? Just that it returns `() -> Shape | None`? So no need for a `getattr()` here, just `slide.shapes.title`.

scanny (approved):
Couple comments, happy to discuss in a quick meet if you like.

Approving in advance of whatever changes you decide to make there, this looks like it works fine for the intended behavior :)

unstructured/partition/pptx.py
newelh: I agree. Boolean didn't feel quite right here, felt like I was handjamming to fit your spec. I think an `_iter_title_shape_elements()`makes sense instead. I can also rename that function.

unstructured/partition/pptx.py
scanny: Ohhh, right, of course. There's no entry for `Shapes.title` in the type-stub yet.

You can add this to the bottom of `typings/pptx/shapes/shapetree.pyi` to fix that up:
```python
    @property
    def title(self) -> Shape | None: ...
```

Will look like this afterward and that should get rid of the mypy error:
```python
class SlideShapes(_BaseGroupShapes):
    def __iter__(self) -> Iterator[BaseShape]: ...
    @property
    def title(self) -> Shape | None: ...
```

scanny (approved):
LGTM :)

PR #1619, opened 2023-10-02
Feat: Adds local embedding model

mallorih:
This PR adds a local embedding model option as an alternative to using our OpenAI embedding brick. This brick uses LangChain's HuggingFacEmbeddings.

**Testing**

```Python
from unstructured.embed.huggingface import HuggingFaceEmbeddingEncoder

hf = HuggingFaceEmbeddingEncoder()
hf.embed_query("Q")
```


pravin-unstructured (approved):
LGTM. Will test it out tomorrow.

unstructured/embed/huggingface.py
```diff
@@ -0,0 +1,71 @@
...
+        return self.hf
+
+    def num_of_dimensions(self):
+        return np.shape(self.examplary_embedding)
+
+    def is_unit_vector(self):
+        return np.isclose(np.linalg.norm(self.examplary_embedding), 1.0)
+
+    def embed_query(self, query):
+        return self.hf.embed_documents([str(query)])
```
ahmetmeleq, line 49:
```suggestion
        return self.hf.embed_query(str(query))
```
Needs to be updated in the OpenAI embedding model likewise

ahmetmeleq:
In terms of the `extras` parameter in `requires_dependencies`, a different naming convention can be implemented in the future to avoid clash with existing extras (as was in your case). Should be alright to skip decorating with `requires_dependencies` for now

ahmetmeleq:
To clarify: Unintentionally updated the changelog, then reverted

ahmetmeleq:
Reminder to update the docs: https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/bricks/embedding.rst

ahmetmeleq:
1. Let's form an extra named `embed-huggingface` and add it as an extra key to setup.py as in:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/setup.py#L160

2. We need an .in requirements file for this to work (we can name it `embed-huggingface.in`). Example:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/requirements/ingest-openai.in

3. Finally, let's add a `@requires_dependencies` decorator to the methods where we use the extra libraries explicitly. Example:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/unstructured/embed/openai.py#L44-L49

ahmetmeleq:
Let's wrap potential connection errors as in:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/unstructured/embed/openai.py#L44

mallorih:
> Let's wrap potential connection errors as in:
> 
> https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/unstructured/embed/openai.py#L44

Do we need this if there is no API connection?

ahmetmeleq:
> > Let's wrap potential connection errors as in:
> > https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/unstructured/embed/openai.py#L44
> 
> Do we need this if there is no API connection?

For our case, the part where the embedding model gets downloaded would be suitable to get wrapped with `EmbeddingEncoderConnectionError`; so if there's any connection failure related to downloading the model, the decorator class will handle the error message

.github/workflows/ci.yml
```diff
@@ -334,6 +334,7 @@ jobs:
         make install-ingest-notion
         make install-ingest-delta-table
         pip install "unstructured[openai]"
+        pip install "unstructured[langchain]"
```
ryannikolaidis, line 337: I think ingest test imports are already covered by the `pip install "unstructured[openai]"` above this.

it looks like your test is added as a unit test, so the install here won't help. ultimately if this is now required for the unit tests to pass, probably means this needs to be required [via tests](https://github.com/Unstructured-IO/unstructured/blob/main/requirements/test.in) so that way anyone running the tests has it, right?

mallorih: Ok, so remove the pip install from ci.yml and add `langchain` to the test.in file?

unstructured/embed/huggingface.py
```diff
@@ -0,0 +1,69 @@
...
+        self.model_name = model_name
+        self.model_kwargs = model_kwargs
+        self.encode_kwargs = encode_kwargs
+        self.cache_folder = cache_folder
+
+        self.initialize()
+
+    @EmbeddingEncoderConnectionError.wrap
+    @requires_dependencies(
+        ["langchain", "huggingface"],
```
ahmetmeleq, line 31: Per the first code cell in https://python.langchain.com/docs/integrations/text_embedding/huggingfacehub
```suggestion
        ["langchain", "sentence_transformers"],
```
The list items here signify what might be lacking in terms of `pip install`s

unstructured/embed/huggingface.py
```diff
@@ -0,0 +1,69 @@
...
+        extras="huggingface",
```
ahmetmeleq, line 32:
```suggestion
        extras="embed-huggingface",
```
This string signifies which extra that we defined can be utilized to install each of the potentially missing `pip install`s altogether. 

For our case, for it to not be confused with the existing huggingface extra, let's name it as `embed-huggingface`. 

Since we name it as `embed-huggingface`, now we need to set it up. To do that, we will need to delete these:
- https://github.com/Unstructured-IO/unstructured/blob/88592e5c522bdbd81f2ec8058c28c7d237690bb8/requirements/langchain.in
- https://github.com/Unstructured-IO/unstructured/blob/88592e5c522bdbd81f2ec8058c28c7d237690bb8/requirements/langchain.txt
- https://github.com/Unstructured-IO/unstructured/blob/88592e5c522bdbd81f2ec8058c28c7d237690bb8/setup.py#L161

(and remove any other change we introduced for an extra named "langchain")

And we need to create them again for `embed-huggingface`, with relevant requirements. For example, `embed-huggingface.in` should look like:
```
-c constraints.in
-c base.txt
langchain
sentence_transformers
```

ahmetmeleq: So after setting up the extra correctly, the CI job (test unit) should be able to install all requirements via the name of the dependency

unstructured/embed/huggingface.py
```diff
@@ -0,0 +1,74 @@
...
+        elements_w_embedding = []
+
+        for i, element in enumerate(elements):
+            element.embeddings = embeddings[i]
+            elements_w_embedding.append(element)
+        return elements
+
+    @EmbeddingEncoderConnectionError.wrap
+    @requires_dependencies(
+        ["langchain", "sentence-transformers"],
```
ahmetmeleq, line 57:
```suggestion
        ["langchain", "sentence_transformers"],
```

qued (changes requested):
We should add an extra to `setup.py` for `embed-huggingface`. This is done by adding a key `embed-huggingface` to the `dict` passed to `setup` under the `extras_require` parameter. The pattern to follow is the one with `load_requirements`.

qued:
Also we should add `make` targets as well.

ahmetmeleq:
> We should add an extra to `setup.py` for `embed-huggingface`. This is done by adding a key `embed-huggingface` to the `dict` passed to `setup` under the `extras_require` parameter. The pattern to follow is the one with `load_requirements`.

Addressed with [add make target and extra to setup](https://github.com/Unstructured-IO/unstructured/pull/1619/commits/26275e4be8c0c43c4401da28fdc5d381926a0143)

ahmetmeleq:
> Also we should add `make` targets as well.

Addressed with [add make target and extra to setup](https://github.com/Unstructured-IO/unstructured/pull/1619/commits/26275e4be8c0c43c4401da28fdc5d381926a0143)

unstructured/embed/huggingface.py
```diff
@@ -0,0 +1,74 @@
...
+    @requires_dependencies(
+        ["langchain", "sentence_transformers"],
+        extras="embed-huggingface",
+    )
+    def get_huggingface_client(self):
+        """Creates a langchain Huggingface python client to embed elements."""
+        if hasattr(self, "hf_client"):
+            return self.hf_client
+
+        from langchain.embeddings import HuggingFaceEmbeddings
```
badGarnet, line 65: I just want to leave a comment, not asking for changes, here that this connector is basically a wrapper around langchain's connector. It adds dependency complexity. If we consider what we do is upstream of langchain this also potentially creates circular dependency (at least logically right now). huggingface provides interface to use its model without the need for a wrapper and I'd at least suggest we keep that in mind. Because as this pattern propagates our ingestion pipeline will be dependent on `langchain` to function and that may just be adding one too many major dependencies.

mallorih: So should we switch in the future? Or do you think it's not too big of a task to switch now? We need this for a contract that has a deadline coming up by the end of this month.

qued: With you importing langchain inside the function like this it probably minimizes the chance of just breaking langchain outright. As long as we aren't creating a circular dependency I'm with @badGarnet that this is probably ok for now, but would want to figure out how to drop the langchain dep in the future.

qued (approved):
Adding an approval since my comments have been addressed, but I think we'll need a full `make pip-compile` and commit all the changes to the `requirements/*.txt` files to get rid of conflicts.

PR #1620, opened 2023-10-02
Fix: Handle case where uri should default to None

jhallas:
Ran into an issue where I got an error that uri was not defined, (the first two if statements did not trigger, nor did the exception)


Klaijan:
This bug is fixed here https://github.com/Unstructured-IO/unstructured/pull/1690. Thank you for your contribution tho!

Issue #1535, opened 2023-09-26
feat/per element language detection


shreyanid:
If a language is not provided by the user, use[ langdetect](https://pypi.org/project/langdetect/) to detect the language of the text. First task is to detect language on a document level.

Use the document-level results from the detect_languages function to determine the need for element level detection. If the document contains multiple languages, or the detected language confidence is too low, evaluate language on a per-element level to ensure more granular language accuracy for each element.

TODO:
- decide a confidence threshold for enabling element level language detection vs applying the document language to all elements
- apply element level language detection to all multi-language documents
- link element language to its element text via metadata field

PR #1627, opened 2023-10-03
detect document language across all partitioners

### LINKED ISSUE #1534: feat/finish implementing document level language detection for other doctypes
If the document languages are not provided by the user, use [langdetect](https://pypi.org/project/langdetect/) to detect the language of the text (on a document level for speed). If confidence in result is high enough, we can assume all elements are in the detected language.

Pattern has been established in this [PR](https://github.com/Unstructured-IO/unstructured/pull/1453) for text partitioning. Apply this pattern - adding the languages parameter for user input, detecting the document language, and adding the resulting document language to the element metadata - to all other non-image-based documents (all but pdf and image) partitioning functions as well as auto partition.

Coniferish:
### Summary
Closes #1534 and #1535
Detects document language using `langdetect` package. 
Creates new kwargs for user to set the document language (`languages`) or detect the language at the element level instead of the default document level (`detect_language_per_element`)


Coniferish:
@shreyanid, I need to add documents for testing for the remaining partitioners, but will get that finished tomorrow. Otherwise, this should be ready for review.

test_unstructured/partition/test_text.py
```diff
@@ -165,9 +165,9 @@ def test_partition_text_from_bytes_file_default_encoding(filename):
         assert element.metadata.filename is None
 
 
-def test_auto_partition_element_metadata_user_provided_languages():
+def test_text_partition_element_metadata_user_provided_languages():
```
Coniferish, line 168: This fixes a mistake in the test from a prior commit

test_unstructured/partition/test_text.py
```diff
@@ -514,3 +514,37 @@ def test_add_chunking_strategy_on_partition_text(filename="example-docs/norwich-
...
     assert chunk_elements == chunks
+
+
+def test_partition_text_element_metadata_has_languages():
+    filename = "example-docs/norwich-city.txt"
+    elements = partition_text(filename=filename)
+    assert elements[0].metadata.languages == ["eng"]
+
+
+# def test_partition_text_detects_quote_in_other_language():
```
Coniferish, line 525: @shreyanid 
`langdetect` doesn't seem to be sensitive enough to detect quotes in other languages, so let me know if these tests can be removed.

Coniferish:
I don't have MS Outlook, so I'm having a little trouble creating a .msg file with multiple languages for testing if anyone can help with that.

cragwolfe:
> I don't have MS Outlook, so I'm having a little trouble creating a .msg file with multiple languages for testing if anyone can help with that.

worst case can punt to a follow-on github issue

unstructured/file_utils/metadata.py
```diff
@@ -150,3 +153,43 @@ def _get_exif_datetime(exif_dict: Dict[str, Any], key: str) -> Optional[datetime
     # using the standard EXIF datetime format
     except ValueError:
         return None
+
+
+def apply_lang_metadata(
+    elements: Union[Iterable[Element], List[Element]],
```
scanny, line 159: `list` is iterable, so this can be simplified to just `Iterable[Element]`.

unstructured/file_utils/metadata.py
```diff
@@ -150,3 +153,43 @@ def _get_exif_datetime(exif_dict: Dict[str, Any], key: str) -> Optional[datetime
...
+        elements = list(elements)
+
+    full_text = " ".join(e.text for e in elements if hasattr(e, "text"))
+    detected_languages = detect_languages(text=full_text, languages=languages)
+    if (
+        detected_languages is not None
+        and len(languages) == 1
+        and detect_language_per_element is False
+    ):
+        # -- apply detected languge to each metadata --
```
scanny, line 185: "language" (typo)

unstructured/partition/docx.py
```diff
@@ -248,6 +280,20 @@ def iter_document_elements(
...
+        # -- and using `elements.extend()` for the results of each of the function calls, but is
+        # -- more perfomant, uses less memory (avoids producing and then garbage-collecting all
+        # -- those small lists), is more flexible for later iterator operations like filter,
+        # -- chain, map, etc. and is perhaps more elegant and simpler to read once you have the
+        # -- concept of what it's doing. You can see the same pattern repeating in the "sub"
+        # -- functions like `._iter_paragraph_elements()` where the "just return when done"
+        # -- characteristic of a generator avoids repeated code to form interim results into lists.
+
+        # full_text = ""
+
```
scanny, line 296: did you mean to leave this commented line in?

unstructured/partition/docx.py
```diff
@@ -470,11 +516,10 @@ def page_is_odd() -> bool:
         # -- predict when two page breaks will be needed and emit one of them. The second will be
         # -- emitted by the rendered page-break to follow.
 
-        if start_type == WD_SECTION_START.EVEN_PAGE:
+        if start_type == WD_SECTION_START.EVEN_PAGE and not page_is_odd():
             # -- on an even page we need two total, add one to supplement the rendered page break
             # -- to follow. There is no "first-document-page" special case because 1 is odd.
-            if not page_is_odd():
-                yield from self._increment_page_number()
+            yield from self._increment_page_number()
```
scanny, line 522: I'd be inclined to leave this as it was. If the type-checker or linter complains we should place an exception comment. The reason being that a start type is either even or odd and can't be both, so the top-level logic is just partitioning these two cases. The following branch (testing whether start-type is odd) should not be executed if start-type is even.

It's a small thing, bit I did it that way on purpose for that reason, even though this new logic seems simpler to the static checker.

unstructured/partition/docx.py
```diff
@@ -173,20 +187,38 @@ def partition_docx(
...
+                metadata_filename,
+                include_page_breaks,
+                metadata_last_modified,
+            ),
+            languages=languages,
+            detect_language_per_element=detect_language_per_element,
         ),
     )
 
+    return elements
```
scanny, line 220: A form like this might be worth considering. Same semantics, just maybe easier to follow in step-wise bite-sized pieces:

```python
    elements = _DocxPartitioner.iter_document_elements(
        filename,
       file,
        metadata_filename,
        include_page_breaks,
        metadata_last_modified,
    )
    elements = apply_lang_metadata(
        elements=elements,
        languages=languages,
    )
    return list(elements)
```

This also has the advantage that it's readily scalable, like if we add an additional step it can just get in line at the right place.

(same comment later on pptx partitioner)

scanny (commented):
Hi John, a few comments about the parts I'm familiar with. Probably need someone else to approve the overall, I'd be flying blind for most of that :)

unstructured/partition/pptx.py
```diff
@@ -89,12 +104,16 @@ def partition_pptx(
...
+                source_file,
+                include_page_breaks,
+                include_slide_notes,
+                metadata_filename,
+                metadata_last_modified,
+            ),
+            languages=languages,
+            detect_language_per_element=detect_language_per_element,
         ),
     )
```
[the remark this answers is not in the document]
Coniferish: Sounds good. Wasn't sure which would be preferred, but will make that change.

unstructured/partition/docx.py
Coniferish: Nope. Thanks!

unstructured/partition/docx.py
Coniferish: Sounds good. Yah, that was changed because of the linter. Will add an exception

unstructured/file_utils/metadata.py
Coniferish: I'm getting this error in the checks if I remove that (in an number of the partitioners, not just `xlsx`):
`unstructured/partition/xlsx.py:176: error: Incompatible types in assignment (expression has type "Iterator[Element]", variable has type "List[Element]")  [assignment]`

Should I add it back because of the linting error?

unstructured/partition/lang.py
```diff
@@ -221,7 +221,14 @@ def detect_languages(
...
     """
-    if text.strip() == "":
+    # Skip language detection for partitioners that use other partitioners.
+    # For example, partition_msg relies on partition_html and partition_text, but the metadata
+    # gets overwritten after elements have been returned by _html and _text,
+    # so `languages` would be detected twice.
+    if languages[0] == "":
+        return [""]
+
+    if text.strip() == "" or len(text) < 20:
```
Coniferish, line 231: @shreyanid
I added this because "Stanley Cups" wasn't being detected as English when partitioning xlsx, but now one of the language tests (test_detect_languages_korean_auto) is failing because the text is only 5 characters long (though very obviously not English). Any suggestions on how to find a balance?

Coniferish: Never mind. Thought of using regex to check if characters were all ascii

unstructured/file_utils/metadata.py
```diff
@@ -150,3 +153,43 @@ def _get_exif_datetime(exif_dict: Dict[str, Any], key: str) -> Optional[datetime
...
+) -> Iterator[Element]:
+    """Detect and apply metadata.languages to each element in `elements`."""
+    # -- Note this function has a stream interface, but reads the full `elements` stream into memory
+    # -- before emitting the first updated element as output.
+
+    # Skip language detection for partitioners that use other partitioners.
+    # For example, partition_msg relies on partition_html and partition_text, but the metadata
+    # gets overwritten after elements have been returned by _html and _text,
+    # so `languages` would be detected twice.
+    if languages == [""]:
```
shreyanid, line 171: So if partition_msg uses partition_html / _text, which `languages` result gets unused, and which one matches this "empty" case? Not sure I am follow how this check applies to the case where languages gets detected twice

Coniferish: The languages detected by `partition_html/text` were unused as all of the metadata gets overwritten just after those partition calls (lines 105-111 in msg.py). You can see in there that I pass `languages=[""]` to those calls so we can just skip the language detection.

unstructured/file_utils/metadata.py
```diff
@@ -150,3 +153,43 @@ def _get_exif_datetime(exif_dict: Dict[str, Any], key: str) -> Optional[datetime
...
+        yield from elements
+        return
+
+    if not isinstance(elements, List):
+        elements = list(elements)
+
+    full_text = " ".join(e.text for e in elements if hasattr(e, "text"))
```
shreyanid, line 178: So the `full_text` is being concatenated in post-processing for every doctype - is this better (or at least not too inefficient comparatively) than using the full text that exists when initially retrieved from the document (ex. partition/text.py)? I understand this needs to be done for certain filetypes that never have the full text available (like docx), but am wondering about the cases where we do have full text available in the partitioner.

Coniferish: A couple of reasons seemed to justify this:
- it seemed reasonable to have all the partitioners use the same approach for detecting languages rather than make the code more complicated by having some use one approach and others use another approach--I did try to keep the pattern you initially used at first, but it just ended up being a lot easier to make them uniform. I believe @scanny also told me at some point that the additional iteration over elements to get the document text wasn't going to add a significant computational load (informally running tests didn't seem to take longer at the least) and it was better to prioritize composability over efficiency in this case.
- Detecting the languages after the elements had been created also made it a lot easier to detect language at the element level if multiple languages were detected at the document level or the `detect_language_per_element` param is set to `True` (and again not have to have very different approaches for the different partitioners)

unstructured/partition/lang.py
Coniferish: @shreyanid This is the part I was referring to in the Ingest Test Fixtures Update PR.

shreyanid: Where is the Ingest Test Fixtures Update PR? Yes, I saw the addition of regex to check for ascii characters, did that solve the short korean text problem?

unstructured/file_utils/metadata.py
shreyanid: That makes sense, since tests didn't show too much extra load, the consistency across partitioners is worth it. Thanks!

unstructured/partition/lang.py
Coniferish: It's here:
https://github.com/Unstructured-IO/unstructured/pull/1652#event-10567029937

Yes, the regex solved the korean and Stanley Cups xlsx problem, but it looks like we have some other existing documents with short text that's not having the language detected accurately

unstructured/partition/csv.py
```diff
@@ -29,6 +30,7 @@ def partition_csv(
     metadata_filename: Optional[str] = None,
     metadata_last_modified: Optional[str] = None,
     include_metadata: bool = True,
+    languages: List[str] = ["auto"],
```
shreyanid, line 33: Does the csv partitioner not have the `detect_language_per_element` param because it only results in one table element? Or is there another reason `detect_language_per_element` isn't part of every partitioner?

Coniferish: Yah, I left it off because the partitioner returns single TableElements. In working on this, though, I did realize there are separate elements for multiple pages in an xlsx doc, though. So I could create a file with different languages on different sheets, but that seemed like an unnecessary/unlikely test/use case

unstructured/partition/lang.py
shreyanid: Ok, feel free to show some of those docs/results to me. Should we potentially default to English if these inaccurate short text documents can be identified?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Enhancements
 
+* **Add document level language detection functionality.** Adds the "auto" default for the languages param to all partitioners. The primary language present in the document is detected using the `langdetect` package. Additional param `detect_language_per_element` is also added for partitioners that return multiple elements. Defaluts to `False`.
```
shreyanid, line 5:
```suggestion
* **Add document level language detection functionality.** Adds the "auto" default for the languages param to all partitioners. The primary language present in the document is detected using the `langdetect` package. Additional param `detect_language_per_element` is also added for partitioners that return multiple elements. Defaults to `False`.
```
nit

test_unstructured/partition/test_text.py
shreyanid: Perhaps we should leave them in with an xfail (or similar result) so we remember that this is functionality that is not yet supported, and can revisit other solutions in the future.

unstructured/partition/lang.py
Coniferish: That's what I was trying to do with the regex pattern. It makes sure there are only English characters in the text and text is less than 20 characters so that if there's short text in English, `langdetect` won't incorrectly detect some other language. But it's not able to catch everything.

unstructured/partition/csv.py
shreyanid: Ok, that makes sense - yeah the excel diff languages across sheets is fine to skip for now :) Could be worth adding a comment about why the `detect_language_per_element` param pattern didn't apply here, or apply it anyway for consistency with a comment that it doesn't get used since only a single element is returned.

unstructured/partition/lang.py
Coniferish: It might be worth pushing this off to a follow-up issue where we also look more closely at other language detection packages and come up with our own tests to compare them all.

shreyanid:
Could you add more tests (doesn't have to be in every doctype) with the `languages` param passed in? Maybe in each file type representative of a diff pattern (docx/pptx, partitioners that use other partitioners, csv/xlsx, etc) will suffice.

unstructured/partition/csv.py
Coniferish: Sounds good. I have one in the tests, but it would be better to have a comment in the actual partitioner

unstructured/file_utils/metadata.py
```diff
@@ -150,3 +153,43 @@ def _get_exif_datetime(exif_dict: Dict[str, Any], key: str) -> Optional[datetime
...
+    detected_languages = detect_languages(text=full_text, languages=languages)
+    if (
+        detected_languages is not None
+        and len(languages) == 1
+        and detect_language_per_element is False
```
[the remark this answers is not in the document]
Coniferish: Yah, if either the length of `languages` is not 1 or `detect_language_per_element` is `True`, it will detect the language for each element

unstructured/partition/lang.py
shreyanid: That sounds good!

Coniferish: https://github.com/Unstructured-IO/unstructured/issues/1660

unstructured/partition/pdf.py
```diff
@@ -133,9 +133,6 @@ def partition_pdf(
     """
     exactly_one(filename=filename, file=file)
 
-    if not isinstance(languages, list):
-        raise TypeError("The language parameter must be a list of language codes as strings.")
```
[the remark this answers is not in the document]
scanny: Could be because type-checker complained, but the type-checker won't help when an end-user sends us a string or whatever. @Coniferish you'll need a `# type: ignore` on the `isinstance()` line if that's what it was.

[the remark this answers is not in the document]
Coniferish: Because I moved it to `detect_languages`. That way it only has to be in one place instead of each of the partitioners. I added a test in `test_text` to check that it was being raised: `test_partition_text_element_metadata_raises_TypeError`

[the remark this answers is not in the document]
Coniferish: Whoops. Realizing we still need it there since pdf doesn't use `detect_languages`. It was probably removed when I was removing it from other partitioners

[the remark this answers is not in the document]
Coniferish: Actually, it's not needed there since both `pdf` and `image` rely on `partition_pdf_or_image` and it's handled there. I'll change the message, though, to add an example

unstructured/partition/lang.py
```diff
@@ -221,7 +223,20 @@ def detect_languages(
     Detects the list of languages present in the text (in the default "auto" mode),
     or formats and passes through the user inputted document languages if provided.
     """
-    if text.strip() == "":
+    if not isinstance(languages, list):
```
shreyanid, line 226: ah, since the language type check is now here, can we make pdf and image partitioners consistent (either include the typing error or remove it in both places)

unstructured/partition/pdf.py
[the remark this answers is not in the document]
Coniferish: Made a commit to remove redundant error catches and updated this message

unstructured/partition/lang.py
Coniferish: Actually, pdf and image don't rely on `detect_languages`, so a check is needed, but it's only needed in `partition_pdf_or_image`. See initial conversation and the commit I just added

shreyanid:
just realized the auto partitioner was not yet updated - we need to pass the languages parameter through to each partitioner

shreyanid:
also please bump dev version in `CHANGELOG` and `unstructured/__version__.py`

unstructured/partition/auto.py
```diff
@@ -133,6 +133,7 @@ def partition(
     ssl_verify: bool = True,
     ocr_languages: Optional[str] = None,  # changing to optional for deprecation
     languages: List[str] = ["eng"],
```
shreyanid, line 135: change default to "auto"

Coniferish: Done. Think I copied that from image.py or pdf.py where the default is "eng". Seems to be present in another file or two. Unsure if it's intentional in those places or not

shreyanid: Ah you're right, for image and pdf we want a default of english because there's no language detection that occurs (since the text is not readily available/strategy is not fast), only passing through to OCR. Maybe we shouldn't include a default here in auto, so that the default gets set by individual partitioners as auto or eng.

Coniferish: I changed the default to `None` and made `languages` optional so it'd be consistent with the other kwargs. 

I have also heard from Matt a couple of times before that we don't want to change the order of the kwargs/want to introduce new kwargs at the end of the parameters in case users are using positional arguments and making `languages` just not have a default value would have made it necessary to put it before all the args with default values.

shreyanid (approved):
changes lgtm! update final changelog version / fix all conflicts and tests before merging

PR #1628, opened 2023-10-03
feat: add file-based access permissions for SharePoint ingest

ahmetmeleq:
This PR:

- defines rbac_data as a SourceMetadata field,
- manages connections to an external api for obtaining rbac data with ConnectorRBAC class,
- serializes rbac data and saves it to the disk,
- matches the rbac_data in the disk to each IngestDoc, using a common field,
- forwards rbac data to Elements, via the partition() function

To test the changes, run `examples/ingest/sharepoint/ingest.sh` with the relevant rbac & connector credentials


ahmetmeleq:
Potential improvements before merging:
- improve rbac data to ingestdoc matching logic (currently checking filenames in a naive way)
- better folder management for writing down the rbac data (currently writes all rbac data directly into the output_dir)
- obtaining rbac_data field as a python object and serializing later, (currently obtains it as a str)
- handling errors caused by upstream connection issues
  - ie requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url: ...

ahmetmeleq:
Note: Reading permissions data as an object rather than a str causes proper formatting in the outputted elements.json file (rather than keeping the permissions data in one line), and results in permission data dominating / polluting the file. Repeating it for each element adds to the problem. Open to suggestions about this

ahmetmeleq:
Note: Sharepoint seems to be having trouble with our parallel requests, giving:
```
Error: {"error":"invalid_request",
"error_description":"AADSTS900023: Specified tenant identifier 'none' is neither a valid DNS name, nor a valid external domain.
...}
```
or
```
office365.runtime.client_request_exception.ClientRequestException: (None, None, "503 Server Error: Service Unavailable for url: 
```

To address this, I've reduced [num-processes in the example .sh to 1](https://github.com/Unstructured-IO/unstructured/pull/1628/commits/035cc3240c7febd7d7ecf02bc321c33d68135575), and it seems like this helps

ahmetmeleq:
Example file for testing:
```
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
cd "$SCRIPT_DIR"/../../.. || exit 1

    sharepoint \
    --client-id $SHAREPOINT_CLIENT_ID \
    --client-cred $SHAREPOINT_CRED \
    --permissions-application-id $SHAREPOINT_RBAC_CLIENT_APPLICATION_ID \
    --permissions-client-cred $SHAREPOINT_RBAC_CLIENT_SECRET \
    --site $SHAREPOINT_SITE \
    --permissions-tenant $SHAREPOINT_RBAC_TENANT \
    --output-dir sharepoint-ingest-output \
    --num-processes 1 \
    --path "Shared Documents" \
    --verbose
```

unstructured/ingest/processor.py
```diff
@@ -76,16 +76,17 @@ def run_partition(self, docs):
...
+        for i in range(len(json_docs)):
+            self.doc_processor_fn(json_docs[i])
+        logger.info(f"Processing {len(docs)} docs")
+        # json_docs = [doc.to_json() for doc in docs]
+        # with mp.Pool(
+        #     processes=self.num_processes,
+        #     initializer=ingest_log_streaming_init,
+        #     initargs=(logging.DEBUG if self.verbose else logging.INFO,),
+        # ) as pool:
+        #     pool.map(self.doc_processor_fn, json_docs)
```
ryannikolaidis, line 89: debugging?

unstructured/ingest/interfaces.py
```diff
@@ -201,6 +202,13 @@ def version(self) -> t.Optional[str]:
             self.update_source_metadata()
         return self.source_metadata.version  # type: ignore
 
+    @property
+    def permissions_data(self) -> t.Optional[t.List[t.Dict[str, t.Any]]]:
+        """Role based access control data, aka permissions or sharing, from the source system."""
```
ryannikolaidis, line 207:
```suggestion
        """Access control data, aka permissions or sharing, from the source system."""
```
probably no reason to assume this permission data is role based

unstructured/ingest/cli/cmds/sharepoint.py
```diff
@@ -25,6 +25,9 @@
 class SharepointCliConfig(BaseConfig, CliMixin):
     client_id: t.Optional[str] = None
     client_cred: t.Optional[str] = None
+    permissions_application_id: t.Optional[str] = None
+    permissions_client_cred: t.Optional[str] = None
+    permissions_tenant: t.Optional[str] = None
```
ryannikolaidis, line 30: none of these overlap with site/client_id/client_cred? or thinking it might make sense to allow users to have a distinct app with access for these things?

unstructured/ingest/cli/cmds/sharepoint.py
```diff
@@ -55,6 +58,24 @@ def add_cli_options(cmd: click.Command) -> None:
...
+                ["--permissions-client-cred"],
+                default=None,
+                type=str,
+                help="Credentials for ingesting permission (rbac) data",
+            ),
+            click.Option(
+                ["--permissions-tenant"],
+                default=None,
+                type=str,
+                help="Sharepoint permission (rbac) tenant name, such as: abcde.onmicrosoft.com",
```
ryannikolaidis, line 77: double checking: so there is an independent tenant specifically for permissions?

unstructured/ingest/cli/cmds/sharepoint.py
```diff
@@ -55,6 +58,24 @@ def add_cli_options(cmd: click.Command) -> None:
...
+            ),
```
ryannikolaidis, line 78: if one of these is specified is it required that the other two must also be specified? if so we should probably validate and throw the appropriate error

unstructured/ingest/connector/sharepoint.py
```diff
@@ -168,7 +184,44 @@ def _fetch_page(self):
...
+                if ingestdoc_docname == permissions_docname and parent_name_matches(
+                    parent_type=parent_type,
+                    permissions_filename=filename,
+                    ingest_doc_filepath=self.file_path,
+                ):
+                    with open(read_dir / filename) as f:
+                        self._permissions_data = json.loads(f.read())
+
     def update_source_metadata(self, **kwargs):
+        self.update_permissions_data()
```
ryannikolaidis, line 224: should updating and setting permissions data be conditional on whether or not the user specified / opted in to it in the command line args?

unstructured/ingest/connector/sharepoint.py
```diff
@@ -35,6 +37,9 @@
 class SimpleSharepointConfig(BaseConnectorConfig):
     client_id: str
     client_credential: str = field(repr=False)
+    permissions_application_id: str
```
rbiseck3, line 40: Should these all be optional since it's an additional step that a user can choose to omit?

unstructured/ingest/connector/sharepoint.py
```diff
@@ -356,3 +417,158 @@ def get_ingest_docs(self):
             site_client = self.connector_config.get_site_client(site_url)
             ingest_docs = ingest_docs + self._ingest_site_docs(site_client)
         return ingest_docs
+
+
+class PermissionsConnector:
```
rbiseck3, line 422: Do get all the benefits of it and keep the pattern, this can be a `@dataclass` with the access token set in a `__post_init__`. This would also guarantee sensitive info isn't serialized.

unstructured/ingest/processor.py
ahmetmeleq: Shouldn't have git added, gonna remove 👍

ahmetmeleq: https://github.com/Unstructured-IO/unstructured/pull/1628/commits/803dcec76115f03761138016845cdc019efcc4c7

unstructured/ingest/cli/cmds/sharepoint.py
ahmetmeleq: None of these overlap unfortunately. Each of these are created separately, and replacing any of them with its document ingestion correspondent causes failure

Only possible reduction of args here is parsing the `--site` argument to obtain `--permissions-tenant`, however I don't recommend it. If we want to do that, we'll have to find examples for all possible `--site` URL formats, and write a parsing logic that encapsulates all cases. Only covering the URL format we have can fail often for other users; and finding other possible URL formats will probably require creating different sites from scratch and trying to ingest from each of them

ahmetmeleq: Only possible reduction of args here is parsing the `--site` argument to obtain `--permissions-tenant`, however I don't recommend it. If we want to do that, we'll have to find examples for all possible `--site` URL formats, and write a parsing logic that covers all cases. Only covering the URL format we have can fail often for other users; and finding other possible URL formats will probably require creating different sites from scratch and trying to ingest from each of them. 

Example: 

Our `--site` format is: https://xxxxx-admin.sharepoint.com/
Our `--permissions-tenant` format is: xxxxx.onmicrosoft.com

Others'  `--site` formats might be: 
- https://xxxxx.sharepoint.com/
- https://xxxxx-<something-else\>.sharepoint.com/
- (any other format that we are unaware)

Others' `--permissions-tenant` formats might be: 
- (any other format that we are unaware)

It seems to me that giving user the ability to mention their tenant URL would be much safer here than assuming a format on their site & tenant URLs, and parsing one to another

unstructured/ingest/cli/cmds/sharepoint.py
ahmetmeleq: Agreed, addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/e2fef65bf38ca763a72a658d325ec90e357b3bcb

unstructured/ingest/connector/sharepoint.py
ahmetmeleq: Update: A `permissions_config` argument of type `SharepointPermissionsConfig` is made optional to `SimpleSharepointConfig` now; and these fields are kept mandatory within the `SharepointPermissionsConfig` class

Optional assignment of a `SharepointPermissionsConfig` to `SimpleSharepointConfig` is made in the runner file 

Check: https://github.com/Unstructured-IO/unstructured/pull/1628/commits/c491ff06e02938023c3896083630d5358635f6a6

unstructured/ingest/connector/sharepoint.py
ahmetmeleq: Agreed and addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/145452256c5f42c2907d6fad6486152bdf4669e9

Somehow, `__post_init__` did not run by itself after `__init__` even though `@dataclass` decorator was there; so I implemented it in a different way

Two questions:

How do we guarantee that sensitive info doesn't get serialized with this approach? Is it because `access_token` is not a property anymore, since it is not defined in init? 

Do we have similar concerns for the info in say `SimpleSharepointConfig`?

unstructured/ingest/runner/sharepoint.py
```diff
@@ -14,6 +14,9 @@ def run(
         site: str,
         client_id: str,
         client_cred: str,
+        permissions_application_id: str,
```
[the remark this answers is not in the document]
ahmetmeleq: Agreed and addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/62f3262fb76bab63e76570151a944e7c3a483e35

unstructured/ingest/connector/sharepoint.py
```diff
@@ -35,6 +37,9 @@
 class SimpleSharepointConfig(BaseConnectorConfig):
     client_id: str
     client_credential: str = field(repr=False)
+    permissions_application_id: str
+    permissions_client_cred: str = field(repr=False)
```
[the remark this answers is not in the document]
ahmetmeleq: Agreed and addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/e2fef65bf38ca763a72a658d325ec90e357b3bcb

ryannikolaidis:
example for sharepoint/ingest.sh is empty in the file diff? meant to put something in there?

ryannikolaidis:
> also worth updating docs to include steps necessary to set up an app that has the right permissions. i.e. how you get values for: permissions_application_id, permissions_client_cred, and permissions_tenant

bump ^. maybe include in the example script since that's usually where we include install instructions?

unstructured/ingest/connector/sharepoint.py
```diff
@@ -31,6 +34,26 @@
...
+    application_id: str = None
+    client_credential: str = None
+    tenant: str = None
+
+    def __post_init__(self):
+        self.provided = False
+        if any([self.application_id or self.client_credential or self.tenant]):
+            if not all([self.application_id and self.client_credential and self.tenant]):
+                raise ValueError(
+                    "Please provide either none or all of the following optional values:\n"
```
rbiseck3, line 48: Can this check be part of converting the click options since this part of the code has no knowledge of the cli parameters? It can be factored into the `from_dict` method when this config is created such as this: [CliEmbeddingConfig from_dict()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/cli/interfaces.py#L269)

unstructured/ingest/cli/cmds/sharepoint.py
```diff
@@ -55,6 +58,24 @@ def add_cli_options(cmd: click.Command) -> None:
                     https://[tenant]-admin.sharepoint.com.\
                     This requires the app to be registered at a tenant level",
             ),
+            click.Option(
```
rbiseck3, line 61: Can we actually add this via a `PermissionConfig` data class similar to how we add read, and partition configs to the list of options available on a click command? Then we can also pull the data out using `from_dict`

ahmetmeleq:
> > also worth updating docs to include steps necessary to set up an app that has the right permissions. i.e. how you get values for: permissions_application_id, permissions_client_cred, and permissions_tenant
> 
> bump ^. maybe include in the example script since that's usually where we include install instructions?

Agreed, addressed with: https://github.com/Unstructured-IO/unstructured/pull/1628/commits/b4ba7500ddb6b516cbcfe0416f8754b4552e4a5f

ahmetmeleq:
> example for sharepoint/ingest.sh is empty in the file diff? meant to put something in there?

Addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/bb03eaf8763899686ba09b992d7d8bf9a44a442f

unstructured/ingest/cli/cmds/sharepoint.py
ahmetmeleq: Please let me know if there are additional concerns on this one

unstructured/ingest/connector/sharepoint.py
ahmetmeleq: Addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/f3b70195d44175c6bc49f56134f29c6b75be1e19

ahmetmeleq:
> should add permission options to the ingest test for sharepoint ingest test so we're validating this.

Addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/bfe29258c722a749c095fcae3acfe34a06813eff

examples/ingest/sharepoint/ingest.sh
```diff
@@ -12,6 +12,8 @@
 # To get the credentials for your Sharepoint app, follow these steps:
 # https://github.com/vgrem/Office365-REST-Python-Client/wiki/How-to-connect-to-SharePoint-Online-and-and-SharePoint-2013-2016-2019-on-premises--with-app-principal
 
+# To setup your application and obtain permissions related variables (--permissions-application-id, --permissions-client-cred, --permissions-tenant), follow these steps:
+# https://tsmatz.wordpress.com/2016/10/07/application-permission-with-v2-endpoint-and-microsoft-graph
```
ryannikolaidis, line 16: since this is from a random wordpress site, can you archive these instructions somewhere internally on our end? want to make sure we don't lose this if that site disappears.

unstructured/ingest/connector/sharepoint.py
```diff
@@ -328,3 +387,160 @@ def get_ingest_docs(self):
...
+                drives_for_site = drives["value"]
+                drive_ids.extend([(site_id, drive["id"]) for drive in drives_for_site])
+
+        print("Obtaining item data from drives for permissions (rbac)")
+        item_ids = []
+        for site, drive_id in drive_ids:
+            drive_items = self.get_drive_items(site, drive_id)
+            if drive_items:
+                item_ids.extend(
+                    # [(site, drive_id, item["id"], item["name"]) for item in drive_items["value"]],
```
ryannikolaidis, line 521: debug? can remove?

examples/ingest/sharepoint/ingest.sh
ahmetmeleq: I've added the contents to our google drive and shared with you now 👍

unstructured/ingest/connector/sharepoint.py
ahmetmeleq: That's right, addressed with https://github.com/Unstructured-IO/unstructured/pull/1628/commits/27be65f0d252341d894c9b8d460a29e72af833d2 now

ahmetmeleq:
Test fixtures update PR on the way as well. After that, and passing the tests, should be good to go

ryannikolaidis:
> Test fixtures update PR on the way as well

are these still on route?

ahmetmeleq:
> > Test fixtures update PR on the way as well
> 
> are these still on route?

We're there now 👍

test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/ideas-page.html.json
```diff
@@ -11,7 +11,95 @@
...
+            ],
+            "shareId": "Q29tbXVuaWNhdGlvbiBzaXRlIFZpc2l0b3Jz",
+            "grantedToV2": {
+              "siteGroup": {
+                "displayName": "Communication site Visitors",
+                "id": "4",
+                "loginName": "Communication site Visitors"
+              }
+            },
+            "grantedTo": {
```
ryannikolaidis, line 49: okay for now. at some point when we have >1 instance of this, we'll want to define our own schema for permissions.

docs/source/source_connectors/sharepoint.rst
```diff
@@ -22,6 +22,9 @@ Run Locally
           --client-id "<Microsoft Sharepoint app client-id>" \
           --client-cred "<Microsoft Sharepoint app client-secret>" \
           --site "<e.g https://contoso.sharepoint.com or https://contoso.admin.sharepoint.com to process all sites within tenant>" \
+          --permissions-application-id "<Microsoft Graph API application id, to process permissions (rbac) data>" \
+          --permissions-client-cred "<Microsoft Graph API application credentials, to process permissions (rbac) data>" \
+          --permissions-tenant "<e.g https://contoso.onmicrosoft.com (tenant URL) to process permission (rbac) data>" \
```
ryannikolaidis, line 27:
```suggestion
          --permissions-application-id "<Microsoft Graph API application id, to process per-file access permissions>" \
          --permissions-client-cred "<Microsoft Graph API application credentials, to process per-file access permissions>" \
          --permissions-tenant "<e.g https://contoso.onmicrosoft.com (tenant URL) to process per-file access permissions>" \
```

docs/source/source_connectors/sharepoint.rst
```diff
@@ -68,6 +75,9 @@ You can also use upstream connectors with the ``unstructured`` API. For this you
...
```
ryannikolaidis, line 80:
```suggestion
          --permissions-application-id "<Microsoft Graph API application id, to process per-file access permissions>" \
          --permissions-client-cred "<Microsoft Graph API application credentials, to process per-file access permissions>" \
          --permissions-tenant "<e.g https://contoso.onmicrosoft.com (tenant URL) to process per-file access permissions>" \
```

examples/ingest/sharepoint/ingest.sh
```diff
@@ -12,6 +12,8 @@
 # To get the credentials for your Sharepoint app, follow these steps:
 # https://github.com/vgrem/Office365-REST-Python-Client/wiki/How-to-connect-to-SharePoint-Online-and-and-SharePoint-2013-2016-2019-on-premises--with-app-principal
 
+# To setup your application and obtain permissions related variables (--permissions-application-id, --permissions-client-cred, --permissions-tenant), follow these steps:
```
ryannikolaidis, line 15:
```suggestion
# To optionally setup your application and obtain permissions related variables (--permissions-application-id, --permissions-client-cred, --permissions-tenant), follow these steps:
```

unstructured/ingest/cli/cmds/sharepoint.py
ryannikolaidis: @ahmetmeleq now that you have CliPermissionsConfig can we remove these options explicitly defined here? and instead add via add_options[ here](https://github.com/Unstructured-IO/unstructured/blob/569561e59bdcead6953f87d1e18dcbe32ada78e7/unstructured/ingest/cli/cmds/sharepoint.py#L95)?

ryannikolaidis (commented):
couple nits on docs and a comment / followup on CliPermissionsConfig

test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/ideas-page.html.json
ahmetmeleq: If users would like to interact with the source of the permission data again based on the ingested state of the permission data (ie, to update the permissions), they might need the original schema again. Considering that, if we want our own schema we can write parsers two ways: original->unstructured & unstructured->original

unstructured/ingest/cli/cmds/sharepoint.py
ahmetmeleq: Oh right, I thought this is for the future for each pipeline node to have their own config. Now I realize how it immediately reduces the number of explicitly defined options in the connector cli config. I added https://github.com/Unstructured-IO/unstructured/pull/1628/commits/ae982a4e4846864ee94ae1ac15c29614fbcc8abb and https://github.com/Unstructured-IO/unstructured/pull/1628/commits/b898ed5d3fa3b90f74872906bb6bf7b9f5f84209 👍

ryannikolaidis: yep, exactly!

test_unstructured_ingest/expected-structured-output/Sharepoint/Shared Documents/ideas-page.html.json
ryannikolaidis: yep, for sure.

ryannikolaidis (approved):
lgtm! nice work, I know this was a huge lift.

PR #1631, opened 2023-10-03
docx: improve page break fidelity

scanny:
Page breaks can and often do occur within a paragraph. The full text of the paragraph is attributed to the page (number) the paragraph starts on.

Improve page-break fidelity such that a paragraph containing a page-break is split into two elements, one containing the text before the page-break and the other the text after. Emit the `PageBreak` element between these two and assign the correct page-number (n and n+1 respectively) to the two textual elements.

This functionality is largely provided upstream by the new `python-docx` v1.0.0 release (1.0.0 from 0.8.11 because it drops Python 2 support). That version also makes obsolete the "include hyperlink text in `Paragraph.text` monkey patch that we had maintained up to now. Remove that monkey-patch.


cragwolfe:
Two test .docx's were added to:

    test_unstructured_ingest/expected-structured-output/box/

But, we already have an example-docs/ folder that essentially serves the purposes of storing docs for unittests (for better or worse, that's the existing convention). Is there a specific reason you do not want to move these to example-docs, @scanny ? Is there a difference between https://github.com/Unstructured-IO/unstructured/blob/main/example-docs/handbook-1p.docx and `test-handbook-1p.dox` is this PR?

scanny:
@cragwolfe The ingest test setup is such that files in `example-docs/` that are used in ingest tests cannot be changed once created. I needed to edit `handbook-1p.docx` to pass a couple of the existing page-break-related "unit" tests but then the ingest tests were picking up the old version and failing on that. For whatever reason, at least some of the ingest tests do not pick up changes to files in `example-docs/` in a PR.

Alan mentioned it might have something to do with the Box connector, but in any case I concluded that files in `example-docs/` were only good for tests when they were and if that changed you need to create a new one.

I put these in a separate test_files directory because I didn't think in general the files that make good input for tests make good example files for new users and being good for both purposes didn't strike me as a constraint we wanted to put on test files. A test file might be very short, have an unrealistic density of page breaks, etc. because that makes it efficient input for a suite of tests. The new `page-breaks.docx` is a good example of that.

I could see us having a broader scope location for test files, like maybe `partition/test-files/` if we wanted, but I didn't see a ready way to negotiate a new convention so just put them near the tests and figured we'd get to consolidating them if we wanted further down the line.

I don't feel particularly strongly about it, but like I say, I don't see test files in general making good example files, so would advocate for a distinct location. On the question of whether we use the same files for ingest tests as we do for other tests, I'd be inclined to avoid it just on the grounds of test independence, but that ship has probably already sailed.

In any case, I'm happy to move them if you want, just let me know.

scanny:
@cragwolfe btw, yes, the new handbook-1p.docx has a _rendered_ page break added, as Word would do, and only that page-break is recognized by the partitioner. That's the only difference but enough to break the ingest tests without a way to fix them other than to create a new file for the unit tests and change the ingest-test expectations to no longer expect a page break for that file.

qued:
As far as I know, none of the files in `example-docs` are used in ingest tests. Despite the name, the main purpose of the `example-docs` folder is to house files used in unit tests.

scanny:
@qued Hmm, that's interesting now that you mention it. Are you saying that, yes, there is a `handbook-1p.docx` used in ingest tests, but it's not the same actual file as the one in `example-docs`? I was seeing an ingest-test difference when I changed that file, but I bet I had changed the code too at that point and maybe all I have to do is just accept the ingest-output difference.

I'll take a look at that tomorrow. I think that was my misconception :)

cragwolfe:
> As far as I know, none of the files in example-docs are used in ingest tests. Despite the name, the main purpose of the example-docs folder is to house files used in unit tests.

yep, though also in in some ingest "local" filesystem tests, e.g.:
https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-local-single-file-with-encoding.sh#L28

qued:
Ok it seems like what I said is not quite right (see correction from @cragwolfe ) but that particular file `handbook-1p.docx` is pulled down from somewhere else when it's used in the ingest test.

cragwolfe (approved):

PR #1634, opened 2023-10-03
chore: process chipper hierarchy

qued:
PR to support schema changes introduced from [PR 232](https://github.com/Unstructured-IO/unstructured-inference/pull/232) in `unstructured-inference`.

Specifically what needs to be supported is:
* Change to the way `LayoutElement` from `unstructured-inference` is structured, specifically that this class is no longer a subclass of `Rectangle`, and instead `LayoutElement` has a `bbox` property that captures the location information and a `from_coords` method that allows construction of a `LayoutElement` directly from coordinates.
* Removal of `LocationlessLayoutElement` since chipper now exports bounding boxes, and if we need to support elements without bounding boxes, we can make the `bbox` property mentioned above optional. 
* Getting hierarchy data directly from the inference elements rather than in post-processing
* Don't try to reorder elements received from chipper v2, as they should already be ordered.

#### Testing:

The following demonstrates that the new version of chipper is inferring hierarchy.

```python
elements = partition_pdf("example-docs/layout-parser-paper-fast.pdf", strategy="hi_res", model_name="chipper")
children = [el for el in elements if el.metadata.parent_id is not None]
print(children)

```
Also verify that running the traditional `hi_res` gives different results:
```python
elements = partition_pdf("example-docs/layout-parser-paper-fast.pdf", strategy="hi_res")

```


christinestraub (approved):
LGTM!

qued:
Ingest test diffs are caused by [this change](https://github.com/Unstructured-IO/unstructured-inference/pull/245), and I think this is the sole source of diffs.

christinestraub:
> Ingest test diffs are caused by [this change](https://github.com/Unstructured-IO/unstructured-inference/pull/245), and I think this is the sole source of diffs.

Yes, that's right.

ajjimeno (approved):
LGTM!

Issue #1286, opened 2023-09-01
feat: support for CompositeElement inverse operations


cragwolfe:
**Is your feature request related to a problem? Please describe.**

Composite Elements were introduced in https://github.com/Unstructured-IO/unstructured/pull/1222 , the idea being that elements should be combined in an intelligent way before being stored downstream.

While this is important, it is also important to be able to recover the original elements that comprise the CompositeElement.

**Describe the solution you'd like**

Add a metadata field to `CompositeElement` named `orig_elements` that is the base64 encoded string of gzip'ed json representation of the original list of elements.

Of course, tests should be added that show, in pseudo-code:

    elements_to_json(elements) == decompress(
                                    composite_element.metadata.orig_elems)

Do not rely on element object equality but instead json serialization for the reason described in https://github.com/Unstructured-IO/unstructured/issues/1285 .

**Describe alternatives you've considered**

One could just nest the original elements directly, but that gets unwieldy pretty fast.

maxburke:
This would be incredibly useful for my usecase

punjabdhaputar:
+1 this would be fantastic for my usecase as well!

scanny:
This feature was added in v0.13.0 of the library.

PR #1637, opened 2023-10-04
fix: chunking fails with detection_class_prob in metadata

ryannikolaidis:
Fixes chunking when `detection_class_prob` appears in Element metadata.

Currently, when `detection_class_prob` appears in Element metadata, Elements will only be combined by `chunk_by_title` if they have the same `detection_class_prob` value (which is rare). This is unlikely a case we ever need to support and most often results in no chunking. 

`detection_class_prob` is now included in the chunking list of metadata keys excluded for similarity comparison. 

## Testing

Unit test added that currently breaks due to changes in to `detection_class_prob` but with this change passes since `detection_class_prob` is excluded from chunking similarity comparison.


qued:
So the current logic is to start a new section when any metadata field changes? This makes sense for page numbers, and maybe 1 or 2 fields that I can't think of right now, but is this something we should whitelist rather than blacklist? It seems more likely we would _not_ want this behavior for most new metadata fields.

qued (approved):
LGTM for a quick fix, but per my comment I think at some point we should switch to whitelisting metadata fields we want to use to trigger new chunks rather than blacklisting ones we don't, to prevent this happening in the future when we add new fields.

cragwolfe:
Related: https://github.com/Unstructured-IO/unstructured/issues/1286 .
IMO, inverse operations of composite elements to get the original will be important. one could "re-index" without saving things twice, for instance, or get high granularity detail about any sub-element in a downstream app.

newelh (approved):

ryannikolaidis:
yes, totally agree on both of these points (whitelist and inverse operations of composite elements). just a quick fix here so that chunking still works out of the box when inference is used (triggering detection_class_prob)

ryannikolaidis:
hopefully test coverage still keeps us honest with those changes.

PR #1650, opened 2023-10-05
Feat: Bag of words for testing metric

mallorih:
This PR adds the `bag_of_words` function to count the frequency of words for evaluation.

**Testing**
```Python
from unstructured.cleaners.core import bag_of_words
string = "The dog loved the cat, but the cat loved the cow."

print(bag_of_words)

### CODE REVIEW

**Comment by shreyanid** [comment]
Could we add a test for text that contains incorrect spaces between words (like how the unstructured output looks sometimes)? for example in "i n t r o d u c t i o n", each letter will be considered its own word right?

File: test_unstructured/cleaners/test_core.py [hunk 1]
```diff
@@ -300,3 +300,20 @@ def test_clean(text, extra_whitespace, dashes, bullets, lowercase, trailing_punc
...
+@pytest.mark.parametrize(
+    ("text", "expected"),
+    [
+        (
+            "The dog loved the cat, but the cat loved the cow",
+            {"the": 4, "cat": 2, "loved": 2, "dog": 1, "but": 1, "cow": 1},
+        ),
+        (
+            "Hello my name is H a r p e r, what's your name?",
+            {"hello": 1, "my": 1, "name": 2, "is": 1, "whats": 1, "your": 1},
```
shreyanid, line 314: Can we keep punctuation within a word? (apostrophes and hyphens)

mallorih:
> Could we add a test for text that contains incorrect spaces between words (like how the unstructured output looks sometimes)? for example in "i n t r o d u c t i o n", each letter will be considered its own word right?

Are we sure we want to keep all punctuation?
```string = "I have a dog and a cat, i love my dog."
>>> bag_of_words(string)
{'have': 1, 'dog': 1, 'and': 1, 'cat,': 1, 'love': 1, 'my': 1, 'dog.': 1}
```

shreyanid:
No, we should not keep all punctuation, only punctuation within a word (hyphens and apostrophes). commas, dashes, between words or end of sentence punctuation should all be removed.

Klaijan:
Came across this pr and wondering what would we do with punctuation at the end of the word, but actually have the same semantic sense as ones within the word, e.g. we would keep `dog's` but remove `dogs'`?

shreyanid:
Yes, `dogs'` should (ideally) fall into the same case as `dog's` because those are both punctuation for only one word, not between words.

unstructured/metrics/text_extraction.py
```diff
@@ -50,3 +52,49 @@ def calculate_edit_distance(
     elif return_as == "distance":
         return distance
     return 0.0
+
+
+# Duplicate code from cleaners.core,
```
shreyanid, line 57: Could we call the functions in cleaners.core instead of duplicating?

unstructured/metrics/text_extraction.py
```diff
@@ -50,3 +52,49 @@ def calculate_edit_distance(
...
+# not sure we want this functionality introduced in the main library.
+def remove_punctuation(s: str, exclude_punctuation: Optional[list]) -> str:
+    """Removes punctuation from a given string."""
+
+    tbl = dict.fromkeys(
+        i for i in range(sys.maxunicode) if unicodedata.category(chr(i)).startswith("P")
+    )
+
+    if exclude_punctuation:
```
shreyanid, line 66: We should exclude punctuation not just by character but by functionality - for example, remove dashes but not hyphens. maybe use the P_ [unicode categories](https://pypi.org/project/unicategories/) to determine which type of punctuation it is

unstructured/metrics/text_extraction.py
mallorih: The function in cleaners.core doesn't allow you to exclude punctuation, and I didn't know if we wanted to introduce that functionality to that function.

test_unstructured/metrics/test_text_extraction.py
```diff
@@ -123,6 +123,21 @@ def test_calculate_edit_distance_with_filename(filename, expected_score, expecte
...
+            "My dog's hair is red, but the dogs' houses are blue.",
+            {
+                "my": 1,
+                "dog's": 1,
+                "hair": 1,
+                "is": 2,
+                "red": 1,
+                "but": 1,
+                "the": 1,
+                "dogs": 1,
```
shreyanid, line 136: Should be dogs'

mallorih: Oh I misunderstood the comment above. Reverting change.

unstructured/metrics/text_extraction.py
shreyanid: Ah ok, if it's doing something different then it's fine to just duplicate/adapt it here without a comment. if it's better to centralize these types of cleaning functions to one file, perhaps we should create a new function like `remove_sentence_punctuation` (or something) in cleaners.core and call it here, so it can be used for other purposes other than metrics as well.

test_unstructured/metrics/test_text_extraction.py
```diff
@@ -59,11 +98,47 @@ def test_calculate_edit_distance_with_filename(filename, expected_score, expecte
...
+            "My dog's hair is red, but the dogs' houses are blue.",
+            {
+                "my": 1,
+                "dog's": 1,
+                "hair": 1,
+                "is": 2,
+                "red": 1,
+                "but": 1,
+                "the": 1,
+                "dog's": 1,
```
shreyanid, line 136: This should be `dogs'` with the apostrophe at the end (freq 1), since `dog's` is already its own category (freq 1)

unstructured/metrics/text_extraction.py
mallorih: Not sure this does what we want. I think it just has a string of all the different types of hypens. It doesn't differentiate between where the hyphen is in a sentence.

shreyanid: Hm. If there isn't already a way to do it using unicode categories, we can use regex or any other pattern matcher.
We want to match a "dash" with no spaces surrounding it (ex. easy-peasy) as being a hyphen that we want to keep in our output words, and everything else is a dash that is between separate words. I added a test for dash / hyphen

mallorih: Gotcha! I just used our `clean_bullet` brick to remove dashes from the beginning of the sentence and leave the hyphens. Is there another use case that I'm forgetting?

shreyanid (approved):
looks great! thanks for taking care of all the punctuation cases :)

PR #1656, opened 2023-10-05
feat: add calculate edit distance feature

Klaijan:
**Executive Summary**

Adds function to calculate edit distance (Levenshtein distance) between two strings. The function can return as: 1. score (similarity = 1 - distance/source_len) 2. distance (raw levenshtein distance)

**Technical details**
- The `weights` param is set to default at (2,1,1) for (insertion, deletion, substitution), meaning that we will penalize the insertion we need to add from output (target) in comparison with the source (reference). In other word, the missing extraction will be penalized higher.
- The function takes in 2 strings in an assumption that both string are already clean and concatenated (CCT)

**Important Note!**
Test case needs to be updated to use CCT once the function is ready. It is now only tested the "functionality" of edit distance, not the edit distance with CCT as its intended to be.


test_unstructured/metrics/test_text_extraction.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+from unstructured.metrics.text_extraction import calculate_edit_distance
+from unstructured.partition.auto import partition
+
+
+@pytest.mark.parametrize(
+    ("filename", "similarity", "percentage"),
```
shreyanid, line 8: I see that similarity and percentage correspond to score and distance outputted by the function, but what do these terms mean, and why are they inverse of each other?

shreyanid: since "score" (similarity) is the default return type from the function, why are two of these so low? is percentage the accuracy measure we actually want to use for marketing?

test_unstructured/metrics/test_text_extraction.py
```diff
@@ -0,0 +1,30 @@
...
+)
+def test_calculate_edit_distance(filename, similarity, percentage):
+    with open("example-docs/fake-text.txt") as f:
+        source_cct = f.read()
+
+    elements = partition(filename=f"example-docs/{filename}")
+    output_cct = "\n\n".join([str(el) for el in elements])
+
+    score = calculate_edit_distance(output_cct, source_cct, return_as="score")
+    distance = calculate_edit_distance(output_cct, source_cct, return_as="percentage")
```
shreyanid, line 23: OH this variable says distance but returns as percentage. let's standardize that :)

unstructured/metrics/text_extraction.py
```diff
@@ -0,0 +1,55 @@
...
+        ValueError: If 'return_as' is not one of the valid return types
+        ["score", "percentage", "distance"].
+
+    Note:
+        This function calculates the edit distance (or similarity score) between
+        two strings using the Levenshtein distance algorithm. The 'weights' parameter
+        allows customizing the cost of insertion, deletion, and substitution
+        operations. The 'return_as' parameter determines the type of result to return:
+        - "score": Returns the similarity score, where 1.0 indicates a perfect match.
+        - "percentage": Returns the normalized edit distance as a percentage.
```
shreyanid, line 39: so in this "percentage" case 0% edit is better? Let's clarify that here

test_unstructured/metrics/test_text_extraction.py
Klaijan: Since we want to report the score of 100% being perfect, I did (1 - normalized distance) to get the score. And that's why they are the complement of each other's %. We can change the name accordingly, but I believe "score" is a good intuitive name of default result. I could change percentage -> distance and distance -> raw_distance if that makes more sense to you.

shreyanid:
> 1. score (similarity = 1 - percentage) 2. percentage (distance/source_len) 3. distance (raw levenshtein distance)

I'm debating if we need all 3 of these, since the names are getting a little confusing. Should we just stick with score (100% is perfect match, 0 is nothing in common) since that's the accuracy measure that's most easily understood and we'd use for marketing purposes? Not sure when we might want the other values - let me know what use cases you had in mind.

Klaijan:
@shreyanid I'd say we can keep the score and a raw distance (which would be rename "distance")

test_unstructured/metrics/test_text_extraction.py
shreyanid: Ok, score makes sense to me (could consider accuracy / correctness as names as well, but I like score too). 

> I could change percentage -> distance and distance -> raw_distance if that makes more sense to you.

Maybe we can do percentage -> edit_percentage (or percent_edit or some similar variation) and distance -> raw_distance ? I don't want them both to have the same name of "distance" since one is normalized and one is not.

shreyanid:
That sounds more reasonable, ty!

unstructured/metrics/text_extraction.py
shreyanid: update: being removed so nvm this comment

test_unstructured/metrics/test_text_extraction.py
shreyanid: update: removing "percentage" and keeping only score and distance/raw_distance (name choice is up to you)

test_unstructured/metrics/test_text_extraction.py
```diff
@@ -0,0 +1,29 @@
...
+        ("eml/fake-email.eml", 0.03, 164),
+        ("README.md", 0.0, 719),
+    ],
+)
+def test_calculate_edit_distance(filename, expected_score, expected_distance):
+    with open("example-docs/fake-text.txt") as f:
+        source_cct = f.read()
+
+    elements = partition(filename=f"example-docs/{filename}")
+    output_cct = "\n\n".join([str(el) for el in elements])
```
shreyanid, line 20: let's join with a single "\n" instead of two

test_unstructured/metrics/test_text_extraction.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from unstructured.metrics.text_extraction import calculate_edit_distance
+from unstructured.partition.auto import partition
+
+
+@pytest.mark.parametrize(
```
shreyanid, line 7: can we add more tests? a few im thinking of (can be existing files or new ones):
- test where unstructured output has spaced out words
- test where unstructured output has strung together words
- test where entire sentences are missing
- test with duplicated text

test_unstructured/metrics/test_text_extraction.py
```diff
@@ -0,0 +1,29 @@
...
+    ("filename", "expected_score", "expected_distance"),
+    [
+        ("fake-text.txt", 0.82, 30),
+        ("eml/fake-email.eml", 0.03, 164),
+        ("README.md", 0.0, 719),
+    ],
+)
+def test_calculate_edit_distance(filename, expected_score, expected_distance):
+    with open("example-docs/fake-text.txt") as f:
```
shreyanid, line 16: are all of the parameterized files being compared to the same fake-text.txt source cct? that explains why the scores are so low! let's compare each file to it's own source cct (can be made by hand if its just a few documents)

test_unstructured/metrics/test_text_extraction.py
Klaijan: Do we want to follow the demo on the unstructured's main page? The demo uses `\n\n` so I followed that. Also, I believe this function should be edited once the CCT function is ready. So once this is done, we'll want to create a ticket to edit this test case :)

test_unstructured/metrics/test_text_extraction.py
Klaijan: We could, but what actually do we want to test against? Do we want to compare the score for those cases aka test the function scoring system? If file open and partitioning is not needed, I'd say we might not want to test against any files but just comparing mock string to mock string (which is partially captured in the first test).

test_unstructured/metrics/test_text_extraction.py
Klaijan: I believe we do not have the CCT function ready yet so I made tests possible as I could. (We have ground truth/gold standard, but not the function to convert Unstructured elements to CCT if I understand correctly).

test_unstructured/metrics/test_text_extraction.py
shreyanid: The demo is just to print reasonably, but for this string comparison purpose a single newline or even single space will suffice. 
Yes, I agree we can edit this once we have a CCT function !

test_unstructured/metrics/test_text_extraction.py
shreyanid: We want to test against a sample source cct string. This test does not need to test file opening and partitioning - yes, let's just compare mock strings (the CCT output)

shreyanid: In your initial test there are a few arbitrary sample files, but the test cases I listed are specific occurrences in CCT strings we want to observe / test the scoring metric for. We can just make some sample strings to represent these common occurrences and the source string will be how we actually expect/want them to be parsed out of a document.

test_unstructured/metrics/test_text_extraction.py
shreyanid: Ok, as mentioned previously we can test this function on some smaller string cases as well as files where it's easy to create a human transcription of the text. We definitely don't want to be mismatching source and output files

test_unstructured/metrics/test_text_extraction.py
Klaijan: Fyi for the above comment - I tested with `\n` and the score is actually worse because the original .txt file for fake-text is double spacing. 

I do believe that the score now is not as important as we don't have the actual test files just yet. From what I understand, we will want to use this metric especially when we make improvements on the partitioning function, therefore the process of partitioning should also be in the test case in the final version. But for now, we can test on mock strings.

From your previous comment, I believe you want to test the scoring function on different scenarios:
1. Same string but spaced out words
2. Same string but not spaces
3. Missing sentences
4. Duplicated text
Which the expected score could be manually calculated or using the very same Levenshtein distance calculation and put in.

test_unstructured/metrics/test_text_extraction.py
shreyanid: Let's either remove "eml/fake-email.eml" and "README.md" from the list of filenames being tested, or create a source cct for them to be compared to (I'd pick the first option, second is too much work for now). We should not be comparing CCT to a different source file.

unstructured/metrics/text_extraction.py
```diff
@@ -0,0 +1,52 @@
...
+    Raises:
+        ValueError: If 'return_as' is not one of the valid return types
+        ["score", "distance"].
+
+    Note:
+        This function calculates the edit distance (or similarity score) between
+        two strings using the Levenshtein distance algorithm. The 'weights' parameter
+        allows customizing the cost of insertion, deletion, and substitution
+        operations. The 'return_as' parameter determines the type of result to return:
+        - "score": Returns the similarity score, where 1.0 indicates a perfect match.
```
shreyanid, line 38: Looks great! Could you update the summary to reflect that we only offer 2 return types now?

shreyanid (approved):
LGTM! Since you added a requirement maybe re-pip compile.

Future PRs can revisit the default weights, as well as use actual CCTs for testing.

Issue #1663, opened 2023-10-06
bug/zero-division-error-pdf


cdpierse:
**Describe the bug**
When trying to partition the pdf at the link below we run into a zero division error. https://www.ibm.com/docs/en/SSHT6S_8.2.0/com.ibm.izsam.doc/doc/pdf/IZSAM8.2_221012.pdf 
**To Reproduce**
Provide a code snippet that reproduces the issue.
```python
from unstructured.partition.auto import partition

elements = partition(
    filename="./IZSAM8.2_221012.pdf",
    strategy="fast",
)
```

Winvoker:
I have the same problem with my pdf file.

PR #1665, opened 2023-10-06
Fix: skip empty annotation bbox 

yuming-long:
Address: https://github.com/Unstructured-IO/unstructured/issues/1663
## Summary
While trying to find how overlap between a element bbox and annotation bbox, we find the intersection of two bboxes and divide it by the size of annotation bbox, this will cause a zero division error if size of annotation bbox is 0.

* this PR fix the zero division error for function `check_annotations_within_element`
* also fix error: `TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'` by stop inserting empty word with None bbox into list of words in function `get_word_bounding_box_from_element`

## Test
reproduce with code and document as the user mentioned and should see no error:
```

elements = partition(
    filename="./IZSAM8.2_221012.pdf",
    strategy="fast",
)
```


unstructured/partition/pdf.py
```diff
@@ -1067,12 +1067,13 @@ def check_annotations_within_element(
...
     for annotation in annotation_list:
-        if annotation["page_number"] == page_number and (
-            calculate_intersection_area(element_bbox, annotation["bbox"])
-            / calculate_bbox_area(annotation["bbox"])
-            > threshold
-        ):
-            annotations_within_element.append(annotation)
+        if annotation["page_number"] == page_number:
+            annotation_bbox_size = calculate_bbox_area(annotation["bbox"])
+            if annotation_bbox_size and (
```
Klaijan, line 1072: From the error, I feel like `None` comes from `word` rather than `annot`. Do you think it makes sense to catch `NoneType` at `map_bbox_and_index` as well?

unstructured/partition/pdf.py
Klaijan, line 1: Could we do `try... except...` at `map_bbox_and_index` and return `None` if calculation is invalid (throw error). Then at https://github.com/Unstructured-IO/unstructured/blob/54f809fc32c36bf7f181e78288cb2d892713379e/unstructured/partition/pdf.py#L520 we catch `None` and refuse to append to `urls_metadata` if so. 

Feel free to comment if this doesn't make sense to you!

yuming-long: i am not sure if in the `except` error case we just return`None` regardless is a good idea. I mean what if we have error other than the `TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'` in `map_bbox_and_index`, and since we just return `None` in that case, we didn't catch the error?

PR #1671, opened 2023-10-06
fix: isolate metadata imports to doctype

qued:
In a different PR, some no-extras tests started failing with import errors when something innocuous was imported from `unstructured.file_utils.metadata`. This turned out to be because of the top-level, doctype-specific imports in that file. Importing a general metadata object shouldn't require installation of modules like `PIL`, `docx`, and `openpyxl`.

To fix, I moved these functions to be imported inside the functions that use them, and added the `requires_dependencies` decorator to the functions.

#### Testing:

You should be able to run something like:
```python
from unstructured.file_utils.metadata import Metadata
```

Without `openpyxl` installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
-## 0.10.20-dev4
+## 0.10.20-dev5
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


newelh (approved):
LGTM!

PR #1673, opened 2023-10-06
chore: adding max_characters to other element type chunking

amanda103:
This PR adds the `max_characters` (hard max) param to non-table element chunking. Additionally updates the `num_characters` metadata to `max_characters` to make it clearer which param we're referencing. 

To test:

```

filename = "example-docs/example-10k-1p.html"
chunk_elements = partition_html(
        filename,
        chunking_strategy="by_title",
        combine_text_under_n_chars=0,
        new_after_n_chars=50,
        max_characters=100,
    )

for chunk in chunk_elements:
     print(len(chunk.text))

# previously we were only respecting the "soft max" (default of 500) for elements other than tables
# now we should see that all the elements have text fields under 100 chars.
```


test_unstructured/partition/test_auto.py
```diff
@@ -1052,18 +1052,26 @@ def test_add_chunking_strategy_on_partition_auto_respects_max_chars():
...
     filename = "example-docs/example-10k-1p.html"
 
-    # default chunk size in chars is 200
-    partitioned_table_elements_200_chars = [
+    table_elements = [
+        e
+        for e in partition(
+            filename,
+        )
+        if isinstance(e, Table)
```
newelh, line 1060: could `e for e in partition(filename) if isinstance(e, Table)` all fit on one line?

unstructured/chunking/title.py
```diff
@@ -133,7 +132,25 @@ def chunk_by_title(
...
+                else:
+                    # Otherwise, create a chunk from the end
+                    chunk_text = text[-max_characters:]
+                    text = text[:-max_characters]
+
+                # Prepend the chunk to the beginning of the list
+                chunked_elements.insert(0, CompositeElement(text=chunk_text, metadata=metadata))
+        else:
+            # If it doesn't exceed, create a single CompositeElement
+            chunked_elements.append(CompositeElement(text=text, metadata=metadata))
```
newelh, line 153: You'll have to verify this behavior matches what you're expecting but it passes tests. I think you can reduce this to a 2 liner 
```
chunks = [text[i:i + max_characters] for i in range(0, len(text), max_characters)]
chunked_elements.extend(CompositeElement(text=chunk, metadata=metadata) for chunk in chunks)
```

amanda103: I think this is almost what we want! This is iterating over text forwards, so if there's a leftover chunk smaller than max_characters it'll end up at the end. The behavior Crag and I talked about was it ending up at the start (so iterating over text backwards).

This is how I tweaked your improvement, but it's a little hairy to look at. Not sure it's worth the fewer lines of code:
`chunks = [text[max(0, i - max_characters + 1) : i + 1] for i in range(len(text) -1, -1, -max_characters)]`

test_unstructured/partition/test_auto.py
amanda103: yup looks that way!

unstructured/chunking/title.py
newelh: Great, feel free to leave as is then.

newelh (approved):

PR #1674, opened 2023-10-06
destination connector method elements input

rbiseck3:
### Description
**Ingest destination connectors support for writing raw list of elements** Along with the default write method used in the ingest pipeline to write the json content associated with the ingest docs, each destination connector can now also write a raw list of elements to the desired downstream location without having an ingest doc associated with it.

### DIFF
--- a/unstructured/ingest/connector/azure_cognitive_search.py
+++ b/unstructured/ingest/connector/azure_cognitive_search.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 
 import azure.core.exceptions
 
+from unstructured.documents.elements import Element
 from unstructured.ingest.error import WriteError
 from unstructured.ingest.interfaces import (
     BaseConnectorConfig,
@@ -13,6 +14,7 @@ from unstructured.ingest.interfaces import (
     WriteConfig,
 )
 from unstructured.ingest.logger import logger
+from unstructured.staging.base import convert_to_dict
 from unstructured.utils import requires_dependencies
 
 
@@ -82,18 +84,11 @@ class AzureCognitiveSearchDestinationConnector(BaseDestinationConnector):
         if page_number := data.get("metadata", {}).get("page_number"):
             data["metadata"]["page_number"] = str(page_number)
 
-    def write(self, docs: t.List[BaseIngestDoc]) -> None:
-        json_list = []
-        for doc in docs:
-            local_path = doc._output_filename
-            with open(local_path) as json_file:
-                json_content = json.load(json_file)
-                for content in json_content:
-                    self.conform_dict(data=content)
-                logger.info(
-                    f"appending {len(json_content)} json elements from content in {local_path}",
-                )
-                json_list.extend(json_content)
+    def write_elements(self, elements: t.List[Element]) -> None:
+        elements_dict = convert_to_dict(elements)
+        self.write_dict(json_list=elements_dict)
+
+    def write_dict(self, json_list: t.List[t.Dict[str, t.Any]]) -> None:
         logger.info(
             f"writing {len(json_list)} documents to destination "
             f"index at {self.write_config.index}",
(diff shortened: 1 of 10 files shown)


unstructured/ingest/connector/fsspec.py
```diff
@@ -234,6 +238,32 @@ def initialize(self):
...
+        s3_folder = self.connector_config.path
+
+        s3_output_path = str(PurePath(s3_folder, filename)) if filename else s3_folder
+        element_dict = convert_to_dict(elements)
+        with tempfile.NamedTemporaryFile(mode="w+", encoding=encoding) as tmp_file:
+            json.dump(element_dict, tmp_file, indent=indent)
+            logger.debug(f"Uploading {tmp_file.name} -> {s3_output_path}")
+            fs.put_file(lpath=tmp_file.name, rpath=s3_output_path)
+
     def write(self, docs: t.List[BaseIngestDoc]) -> None:
```
ryannikolaidis, line 267: unable to do the write/write_dict/write_elements pattern for this as you did the others?

unstructured/ingest/connector/azure_cognitive_search.py
```diff
@@ -120,3 +115,17 @@ def write(self, docs: t.List[BaseIngestDoc]) -> None:
...
+            local_path = doc._output_filename
+            with open(local_path) as json_file:
+                json_content = json.load(json_file)
+                for content in json_content:
+                    self.conform_dict(data=content)
+                logger.info(
+                    f"appending {len(json_content)} json elements from content in {local_path}",
+                )
+                json_list.extend(json_content)
+        self.write_dict(json_list=json_list)
```
ryannikolaidis, line 131: is this pattern generalizable so we don't have to redefine everywhere?

unstructured/ingest/connector/azure_cognitive_search.py
```diff
@@ -82,18 +84,11 @@ def conform_dict(self, data: dict) -> None:
...
-                json_content = json.load(json_file)
-                for content in json_content:
-                    self.conform_dict(data=content)
-                logger.info(
-                    f"appending {len(json_content)} json elements from content in {local_path}",
-                )
-                json_list.extend(json_content)
+    def write_elements(self, elements: t.List[Element], *args, **kwargs) -> None:
+        elements_dict = convert_to_dict(elements)
+        self.write_dict(json_list=elements_dict)
```
ryannikolaidis, line 89: same as previous comment, can we make this a function defined on BaseIngestDoc?

unstructured/ingest/connector/fsspec.py
rbiseck3: There's no need for it here since no manipulation is required. The file to upload already exists on disk and simply running it through `fs.put_file()` is all that's needed.

unstructured/ingest/connector/azure_cognitive_search.py
rbiseck3: Looking at it, other than the naming of it, the logic within each is highly specific to each connector and doesn't necessarily have to follow this pattern for any other downstream connectors.

rbiseck3:
@ryannikolaidis given that there's only 3 implementations of this so far, and how varied those 3 are in the implementation, would make sense until we have some more use cases before we look for generalizations in the code.

unstructured/ingest/connector/azure_cognitive_search.py
ryannikolaidis: oh sorry, selected the wrong code for the comment here.

ryannikolaidis:
> @ryannikolaidis given that there's only 3 implementations of this so far, and how varied those 3 are in the implementation, would make sense until we have some more use cases before we look for generalizations in the code.

I have a hunch that once you get in and code, most implementations should fall into that write/write_dict/write_elements pattern and would definitely like to avoid another copy and paste across every single connector if there are patterns we can abstract (even if they are optionally overwritten where needed or mixed in). maybe worth taking a second to explore you're comment on s3 as an option.

overall I think the code itself makes sense if you want to start moving forward. that's just my biggest caution here.

ryannikolaidis (approved):
lgtm! one request: make sure we have this included in the README destination connector steps when the doc PR lands

PR #1579, opened 2023-09-29
Refactor: support entire page OCR with `ocr_mode` and `ocr_languages`

yuming-long:
## Summary
Second part of OCR refactor to move it from inference repo to unstructured repo, first part is done in https://github.com/Unstructured-IO/unstructured-inference/pull/231. This PR adds OCR process logics to entire page OCR, and support two OCR modes, "entire_page" or "individual_blocks".

The updated workflow for `Hi_res` partition:
* pass the document as data/filename to inference repo to get `inferred_layout` (DocumentLayout)
* pass the document as data/filename to OCR module, which first open the document (create temp file/dir as needed), and split the document by pages (convert PDF pages to image pages for PDF file)
* if ocr mode is `"entire_page"`
    *  OCR the entire image
    * merge the OCR layout with inferred page layout
 * if ocr mode is `"individual_blocks"`
    *  from inferred page layout, find element with no extracted text, crop the entire image by the bboxes of the element
    * replace empty text element with the text obtained from OCR the cropped image
* return all merged PageLayouts and form a DocumentLayout subject for later on process

This PR also bump `unstructured-inference==0.7.2` since the branch relay on OCR refactor from unstructured-inference.
  
## Test
```
from unstructured.partition.auto import partition

entrie_page_ocr_mode_elements = partition(filename="example-docs/english-and-korean.png", ocr_mode="entire_page", ocr_languages="eng+kor", strategy="hi_res")
individual_blocks_ocr_mode_elements = partition(filename="example-docs/english-and-korean.png", ocr_mode="individual_blocks", ocr_languages="eng+kor", strategy="hi_res")
print([el.text for el in entrie_page_ocr_mode_elements])
print([el.text for el in individual_blocks_ocr_mode_elements])
```
latest output:
```
# entrie_page
[...]

PR #1675, opened 2023-10-06
Benjamin/bump unstructured inference 0.7.1

benjats07:
This PR updates unstructured-inference to 0.7.1, that download the quantized yolox model instead of dynamically generating it, also pins onnxruntime<1.16


yuming-long:
we are bump `0.7.2` in this PR: https://github.com/Unstructured-IO/unstructured/pull/1579 :)

PR #1680, opened 2023-10-07
chore: don't cleanup ingest test outputs (non-CI)

cragwolfe:
When running test-ingest test fixtures locally (but not in CI), keep output .json's and other workdir artifacts around for the convenience of debugging.

**Test Instructions**

Run 

    bash -x ./test_unstructured_ingest/test-ingest-azure.sh

and witness output .json's are visible. Yay! Now, to instead clean up output .json's and workdir, run:

    UNSTRUCTURED_CLEANUP_DEV_FIXTURES=1 bash -x ./test_unstructured_ingest/test-ingest-azure.sh
    
and witness the files have been cleaned up. Yay!


ryannikolaidis (approved):
🙏

PR #1681, opened 2023-10-07
serde tests round-trip through JSON

scanny:
Each partitioner has a test like `test_partition_x_with_json()`. What these do is serialize the elements produced by the partitioner to JSON, then read them back in from JSON and compare the before and after elements.

Because our element equality (`Element.__eq__()`) is shallow, this doesn't tell us a lot, but if we take it one more step, like `List[Element] -> JSON -> List[Element] -> JSON` and then compare the JSON, it gives us some confidence that the serialized elements can be "re-hydrated" without losing any information.

This actually showed up a few problems, all in the serialization/deserialization (serde) code that all elements share.


test_unstructured/partition/csv/test_tsv.py
```diff
@@ -8,9 +8,9 @@
 )
 from unstructured.cleaners.core import clean_extra_whitespace
 from unstructured.documents.elements import Table
-from unstructured.partition.json import partition_json
 from unstructured.partition.tsv import partition_tsv
-from unstructured.staging.base import elements_to_json
+
+from ...unitutils import assert_round_trips_through_JSON, example_doc_path
```
cragwolfe, line 13: please avoid relative imports

test_unstructured/unitutils.py
```diff
@@ -0,0 +1,38 @@
+"""Utilities that ease unit-testing."""
```
cragwolfe, line 1: name nit on the module name: it does not read naturally, nor is it obvious what the `uni` means.

unstructured/documents/elements.py
```diff
@@ -193,22 +192,23 @@ class ElementMetadata:
...
-    if UNSTRUCTURED_INCLUDE_DEBUG_METADATA:
-        detection_origin: Optional[str] = None
-
-    def __setattr__(self, key, value):
-        if not UNSTRUCTURED_INCLUDE_DEBUG_METADATA and key == "detection_origin":
-            return
-        else:
-            super().__setattr__(key, value)
+    # -- The detection mechanism responsible for this element, for debugging purposes. Should
+    # -- only be populated when UNSTRUCTURED_INCLUDE_DEBUG_METADATA flag is True. Note
```
cragwolfe, line 196: er. but `UNSTRUCTURED_INCLUDE_DEBUG_METADATA` references are gone?

test_unstructured/partition/csv/test_tsv.py
scanny: I do. This is necessary because this is in the tests and the tests are not an installed module so have no qualified name. This module would never appear in the library code nor would we need relative imports..

test_unstructured/unitutils.py
scanny: it's "unit-utils", I can punctuate it if you think that's better, like `unit_utils`.

unstructured/documents/elements.py
scanny: This is in the definition of `ElementMetadata`. The initial implementation made the _definition_ of the `detection_origin` field on `ElementMetadata` conditional based on that flag. That means there are two distinct definitions of `ElementMetadata` and every part of the code that uses it (many of course) needs to become aware that there are two definitions and add code to figure out which one is in effect when they use it.

Then this `.__setattr__()` override was needed to prevent raising an exception when the field was not defined.

Benjamin and I discussed and agreed that there's no problem with having the `detection_origin` field defined on `ElementMetadata` just like any other optional metadata field and then it can be considered or not based on the state of that flag. That limits the coupling to those parts of the code that care about that, and allowed removal of some of the broad special-case, effect-at-a-distance, general-complexifying consequences of a conditional definition.

test_unstructured/unitutils.py
scanny: I changed this to `unit_utils.py`.

test_unstructured/partition/csv/test_tsv.py
scanny: @cragwolfe I fixed this, I missed that we were already importing things from `test_unstructured/...`. Note that this only works when tests are run from the root directory of the repo, but that seems to not be a problem so far.

unstructured/documents/elements.py
scanny: @cragwolfe yep, put this back in.

cragwolfe (approved):
LGTM!

PR #1682, opened 2023-10-07
up version of google drive tests

potter-potter:
just upping the version on the google drive tests
probably an internal google change and not a human change.

[https://github.com/googleapis/google-api-python-client/blob/7df5db27b136225b7bfb2cad041242316f55511d/tests/data/drive.json#L319
](url)

```

"version": {
     "type": "string",
     "description": "A monotonically increasing version number for the file. This reflects every change made to the file on the server, even those not visible to the user.",
     "format": "int64"
    },
```


ryannikolaidis:
maybe we just exclude version [here](https://github.com/Unstructured-IO/unstructured/blob/1e32da6389fa01c2ae23f16865e6859b8ac23e21/test_unstructured_ingest/test-ingest-google-drive.sh#L36)?

ryannikolaidis (approved):

ryannikolaidis (commented):
thank you!

PR #1685, opened 2023-10-09
roman/cli infer table arg

rbiseck3:
### Description
Add new parameter to map to `skip_infer_table_types` partition arg. Applies to partition config which is set on all connectors.


unstructured/ingest/cli/interfaces.py
```diff
@@ -139,6 +139,12 @@ class CliPartitionConfig(PartitionConfig, CliMixin):
     @staticmethod
     def add_cli_options(cmd: click.Command) -> None:
         options = [
+            click.Option(
+                ["--skip-infer-table-types"],
+                type=DelimitedString(),
+                default=None,
+                help="Option list of document types to skip table extraction on",
```
MthwRobinson, line 146: Small type: "Option" -> "Optional"

MthwRobinson (approved):
One small typo, otherwise LGTM from my end

ryannikolaidis:
can we add a quick test? either unit or simply throw that option in an ingest test with a doc that with results that differ based on the setting?

ryannikolaidis (changes requested):
can we add a quick test? either unit or simply throw that option in an ingest test with a doc that with results that differ based on the setting?

test_unstructured_ingest/test-ingest-local-single-file-with-pdf-infer-table-structure.sh
```diff
@@ -22,6 +22,7 @@ PYTHONPATH=. ./unstructured/ingest/main.py \
     --num-processes "$max_processes" \
     --metadata-exclude coordinates,filename,file_directory,metadata.data_source.date_processed,metadata.last_modified,metadata.detection_class_prob,metadata.parent_id,metadata.category_depth \
     --output-dir "$OUTPUT_DIR" \
+    --skip-infer-table-types "" \
```
ryannikolaidis, line 25: what is the difference in specifying empty string here vs the default None?

ryannikolaidis: add a test that validates something other than what happens by default? i.e. should require a change to expected outputs or new test?

rbiseck3: Currently the partition method itself has this as the default for this method: 
```python
skip_infer_table_types: List[str] = ["pdf", "jpg", "png", "xls", "xlsx"],
```
So the default for the CLI is None meaning use the default in the method. If you pass in an empty string, than a default list is used an nothing is omitted. This was actually omitting all of these any time the `--pdf-infer-table-structure` flag was set to true. I took this approach to not have to worry about keeping the default value in sync in case it ever changes on the partition method.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.10.21-dev0
+
+### Enhancements
+* **Expose skip_infer_table_types in ingest CLI** For each connector a new `--skip-infer-table-types` parameter was added to map to the `skip_infer_table_types` partition argument.
```
ryannikolaidis, line 4: Add the why..maybe something like:

```suggestion
* **Expose skip_infer_table_types in ingest CLI** For each connector a new `--skip-infer-table-types` parameter was added to map to the `skip_infer_table_types` partition argument. This gives more granular control to unstructured-ingest users, allowing them to specify which file types for which we should attempt table extraction.
```

ryannikolaidis:
quick suggestion on the changelog otherwise gtg!

ryannikolaidis (approved):
awesome, thanks for grabbing this!

PR #1687, opened 2023-10-09
build(test): skip full python matrix for most ingest tests

ryannikolaidis:
We’re probably unfairly (to the test) making a large volume of new connections and requests to test services when all of our ingest tests run across the full python test matrix and when a lot of PRs a firing at once. Lets limit the full matrix run to a select few, but still have all ingest tests run on python v3.10. This is done by checking the version and skipping in ingest-test.sh.

Bonus: Bumps ingest test fixture workflow to use 3.10. This technically shouldn't make a difference, but since we're making 3.10 the default of the matrix strategy, it probably makes sense to use 3.10 for the ingest fixture generation as well for consistency.

## Testing
- [example](https://github.com/Unstructured-IO/unstructured/actions/runs/6460319121/job/17537900978?pr=1687) running all tests in 3.10
- [example](https://github.com/Unstructured-IO/unstructured/actions/runs/6460319121/job/17537899999?pr=1687) skipping/running the expected tests in 3.8

### DIFF
--- a/test_unstructured_ingest/test-ingest.sh
+++ b/test_unstructured_ingest/test-ingest.sh
@@ -72,6 +72,7 @@ for test in "${all_tests[@]}"; do
   CURRENT_TEST="$test"
   if [[ "$python_version" != "Python 3.10"* ]] && [[ ! "${full_python_matrix_tests[*]}" =~ $test ]] ; then
     echo "--------- SKIPPING SCRIPT $test ---------"
+    continue
   fi
   if [[ "$test" == "test-ingest-notion.sh" ]]; then
     echo "--------- RUNNING SCRIPT $test --- IGNORING FAILURES"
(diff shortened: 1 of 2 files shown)


test_unstructured_ingest/test-ingest.sh
```diff
@@ -45,29 +45,46 @@ scripts=(
...
 
 trap print_last_run EXIT
 
-for script in "${scripts[@]}"; do
-  CURRENT_SCRIPT=$script
-  if [[ "$CURRENT_SCRIPT" == "test-ingest-notion.sh" ]]; then
-    echo "--------- RUNNING SCRIPT $script --- IGNORING FAILURES"
+python_version=$(python --version 2>&1)
+
+for test in "${all_tests[@]}"; do
```
rbiseck3, line 71: can we add some documentation here on what this logic is meant to be handling? It's getting pretty complicated and reading bash logic comparisons are not the easiest.

rbiseck3:
Small ask for some additional documentation, but otherwise looks good. Always a fan of anything that helps expedite the ingest tests.

rbiseck3 (approved):

PR #1688, opened 2023-10-09
feat: element type frequency

Klaijan:
**Executive Summary**

Add function that returns frequency of given element types and depth.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * **Adds `edit_distance` calculation metrics** In order to benchmark the cleaned, extracted text with unstructured, `edit_distance` (`Levenshtein distance`) is included.
 * **Adds detection_origin field to metadata** Problem: Currently isn't an easy way to find out how an element was created. With this change that information is added. Importance: With this information the developers and users are now able to know how an element was created to make decisions on how to use it. In order tu use this feature
 setting UNSTRUCTURED_INCLUDE_DEBUG_METADATA=true is needed.
+* **Adds a function that calculates frequency of the element type and its depth** To capture the accuracy of element type extraction, this metrics introduces one way to measure the element type accuracy by take in account the depth of the element.
```
shreyanid, line 17:
```suggestion
* **Adds a function that calculates frequency of the element type and its depth** To capture the accuracy of element type extraction, this function counts the occurrences of each unique element type with its depth for use in element metrics.
```
small changes/simplification, don't need as much emphasis on depth

unstructured/metrics/element_type.py
```diff
@@ -0,0 +1,19 @@
+from typing import Dict, List, Union
+
+
+def get_element_type_frequency(
+    elements: List,
+) -> Union[Dict[str, Dict[str, int]], Dict]:
+    frequency: Dict = {}
```
shreyanid, line 7: Can we have this dictionary initialize all element categories to 0 (like a default dict)? Will help with comparison between 2 such dictionaries since they will both have the same keys

unstructured/metrics/element_type.py
```diff
@@ -0,0 +1,19 @@
...
+    if len(elements) == 0:
+        return frequency
+    for element in elements:
+        category = element.category
+        category_depth = element.metadata.category_depth
+        if category not in frequency:
+            frequency[category] = {}
+        if str(category_depth) not in frequency[category]:
```
shreyanid, line 15: Debating if a similar value initialization should be done for depth, and in general what the best representation of depth is for this use case. 

I like that the nested category depth info makes it easy to consider element types without depth if desired, but the alternative of the main key being a tuple (`element category`, `depth`) would also possibly have the same benefit. Drawback of a tuple is that every category needs to hold some value in the depth field of the tuple whether or not it applies.

For comparisons, I was thinking it would be easiest if both dictionaries being compared had the exact same keys, but in the case of depth we don't have a limit on it and could potentially nest infinitely, so we don't want to represent all those values with a default.

What do you think makes the most sense for a representation + default value decision given that the primary use case is to (easily) compare the resulting element frequencies between 2 input element lists? I'm leaning default for element categories (prior comment), no default for category depth, and let category depth be nested so it isn't clunky for categories that don't have a depth (which is most of them).

test_unstructured/metrics/test_element_type.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.mark.parametrize(
+    ("filename", "frequency"),
+    [
+        (
+            "fake-email.txt",
+            {
+                "UncategorizedText": {"None": 6},
+                "ListItem": {"None": 12},
+                "Title": {"None": 5},
+                "NarrativeText": {"None": 2},
```
shreyanid, line 16: Oh hmm, if the result is still going to be a nested dictionary whether or not the category has a depth... it's looking a little clunky. I'm leaning baking both the category name and depth into a tuple now for easier indexing (with a tuple, you can just index for the first element and ignore the second field, but with a nested dictionary you are forced to index each level). Will edit my previous comment lol

unstructured/metrics/element_type.py
Klaijan: So you mean for the case that has `category_depth=None`, you'd want the value in the tuple to be just pure number? What about when same category have both depth and no depth in the same document?

test_unstructured/metrics/test_element_type.py
```diff
@@ -0,0 +1,87 @@
...
+from unstructured.partition.auto import partition
+
+
+@pytest.mark.parametrize(
+    ("filename", "frequency"),
+    [
+        (
+            "fake-email.txt",
+            {
+                "UncategorizedText": [("None", 6)],
```
shreyanid, line 13: I meant a tuple as the key in the dictionary of frequencies.
ex. `("UncategorizedText", "None"): 6` or ex. `("Title", 1): 3`

the type of the depth as string or numerical (to cover both None and 0, 1, etc) is up to you

Klaijan: But in that case, how would the Element types that aren't exist in the doc be initialize? `("UncategorizedText",)` like this?

shreyanid: In this way, each tuple key to the dictionary represents a unique element type. Overall, the type of the dictionary will be `Dict[Tuple[str, int], int]` (or possibly str for second value in tuple to capture None)

Would it then no longer be possible to create all the keys for the dictionary? Probably. Let's instead prioritize having meaningful keys instead of initializing the dictionary with all elements categories.

Klaijan: So we revert back to not having all the Element Type initialized, then assign value to each tuple key.

shreyanid (approved):
LGTM!

PR #1689, opened 2023-10-09
roman/update ingest pipeline docs

rbiseck3:
### Description
* Update all existing connector docs to use new pipeline approach

### Additional changes:
* Some defaults were set for the runners to match those in the configs to make those easy to handle, i.e. the biomed runner:
```python
max_retries: int = 5,
max_request_time: int = 45,
decay: float = 0.3,
```


unstructured/ingest/runner/confluence.py
```diff
@@ -13,8 +13,8 @@ def run(
         url: str,
         user_email: str,
         api_token: str,
-        max_num_of_spaces: int,
-        max_num_of_docs_from_each_space: int,
+        max_num_of_spaces: int = 500,
+        max_num_of_docs_from_each_space: int = 100,
```
ryannikolaidis, line 17: same as above

ryannikolaidis: should we have cli.cmds.confluence import these defaults from the runner so they are only explicitly assigned in one place? even if it just means we have a MAX_NUM_OF_SPACES constant at the top of the confluence runner module, but defer if there's something more elegant.

also this may be a bigger question of how we do this for all of these. if so, probably better for another PR..but also not necessarily the most urgent

ryannikolaidis:
do we have doc pages for the various required configs? if not we should

rbiseck3:
> do we have doc pages for the various required configs? if not we should 

@ryannikolaidis, should that be added to a README somewhere or the sphinx docs?

ryannikolaidis:
> > do we have doc pages for the various required configs? if not we should
> 
> @ryannikolaidis, should that be added to a README somewhere or the sphinx docs?

sphinx docs for sure. maybe both? it does look like the README could probably benefit from a bit more details. and oh, now that I'm looking at it the entire [design section](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/README.md?plain=1#L94-L112) needs a bump with these changes, right? can probably re-use a good portion of what you added in the PR description.

ryannikolaidis:
bump ^ still moving forward? would be good to get things up to date so we can point users at docs

ryannikolaidis:
> > > do we have doc pages for the various required configs? if not we should
> > 
> > 
> > @ryannikolaidis, should that be added to a README somewhere or the sphinx docs?
> 
> sphinx docs for sure. maybe both? it does look like the README could probably benefit from a bit more details. and oh, now that I'm looking at it the entire [design section](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/README.md?plain=1#L94-L112) needs a bump with these changes, right? can probably re-use a good portion of what you added in the PR description.

@rbiseck3 where are we with this? definitely should bump the README which is out of date with the new refactor

ryannikolaidis (approved):
lgtm! 

from conversation outside this PR: to follow will be a bigger overhaul on ingest docs to more comprehensively cover capabilities

PR #1701, opened 2023-10-10
feat: calculate metric for percent of text missing

shreyanid:
### Summary
Missing text is a particularly important metric of quality for the Unstructured library because information from the document is not being captured and therefore not usable by downstream applications.

Add function to calculate the percent of text missing relative to the source transcription. Function takes 2 text strings (output and source) as input, and returns the percentage of text missing as a decimal.

### Technical Details
- The 2 input strings are both assumed to already contain clean and concatenated text (CCT)
- Implementation compares the bags of words (frequency counts for each word present in the text) of each input text
- Duplicated/extra text is not penalized
- Value is limited to the range [0, 1]

### Test
- Several edge cases are covered in the test function (missing text, duplicated text, spaced out words, etc).
- Can test other cases or text inputs by calling the function with 2 CCT strings as "output" and "source"

### DIFF
--- a/unstructured/metrics/text_extraction.py
+++ b/unstructured/metrics/text_extraction.py
@@ -82,3 +82,50 @@ def bag_of_words(text: str) -> Dict[str, int]:
                     bow[incorrect_word] = 1
             i = j
     return bow
+
+
+def calculate_percent_missing_text(
+    output: str,
+    source: str,
+) -> float:
+    """
+    Creates the bag of words (BOW) found in each input text and their frequencies, then compares the 
+    output BOW against the source BOW to calculate the % of text from the source text missing from
+    the output text.
+
+    Takes "clean, concatenated text" (CCT) from a document output and the ground truth source text 
+    as inputs.
+
+    If the output text contains all words from the source text and then some extra, result will be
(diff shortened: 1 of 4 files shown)


unstructured/metrics/text_extraction.py
```diff
@@ -82,3 +89,47 @@ def bag_of_words(text: str) -> Dict[str, int]:
...
+    total_missing_word_count = 0
+
+    for source_word, source_count in source_bow.items():
+        total_source_word_count += source_count
+        if source_word not in output_bow:
+            # entire count is missing
+            total_missing_word_count += source_count
+        else:
+            output_count = output_bow[source_word]
+            total_missing_word_count += max(source_count - output_count, 0)
```
Klaijan, line 128: Just to clarify, we are not penalizing any extra words extracted right?

shreyanid: Correct, we are not penalizing duplicated text. If the unstructured output text has everything from the source text and then more, we count that is 0% missing text.

Klaijan (approved):
LGTM!

mallorih (approved):

PR #1705, opened 2023-10-11
Add basic call to scarf to get anonymous analytics

tabossert:
There is a built in option to not send data by setting an env var, SCARF_NO_ANALYTICS=true.

DoD:
- When importing or running unstructured package it will make a get call to scarf 
- When env variable is set to not track, call is not made


unstructured/logger.py
```diff
@@ -13,6 +15,7 @@ def detail(self, message, *args, **kws):
     if self.isEnabledFor(DETAIL):
         self._log(DETAIL, message, args, **kws)
 
+scarf_analytics()
```
cragwolfe, line 18:
```suggestion
# Note(Trevor,Crag): to opt out of scarf analytics, set the environment variable:
# SCARF_NO_ANALYTICS=true. See the README for more info.
scarf_analytics()
```

cragwolfe (approved):

PR #1719, opened 2023-10-11
fix: _add_embeddings_to_elements bug resulting in duplicated elements

ryannikolaidis:
Currently when the  OpenAIEmbeddingEncoder adds embeddings to Elements in `_add_embeddings_to_elements` it overwrites each Element's `to_dict` method, mistakenly resulting in each Element having identical values with the exception of the actual embedding value. This was due to the way it leverages a nested `new_to_dict` method to overwrite. Instead, this updates the original definition of Element itself to accommodate the `embeddings` field when available. This also adds a test to validate that values are not duplicated.


unstructured/documents/elements.py
```diff
@@ -490,6 +490,8 @@ def to_dict(self) -> dict:
         out["element_id"] = self.id
         out["type"] = self.category
         out["text"] = self.text
```
rbiseck3, line 492: Can we instead add this to the Element class:

embedding: Optional[List[float]] = None


That way we can also have some typing around it.

rbiseck3 (changes requested):

unstructured/documents/elements.py
```diff
@@ -454,9 +454,11 @@ def __init__(
         coordinate_system: Optional[CoordinateSystem] = None,
         metadata: Optional[ElementMetadata] = None,
         detection_origin: Optional[str] = None,
+        embeddings: Optional[List[float]] = None,
     ):
         metadata = metadata if metadata else ElementMetadata()
         self.text: str = text
+        self.embeddings: Optional[str] = embeddings
```
rbiseck3, line 461: Why is the typing here set to `Optional[str]`?

ryannikolaidis: Typo

ryannikolaidis:
```suggestion
        self.embeddings: Optional[List[float]] = embeddings
```

rbiseck3 (approved):

PR #1723, opened 2023-10-11
feat: calculate element type percent match

Klaijan:
**Executive Summary**
Adds function to calculate the percent match between two element type frequency output from `get_element_type_frequency` function. 

**Technical Detail**
- The function takes two `Dict` input which both should be output from `get_element_type_frequency`
- Implementors can define weight `category_depth_weight` they want to give to the matching `type` but different in `category_depth` case
- The function loops through output item list first to find exact match and count total exact match, and collect the remaining value for both output and source in new list (of `dict` type). Then it loops through existing source item list that has not been an exact match, to find `type` match which then weigh with the factor of `category_depth_weight` defined earlier, default at 0.5)

**Output**
output
```
{
  ("Title", 0): 2,
  ("Title", 1): 1,
  ("NarrativeText", None): 3,
  ("UncategorizedText", None): 1,
}
```

source
```
{
  ("Title", 0): 1,
  ("Title", 1): 2,
  ("NarrativeText", None): 5,
}
```

With this output and source, and weight of 0.5, the % match will yield 5.5 / 8 -- for 5 exact match, and 1 partial match with 0.5 weight.


test_unstructured/metrics/test_element_type.py
```diff
@@ -35,3 +38,37 @@ def test_get_element_type_frequency(filename, frequency):
...
+                ("ListItem", 2): 3,
+                ("Table", None): 1,
+            },
+        ),
+    ],
+)
+def test_calculate_element_type_percent_match(filename, frequency):
+    elements = partition(filename=f"example-docs/{filename}")
+    elements_freq = get_element_type_frequency(elements_to_json(elements))
+    assert calculate_element_type_percent_match(elements_freq, frequency) == 1.0
```
shreyanid, line 72: so in these tests, are the elements always 100% match? Can we get some examples (we can just modify the "ground truth" parametrized element frequencies) where we don't have 100%, either through missing or misclassified elements?

unstructured/metrics/element_type.py
```diff
@@ -20,3 +20,52 @@ def get_element_type_frequency(
...
+    total_match_element_count = 0
+
+    unmatched_depth_output = {}
+    unmatched_depth_source = {}
+
+    for k, v in output_copy.items():
+        if k in source_copy:
+            match_count = min(output_copy[k], source_copy[k])
+            total_match_element_count += match_count
+            total_source_element_count += match_count
```
shreyanid, line 51: should this be source_count and not match_count?

Klaijan: It'll continue counting the rest of the source later since we're looping through the output list now so we can only track the matched one in this loop.

test_unstructured/metrics/test_element_type.py
Klaijan: We could mock fake output, but it'd be best if we compare the mock one against the gold standard. So if we want to see a comparison from same file but two sources, let's create a gold standard for these files.

shreyanid: Ok sounds good. It can possibly just be the unstructured output with manual human edits to look like the gold standard, we don't need to make it from scratch.

I just want to see a case where the unstructured output somehow doesn't 100% match the gold standard, however that test can be added is up to you :)

unstructured/metrics/element_type.py
```diff
@@ -20,3 +20,71 @@ def get_element_type_frequency(
...
+        if k in unmatched_depth_output:
+            match_count = min(unmatched_depth_output[k], unmatched_depth_source[k])
+            total_match_element_count += match_count * category_depth_weight
+
+    return min(max(total_match_element_count / total_source_element_count, 0.0), 1.0)
+
+
+def _convert_to_frequency_without_depth(d: Dict) -> Dict:
+    """
+    Takes in element frequency with depth of format (type, tepth): value
```
shreyanid, line 80:
```suggestion
    Takes in element frequency with depth of format (type, depth): value
```
typo

shreyanid (approved):
LGTM!

Issue #1660, opened 2023-10-06
enhancement: evaluate language detection packages to see which is best for detecting short text


Coniferish:
Stemming from conversations [here](https://github.com/Unstructured-IO/unstructured/pull/1627#discussion_r1346498859) and [here](https://github.com/Unstructured-IO/unstructured/pull/1652), it would be worthwhile to do our own comparison of language detection packages to see which is best for detecting the language of short text. Speed and size of the packages should also be considered. Packages of interest include langdetect (which we are currently using), fasttext (if it is compatible with py 3.11), langid, and lingua. We are also currently using a regex pattern and arbitrary text length limit to default to "eng", so this should also be considered/reconsidered.

See `detect_languages` in `lang.py`

Coniferish:
Notes/research from 10/2023

Package | Url | Stars | Last updated | License | Python compatibility | Detects multiple languages in the same text | Deterministic?
-- | -- | -- | -- | -- | -- | -- | --
Langdetect | https://github.com/Mimino666/langdetect | 1.5k | 2021 |   |   |   |  
Polyglot | https://github.com/aboSamoor/polyglot | 2.2k | 2020 | MIT |   |   |  
python-polyglot | https://github.com/lainq/polyglot | 27 | 2021 |   |   |   |  
cld3 | https://github.com/google/cld3 | 706 | 2022 | Apache 2.0 |   |   |  
pycld3 | https://github.com/bsolomon1124/pycld3 | 135 | 2021 | Apache 2.0 |   |   |  
pycld2 | https://github.com/aboSamoor/pycld2 | 147 | 2022 | Apache 2.0 |   |   |  
Fasttext | https://github.com/facebookresearch/fastText | 25.1k | 2023 | MIT |   |   |  
~~spaCy~~ | https://github.com/explosion/spaCy |   |   |   |   |   |  
lingua | https://github.com/wichert/lingua | 44 | 2022 |   |   |   |  
lingua-py | https://github.com/pemistahl/lingua-py | 635 | 2023 | apache 2.0 |   |   |  
~~googletrans~~ | https://github.com/ssut/py-googletrans |   |   |   |   |   |  
~~textblob~~ | https://github.com/sloria/TextBlob | 8.7k | 2023 |   |   |   |  
langid | https://github.com/saffsd/langid.py | 2.2k | 2017 | BSD-2-Clause |   |   |  
[...]

PR #1652, opened 2023-10-05
detect document language across all partitioners <- Ingest test fixtures update

ryannikolaidis:
This pull request includes updated ingest test fixtures.
Please review and merge if appropriate.

PR #1729, opened 2023-10-12
roman/bugfix fix default language ingest option

rbiseck3:
### Description
Set language to None by default. Update ingest test to use local file used in language unit tests to validate.


test_unstructured_ingest/expected-structured-output/airtable-diff/app5YQxSfp220fWtm/tblxdPc7L2meGIZLE.json
```diff
@@ -15,7 +15,9 @@
       },
       "filetype": "text/csv",
       "languages": [
-        "eng"
+        "cat",
+        "ita",
+        "fra"
```
ryannikolaidis, line 20: 🤣 I mean...fair

test_unstructured_ingest/expected-structured-output/Sharepoint/SitePages/This-is-a-title.html.json
```diff
@@ -16,7 +16,8 @@
       "filename": "This-is-a-title.html",
       "filetype": "text/html",
       "languages": [
-        "eng"
+        "cat",
+        "fra"
```
ryannikolaidis, line 20: hmm, so obv not a bug for this PR since ingest is just enabling it, but this one definitely looks like `eng` to me??

`"text": "This is a plain text site page for testing purposes"`

rbiseck3: @Coniferish might have some better insight?

ryannikolaidis: and cc @shreyanid?

ryannikolaidis (approved):
noted some curiosities with respect to the auto languages, but that's out of the purview of this PR. changes here are gtg!

test_unstructured_ingest/expected-structured-output/Sharepoint/SitePages/This-is-a-title.html.json
Coniferish: Yah, `langdetect` doesn't do well with short text, so I added logic in `detect_languages` to default to English, but conservatively set the length to 20 characters. We have a follow-up PR [here](https://github.com/Unstructured-IO/unstructured/issues/1660) and discussed some of that in this Ingest PR: https://github.com/Unstructured-IO/unstructured/pull/1652

test_unstructured_ingest/expected-structured-output/local-single-file-with-encoding/fake-html-cp1252.html.json
```diff
@@ -19,7 +22,10 @@
       "data_source": {},
       "filetype": "text/html",
       "languages": [
-        "eng"
+        "por",
+        "cat",
+        "eng",
+        "vie"
```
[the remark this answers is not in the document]
Coniferish: Though I'm confused how this one isn't being detected as English given the length check is `<20` and this is 19 characters...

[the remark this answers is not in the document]
Coniferish: I think something might still be wrong with the ingest... This really shouldn't be happening.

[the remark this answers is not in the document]
Coniferish: Unless this is part of a larger text document that is primarily Latin or some other language. The default is to detect language at the _document_ level rather than the _element_ level. So if a document is primarily in Latin, every element will be detected as such unless `detect_language_per_element` is set to `True`

test_unstructured_ingest/expected-structured-output/Sharepoint/SitePages/This-is-a-title.html.json
Coniferish: But this one is actually because language detection happens at the document level by default and the majority of this text is in Latin.

PR #1730, opened 2023-10-12
roman/fsspec compression support

rbiseck3:
### Description 
Opened to replace original PR: [1443](https://github.com/Unstructured-IO/unstructured/pull/1443)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.10.24-dev0
+
+### Enhancements
+
+* **Ingest compression utilities and fsspec connector support** Generic utility code added to handle files that get pulled from a source connector that are either tar or zip compressed and uncompress them locally. This is then processed using a local source connector. Currently this functionality has been incorporated into the fsspec connector and all those inheriting from it (currently: Azure Blob Storage, Google Cloud Storage, S3, Box, and Dropbox).
+
+### Features
+
+### Fixes
+
 ## 0.10.23
 
 ### Enhancements
--- a/unstructured/ingest/interfaces.py
+++ b/unstructured/ingest/interfaces.py
@@ -76,10 +76,14 @@ class ProcessorConfig(BaseConfig):
 
 
 @dataclass
-class FsspecConfig(BaseConfig):
+class FileStorageConfig(BaseConfig):
     remote_url: str
     uncompress: bool = False
     recursive: bool = False
+
+
+@dataclass
+class FsspecConfig(FileStorageConfig):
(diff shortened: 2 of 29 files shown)


unstructured/ingest/interfaces.py
```diff
@@ -61,6 +73,53 @@ class ProcessorConfig(BaseConfig):
     num_processes: int = 2
 
 
+@dataclass
+class FsspecConfig(BaseConfig):
```
ryannikolaidis, line 77: what's the thinking around moving this to interfaces?

ryannikolaidis: got it, I guess the icky-ness is that this is for a specific implementation: fsspec. though...really does this have to be fsspec? seems like these options are all reasonable for any file storage type connector? e.g. onedrive and google drive which are both file storage sources but not fsspec

ryannikolaidis: ...not that we need to support these in this PR, but that the name of this config could be more abstract if it could be used to satisfy those as well in a later PR

rbiseck3: Small modification made to make it more generic. Some of the code is still specific to fsspec such as all the protocol parsing but I moved out what could be generic into a FileStorageConfig.

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
-## 0.10.22-dev5
+## 0.10.22-dev6
 
 ### Enhancements
 
 * **Expose skip_infer_table_types in ingest CLI.** For each connector a new `--skip-infer-table-types` parameter was added to map to the `skip_infer_table_types` partition argument. This gives more granular control to unstructured-ingest users, allowing them to specify the file types for which we should attempt table extraction.
 * **Add flag to ingest CLI to raise error if any single doc fails in pipeline** Currently if a single doc fails in the pipeline, the whole thing halts due to the error. This flag defaults to log an error but continue with the docs it can.
+* **Ingest compression support** Generic utility code added to handle files that get pulled from a source connector that are either tar or zip compressed and uncompress them locally. This is then processed using a local source connector. Currently this functionality has been incorporated into the Fsspec connector and all those inheriting from it.
```
ryannikolaidis, line 7:
```suggestion
* **Ingest compression utilities and fsspec connector support** Generic utility code added to handle files that get pulled from a source connector that are either tar or zip compressed and uncompress them locally. This is then processed using a local source connector. Currently this functionality has been incorporated into the fsspec connector and all those inheriting from it.
```

ryannikolaidis (approved):
nit on the changelog otherwise this looks great!

PR #1736, opened 2023-10-12
roman/ingest continue on error

rbiseck3:
### Description
Add flag to raise an error on failure but default to only log it and continue with other docs


ryannikolaidis:
is there a quick way to validate this is doing what we think? thinking out loud: maybe local ingest two files, one which is an unsupported (maybe fake type we'll never support) and one is valid and confirm that the valid one makes it to completion? (don't even need to diff, can just validate number of files output)?

basically just want to make sure we never regress this.

rbiseck3:
@ryannikolaidis 
I manually verified it locally by injecting an error in various steps using 
```python
import random
if random.random() > 0.5:
  raise ValueError("Test Error")
```

Not sure how to deterministically force a partition step to fail in the ingest tests.

ryannikolaidis:
> @ryannikolaidis I manually verified it locally by injecting an error in various steps using
> 
> ```python
> import random
> if random.random() > 0.5:
>   raise ValueError("Test Error")
> ```
> 
> Not sure how to deterministically force a partition step to fail in the ingest tests.

oh I would just make a test directory with two files: `small.txt` and `foo.not-a-real-extension`. then have it process the directory. it will hit unknown file type and error when trying to partition `foo.not-a-real-extension`.

ryannikolaidis (approved):
lgtm!

PR #1738, opened 2023-10-12
Added AWS Bedrock embeddings

jackretterer:
Summary: Added support for AWS Bedrock embeddings. Leverages "amazon.titan-tg1-large" for the embedding model.

Test

- find your aws secret access key and key id; make sure the account has access to bedrock's tian embed model
- follow the instructions in https://github.com/Unstructured-IO/unstructured/blob/d5e797cd44bfe3d423750b4442148d03f942d27d/docs/source/bricks/embedding.rst#bedrockembeddingencoder


unstructured/embed/bedrock.py
```diff
@@ -0,0 +1,72 @@
...
+        return np.shape(self.examplary_embedding)
+
+    def is_unit_vector(self):
+        return np.isclose(np.linalg.norm(self.examplary_embedding), 1.0)
+
+    def embed_query(self, query):
+        return np.array(self.bedrock_client.embed_query(query))
+
+    def embed_documents(self, elements: List[Element]) -> List[Element]:
+        embeddings = [np.array(self.bedrock_client.embed_query(str(e))) for e in elements]
```
badGarnet, line 35:
```suggestion
        embeddings = [np.array(self.embed_query(str(e))) for e in elements]
```

unstructured/embed/bedrock.py
```diff
@@ -0,0 +1,72 @@
...
+
+            def new_to_dict(self):
+                d = original_method()
+                d["embeddings"] = self.embeddings
+                return d
+
+            element.embeddings = embeddings[i]
+            elements_w_embedding.append(element)
+            element.to_dict = types.MethodType(new_to_dict, element)
+        return elements
```
badGarnet, line 54: I see this is using https://github.com/Unstructured-IO/unstructured/blob/97f832619ab60022b29c50ba89703d9922cfbe5f/unstructured/embed/openai.py#L37-L52
this part of the code essentially depend on `element` data definition so can be shared by different embeddings. So we can consider refactor this either:
- as a method for elements
- as a utils func in `unstructured/embed/utils.py` so both here and openai version can import the same function

unstructured/embed/bedrock.py
```diff
@@ -0,0 +1,72 @@
...
+    def _add_embeddings_to_elements(self, elements, embeddings) -> List[Element]:
+        assert len(elements) == len(embeddings)
+        elements_w_embedding = []
+
+        for i, element in enumerate(elements):
+            original_method = element.to_dict
+
+            def new_to_dict(self):
+                d = original_method()
+                d["embeddings"] = self.embeddings
```
badGarnet, line 48: yeah; given this operation here I mean leaning even more toward moving this function to be part of element (here `self` refers to an element... NOT this class itself

unstructured/embed/bedrock.py
badGarnet, line 54: I see this is using https://github.com/Unstructured-IO/unstructured/blob/97f832619ab60022b29c50ba89703d9922cfbe5f/unstructured/embed/openai.py#L37-L52
this part of the code essentially depend on `element` data definition so can be shared by different embeddings. So we can consider refactor this either:
- as a method for elements
- as a utils func in `unstructured/embed/utils.py` so both here and openai version can import the same function

unstructured/embed/bedrock.py
badGarnet: actually langchain has a bulk embed api: https://api.python.langchain.com/en/latest/embeddings/langchain.embeddings.bedrock.BedrockEmbeddings.html#langchain.embeddings.bedrock.BedrockEmbeddings.embed_documents
so better to do:
```
embeddings = self.bedrock_client.embed_documents([str(e) for e in elements])
```

unstructured/embed/bedrock.py
```diff
@@ -0,0 +1,63 @@
...
+            bedrock_runtime = boto3.client(
+                service_name="bedrock-runtime",
+                aws_access_key_id=self.aws_access_key_id,
+                aws_secret_access_key=self.aws_secret_access_key,
+                region_name=self.region_name
+            )
+
+            bedrock_client = BedrockEmbeddings(client=bedrock_runtime)
+            self.examplary_embedding = np.array(bedrock_client.embed_query("Q"))
+            return bedrock_client
```
ahmetmeleq, line 63: If we already have a client and this method is called, we can return 
```suggestion
    def get_bedrock_client(self):
        if getattr(self, "bedrock_client", None):
            return self.bedrock_client

        bedrock_runtime = boto3.client(
            service_name="bedrock-runtime",
            aws_access_key_id=self.aws_access_key_id,
            aws_secret_access_key=self.aws_secret_access_key,
            region_name=self.region_name
        )

        bedrock_client = BedrockEmbeddings(client=bedrock_runtime)
        self.examplary_embedding = np.array(bedrock_client.embed_query("Q"))
        return bedrock_client
```

ahmetmeleq:
Reminder to update the docs: https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/bricks/embedding.rst

ahmetmeleq:
1. Let's form an extra named `embed-aws-bedrock` and add it as an extra key to setup.py as in:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/setup.py#L160

2. We need an .in requirements file for this to work (we can name it `embed-aws-bedrock.in`). Example:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/requirements/ingest-openai.in

3. Finally, let's add a `@requires_dependencies` decorator to the methods where we use the extra libraries explicitly. Example:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/unstructured/embed/openai.py#L44-L49

ahmetmeleq:
Let's wrap potential connection errors as in:
https://github.com/Unstructured-IO/unstructured/blob/282b8f700d9471b9430e8a37af73ca96b980e0f0/unstructured/embed/openai.py#L44

unstructured/embed/bedrock.py
```diff
@@ -0,0 +1,65 @@
...
+        return np.shape(self.examplary_embedding)
+
+    def is_unit_vector(self):
+        return np.isclose(np.linalg.norm(self.examplary_embedding), 1.0)
+
+    def embed_query(self, query):
+        return np.array(self.bedrock_client.embed_query(query))
+
+    def embed_documents(self, elements: List[Element]) -> List[Element]:
+        embeddings = [np.array(self.embed_query(str(e))) for e in elements]
```
ahmetmeleq, line 34: Will this work? We shouldn't use query embedding while embedding documents
```suggestion
        embeddings = self.bedrock_client.embed_documents([str(e) for e in elements])
```

ahmetmeleq:
Can we add a mock test as in: https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured/embed/test_openai.py

docs/requirements.txt
```diff
@@ -116,7 +116,7 @@ sphinxcontrib-serializinghtml==1.1.5
     # via
     #   -r requirements/build.in
     #   sphinx
-urllib3==1.26.17
+urllib3==1.26.18
```
ahmetmeleq, line 119:
```suggestion
urllib3==1.26.17
```

ahmetmeleq:
LGTM!

ahmetmeleq (approved):
LGTM!

PR #1741, opened 2023-10-12
feat: staging function to extract element text into one string

shreyanid:
### Summary
In order to enable larger scale testing of the new text extraction metrics, create a helper function to get the clean, concatenated text (CCT) from partitioned elements.

### Test
Partition any file, then pass the resulting elements into the new `elements_to_text` function. Can test getting the output as string or as text file.

```
from unstructured.staging.base import elements_to_text

elements = partition(filename="example-docs/chevron-page.pdf", strategy="hi_res")
elements_text = elements_to_text(elements, "output-text-file.txt")
print(elements_text)
```


unstructured/staging/base.py
```diff
@@ -39,6 +39,35 @@ def _get_metadata_table_fieldnames():
 ] + _get_metadata_table_fieldnames()
 
 
+def convert_to_text(elements: List[Element]) -> str:
+    """Converts a list of elements into clean, concatenated text."""
+    clean_concat_text: str = ""
+    for element in elements:
```
rbiseck3, line 45: Two questions: why do we need to run `to_dict()` rather than simply get the text from the element using `element.text`? And do we always want to concatenate using a newline character?

unstructured/staging/base.py
```diff
@@ -39,6 +39,35 @@ def _get_metadata_table_fieldnames():
...
+        chunk = element.to_dict()
+        clean_concat_text += chunk["text"]
+        clean_concat_text += "\n"
```
rbiseck3, line 48: Potentially simpler code? 
```python
return "\n".join([e.text for e in elements])
```

Might not need an entire function for it.

unstructured/staging/base.py
```diff
@@ -39,6 +39,35 @@ def _get_metadata_table_fieldnames():
...
+    return clean_concat_text
```
badGarnet, line 49:
```suggestion
    return "\n".join([element.text for element in elements])
```

unstructured/staging/base.py
```diff
@@ -39,6 +39,35 @@ def _get_metadata_table_fieldnames():
 ] + _get_metadata_table_fieldnames()
 
 
+def convert_to_text(elements: List[Element]) -> str:
+    """Converts a list of elements into clean, concatenated text."""
+    clean_concat_text: str = ""
```
shreyanid, line 44: question for reviewer: will adding all the text to one string like this cause a problem for very large files?

unstructured/staging/base.py
rbiseck3: Slight optimization to omit elements with empty text:
```python
"\n".join([e.text for e in elements if e])
```

unstructured/staging/base.py
shreyanid: Simplification to use element.text added now from other comments :)
On newline, I'd say part of the benefit of separating out the actual text conversion logic (even though its only one line) is that we can change maybe the joining character or any other text formatting we may want to do. For now, we do always want to concatenate with a single new line character, so it's not currently a parametrized value.

rbiseck3 (approved):

PR #1743, opened 2023-10-12
fix: default to None for the languages metadata field

### LINKED ISSUE #1714: fix/remove english as default language if not detected
Instead of defaulting the document/element language to English if unable to be detected, remove any default from being applied. 

"auto" will still be the default into the partitioners to trigger language detection appropriately, but the default return value in `detect_languages` in lang.py will be modified, and nothing/None will be added to the `languages` metadata field.

Coniferish:
### Summary
Closes #1714
Changes the default value for `languages` to `None` for elements that don't have text or the language can't be detected.

### Testing
```
elements = partition(filename=filename, detect_language_per_element=True)

# PageBreak elements don't have text and will be collected here
none_langs = [element for element in elements if element.metadata.languages is None]
none_langs[0].text
```


unstructured/partition/lang.py
```diff
@@ -220,7 +220,7 @@ def _convert_to_standard_langcode(lang: str) -> str:
 def detect_languages(
     text: str,
     languages: Optional[List[str]] = ["auto"],
-) -> List[str]:
+) -> List[str] | None:
```
rbiseck3, line 223: I don't think this syntax was supported in python3.8, instead you have to use `Optional[List[str]]`

rbiseck3: (I could be wrong)

rbiseck3 (approved):

unstructured/partition/lang.py
```diff
@@ -234,12 +234,13 @@ def detect_languages(
...
+    # Also return None if there is no text.
+    if languages[0] == "" or text.strip == "":
+        return None
 
-    # Default to "eng" if text is empty or it has only ascii characters and is short
-    if text.strip() == "" or (re.match(r"^[\x00-\x7F]+$", text) and len(text) < 20):
-        return ["eng"]  # english as default
+    # Default to "eng" if text has only ascii characters and is short
+    if re.match(r"^[\x00-\x7F]+$", text) and len(text) < 20:
+        return ["eng"]
```
shreyanid, line 243: Should this default also be None? Since the short pieces of text all being classified as English is what prompted this removal of the default in the first place.

Coniferish: I guess I misunderstood then. It can be, but in that case it should probably just be simplified to returning `None` if the text isn't long enough to reliably detect the language. It'd be worth investigating what that minimum is, though, if there is one for reliably detecting languages with langdetect, but I suspect that also greatly depends on the actual words that are being used in the tests--doesn't seem worthwhile for us to try and come up with those tests ourselves, but if we can find some existing info on it, that'd be nice.

As it's written now, it assumes the following:
 - the vast majority of short text is English
 - it's preferred to have some short text _incorrectly_ identified as English than to have all short text have no language detected

Changing it assumes:
 - users prefer not having language detected more than incorrect language detection
 - users are okay with language not being detected for all elements under a certain length (or we can just get rid of this snippet entirely and solely rely on what langdetect returns, but we've seen it return 3+ languages for short English text)
 
 I'm fine with either decision.

shreyanid: Thanks for outlining the tradeoff. In regards to
> it's preferred to have some short text incorrectly identified as English than to have all short text have no language detected
I'm not sure that that would actually be preferred, because in the case of short non-English text (which is frequently occurring in unstructured element text outputs), labeling it as English will feel glaringly wrong.

> users prefer not having language detected more than incorrect language detection
While we figure out the most accurate way to perform language detection, I'm leaning removing any language being assigned to short elements. Maybe we decrease the character threshold to 10, or switch to a word count of 6?
I agree with you that we shouldn't solely rely on what langdetect returns though

Coniferish: Re slack:
"...I know we often have very short element text in unstructured output, so there may barely be any language detection / the impact of incorrectly identified English text will be much more visible. for ex. if a document is in Korean and every element is 3-5 characters, all of them getting labeled as English will feel really wrong"

I already handled that before with the regex pattern. It matches for ascii characters, so any text that has text other than ascii characters will not default to English.

Coniferish: @qued Do you have any advice/input on this?

Coniferish: To clarify the point about ascii characters, if an element has text that contains special characters (like ñ, å, or Korean/Mandarin/etc.) it will _not_ default to English

shreyanid: Ok, I see what you mean about the ascii character check - if a language uses non-ascii characters (accented letters or just entirely separate charsets), it won't count as English, which removes many of the common languages that could result in short text. So 

> short text incorrectly identified as English

is actually a very rare case? 
In that case, I am fine with leaving this as default English. If we can find some cases of this incorrectly labeling text as English, let's add those test cases for visibility of the problem.

Coniferish: > Maybe we decrease the character threshold to 10, or switch to a word count of 6? I agree with you that we shouldn't solely rely on what langdetect returns though

I think switching to word count instead of characters is a good idea if we keep this snippet

Coniferish: > is actually a very rare case?
> In that case, I am fine with leaving this as default English. If we can find some cases of this incorrectly labeling text as English, let's add those test cases for visibility of the problem.

Yes, should be rather rare and only happen if a snippet of another language only has ascii characters. Ex "Die deutsche Sprache bzw. Deutsch ist eine westgermanische Sprache."

shreyanid (approved):
Changes look great, thank you!

Issue #1461, opened 2023-09-19
docx: Footer elements should appear at section start, not end


scanny:
`partition_docx()` emits `Footer` elements when present but does so at the _end_ of the section they apply to rather than its start.

The semantics of a `Footer` element is "this footer applies to all pages that follow until superceded by a new footer". Accordingly, the footer should show at the _beginning_ of a section, including the first section (page 1).

Move the call to `._iter_section_footers()` from the bottom of the section loop to the top.

scanny:
Closing as will not fix. It's unclear where we want the footer element to appear and it seems more likely it's already where we want it. We can reopen or create a new issue once a clear preference emerges, possibly creating footer elements for each page within the section a footer is defined for.

PR #1747, opened 2023-10-13
fix: #1461 Footer should appear at section start

scanny:
It's an easy intuitive mistake to make, that a header element appears at the beginning of a section and footers appear at the end. However a footer is just like a header in that it applies to all pages that follow it, until superceded by a new footer, if any.

Emit footers at the beginning of the section they're defined in, not at the end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Enhancements
 
-* **Improve natural reading order** Some `OCR` elements with only spaces in the text have full-page width in the bounding box, which causes the `xycut` sorting to not work as expected. Now the logic to parse OCR results removes any elements with only spaces (more than one space).
+* **Improve natural reading order.** Some `OCR` elements with only spaces in the text have full-page width in the bounding box, which causes the `xycut` sorting to not work as expected. Now the logic to parse OCR results removes any elements with only spaces (more than one space).
 * **Ingest compression utilities and fsspec connector support** Generic utility code added to handle files that get pulled from a source connector that are either tar or zip compressed and uncompress them locally. This is then processed using a local source connector. Currently this functionality has been incorporated into the fsspec connector and all those inheriting from it (currently: Azure Blob Storage, Google Cloud Storage, S3, Box, and Dropbox).
 * **Ingest destination connectors support for writing raw list of elements** Along with the default write method used in the ingest pipeline to write the json content associated with the ingest docs, each destination connector can now also write a raw list of elements to the desired downstream location without having an ingest doc associated with it.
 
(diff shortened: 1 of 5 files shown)


cragwolfe:
> It's an easy intuitive mistake to make, that a header element appears at the beginning of a section and footers appear at the end. 

Yes, why is this intuition wrong? As the order of elements returned by partition should reflect "natural reading order," isn't better that a page footers be the last element for a page?

scanny:
@cragwolfe The reason it's mistaken is that footers are emitted (at most) "once-per-section", not "once-per-page". Otherwise, in a common case where you had say a seven page document with one (docx) section, the footer would first appear as the last element in the sequence.

The "at-most-once" bit is because a docx header/footer has an ".is_linked_to_previous"  attribute which when set means "don't change the header/footer, use the same one as last section".

We _could_ consider adding repeated header/footer elements before and after every page respectively, but that's not how it's been so far. This would be non-trivial, but certainly possible, depending on how far we wanted to go with applying first-page, even-page, odd-page headers in the right places and possibly evaluating fields like page-numbers that are computed by Word at "print" time.

The current semantic is "Here is a new header/footer element. This applies to each page going forward until it is superseded by a new one of the same type." with type meaning first-page, even-page, primary, and header or footer, so six independent types in all.

cragwolfe:
> Otherwise, in a common case where you had say a seven page document with one (docx) section, the footer would first appear as the last element in the sequence.

Is that not a more natural reflection of reading order?

scanny:
Well, I suppose intelligent folks could disagree, but if the footer first appeared on page 1 and I didn't hear about it until page 7, that seems late to me :)

I should probably note here that when we say "section" in a DOCX context it means like a "chapter", not the same kind of "section" (Title + a few NarrativeText elements) we talk about in a chunking context. So a docx section can easily be 20 pages or 100 even.

The way it is (and has been) that a header or footer is emitted once per section, it's a downstream responsibility to place that text where it would appear on each page, if that's desired (no idea if or when it is).

The mechanism for this would be something like:

1. keep running track of the current header(s) and footer(s), changing them out as new ones appear in the stream.
2. detect page breaks (page starts and ends really, but derived from page-breaks).
3. When a page break occurs
    a. insert the current applicable footer (first-page, even-page, primary)
    b. insert the page-break (if we do that)
    c. insert the currently applicable header, before ("at the top of") the new page that starts with the next element.

If you get them both up front, as this PR provides, you can easily do step 1. If you don't get it until the end of the section then you would have to "back-up" in the element stream and insert (maybe FooterText) elements in the right places.

It seems clear that I would want them in advance so I can pop their values in while I'm processing the page-break, rather than add a dedicated post-processing step to put those in after. And if we're actually streaming the elements they would be gone so back-filling them wouldn't even be an option.

As I mentioned earlier, we could consider inserting header and footer text "in-stream" on each page-break boundary, but that hasn't been done so far.

scanny:
Looks like the real solution to this problem is that what we're emitting is header/footer _definitions_ where what we would like to be emitting is header/footer _renderings_. A footer definition happens once or a few times per document whereas a footer rendering would happen once per page. Accomplishing a rendering is certainly a possibility, but is not addressed by this PR and looks like it's not in the cards at the moment.

Closing this PR for now. It's only a few lines of work and we can always reopen it later if need be but I think we'd just start from scratch to emit header/footer renderings.

PR #1753, opened 2023-10-13
fix: zero divide in under_non_alpha_ratio

qued:
The function `under_non_alpha_ratio` in `unstructured.partition.text_type` was producing a divide-by-zero error. After investigation I found this was a possibility when the function was passed a string of all spaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
-* **Fixes `under_non_alpha_ratio` dividing by zero** Although this function guarded against a specific cause of division by zero, there were edge cases slipping through like strings tih only whitespace. This update more generally prevents the function from performing a division by zero.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Fixes
 
+* **Fixes `under_non_alpha_ratio` dividing by zero** Although this function guarded against a specific cause of division by zero, there were edge cases slipping through like strings tih only whitespace. This update more generally prevents the function from performing a division by zero.
```
cragwolfe, line 11:
```suggestion
* **Fixes `under_non_alpha_ratio` dividing by zero** Although this function guarded against a specific cause of division by zero, there were edge cases slipping through like strings with only whitespace. This update more generally prevents the function from performing a division by zero.
```

cragwolfe (approved):

PR #1761, opened 2023-10-14
fix: use nx to avoid recursion limit

qued:
Fixes recursion limit error that was being raised when partitioning Excel documents of a certain size.

Previously we used a recursive method to find subtables within an excel sheet. However this would run afoul of Python's recursion depth limit when there was a contiguous block of more than 1000 cells within a sheet. This function has been updated to use the NetworkX library which avoids Python recursion issues.

* Updated `_get_connected_components` to use `networkx` graph methods rather than implementing our own algorithm for finding contiguous groups of cells within a sheet.
* Added a test and example doc that replicates the `RecursionError` prior to the change.
*  Added `networkx` to `extra_xlsx` dependencies and `pip-compile`d.

#### Testing:
The following run from a Python terminal should raise a `RecursionError` on `main` and succeed on this branch:
```python
import sys
from unstructured.partition.xlsx import partition_xlsx
old_recursion_limit = sys.getrecursionlimit()
try:
    sys.setrecursionlimit(1000)
    filename = "example-docs/more-than-1k-cells.xlsx"
    partition_xlsx(filename=filename)
finally:
    sys.setrecursionlimit(old_recursion_limit)

```
Note: the recursion limit is different in different contexts. Checking my own system, the default in a notebook seems to be 3000, but in a terminal it's 1000. The documented Python default recursion limit is 1000.


unstructured/partition/xlsx.py
```diff
@@ -201,44 +202,26 @@ def _get_connected_components(
         overlapping components to return distinct components.
     """
     max_row, max_col = sheet.shape
-    visited = set()
+    graph: nx.Graph = nx.grid_2d_graph(max_row, max_col)
+    node_array = np.indices((max_row, max_col)).T
+    empty_cells = sheet.isna().T
+    nodes_to_remove = [tuple(pair) for pair in node_array[empty_cells]]
+    graph.remove_nodes_from(nodes_to_remove)
+    connected_components_as_nodes = list(nx.connected_components(graph))
```
badGarnet, line 210: do we need to convert it into list? seems we are just iterating over this once. So maybe keep it as generator to save some memory?

qued: Dropped the conversion to list in [1f43edd](https://github.com/Unstructured-IO/unstructured/pull/1761/commits/1f43eddf680cd5ee08b8b347ed93d4a5fc03dd36).

badGarnet (approved):
main can run the test code; but with that using library code is preferred here and it is more readable

badGarnet:
confirming 1000 limit triggers recursion depth error

PR #1765, opened 2023-10-16
feat: utils method to draw bboxes on pdf and images

LaverdeS:
This PR adds functionality to `unstructured` utils to draw bounding boxes on page images. The new method `draw_bboxes_on_pdf_or_image` takes a `file_path` (PDF or image) and a list of the extracted elements as input , and draws the element bounding boxes superimposed in the original document image (separated by pages and colours for the different element types). Is initially intended to be used for visual inspection of overlapping bounding boxes cases and fulfils similar functions as `page.annotate()` from `unstructured-inference`.

To test it, the following snippet can be used:
```
from unstructured.utils import draw_bboxes_on_pdf_or_image

model_name = "yolox_quantized"
target_file_path = "/sample-docs/layout-parser-paper-fast.pdf"
elements = partition(filename=target_file_path, strategy='hi_res', model_name=model_name)
draw_bboxes_on_pdf_or_image(target_file_path, elements, save_images=True, save_coords=True, plot=True)
```


pravin-unstructured (commented):
This makes sense. For larger PDFS I presume this takes a long time?

LaverdeS:
> This makes sense. For larger PDFS I presume this takes a long time?

@pravin-unstructured 
For `layout-parser-paper.pdf` which has 16 pages in `example-docs`, it takes:
```
CPU times: user 20.5 s, sys: 2.78 s, total: 23.3 s
Wall time: 13.3 s
```
For `layout-parser-paper-fast.pdf` which has 2 pages in `example-docs`, it takes:
```
CPU times: user 2.57 s, sys: 2.43 s, total: 5 s
Wall time: 1.67 s
```
For `layout-parser-paper-fast-1p.jpg` which has 1 page (and no need of pdf2image conversion) in `example-docs`, it takes:
```
CPU times: user 1.34 s, sys: 2.39 s, total: 3.73 s
Wall time: 920 ms
```

christinestraub:
This method serves the same purpose as [visualization.py](https://github.com/Unstructured-IO/unstructured/blob/main/examples/layout-analysis/visualization.py), right?

LaverdeS:
> @LaverdeS This method serves the same purpose as [visualization.py](https://github.com/Unstructured-IO/unstructured/blob/main/examples/layout-analysis/visualization.py), right?

@christinestraub oh yeah, just that `visualization.py` and `/custom-layout-order` are part of `examples/`. But they serve the same purpose. I created this PR for having it easy for visual inspection with colours and background to detect overlapping bounding boxes in complex document pages with a lot of bounding boxes, or to quickly inspect what element types are presenting what types of errors in the bbox prediction, for instance. There is a similar method in `unstructured-inference` for pages `page.annotate()` but we wanted one integrated in `unstructured`.

unstructured/utils.py
```diff
@@ -21,13 +22,33 @@
...
+from matplotlib import colors, patches
+from pdf2image import convert_from_path
+from PIL import Image
 from typing_extensions import ParamSpec
 
 from unstructured.__version__ import __version__
 
 DATE_FORMATS = ("%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d+%H:%M:%S", "%Y-%m-%dT%H:%M:%S%z")
-
+TYPE_TO_COLOUR_MAP = {
```
benjats07, line 36: I remember some of these colors are difficult to see/read (like cyan), would like to collapse all these colors into a list of "eye-popping" colors. In the end this is for understanding an output, not exactly to identify classes. For that, I would add a string at the start of the text inside the bbox.

benjats07 (changes requested):

unstructured/utils.py
LaverdeS: Oh I tried to pick the most eye-popping colours without repetition. I changed `cyan` for `indigo`. You can always set `TYPE_TO_COLOUR_MAP` to diff values and see how it gets. There is a string with the label and element-index also being drawn per element. Is this what you mean? let me know if you consider there are more necessary changes to finally merge this PR.

unstructured/utils.py
```diff
@@ -280,3 +301,105 @@ def scarf_analytics():
...
+
+def draw_bboxes_on_pdf_or_image(
+    file_path,
+    elements,
+    desired_width=20,
+    save_images=False,
+    save_coordinates=False,
+    output_folder=None,
+    plot=False,
+):
```
qued, line 314: Some general thoughts:
1. We definitely should not have to pick up `matplotlib` as a base level dependency. With that in mind this should probably be moved somewhere else like a `visualize.py` file, and the import should be moved inside the function. The tests should live in their own file as well.
1. I think this function does too much. At the very least, the logic of plotting should have its own function with inputs only an object representing a single image and an object representing a single bounding box, and output should be an object representing a single image on which the box has been plotted. The business of loading and saving the image, resizing the image, and plotting multiple boxes across multiple pages are separate concerns. Then we can have a function of plotting multiple boxes to a single page/image, that takes the minimum necessary parameters and calls the single image/box function. It's okay to have a one-stop-shop function like this, but if I already have an image in memory that I want to annotate I shouldn't have to write a new function.
1. We need the normal code quality perks like type hints.
1. I prefer the pattern of plotting directly to the `PIL` `Image` and returning that object rather than using a heavier dependency like `matplotlib`. I could potentially live with `matplotlib`, but this function seems to be written around using a notebook with calls like `plt.show()`, where using only `PIL` is notebook friendly while also usable in e.g. a script. In any case it's nicer to return an object with a notebook-friendly repr method than to have a `plot` parameter. `PIL.Image.Image` satisfies this, and so does `matplotlib.figure.Figure` IIRC.
1. Couldn't we drop the `save_images` parameter and assume anything other than `None` in `output_folder` indicates the user wants to save the images? I realize using both parameters is a pattern we've used elsewhere but it doesn't seem ideal.
1. I'm not sure `save_coordinates` belongs here. Probably it was relevant to your use case, so I'm open minded, but we have existing functions that output JSON with coordinates, and my feeling is that it would be better to incorporate those with this or vice-versa.
1. I'm not sure about the resizing functionality here either. Again, you were probably using this, but it's like we're opening the door to a function that supports full fledged image formatting. If we output an image or figure, then all that functionality is inherent in the output object itself, which seems good enough to me.

ajjimeno:
The function `draw_bboxes_on_pdf_or_image` should take into account the dpis used when processing the document, since this affects the dimensions of the generated image. If the dimensions do not match the ones used by the `unstructured` process that generated the elements with their bounding boxes, the bounding boxes will be off.

krutko-ds:
@qued This function would be useful for me as well, is it possible to merge without meeting all the code quality expectations?

PR #1771, opened 2023-10-17
feat(chunking): add overlap

scanny:
Had to fix a few bugs first, which accounts for most of the code changes. Reads best commit-by-commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 
 ### Enhancements
 
+* **Add overlap option for split chunks.** Users can supply an `overlap: int` value on a partition function to start continuation chunks the specified number of characters "early", such that the last n characters of the prior chunk also appear as the first n characters of the following chunk, to improve semantic context.
+
 ### Features
 
 ### Fixes
+
+* **Fix out-of-order sequencing of split chunks.** Fixes behavior where "split" chunks were inserted at the beginning of the chunk sequence. This would produce a chunk sequence like [5a, 5b, 3a, 3b, 1, 2, 4] when sections 3 and 5 exceeded `max_characters`.
+* **Fix breaking chunks on regex-metadata matches.** Fixes "over-chunking" when `regex_metadata` was used, where every element that contained a regex-match would start a new chunk.
 
 ## 0.10.24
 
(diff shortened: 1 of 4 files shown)


unstructured/chunking/title.py
```diff
@@ -287,6 +310,7 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> List[Element]:
                     combine_text_under_n_chars=params.get("combine_text_under_n_chars", 500),
                     new_after_n_chars=params.get("new_after_n_chars", 500),
                     max_characters=params.get("max_characters", 500),
+                    overlap=params.get("overlap", 0),
```
amanda103, line 313: Just trying to think about the way this behaves with the other inputs:
1. what happens if overalp == max characters?
2. overlap > max characters - should error before going into chunking

Any other things we should be handling?

scanny: Yes, that's a good point. I can add a validation step for that.

CHANGELOG.md
```diff
@@ -2,10 +2,15 @@
 
 ### Enhancements
 
+* **Add overlap option for split chunks.** Users can supply an `overlap: int` value on a partition function to start continuation chunks the specified number of characters "early", such that the last n characters of the prior chunk also appear as the first n characters of the following chunk, to improve semantic context.
+
 ### Features
 
 ### Fixes
 
+* **Fix out-of-order sequencing of split chunks.
```
cragwolfe, line 11:
```suggestion
* **Fix out-of-order sequencing of split chunks.**
```

CHANGELOG.md
```diff
@@ -2,10 +2,15 @@
...
+* **Fix breaking chunks on regex-metadata matches.
```
cragwolfe, line 12:
```suggestion
* **Fix breaking chunks on regex-metadata matches.**
```
Please say more. what does this mean, why does it matter?

CHANGELOG.md
scanny: k, fixed

CHANGELOG.md
scanny: k, fixed

adanielyan:
Any idea when this will be merged to main?

adanielyan:
Appreciate a quick response, @scanny!

My use case is related to populating a vector database with embeddings extracted from PDF files.

Issue #1414, opened 2023-09-14
feat: create downstream vectordb (pinecone) connector


rbiseck3:
Create downstream connector for Pinecone vector database. This will depend on [#1319](https://github.com/Unstructured-IO/unstructured/issues/1319) for full completeness as embeddings are required to add the content into a vector db. Does not seem that Pinecone has a supported docker approach for local testing so will have to use a Pinecone instance with an associated api key to test against.

PR #1774, opened 2023-10-17
feat: add pinecone destination connector

### LINKED ISSUE #2039: bug/ingest-chunking-parameters-not-passed
**Describe the bug**
When ingest is being run with chunking parameters, the parameters except `--chunk-elements ` are not successfully passed to the Chunking node, resulting in Chunking node acting in unexpected ways.

**To Reproduce**
Clone https://github.com/Unstructured-IO/unstructured/tree/ahmet/pinecone-connector, run `/test_unstructured_ingest/dest/pinecone.sh` with these modifications:
```
  local \
  ...
  --chunk-elements \
  --chunk-max-characters 10 \
  ...
  pinecone \
  ...
```

**Expected behavior**
- Same number of vectors will be upserted each time the test is run, no matter the value of `--chunk-max-characters` parameter. 
- When the CliChunkingConfig is logged, values will be `CliChunkingConfig(chunk_elements=True, multipage_sections=True, combine_text_under_n_chars=500, max_characters=1500)` no matter what the provided chunking parameters are.

**Environment Info**
`make install` + `pip install pinecone-client`

ahmetmeleq:
Closes https://github.com/Unstructured-IO/unstructured/issues/1414
Closes #2039 

This PR:
- Uses Pinecone python cli to implement a destination connector for Pinecone and provides the ingest readme  requirements [(here)](https://github.com/Unstructured-IO/unstructured/tree/main/unstructured/ingest#the-checklist) for the connector
- Updates documentation for the s3 destination connector
- Alphabetically sorts setup.py contents
- Updates logs for the chunking node  in ingest pipeline
- Adds a baseline session handle implementation for destination connectors, to be able to parallelize their operations
- For the [bug](https://github.com/Unstructured-IO/unstructured/issues/1892) related to persisting element data to ingest embedding nodes; this PR tests the [solution](https://github.com/Unstructured-IO/unstructured/pull/1893) with its ingest test
- Solves a bug on ingest chunking params with [bugfix on chunking params and implementing related test](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/69e1949a6fa806d14014eb68074929595f92360e)


ahmetmeleq:
to do: update the code with clean index / create index functions in the branch, make the test runnable

docs/source/destination_connectors/pinecone.rst
```diff
@@ -0,0 +1,74 @@
+S3
```
rbiseck3, line 1: title is incorrect (still s3)

setup.py
```diff
@@ -127,39 +127,40 @@ def load_requirements(file_list: Optional[Union[str, List[str]]] = None) -> List
         "tsv": tsv_reqs,
         "xlsx": xlsx_reqs,
         # Extra requirements for data connectors
-        "s3": load_requirements("requirements/ingest/s3.in"),
```
rbiseck3, line 130: Don't want to revert all of these.

test_unstructured_ingest/test-ingest-pinecone.sh
```diff
@@ -0,0 +1,120 @@
...
+
+if [ "$response_code" -lt 400 ]; then
+  echo "Index creation success: $response_code"
+else
+  echo "Index creation failure: $response_code"
+  exit 1
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+  s3 \
```
rbiseck3, line 89: To avoid network calls, I've been using the local source connector to generate the content to push up. Take a look at the tests I created in this PR: [1777](https://github.com/Unstructured-IO/unstructured/pull/1777)

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,126 @@
...
+    ConfigSessionHandleMixin,
+    WriteConfig,
+    WriteConfigSessionHandleMixin,
+)
+from unstructured.ingest.logger import logger
+from unstructured.utils import requires_dependencies
+
+
+@requires_dependencies(["pinecone"], extras="pinecone")
+def create_pinecone_object(api_key, index_name, environment):
```
rbiseck3, line 21: Would make sense for this to be a method of the `PineconeWriteConfig` class. Just an idea.

unstructured/ingest/pipeline/interfaces.py
```diff
@@ -71,6 +71,8 @@ def __call__(self, iterable: t.Optional[t.Iterable[t.Any]] = None) -> t.Any:
             else:
                 self.result = self.run()
         else:
+            # Debugging tip: You can comment out the "with mp.Pool" block, and use the line below:
+            # self.result = [self.run(item) for item in iterable]
```
rbiseck3, line 75: This isn't needed if you set the number of processes to 1.

unstructured/ingest/interfaces.py
```diff
@@ -653,3 +653,20 @@ def session_handle(self):
...
+
+class WriteConfigSessionHandleMixin(ConfigSessionHandleMixin):
+    _session_handle: t.Optional[BaseSessionHandle] = None
+
+    @property
+    def session_handle(self):
+        """If a session handle is not assigned, creates a new one and assigns it."""
+        if self._session_handle is None:
+            self._session_handle = self.create_session_handle()
+            print("CREATE123: SESSION HANDLE")
```
rbiseck3, line 666: Will want to get rid of these print statements.

setup.py
ahmetmeleq: Didn't understand, this is alphabetical reordering. Didn't intend to revert / reorder, is anything wrong in the current state?

unstructured/ingest/interfaces.py
ahmetmeleq: Agreed, addressed with [doc, comment and logging updates](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/6d84efc018bb5a2d466c06caa56705e9531d2af7)

unstructured/ingest/pipeline/interfaces.py
ahmetmeleq: Agreed, addressed with [doc, comment and logging updates](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/6d84efc018bb5a2d466c06caa56705e9531d2af7)

docs/source/destination_connectors/pinecone.rst
ahmetmeleq: Agreed, addressed with [doc, comment and logging updates](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/6d84efc018bb5a2d466c06caa56705e9531d2af7)

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Agreed, addressed with [test and session creation updates](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/84e65e5389f05c57508311755c837d8d0b0815dc)

test_unstructured_ingest/test-ingest-pinecone.sh
ahmetmeleq: Agreed, addressed with [test and session creation updates](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/84e65e5389f05c57508311755c837d8d0b0815dc)

setup.py
rbiseck3: It's changing the path back: `requirements/ingest/azure.in` -> `requirements/ingest-azure.in`

test_unstructured_ingest/test-ingest-pinecone.sh
```diff
@@ -0,0 +1,119 @@
...
+  echo "Index creation success: $response_code"
+else
+  echo "Index creation failure: $response_code"
+  exit 1
+fi
+
+PYTHONPATH=. ./unstructured/ingest/main.py \
+  local \
+  --num-processes "$max_processes" \
+  --metadata-exclude coordinates,filename,file_directory,metadata.data_source.date_processed,metadata.last_modified,metadata.detection_class_prob,metadata.parent_id,metadata.category_depth \
```
rbiseck3, line 92: For destination ingest tests, we don't want to omit any of the metadata so it get's populated in the schema if one is set. We're not checking the content in destination tests so this shouldn't affect that.

test_unstructured_ingest/test-ingest-pinecone.sh
```diff
@@ -0,0 +1,119 @@
...
+  --work-dir "$WORK_DIR" \
+  --chunk-elements \
+  --chunk-multipage-sections \
+  --embedding-api-key "$OPENAI_API_KEY" \
+  pinecone \
+  --api-key "$PINECONE_API_KEY" \
+  --index-name "$PINECONE_INDEX" \
+  --environment "$PINECONE_ENVIRONMENT"
+
+sleep 15
```
rbiseck3, line 107: Is there a more deterministic way to check that the content gets written other than using a single 15 second sleep? Take a look at the azure cognitive search test.

test_unstructured_ingest/test-ingest-pinecone.sh
ahmetmeleq: Agreed, addressed with [do not exclude metadata](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/cf7731584d3074e44ed670956415828b54a85d29)

setup.py
ahmetmeleq: Ah I see, that slipped my eyes

Addressed with: [fix path typos on setup.py](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/025776976a0458dd007734de7ef26b021436b816)

test_unstructured_ingest/test-ingest-pinecone.sh
```diff
@@ -0,0 +1,131 @@
...
+
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+cd "$SCRIPT_DIR"/.. || exit 1
+OUTPUT_FOLDER_NAME=s3-pinecone-dest
+OUTPUT_DIR=$SCRIPT_DIR/structured-output/$OUTPUT_FOLDER_NAME
+WORK_DIR=$SCRIPT_DIR/workdir/$OUTPUT_FOLDER_NAME
+DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+
+if [ -z "$OPENAI_API_KEY" ] && [ -z "$PINECONE_API_KEY" ]; then
```
ryannikolaidis, line 13: why openai? did you prove that we could use huggingface over [here](https://github.com/Unstructured-IO/unstructured/pull/1918)?

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,125 @@
...
+    WriteConfig,
+    WriteConfigSessionHandleMixin,
+)
+from unstructured.ingest.logger import logger
+from unstructured.utils import requires_dependencies
+
+
+@dataclass
+class PineconeSessionHandle(BaseSessionHandle):
+    service: "pinecone.Index"  # noqa: F821
```
ryannikolaidis, line 22: sweet, thanks for pushing this.

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,125 @@
...
+    def write_dict(self, *args, dict_list: t.List[t.Dict[str, t.Any]], **kwargs) -> None:
+        logger.info(
+            f"Inserting / updating {len(dict_list)} documents to destination "
+            f"index at {self.connector_config.index_name}",
+        )
+
+        # this is advised to be 100 at maximum in pinecone docs, however when we
+        # chunk content, we hit to the object size limits, so we decrease the batch
+        # size even more here
+        pinecone_batch_size = 10
```
ryannikolaidis, line 85: does this need to scale with chunk size?

unstructured/ingest/pipeline/reformat/chunking.py
```diff
@@ -45,13 +45,13 @@ def run(self, elements_json: str) -> Optional[str]:
...
             ):
-                logger.debug(f"File exists: {json_path}, skipping embedding")
+                logger.debug(f"File exists: {json_path}, skipping chunking")
                 return str(json_path)
             elements = elements_from_json(filename=elements_json)
             chunked_elements = self.chunking_config.chunk(elements=elements)
             elements_dict = convert_to_dict(chunked_elements)
             with open(json_path, "w", encoding="utf8") as output_f:
-                logger.info(f"writing embeddings content to {json_path}")
+                logger.info(f"writing chunking content to {json_path}")
```
ryannikolaidis, line 54: good catch

test_unstructured_ingest/test-ingest-pinecone.sh
ahmetmeleq: After that one goes in, we can use it 👍 (it adds support for other embedding modules) 

I'll update accordingly

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Has an inverse relationship with anything that increases the amount of data in an element, including chunk_size

So technically we could break this if we had huge documents with huge chunk size params

But even before breaking this part, chunk size might break embeddings, since they might have fixed context windows (not sure, provider organization might have handled it as well).

ryannikolaidis: sorry, yea, get the inverse relationship, just meant that the fixed value seems suspicious. do you know the limits at which this falls apart? as long as we can handle reasonable windows, probably good enough for now and we can stub a ticket for more investigation later

test_unstructured_ingest/test-ingest-pinecone.sh
potter-potter: This will be nice!

unstructured/ingest/interfaces.py
```diff
@@ -670,3 +670,18 @@ def session_handle(self):
     @session_handle.setter
     def session_handle(self, session_handle: BaseSessionHandle):
         self._session_handle = session_handle
+
+
+class WriteConfigSessionHandleMixin(ConfigSessionHandleMixin):
```
potter-potter, line 675: This is very similar to the dataclass above. For my understanding, why did you not choose dataclass here?

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,132 @@
...
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+
+if [ -z "$OPENAI_API_KEY" ] && [ -z "$PINECONE_API_KEY" ]; then
+   echo "Skipping Pinecone ingest test because neither OPENAI_API_KEY nor PINECONE_API_KEY env vars are set."
+   exit 0
+fi
+
+PINECONE_ENVIRONMENT="gcp-starter"
+PINECONE_INDEX="ingest-test"
+PINECONE_PROJECT_ID="bfa06d5"
```
potter-potter, line 21: Are you able to run this multiple times with this id? My id shifted the second time I ran it. May have to be a random id?

potter-potter: I think that url is unique to your user. But I'm guessing it won't be able to use that for CI. Mine is consistently `ingest-test-ec7b203.svc.gcp-starter.pinecone.io`

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,125 @@
...
+            with open(local_path) as json_file:
+                dict_content = json.load(json_file)
+
+                # assign element_id and embeddings to "id" and "values"
+                # assign everything else to "metadata" field
+                dict_content = [
+                    {
+                        "id": element.pop("element_id", None),
+                        "values": element.pop("embeddings", None),
+                        "metadata": {k: json.dumps(v) for k, v in element.items()},
```
potter-potter, line 117: This last part creates a big long string of metadata. Are we sure thats what we want to do?

Looks like this:

test_unstructured_ingest/test-ingest-pinecone.sh
ahmetmeleq: Addressed with [ingest test uses huggingface embedder](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/791bf03d00cfca5a0b6e74066b86f5af7c90ad4e) 👍

unstructured/ingest/interfaces.py
ahmetmeleq: image

It goes as: 
- step 1

- step 2
  - We said we won't implement per-doc session handlers, but then we will be discarding `IngestDocSessionHandleMixin`s since they are supposed to be used in per-doc operations. And `IngestDocSessionHandleMixin`s have a role of keeping sessions as properties, getting them and setting them. Now we need something else to take that role from `IngestDocSessionHandleMixin`s, and provide it for our batches


Part 1:
  - We already needed to implement a `ConfigSessionHandle` class, which'll implement a `create_session_handle` method via `ConfigSessionHandleMixin`. So we already implemented that part in [here](https://github.com/Unstructured-IO/unstructured/blob/791bf03d00cfca5a0b6e74066b86f5af7c90ad4e/unstructured/ingest/connector/pinecone.py#L43) .
  - The remaining part was having something to keep session handles as a property, and to get & set that property. Normally, IngestDocs do that.

Part 2:
  - Now, in source connectors we perform session related operations per doc (download each doc) , when in destination connectors we perform session related operations per batch (upload each batch).
  - So I didn't want to just choose a random ingest doc within the batch and create a session handle via that random doc. I wanted the design to be obvious that our session handlers are created per batch, rather than a specific doc. That meant we need something else than IngestDocs to get, set and keep our session_handle property.
  - So I thought the `ConfigSessionHandle` class which we've already implemented can keep our session_handle property. No need to create another class.

Part 3:
- Therefore `PineconeWriteConfig` should have inherited both from `ConfigSessionHandleMixin` and `IngestDocSessionHandleMixin`. But `IngestDocSessionHandleMixin` [expects a connector_config](https://github.com/Unstructured-IO/unstructured/blob/f434c3d79faaba78abc4d1bb3d0b7921c3cd70d4/unstructured/ingest/interfaces.py#L662) while PineconeWriteConfig is itself a config! (this would be self reference)
- So I created a new class where the only difference is that create_session_handle is called from the object itself, rather than its connector_config. This way we acknowledge that PineconeWriteConfig (and potentially other destination WriteConfigs in the future) will take the role of both an ingest doc, and a connector config when managing session handles

ahmetmeleq: image

It goes as: 

Part 1:
  - We already needed to implement a `ConfigSessionHandle` class, which'll implement a `create_session_handle` method via `ConfigSessionHandleMixin`. So we already implemented that part in [here](https://github.com/Unstructured-IO/unstructured/blob/791bf03d00cfca5a0b6e74066b86f5af7c90ad4e/unstructured/ingest/connector/pinecone.py#L43) .
  - The remaining part was having something to keep session handles as a property, and to get & set that property. Normally, IngestDocs do that.

Part 2:
  - Now, in source connectors we perform session related operations per doc (download each doc) , when in destination connectors we perform session related operations per batch (upload each batch).
  - So I didn't want to just choose a random ingest doc within the batch and create a session handle via that random doc. I wanted the design to be obvious that our session handlers are created per batch, rather than a specific doc. That meant we need something else than IngestDocs to get, set and keep our session_handle property.
  - So I thought the `ConfigSessionHandle` class which we've already implemented can keep our session_handle property. No need to create another class.

Part 3:
- Therefore `PineconeWriteConfig` should have inherited both from `ConfigSessionHandleMixin` and `IngestDocSessionHandleMixin`. But `IngestDocSessionHandleMixin` [expects a connector_config](https://github.com/Unstructured-IO/unstructured/blob/f434c3d79faaba78abc4d1bb3d0b7921c3cd70d4/unstructured/ingest/interfaces.py#L662) while PineconeWriteConfig is itself a config! (this would be self reference)
- So I created a new class where the only difference is that create_session_handle is called from the object itself, rather than its connector_config. This way we acknowledge that PineconeWriteConfig (and potentially other destination WriteConfigs in the future) will take the role of both an ingest doc, and a connector config when managing session handles

unstructured/ingest/connector/pinecone.py
ahmetmeleq: ## Some research on the limits

1. For each vector, the maximum metadata size allowed is 40,960 bytes [1]. This per-vector limitation makes chunks bigger than 40,000 characters fail when doing upserts. While that is the case, 38,000 character chunks had successful upserts. (when using the example file book-war-and-peace-1225p.txt) _So the maximum eligible chunk_size was between 38,000 and 40,000_ for the embedding model that I was using (langchain-huggingface). When the embedding model is changed to have higher vector dimensions, we might come up to another limitation ie per-vector total object size limitation (regardless of metadata).

2. As for the allowed message length, the allowed maximum is 4,194,304 bytes [2]. That makes ~ x100 for the maximum vector metadata size allowed. Also for the request size, the allowed size is 2MBs [3]. These two combined limitations _allowed for 85 sized batches while not allowing for 90_. That is a bit lower than the recommended batch size (100) in [pinecone docs](https://docs.pinecone.io/reference/upsert).

> [1]: HTTP response body: {"code":3,"message":"Metadata size is ... bytes, which exceeds the limit of 40960 bytes per vector","details":[]}
> [2]: HTTP response body: {"code":11,"message":"Error, message length too large: found ... bytes, the limit is: 4194304 bytes","details":[]}
> [3]: HTTP response body: {"code":3,"message":"Request size ... exceeds the maximum supported size of 2MB","details":[]}

## Resulting Changes

I think allowing `--batch-size` as a cli param still makes sense, so I've added it with [make batch size a parameter](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/766d4853daa708a052a4aa7bc1f6e433da49c3df) and I've set the default to 50.

So now we have batch size as a cli param, with the default value set to 50, and at least with `langchain-huggingface` embedder and `book-war-and-peace-1225p.txt` , it can support up to 85 as a value.

ahmetmeleq:
https://github.com/Unstructured-IO/unstructured/pull/1774#discussion_r1382616821
@potter-potter 
Yes, I'm using the CI user while developing as well - didn't see any problems in CI for now ([logs for reference](https://github.com/Unstructured-IO/unstructured/actions/runs/6788194435/job/18452715769#step:11:1425))

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Agreed, addressed with [pass metadata fields individually](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/f2786ce167d91a69f69f28e50f8a2a1f1e0b146c)

unstructured/ingest/interfaces.py
potter-potter: Thanks for this detailed explanation.

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,128 @@
...
+            f"Inserting / updating {len(dict_list)} documents to destination "
+            f"index at {self.connector_config.index_name}",
+        )
+
+        # this is advised to be 100 at maximum in pinecone docs, however when we
+        # chunk content, we hit to the object size limits, so we decrease the batch
+        # size even more here
+        pinecone_batch_size = self.write_config.batch_size
+
+        num_processes = 1
```
potter-potter, line 88: Was this set here for development? Or is there a problem with multi-processing?

docs/source/ingest/destination_connectors/pinecone.rst
```diff
@@ -0,0 +1,74 @@
...
+        command = [
+          "unstructured-ingest",
+          "s3",
+          "--remote-url", "s3://utic-dev-tech-fixtures/small-pdf-set/",
+          "--anonymous",
+          "--output-dir", "s3-small-batch-output \",
+          "--num-processes", "2",
+          "--verbose",
+          "--strategy", "fast",
+          "--embedding-api-key", "<your openai api key here>",
```
ryannikolaidis, line 52: not just openai
..also, shouldn't you also need to specify --embedding-provider?

unstructured/ingest/cli/interfaces.py
```diff
@@ -468,13 +468,13 @@ def get_cli_options() -> t.List[click.Option]:
                 default=False,
             ),
             click.Option(
-                ["--chunk-combine-under-n-chars"],
+                ["--chunk-combine-text-under-n-chars"],
```
ryannikolaidis, line 471: please update chunking.rst

unstructured/ingest/cli/interfaces.py
```diff
@@ -468,13 +468,13 @@ def get_cli_options() -> t.List[click.Option]:
...
                 type=int,
                 default=500,
                 show_default=True,
             ),
             click.Option(
-                ["--chunk-new-after-n-chars"],
+                ["--chunk-max-characters"],
```
ryannikolaidis, line 477: i think we need to support `max_characters` in addition to `new_after_n_chars`. i.e. we should keep new-after and add max-characters 

note, these are not the same thing:


    new_after_n_chars
        Cuts off new sections once they reach a length of n characters (soft max). Defaults to
        `max_characters` when not specified, which effectively disables any soft window.
        Specifying 0 for this argument causes each element to appear in a chunk by itself (although
        an element with text longer than `max_characters` will be still be split into two or more
        chunks).
    max_characters
        Chunks elements text and text_as_html (if present) into chunks of length
        n characters (hard max)

unstructured/ingest/connector/pinecone.py
ryannikolaidis: +1 👆

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,135 @@
...
+DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+
+if [ -z "$PINECONE_API_KEY" ]; then
+   echo "Skipping Pinecone ingest test because PINECONE_API_KEY env var is not set."
+   exit 0
+fi
+
+PINECONE_ENVIRONMENT="gcp-starter"
+PINECONE_INDEX="ingest-test"
```
ryannikolaidis, line 20: does this mean every test running is writing to the same index? if so: this will very likely result in collisions when multiple run at the same time?

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,128 @@
...
+                processes=num_processes,
+            ) as pool:
+                pool.map(
+                    self.write_config.upsert_batch,
+                    [
+                        dict_list[i : i + pinecone_batch_size]  # noqa: E203
+                        for i in range(0, len(dict_list), pinecone_batch_size)
+                    ],  # noqa: E203
+                )
+
```
ryannikolaidis, line 104: I don't think this is going to actually leverage the session as you're intending.

Note: [this logic](https://github.com/Unstructured-IO/unstructured/blob/4b49fbd27f772a02efbc846d9095f2086fedbd11/unstructured/ingest/pipeline/source.py#L9-L25) is what allows the session to persist across a subprocess, by explicitly assigning the session on the doc from the global variable.

ryannikolaidis: have you actually tested this? make sure that num_processes > number of batches and confirm that only num_processes number of sessions are created (can log from the create session method)

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Was for development indeed. Now, I've exposed an additional `num_processes` for the writer, and have set the test to do parallel uploads: 
[expose writer num_processes, apply parallelization in ingest test](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/1ff1fd665ce0499be71b91149927f39288d7f17d)

Here are some logs in local (processing 3 files here) for reference. Batch with the smaller ("remainder") amount being uploaded earlier means that the batches are not being processed in order; pointing to successful parallelization:
```
2023-11-09 13:06:32,485 MainProcess INFO     appending 558 json elements from content in .../unstructured/test_unstructured_ingest/structured-output/s3-pinecone-dest/book-war-and-peace-1225p.txt.json
2023-11-09 13:06:32,503 MainProcess INFO     appending 174 json elements from content in .../unstructured/test_unstructured_ingest/structured-output/s3-pinecone-dest/example-10k-230p.html.json
2023-11-09 13:06:32,503 MainProcess INFO     appending 2 json elements from content in .../unstructured/test_unstructured_ingest/structured-output/s3-pinecone-dest/DA-1p.pdf.json
2023-11-09 13:06:32,503 MainProcess INFO     Upserting 734 elements to destination index at ingest-test
results: {'upserted_count': 80}
results: {'upserted_count': 80}
results: {'upserted_count': 80}
results: {'upserted_count': 80}
results: {'upserted_count': 80}
results: {'upserted_count': 14}
results: {'upserted_count': 80}
results: {'upserted_count': 80}
results: {'upserted_count': 80}
results: {'upserted_count': 80}
```

unstructured/ingest/connector/pinecone.py
ahmetmeleq: I think I somehow tested it in a wrong way before. 

I thought it was working but after your comment I re-tested and saw that it's not working.

Now, after the commit [fix session handles](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/007ad367942b5efd17f396d2edbfc77fed1cf9c5) : 
- when `number of batches > num_processes`, only `num_processes` sessions get created
- when `num_processes > number of batches`, only `number of batches` sessions get created

These test results fit to my understanding of how session handles work now. Please let me know if there are any concerns on the results

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: Agreed, addressed with commit: [random index names to avoid test run collisions](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/35be64a36682c00b4b0b8b571043c69964290ea2)

unstructured/ingest/cli/interfaces.py
ahmetmeleq: Agreed, addressed with commit: [add support for new_after_n_chars](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/1e8f34e9c5e48fec933ed47290b571703afd4552)

Ingest test for the pinecone connector is also updated to test if new_after_n_chars works as intended

docs/source/ingest/destination_connectors/pinecone.rst
ahmetmeleq: Agreed, addressed with [update docs](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/80fed3bd90fd9a1a40f3d101e37da761ff6769ad) . Didn't include `--embedding-api-key` since it is not mandatory for all embedding providers, please let me know if we'd like that included.

potter-potter:
Sessions is working now as I expected. Nice job.

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: @ryannikolaidis , per the parallel index creation collision concerns:

### Multiple indexes are not supported in pinecone free tier

Even after we randomized our index names, we won't be able to create multiple indexes with the free tier:

1. One index per free (starter) project
> --requestCapacity Reached. Starter Projects support a single index. Create a new project to add more. Your Starter Project remains free post-upgrade.

2. One project per free org
image

#### To be able to keep in the free tier (one index at a time)
- Keeping pinecone test for python 3.10 only (rather than all matrix) should help. 
- I also added an index creation retry logic so that index creation can be retried for up to 2 minutes when parallel tests are being triggered at the same time by different contributors. 
  - Pinecone test runs in 90 seconds in my machine and deletion of an index takes 10 to 15 seconds, totalling to <105 seconds. 120 seconds should provide enough leeway while still keeping the retry time as low as possible. It'd support:
    - up to 2 contributors triggering the CI at the exact same time,
    - unlimited contributors triggering CI, as long as it is with >105 second intervals

**Note**: Multiple index creation collisions are not something we've encountered in CI yet since pinecone test only runs in python 3.10. I assume we will encounter it when multiple contributors trigger CI at the same time, thus made the changes mentioned above.

ryannikolaidis: right, this is a separate concern, and sounds like we need to bump to a paid tier.
probably best not to burn CI time and people time for the sake of staying on the free tier.

> pinecone test only runs in python 3.10.
right, that and the fact that this hasn't merged, so this branch is literally the only branch test it..

docs/source/ingest/destination_connectors/pinecone.rst
ryannikolaidis: it looks like you updated in the shell script part of the doc, but not here in the python bit as well

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: Updated to the paid tier now 👍

docs/source/ingest/destination_connectors/pinecone.rst
```diff
@@ -0,0 +1,80 @@
...
+upstream s3 connector. This will create new files on your local.
+
+.. tabs::
+
+   .. tab:: Shell
+
+      .. code:: shell
+
+        unstructured-ingest \
+            s3 \
```
rbiseck3, line 24: Can we update this to use the same approach we already take using the local connector for the first half of this.

rbiseck3: Example: [mongodb.rst](https://github.com/Unstructured-IO/unstructured/blob/roman/mongo-db-atlas-dest-conn/docs/source/ingest/destination_connectors/mongodb.rst?plain=1#L24)

docs/source/ingest/destination_connectors/pinecone.rst
```diff
@@ -0,0 +1,80 @@
...
+            --batch-size <number of elements to be uploaded per batch, ie. 80> \
+            --num-processes <number of processes to be used to upload, ie. 2>
+
+   .. tab:: Python
+
+      .. code:: python
+
+        import subprocess
+
+        command = [
```
rbiseck3, line 46: This should also be using the pythonic approach to running the code, example: [mongodb.rst](https://github.com/Unstructured-IO/unstructured/blob/roman/mongo-db-atlas-dest-conn/docs/source/ingest/destination_connectors/mongodb.rst?plain=1#L45)

examples/ingest/pinecone/ingest.sh
```diff
@@ -0,0 +1,30 @@
...
+# Structured outputs are stored in s3-small-batch-output-to-pinecone/
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "$SCRIPT_DIR"/../../.. || exit 1
+
+
+# As an example we're using the s3 source connector,
+# however ingesting from any supported source connector is possible.
+PYTHONPATH=. ./unstructured/ingest/main.py \
+        s3 \
```
rbiseck3, line 15: Same comment about using a local connector for the source

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,148 @@
...
+
+  else
+    echo "Index $PINECONE_INDEX does not exist, nothing to delete"
+  fi
+
+  # Local file cleanup
+  cleanup_dir "$WORK_DIR"
+  cleanup_dir "$OUTPUT_DIR"
+  if [ "$CI" == "true" ]; then
+    cleanup_dir "$DOWNLOAD_DIR"
```
rbiseck3, line 55: `DOWNLOAD_DIR` is irrelevant here since we're switched to using a local connector for the source.

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,148 @@
...
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+
+if [ -z "$PINECONE_API_KEY" ]; then
+   echo "Skipping Pinecone ingest test because PINECONE_API_KEY env var is not set."
+   exit 0
+fi
+
+RANDOM_SUFFIX=$((RANDOM % 100000 + 1))
+
+PINECONE_ENVIRONMENT="us-east1-gcp"
```
rbiseck3, line 21: Might want `PINECONE_ENVIRONMENT` and `PINECONE_PROJECT_ID` set from CI vars to allow us to point that to different environments/projects with ease.

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,148 @@
...
+WORK_DIR=$SCRIPT_DIR/workdir/$OUTPUT_FOLDER_NAME
+DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+
+if [ -z "$PINECONE_API_KEY" ]; then
+   echo "Skipping Pinecone ingest test because PINECONE_API_KEY env var is not set."
+   exit 0
+fi
+
+RANDOM_SUFFIX=$((RANDOM % 100000 + 1))
```
rbiseck3, line 19: Use this to create a random value:
```shell
RANDOM_SUFFIX=$(uuidgen)
```

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR"/cleanup.sh
+function cleanup {
+
+  # Get response code to check if index exists
+  response_code=$(curl \
+    -s -o /dev/null \
+    -w "%{http_code}" \
+    --request GET \
```
rbiseck3, line 33: Can a `HEAD` request suffice if all we're doing is checking for existence via the response code?

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,148 @@
...
+  "dimension": 384,
+  "metric": "cosine",
+  "pods": 1,
+  "replicas": 1,
+  "pod_type": "p1.x1"
+}
+')
+
+  if [ "$response_code" -ge 400 ]; then
+  echo "Response code: $response_code . Sleeping $create_sleep_amount seconds to retry index creation."
```
rbiseck3, line 89: Just curious, why would the index creation itself cause issue and require a wait? I could see us needing to wait for the index to be up and ready after the initial create request.

unstructured/ingest/cli/cmds/pinecone.py
```diff
@@ -0,0 +1,66 @@
...
+import click
+
+from unstructured.ingest.cli.interfaces import (
+    CliMixin,
+)
+from unstructured.ingest.interfaces import BaseConfig
+
+
+@dataclass
+class PineconeCliWriteConfig(BaseConfig, CliMixin):
```
rbiseck3, line 13: Since this is actually extending configs from the Pinecone Connector file, I've been importing it from there and extending them directory: 
```suggestion
from unstructured.ingest.connector.pinecone import SimplePineconeConfig
class PineconeCliWriteConfig(SimplePineconeConfig, CliMixin):
...
```

unstructured/ingest/cli/cmds/pinecone.py
```diff
@@ -0,0 +1,66 @@
...
+                type=str,
+                help="The environment where the index lives. Eg. 'gcp-starter' or 'us-east1-gcp'",
+            ),
+            click.Option(
+                ["--batch-size"],
+                default=50,
+                type=int,
+                help="Number of records per batch",
+            ),
+            click.Option(
```
rbiseck3, line 49: This should be pulled from the overall pipeline config, rather than a new one here.

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,127 @@
...
+
+@dataclass
+class PineconeWriteConfig(IngestDocSessionHandleMixin, WriteConfig):
+    connector_config: SimplePineconeConfig
+    batch_size: int = 50
+    num_processes: int = 1
+
+    @DestinationConnectionError.wrap
+    @requires_dependencies(["pinecone"], extras="pinecone")
```
rbiseck3, line 55: A method responsible for writing the content should live within the destination connector itself rather than on the write config. Ideally this is only responsible for the values and possible some parsing of them.

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: @ryannikolaidis would we want those in CI vars, or would we prefer them explicitly defined in the script? 

I remember us leaving some non-secret CI variables out before, defining them explicitly in scripts. What would we like to do here?

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: Agreed, addressed with: [remove download_dir, remove index creation loop](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/387a7ad9c88f171ac3b636fc40a75aea7a6f08f3)

docs/source/ingest/destination_connectors/pinecone.rst
ahmetmeleq: Addressed with: [update docs](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/c3266f0bc7a121c7b4bb813c86594656da4efe4a)

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: Such index names did not work, giving a 400 response; example:
`ingest-test-3FE880DE-029B-484F-91FC-417BE200EF03`

Is there any reason we shouldn't use `RANDOM`? If we have to modify it to `uuidgen`, we'll need custom modification in the id after generating it

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: Left from when we had one index limit - just tested this and it does work with multiple tests running. Removed with: [remove download_dir, remove index creation loop](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/387a7ad9c88f171ac3b636fc40a75aea7a6f08f3)

examples/ingest/pinecone/ingest.sh
ahmetmeleq: with [update example](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/f884123c1465d6b31031c426d69497db2847330a)

docs/source/ingest/destination_connectors/pinecone.rst
ahmetmeleq: Agreed, addressed with [pythonic approach in docs](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/cbd734f4303c76884228f858763c52ca21be6b38)

unstructured/ingest/cli/cmds/pinecone.py
ahmetmeleq: This didn't work, giving:
`Error: type object 'PineconeCliWriteConfig' has no attribute 'from_dict'`
when I ran the test

unstructured/ingest/cli/cmds/pinecone.py
ahmetmeleq: I think user might want to set different number of processes for partitioning versus uploading to a destination, if for example destination has rate limits. User might want to have magnitudes of higher number of threads for partitioning compared to uploading content. What do you think?

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: This one gives a 405 method not allowed

ahmetmeleq: However this inquiry caused me to update a log by: [update log](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/fa083ff7c08da6b72f869a3c4d9ffc4ae2ccbdfa)

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Agreed, addressed with [move upsert method](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/29758f88ae6b539d315208c4da57e7ec23a937ce)

The commit does not seem to break session handles as well: 
- when `number of elements to upload (1404) > num_processes (12, 5, 3)` ,  `num_processes` sessions were created . 
- when `num_processes (5) > number of elements to upload (2)` ,  1 session was created.

unstructured/ingest/cli/cmds/pinecone.py
ahmetmeleq: Might be because SimplePineconeConfig implements session handle creation, that's the only difference I can see with the MongoDB configs

docs/source/ingest/destination_connectors/pinecone.rst
```diff
@@ -0,0 +1,79 @@
...
+First you'll need to install the Pinecone dependencies as shown here.
+
+.. code:: shell
+
+  pip install "unstructured[pinecone]"
+
+Run Locally
+-----------
+The upstream connector can be any of the ones supported, but for convenience here, showing a sample command using the
+upstream s3 connector. This will create new files on your local.
```
rbiseck3, line 15:
```suggestion
upstream local connector. This will create new files on your local.
```

docs/source/ingest/destination_connectors/pinecone.rst
```diff
@@ -0,0 +1,79 @@
...
+                embedding_config=EmbeddingConfig(
+                  provider="langchain-huggingface",
+                ),
+                writer_type="pinecone",
+                writer_kwargs={
+                    "api_key": os.getenv("PINECONE_API_KEY"),
+                    "index_name": os.getenv("PINECONE_INDEX_NAME"),
+                    "environment": os.getenv("PINECONE_ENVIRONMENT_NAME"),
+                    "batch_size": 80,
+                    "num_processes": 2,
```
rbiseck3, line 69: This should leverage the same `num_processes` being set in the `ProcessorConfig`. Actually not sure if this is causing a duplicate key error in the CLI itself...

docs/source/ingest/destination_connectors/s3.rst
```diff
@@ -0,0 +1,68 @@
...
+upstream s3 connector. This will create new files on your local.
+
+.. tabs::
+
+   .. tab:: Shell
+
+      .. code:: shell
+
+        unstructured-ingest \
+            s3 \
```
rbiseck3, line 24: Should use local connector here for the source connector

docs/source/ingest/destination_connectors/s3.rst
```diff
@@ -0,0 +1,68 @@
...
+
+   .. tab:: Python
+
+      .. code:: python
+
+        import subprocess
+
+        command = [
+          "unstructured-ingest",
+          "s3",
```
rbiseck3, line 42: Can this use the pythonic approach you added in the pinecone documentation?

test_unstructured_ingest/dest/pinecone.sh
rbiseck3: Only because that's what we use everywhere else. If that doesn't work, mind leaving a comment here why we're using a different approach?

test_unstructured_ingest/dest/pinecone.sh
rbiseck3: We can set a default but allow a custom value to be passed in via CI secrets.

unstructured/ingest/cli/cmds/pinecone.py
rbiseck3: You need this to also be decorated with `@dataclass`

unstructured/ingest/cli/cmds/pinecone.py
rbiseck3: If we want to support both, it should default to what is set by the processor config and overwrite with this value if one is provided.

docs/source/ingest/destination_connectors/s3.rst
ahmetmeleq: Agreed, addressed with [s3 docs pythonic approach and local connector](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/4f65b4965b0278957491c738c2ba08a33b96b3fd)

docs/source/ingest/destination_connectors/s3.rst
ahmetmeleq: Addressed with [s3 docs pythonic approach and local connector](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/4f65b4965b0278957491c738c2ba08a33b96b3fd)

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: Yes, added with [add comment on why we use random rather than uuidgen](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/738d75c8faf488b09f4eeddaa36652d2db71db8d)

test_unstructured_ingest/dest/pinecone.sh
potter-potter: This is an interesting question. Looks like CI vars is mostly secrets and tokens. I would lean towards adding less stuff to ci.yml (unless it's sensitive data of course) but we should be consistent so we don't have to think about this. @ahmetmeleq where are the examples of non-secret CI variables defined in script?

ahmetmeleq: I agree that fewer variables in CI is better, and we had already decided to leave this kind of variables out of CI in the past. 

I am confused about the change in our approach now, and didn't understand what makes it necessary to put these in CI. I am open to putting these in the CI but I don't want to change approaches often.

Here are the examples I've found to non-secret variables being defined in scripts:

Destination:
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/azure-cognitive-search.sh#L14
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/azure.sh#L19-L23
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/box.sh#L14
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/delta-table.sh#L11
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/dropbox.sh#L13
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/gcs.sh#L13
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/mongodb.sh#L13
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/s3.sh#L13
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/dest/sharepoint-embed-cog-index.sh#L13
Source:
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/confluence-diff.sh#L44
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/jira.sh#L66-L68
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/notion.sh#L39
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/airtable-diff.sh#L30-L31
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/azure.sh#L35-L36
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/delta-table.sh#L42
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/discord.sh#L42
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/dropbox.sh#L49
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/gcs.sh#L49
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/google-drive.sh#L49
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/onedrive.sh#L47
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/outlook.sh#L46
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/s3.sh#L38
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/sharepoint-with-permissions.sh#L55
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/slack.sh#L43
- https://github.com/Unstructured-IO/unstructured/blob/2c2d5b65ca8c2178de33016a4dc331da32bd87c3/test_unstructured_ingest/src/wikipedia.sh#L37

ryannikolaidis: I vote: Make the default the current hardcoded variable and optionally set to a value if specified, but let's not bother passing through secrets until we have a compelling reason.

ahmetmeleq: Thank you, now addressed with [check if test variables are defined before setting](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/fe818e4767bf4e046bc28efa2ae54d64c62e9c6a)

unstructured/ingest/cli/cmds/pinecone.py
ahmetmeleq: It was already decorated, but I've found why it happens now:
- the connector config does not inherit from `BaseConfig`, and we need a descendant of `BaseConfig` to provide `BaseConfig`'s functionality
- so I've replaced `CliMixin` with `CliConfig` (which inherits from both `CliMixin` and `BaseConfig`, therefore providing the needed `BaseConfig` functionality)

addressed with [update parent classes for cliconfig](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/a2b2fc379f38280c800a9215ecc8da2ca5400a90)

unstructured/ingest/cli/cmds/pinecone.py
ahmetmeleq: I still don't understand why the writer's number of processes should default to processor's number of processes. 

If a user didn't specify a `num_processes` for the writer, why do we opt them into the same number of processes as in for the processor? 

User will have to opt out of that number whenever they want a different value if we do it that way. Whereas if we have a separate argument, it'll behave in an intuitive way where that separate writer argument has its own default value.

Is this thread a technical concern such as stated in https://github.com/Unstructured-IO/unstructured/pull/1774/#discussion_r1399725686 (another thread down below), rather than a user related concern? To quote that thread:

> This should leverage the same num_processes being set in the ProcessorConfig. Actually not sure if this is causing a duplicate key error in the CLI itself...

If our concern is this quote, the test already checks if cli works with this `num_processes` argument for the writer, I cannot see any issue with that. To test even further I now made them have different values with [different number of processes for processor and writer in test](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/940f72d5fe66129a7b5ed50c380a5a6560557ce5) .

Please let me know if I am missing anything here.

ahmetmeleq:
Active thread link: https://github.com/Unstructured-IO/unstructured/pull/1774#discussion_r1396439468

docs/source/ingest/destination_connectors/pinecone.rst
ahmetmeleq: Assuming it's the same concern, check https://github.com/Unstructured-IO/unstructured/pull/1774#issuecomment-1823804128

unstructured/ingest/connector/pinecone.py
```diff
@@ -114,8 +119,19 @@ def write(self, docs: t.List[BaseIngestDoc]) -> None:
                         "id": element.pop("element_id", None),
                         "values": element.pop("embeddings", None),
                         "metadata": {
-                            k: (json.dumps(v) if isinstance(v, list) else v)
-                            for k, v in flatten_dict(element["metadata"], separator="-").items()
+                            # Since pinecone does not allow lists of any type other than str,
```
potter-potter, line 122: Yeah. I had something similar in ChromaDB. I added a list flattener to `flatten_dict` and made it false by default. Maybe this way is better.

ahmetmeleq: I think it depends on whether the fields under that complex list object should be indexable or not. If those subfields of the list object should definitely be indexable, then I think your approach makes more sense. 

But if it's likely that users won't index the fields under the complex list object, then I think it's easier to retrieve the complex list object to its original form when it's a `json.dumps` string, because they can simply do `json.loads ` to retrieve the object to its original form. Whereas if it is a flattened list, it'll need some custom code to de-flatten the object

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,142 @@
...
+  # Reason of using RANDOM rather than uuidgen here is that index names generated with uuidgen
+  # were not compatible, giving a 400 response probably due to the length of or character variety in
+  # the index name.
+  RANDOM_SUFFIX=$((RANDOM % 100000 + 1))
+  PINECONE_INDEX="ingest-test-$RANDOM_SUFFIX"
+fi
+
+if [ -z "$PINECONE_ENVIRONMENT" ]; then
+  PINECONE_ENVIRONMENT="us-east1-gcp"
+fi
```
rbiseck3, line 29: You can set these using defaults [reference](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html):
```suggestion
PINECONE_ENVIRONMENT=${PINECONE_ENVIRONMENT:-"us-east1-gcp"}
```

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,143 @@
...
+    BaseIngestDoc,
+    BaseSessionHandle,
+    ConfigSessionHandleMixin,
+    IngestDocSessionHandleMixin,
+    WriteConfig,
+)
+from unstructured.ingest.logger import logger
+from unstructured.staging.base import flatten_dict
+from unstructured.utils import requires_dependencies
+
```
rbiseck3, line 19: For things that need to be typed but don't want to include the import here, can we leverage the typing `TYPE_CHECKING` value:
```python
if t.TYPE_CHECKING:
    from pinecone import Index
```

This will help with type checks performed by mypy or an IDE.

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclass
+class PineconeDestinationConnector(BaseDestinationConnector):
+    write_config: PineconeWriteConfig
+    connector_config: SimplePineconeConfig
+
+    def initialize(self):
+        pass
+
+    def check_connection(self):
```
rbiseck3, line 63: Can we include anything in here to actually check the connection?

unstructured/ingest/connector/pinecone.py
rbiseck3: two comments:
* @potter-potter, there's actually already a supported method to flatten dictionaries supported in the unstructured library: [base.py#L178](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/staging/base.py#L178) 
* @ahmetmeleq, given the availability of indexing capabilities supported byb pinecone, my opinion is we should flatten the entire json and preserve anything that meets these fields, as well as the original JSON being dumped as a string to a new field in the pinecone metadata so a user can  extract it out with ease.

unstructured/ingest/interfaces.py
```diff
@@ -211,9 +213,25 @@ class PermissionsConfig(BaseConfig):
...
 
 
+# module-level variable to store session handle
+global_write_session_handle: t.Optional[BaseSessionHandle] = None
+
+
 @dataclass
 class WriteConfig(BaseConfig):
-    pass
+    def global_session(self):
```
rbiseck3, line 222: Shouldn't this live in the `BaseDestinationConnector`?

unstructured/ingest/pipeline/pipeline.py
```diff
@@ -86,6 +86,17 @@ def run(self):
         copier(iterable=partitioned_jsons)
 
         if self.write_node:
+            logger.info(
+                f"uploading elements from {len(partitioned_jsons)} "
+                "document(s) to the destination"
+            )
+
+            write_config = self.write_node.dest_doc_connector.write_config
+            if hasattr(write_config, "num_processes"):
```
rbiseck3, line 95: Can we add this log inside the pinecone destination connector so we don't have to use `hasattr` to dynamically check if that exists?

unstructured/ingest/connector/pinecone.py
potter-potter: @rbiseck3 That's the one (base.py) I'm using in chromadb. Just had to add a flatten list to it. (and made it false by default)

test_unstructured_ingest/dest/pinecone.sh
potter-potter: Oh neat. The walrus (and 1-tooth family) lives in shell as well. I will definitely be using that)

ahmetmeleq: Added with [walrus syntax := instead of if [-z $...] for default parameters](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/127837965117da83c7e5c17c6628c1990a051bd2)

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Thank you, this is helpful. Learned a bunch on type checks and imports as well while reading on this.

Adressed with [better type checking for session handles](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/0ec7cae1489f44ba5d2fde68596993b84bc4d380)

docs/source/ingest/configs/embedding_config.rst
```diff
@@ -10,5 +10,6 @@ the dataset.
 
 Configs
 ---------------------
+* ``embedding_provider``: An unstructured embedding provider to use while doing embedding. A few examples: langchain-openai, langchain-huggingface, langchain-aws-bedrock.
 * ``api_key``: If an api key is required to generate the embeddings via an api (i.e. OpenAI)
```
potter-potter, line 14: Will this conflict with the pinecone `api_key` ?

potter-potter: Whoops. This is in source...

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Since we use session handles per process, and since there's no global initialization step for this connector (global meaning, 'above / before all processes') I intentionally haven't put anything in `initialize` (such as creating a session or checking connection). Please let me know if we'd like to do anything there.

docs/source/ingest/configs/embedding_config.rst
potter-potter: I think its supposed to be embedding_api_key?

test_unstructured_ingest/dest/pinecone.sh
```diff
@@ -0,0 +1,133 @@
...
+     -w "%{http_code}" \
+     --request POST \
+     --url "https://controller.$PINECONE_ENVIRONMENT.pinecone.io/databases" \
+     --header "accept: text/plain" \
+     --header "content-type: application/json" \
+     --header "Api-Key: $PINECONE_API_KEY" \
+     --data '
+{
+  "name": "'"$PINECONE_INDEX"'",
+  "dimension": 384,
```
potter-potter, line 71: This is awesome. I was able to change the dimensions to accommodate the openai partitioner.

docs/source/ingest/configs/embedding_config.rst
ahmetmeleq: The documentation here should be updated. All embedding fields now have an embedding prepend:
https://github.com/Unstructured-IO/unstructured/blob/30cbc420a0809d4fcde2397d9b096b1e8e638203/unstructured/ingest/cli/interfaces.py#L408-L426

Therefore, addressed with [update embedding docs to have embedding prepend for cli args](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/ca0785e97860cba9191a5b68a5a8cdff4a491146)

(answer to the actual question: even in that case, it wouldn't - since pinecone api_key is a destination connector option while embedding api_key is a source connector option)

test_unstructured_ingest/dest/pinecone.sh
ahmetmeleq: 🚀 🚀

unstructured/ingest/pipeline/pipeline.py
ahmetmeleq: I think this log should ideally reside in pipeline, 

however since we had disagreement on `num_processes` and since the option is not yet available for other connectors, addressed with: [move log for number of (upload) processes from pipeline to connector](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/1cf12903494929b4ae59c6a7210461dbc7898685)

unstructured/ingest/interfaces.py
ahmetmeleq: Sessions need to be created after subprocesses are created, and within an object that is not shared between subprocesses. For example for source connectors, a global session is created in `Reader(SourceNode)` rather than `DocFactory(DocFactory)` node:
https://github.com/Unstructured-IO/unstructured/blob/30cbc420a0809d4fcde2397d9b096b1e8e638203/unstructured/ingest/pipeline/source.py#L14-L15

---

We don't have ingest docs within destination connectors, and therefore no analogue to the Reader step

So I thought creating a new `WriteConfig` object for each subprocess would be a good (and intuitive) analogue to our ingest doc based session handle approach in source connectors, that's why global session handle creation resides in WriteConfig for destination connectors.  

The pattern is that WriteConfigs have a module level (global) session handle to share within the process, in addition to WriteConfigs carrying inheritance from IngestDocSessionHandleMixin:
https://github.com/Unstructured-IO/unstructured/blob/ca0785e97860cba9191a5b68a5a8cdff4a491146/unstructured/ingest/connector/pinecone.py#L52

https://github.com/Unstructured-IO/unstructured/blob/35701dcd510f4d5f4f02b48ddcdb44eee894322d/unstructured/ingest/interfaces.py#L221-L226

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Agreed, and addressed with [make all element fields indexable, add element_serialized](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/e1a6365f7f674f57c9e3a06b3190abb33d59341c)

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,143 @@
...
+            with open(local_path) as json_file:
+                dict_content = json.load(json_file)
+
+                # assign element_id to "id", embeddings to "values", and other fields to "metadata"
+                dict_content = [
+                    # While flatten_dict enables indexing on various fields,
+                    # element_serialized enables easily reloading the element object to memory.
+                    # element_serialized is formed without text/embeddings to avoid data bloating.
+                    {
+                        "id": element.pop("element_id", None),
```
rbiseck3, line 125: Setting the id, which I assume needs to be a unique identifier, is a bit of an issue using `element_id`. `element_id` is derived from the text and nothing else, meaning one element with text `hello world` but different metadata (such as location on the page) will have the same `element_id` as another element with the same text but exists elsewhere on the page. I think that needs to be addressed in the core library, but for now it prevents us from using it as the unique identified in a database because it will cause those elements to overwrite themselves and only one will exist. For now I used a random UUID generated for the id in other connectors if one isn't automatically generated for you by the provider.

unstructured/ingest/interfaces.py
rbiseck3: This works for now. We might want to find something else in the future as I'd like to avoid having a lot of business logic tied to the config base classes.

unstructured/ingest/connector/pinecone.py
ahmetmeleq: Agreed; I now applied the same approach as in azure cognitive search, and addressed with: [unique ids for pinecone entries rather than using element ids](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/e353c6b2495539883f5796949ca6c51de9db9109)

unstructured/ingest/connector/pinecone.py
```diff
@@ -0,0 +1,144 @@
...
+
+@dataclass
+class PineconeDestinationConnector(BaseDestinationConnector):
+    write_config: PineconeWriteConfig
+    connector_config: SimplePineconeConfig
+
+    def initialize(self):
+        pass
+
+    def check_connection(self):
```
rbiseck3, line 67: If this fails, it should raise the custom `SourceConnectionError` error.

ahmetmeleq: This is already ensured with this part (since `check_connection` calls `create_pinecone_object`):
https://github.com/Unstructured-IO/unstructured/blob/7e0c7e7a97237679ccd4d5a1c34385a523eb001d/unstructured/ingest/connector/pinecone.py#L30-L32

However, I've added an additional wrapper with [an additional error wrapper for check connection](https://github.com/Unstructured-IO/unstructured/pull/1774/commits/b54e5ce3ed31915a6dd8f69e95034e9b925c307a) in case check_connection gets modified (and does not call create_pinecone_object) in the future

rbiseck3 (approved):

PR #1775, opened 2023-10-17
Add check for duplicate click options

rbiseck3:
### Description
Given that many of the options associated with the `Click` based cli ingest commands are added dynamically from a number of configs, a check was incorporated to make sure there were no duplicate entries to prevent new configs from overwriting already added options.

### Issues that were found and fixes:
* duplicate api-key option set on Notion command conflicts with api key used for unstructured api. Added notion prefix.
* retry logic configs had duplicates in biomed. Removed since this is not handled by the pipeline.


ryannikolaidis (approved):
lgtm

PR #1777, opened 2023-10-17
expand fsspec downstream connectors

rbiseck3:
### Description
Replacing PR [1383](https://github.com/Unstructured-IO/unstructured/pull/1383)


unstructured/ingest/interfaces.py
```diff
@@ -132,6 +132,13 @@ def __post_init__(self):
             self.file_path = ""
             return
 
+        # dropbox paths can start with slash
```
ryannikolaidis, line 135: oh fun. good catch

test_unstructured_ingest/test-ingest-box-dest.sh
```diff
@@ -0,0 +1,54 @@
...
+#    --verbose \
+#    --reprocess \
+#    --input-path example-docs/fake-memo.pdf \
+#    --work-dir "$WORK_DIR" \
+#    box \
+#    --box-app-config "$BOX_APP_CONFIG_PATH" \
+#    --remote-url "$DESTINATION_BOX" \
+#
+## Simply check the number of files uploaded
+#expected_num_files=1
```
ryannikolaidis, line 54: do we not support box yet? leaving this in for when we do? ...if so, would leave details at the top

test_unstructured_ingest/test-ingest-box-dest.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+#TODO currently box api/sdk does not work to create folders and check for content similar to other fsspec ingest tests
```
ryannikolaidis, line 2: for the purposes of testing: can we just attempt to download the content in the test and confirm that we can fetch it?

rbiseck3: The problem here is I can't even create a temporary path with confidence to use for testing in the same way I create it everywhere else. We could technically push to a new path but then have no way to clean it up at the end, bloating how much we store in the box account.

ryannikolaidis (commented):
looks good. a couple quick questions above otherwise gtg

test_unstructured_ingest/test-ingest-gcs-dest.sh
```diff
@@ -0,0 +1,61 @@
...
+    --reprocess \
+    --input-path example-docs/fake-memo.pdf \
+    --work-dir "$WORK_DIR" \
+    gcs \
+    --token "$GCP_INGEST_SERVICE_KEY_FILE" \
+    --remote-url "$DESTINATION_GCS"
+
+# Simply check the number of files uploaded
+expected_num_files=1
+num_files_in_gcs=$(gcloud storage ls "$DESTINATION_GCS"/example-docs/ | wc -l )
```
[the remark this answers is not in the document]
rbiseck3: Created one here: [drop dependency on cloud CLIs in CI](https://github.com/Unstructured-IO/unstructured/issues/1941)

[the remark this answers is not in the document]
rbiseck3: Will also have to investigate how to clean up the created folders without the CLIs.

test_unstructured_ingest/test-ingest-box-dest.sh
rbiseck3: Well across the board, it's aws cli for s3, `az` for azure and `gcloud` for google gloud storage.

rbiseck3: For dropbox it had to be done through their api using `curl`

test_unstructured_ingest/test-ingest-gcs-dest.sh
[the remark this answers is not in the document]
ryannikolaidis: Perfect, thanks!

ryannikolaidis (approved):
Lgtm!

PR #1779, opened 2023-10-17
fix: chunks break on regex-meta changes and regex-meta start/stop not adjusted

scanny:
**Executive Summary.** Introducing strict type-checking as preparation for adding the chunk-overlap feature revealed a type mismatch for regex-metadata between chunking tests and the (authoritative) ElementMetadata definition. The implementation of regex-metadata aspects of chunking passed the tests but did not produce the appropriate behaviors in production where the actual data-structure was different. This PR fixes these two bugs.

1. **Over-chunking.** The presence of `regex-metadata` in an element was incorrectly being interpreted as a semantic boundary, leading to such elements being isolated in their own chunks.

2. **Discarded regex-metadata.** regex-metadata present on the second or later elements in a section (chunk) was discarded.

**Technical Summary**

The type of `ElementMetadata.regex_metadata` is `Dict[str, List[RegexMetadata]]`. `RegexMetadata` is a `TypedDict` like `{"text": "this matched", "start": 7, "end": 19}`.

Multiple regexes can be specified, each with a name like "mail-stop", "version", etc. Each of those may produce its own set of matches, like:

```python
>>> element.regex_metadata
{
    "mail-stop": [{"text": "MS-107", "start": 18, "end": 24}],
    "version": [
        {"text": "current: v1.7.2", "start": 7, "end": 21},
        {"text": "supersedes: v1.7.0", "start": 22, "end": 40},
    ],
}
```

*Forensic analysis*
* The regex-metadata feature was added by Matt Robinson on 06/16/2023 commit: 4ea71683. The regex_metadata data structure is the same as when it was added.

* The chunk-by-title feature was added by Matt Robinson on 08/29/2023 commit: f6a745a7. The mistaken regex-metadata data structure in the tests is present in that commit.

Looks to me like a mis-remembering of the regex-metadata data-structure and insufficient type-checking rigor (type-checker strictness level set too low) to warn of the mistake.

**Over-chunking Behavior**

The over-chunking looked like this:

Chunking three elements with regex metadata should combine them into a single chunk (`CompositeElement` object), subject to maximum size rules (default 500 chars).

```python
elements: List[Element] = [
    Title(
        "Lorem Ipsum",
        metadata=ElementMetadata(
            regex_metadata={"ipsum": [RegexMetadata(text="Ipsum", start=6, end=11)]}
        ),
    ),
    Text(
        "Lorem ipsum dolor sit amet consectetur adipiscing elit.",
        metadata=ElementMetadata(
            regex_metadata={"dolor": [RegexMetadata(text="dolor", start=12, end=17)]}
        ),
    ),
    Text(
        "In rhoncus ipsum sed lectus porta volutpat.",
        metadata=ElementMetadata(
            regex_metadata={"ipsum": [RegexMetadata(text="ipsum", start=11, end=16)]}
        ),
    ),
]

chunks = chunk_by_title(elements)

assert chunks == [
    CompositeElement(
        "Lorem Ipsum\n\nLorem ipsum dolor sit amet consectetur adipiscing elit.\n\nIn rhoncus"
        " ipsum sed lectus porta volutpat."
    )
]
```

Observed behavior looked like this:

```python
chunks => [
    CompositeElement('Lorem Ipsum')
    CompositeElement('Lorem ipsum dolor sit amet consectetur adipiscing elit.')
    CompositeElement('In rhoncus ipsum sed lectus porta volutpat.')
]

```

The fix changed the approach from breaking on any metadata field not in a specified group (`regex_metadata` was missing from this group) to only breaking on specified fields (whitelisting instead of blacklisting). This avoids overchunking every time we add a new metadata field and is also simpler and easier to understand. This change in approach is discussed in more detail here #1790.

**Dropping regex-metadata Behavior**

Chunking this section:

```python
elements: List[Element] = [
    Title(
        "Lorem Ipsum",
        metadata=ElementMetadata(
            regex_metadata={"ipsum": [RegexMetadata(text="Ipsum", start=6, end=11)]}
        ),
    ),
    Text(
        "Lorem ipsum dolor sit amet consectetur adipiscing elit.",
        metadata=ElementMetadata(
            regex_metadata={
                "dolor": [RegexMetadata(text="dolor", start=12, end=17)],
                "ipsum": [RegexMetadata(text="ipsum", start=6, end=11)],
            }
        ),
    ),
    Text(
        "In rhoncus ipsum sed lectus porta volutpat.",
        metadata=ElementMetadata(
            regex_metadata={"ipsum": [RegexMetadata(text="ipsum", start=11, end=16)]}
        ),
    ),
]
```

..should produce this regex_metadata on the single produced chunk:

```python
assert chunk == CompositeElement(
    "Lorem Ipsum\n\nLorem ipsum dolor sit amet consectetur adipiscing elit.\n\nIn rhoncus"
    " ipsum sed lectus porta volutpat."
)
assert chunk.metadata.regex_metadata == {
    "dolor": [RegexMetadata(text="dolor", start=25, end=30)],
    "ipsum": [
        RegexMetadata(text="Ipsum", start=6, end=11),
        RegexMetadata(text="ipsum", start=19, end=24),
        RegexMetadata(text="ipsum", start=81, end=86),
    ],
}
```

but instead produced this:

```python
regex_metadata == {"ipsum": [{"text": "Ipsum", "start": 6, "end": 11}]}
```

Which is the regex-metadata from the first element only.

The fix was to remove the consolidation+adjustment process from inside the "list-attribute-processing" loop (because regex-metadata is not a list) and process regex metadata separately.

### DIFF
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -64,10 +64,6 @@ def test_split_elements_by_title_and_table():
     ]
 
 
-@pytest.mark.xfail(reason="regex_metadata was wrong type", raises=AssertionError, strict=True)
-# -- `ElementMetadata.regex_metadata` is `Dict[str, List[RegexMetadata]]`, not `List[RegexMetadata]`
-# -- when this is fixed, this test fails by isolating a chunk for "Today is a bad day", which is
-# -- where the regex-metadata appears.
 def test_chunk_by_title():
     elements: List[Element] = [
         Title("A Great Day", metadata=ElementMetadata(emphasized_text_contents=["Day"])),
@@ -218,11 +214,6 @@ def test_chunk_by_title_does_not_break_on_regex_metadata_change():
     ]
 
 
-@pytest.mark.xfail(
-    reason="bug: regex_metadata of second and later section elements is discarded",
-    raises=AssertionError,
-    strict=True,
-)
 def test_chunk_by_title_consolidates_and_adjusts_offsets_of_regex_metadata():
     """ElementMetadata.regex_metadata of chunk is union of regex_metadatas of its elements.
 
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -218,6 +218,58 @@ def test_chunk_by_title_does_not_break_on_regex_metadata_change():
     ]
 
 
+@pytest.mark.xfail(
+    reason="bug: regex_metadata of second and later section elements is discarded",
+    raises=AssertionError,
+    strict=True,
+)
+def test_chunk_by_title_consolidates_and_adjusts_offsets_of_regex_metadata():
+    """ElementMetadata.regex_metadata of chunk is union of regex_metadatas of its elements.
+
+    The `start` and `end` offsets of each regex-match are adjusted to reflect their new position in
+    the chunk after element text has been concatenated.
+    """
+    elements: List[Element] = [
+        Title(
+            "Lorem Ipsum",
+            metadata=ElementMetadata(
+                regex_metadata={"ipsum": [RegexMetadata(text="Ipsum", start=6, end=11)]}
+            ),
+        ),
+        Text(
+            "Lorem ipsum dolor sit amet consectetur adipiscing elit.",
+            metadata=ElementMetadata(
+                regex_metadata={
+                    "dolor": [RegexMetadata(text="dolor", start=12, end=17)],
+                    "ipsum": [RegexMetadata(text="ipsum", start=6, end=11)],
+                }
+            ),
+        ),
+        Text(
+            "In rhoncus ipsum sed lectus porta volutpat.",
+            metadata=ElementMetadata(
+                regex_metadata={"ipsum": [RegexMetadata(text="ipsum", start=11, end=16)]}
+            ),
+        ),
+    ]
+    chunks = chunk_by_title(elements)
+
+    assert len(chunks) == 1
+    chunk = chunks[0]
+    assert chunk == CompositeElement(
+        "Lorem Ipsum\n\nLorem ipsum dolor sit amet consectetur adipiscing elit.\n\nIn rhoncus"
+        " ipsum sed lectus porta volutpat."
+    )
+    assert chunk.metadata.regex_metadata == {
+        "dolor": [RegexMetadata(text="dolor", start=25, end=30)],
+        "ipsum": [
+            RegexMetadata(text="Ipsum", start=6, end=11),
+            RegexMetadata(text="ipsum", start=19, end=24),
+            RegexMetadata(text="ipsum", start=81, end=86),
+        ],
+    }
+
+
 def test_chunk_by_title_groups_across_pages():
     elements: List[Element] = [
         Title("A Great Day", metadata=ElementMetadata(page_number=1)),
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -183,6 +183,46 @@ def test_chunk_by_title_separates_by_page_number():
     ]
 
 
+@pytest.mark.xfail(
+    reason="bug: chunks break on regex_metadata differences", raises=AssertionError, strict=True
+)
+def test_chunk_by_title_does_not_break_on_regex_metadata_change():
+    """Sectioner is insensitive to regex-metadata changes.
+
+    A regex-metadata match in an element does not signify a semantic boundary and a section should
+    not be split based on such a difference.
+    """
+    elements: List[Element] = [
+        Title(
+            "Lorem Ipsum",
+            metadata=ElementMetadata(
+                regex_metadata={"ipsum": [RegexMetadata(text="Ipsum", start=6, end=11)]}
+            ),
+        ),
+        Text(
+            "Lorem ipsum dolor sit amet consectetur adipiscing elit.",
+            metadata=ElementMetadata(
+                regex_metadata={"dolor": [RegexMetadata(text="dolor", start=12, end=17)]}
+            ),
+        ),
+        Text(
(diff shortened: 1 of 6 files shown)


test_unstructured/chunking/test_title.py
```diff
@@ -91,12 +99,12 @@ def test_chunk_by_title():
 
     assert chunks[0].metadata == ElementMetadata(emphasized_text_contents=["Day", "day"])
     assert chunks[3].metadata == ElementMetadata(
-        regex_metadata=[{"text": "A", "start": 11, "end": 12}],
+        regex_metadata={"a": [RegexMetadata(text="A", start=11, end=12)]}
```
cragwolfe, line 102: why did the data structure of the value change here?

would this get json serialized / deserialized correctly?

unstructured/chunking/title.py
```diff
@@ -1,7 +1,14 @@
+"""Implementation of chunking.
```
cragwolfe, line 1: Implementation of chunking by title? (since module name is title.py)

unstructured/chunking/title.py
```diff
@@ -115,22 +120,37 @@ def chunk_by_title(
         metadata = first_element.metadata
         start_char = 0
         for element in section:
+            # -- concatenate all element text in section into `text` --
             if isinstance(element, Text):
+                # -- add a blank line between "squashed" elements --
                 text += "\n\n" if text else ""
```
cragwolfe, line 126: not related to this PR, but is this mistakenly prepending two blank lines to the start of every chunk? (if so, can just create the github issue to track)

unstructured/chunking/title.py
```diff
@@ -115,22 +120,37 @@ def chunk_by_title(
...
+                    value = cast(List[Any], value)
+                    # -- get existing (list) value from chunk_metadata --
                     _value = getattr(metadata, attr, []) or []
-
-                    if attr == "regex_metadata":
-                        for item in value:
-                            item["start"] += start_char
-                            item["end"] += start_char
-
+                    # TODO: this mutates the original, work on a copy instead.
```
cragwolfe, line 136: please create a github issue for this

test_unstructured/chunking/test_title.py
scanny: The type of `ElementMetadata.regex_metadata` is `Dict[str, List[RegexMetadata]]`. `RegexMetadata` is a `TypedDict` like `{"text": "this matched", "start": 7, "end": 19}`.

My assumption is you can specify multiple regexes, each with a name like "mail-stop", "version", etc., and each of those may produce its own set of matches, like:

```python
>>> element.regex_metadata
{
    "mail-stop": [{"text": "MS-107", "start": 18, "end": 24}],
    "version": [
        {"text": "current: v1.7.2", "start": 7, "end": 21},
        {"text": "supersedes: v1.7.2", "start": 22, "end": 40},
    ],
}
```

My guess is that this "multiple-regexes" behavior was originally "single-regex" and when it was changed the tests and chunking code wasn't updated. But that's just my guess. Not sure who might remember about that.

scanny: On the serializing/deserializing question, I don't see why it wouldn't work fine, these are all primitive data structures and JSON serializable, but let me take a closer look.

unstructured/chunking/title.py
scanny: No. Text is initialized to `""` a couple lines above this, before iterating through the section elements. So the `else` part of this expression (`+= ""`) governs until after the text buffer has some text in it.

unstructured/chunking/title.py
scanny: This is actually fixed in the next PR (which removes this comment), but I can take this out if you want.

test_unstructured/chunking/test_title.py
scanny: Okay, looking at the serde code, it looks to me like this should round-trip fine. However, there are no tests confirming this that I can find, at least not in `test_unstructured/documents/test_elements.py`. If you want I can ticket this and write a more thorough test. I'll whip up an informal one here just to satisfy myself it actually works, but I'm thinking that's probably outside the scope of this PR to actually add it in, do you agree?

Note that the changes in this PR don't affect the type of `ElementMetadata.regex_metadata` or its serde in any way, they just turned up the type error when I cleaned up typing in this neighborhood and that's what led to these two changes.

They _would_ however (indirectly) affect the serde of the chunks (`CompositeElement` objects mostly) in that regex meta would have been stripped from second and later section elements, so the `CompositeElement` would only contain matches present in the first element.

scanny: Yep, round-trips fine in my test.

scanny: I added a test to confirm JSON round-trip of `ElementMetadata.regex_metadata`.

unstructured/chunking/title.py
scanny: Yep, good call, fixed this.

scanny:
Ok Crag, all the fixes are in for this. Let me know if there's anything else.

test_unstructured/chunking/test_title.py
cragwolfe: > But this account is just my guess. Not sure who might remember about that.

git blame can provide that info :)

scanny: @cragwolfe okay, I did the forensics, turns out it was just a mistake from the start. Good advertisement for strict data typing :)

* 06/16/2023 4ea71683 - regex_metadata feature was added by Matt Robinson in substantially its current form: (`Dict[str, List[RegexMetadata]]`).
* 08/29/2023 f6a745a7 - chunk by titles was added by Matt with the wrong type for regex_metadata in the tests, as we found it.

cragwolfe (approved):
Great description and added tests. Thank you!

PR #1787, opened 2023-10-18
Fix: TypeError - 'Link' object is not iterable

mwbyeon:
In the ebooklib code below, `book.toc` may not iterable and may be a Link class.
https://github.com/aerkalov/ebooklib/blob/1cb3d2c251f82c4702c2aff0ed7aea375babf251/ebooklib/epub.py#L1567-L1588

In this case, it throws "TypeError: 'Link' object is not iterable" exception, 
so fixed to convert to a list if it is not iterable.

About 1% of epub files got this error.

### DIFF
--- a/unstructured/partition/epub.py
+++ b/unstructured/partition/epub.py
@@ -82,7 +82,7 @@ def partition_epub(
 
     # open issue that might resolve the chapter mapping of text
     # https://github.com/aerkalov/ebooklib/issues/289
-    if not isintance(book.toc, Iterable):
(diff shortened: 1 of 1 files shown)


qued:
@mwbyeon We realized that `ebooklib` used a license that is incompatible with our Apache 2.0 license and had to remove it from our dependencies. Does that make this PR moot?

mwbyeon:
@qued  If you're not using `ebooklib`, this PR is meaningless, so let's close it.

PR #1791, opened 2023-10-18
fix: fsspec destination path malformed

rbiseck3:
### Description  
The way the destination path for where to upload content in the base fsspec destination connector didn't take into account all edge cases and this would cause the path to be malformed and potentially break the upload. Fix put in place to more robustly create that destination path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.25-dev6
+## 0.10.25-dev7
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


ryannikolaidis:
do we have a test that would have broken otherwise? or maybe what @ahmetmeleq is doing that ran into this issue will cover that? something we can at least test manually right now?

rbiseck3:
@ryannikolaidis, I pulled this fix in from other bugfix PRs to get this in sooner.

ryannikolaidis:
> @ryannikolaidis, I pulled this fix in from other bugfix PRs to get this in sooner.

so will this be covered by a test somewhere? if so, where?

ryannikolaidis:
> > @ryannikolaidis, I pulled this fix in from other bugfix PRs to get this in sooner.
> 
> so will this be covered by a test somewhere? if so, where?

bump

PR #1795, opened 2023-10-18
build: ingest fixtures workflow to include metrics dir (rework of 1789)

Klaijan:
Add test_unstructured_ingest/metrics path for evaluation metrics master file.


badGarnet:
there seem to be nothing here? Is this PR redundant?

PR #1799, opened 2023-10-19
Roman/optimize ingest ci

rbiseck3:
### Description
Currently the CI caches the CI dependencies but uses the hash of all files in `requirements/`. This isn't completely accurate since the ingest dependencies are installed in a later step and don't affect the cached environment. As part of this PR:
* ingest dependencies were isolated into their own folder in `requirements/ingest/`
* A new cache setup was introduced in the CI to restore the base cache -> install ingest dependencies -> cache it with a new id
* new make target created to install all ingest dependencies via `pip install -r ...`
* updates to Dockerfile to use `find ...` to install all dependencies, avoiding the need to update this when new deps are added. 
* update to pip-compile script to run over all `*.in` files in `requirements/`


tabossert (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -229,14 +229,56 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache/restore@v3
+        id: base-virtualenv-cache
+        with:
+          path: |
+            .venv
+            nltk_data
+          key: unstructured-${{ runner.os }}-${{ matrix.python-version }}-${{ env.GHA_CACHE_KEY_VERSION }}-${{ hashFiles('requirements/*.txt') }}
+      # Due to the dependency on setup, the cache should exist before this is ran. Set failed if it wasn't found.
```
ryannikolaidis, line 249: given that we're WAY over the [10gb cache limit](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). we generally try to do the right thing with a setup job and leveraging that in subsequent jobs, but I suspect at any point these can be ejected (which is probably a good reason for some wonkiness), so I believe that's why when it's not found IN the job, we typically just go ahead and rebuild. e.g. like we do [here](https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ci.yml#L35-L40).

.github/workflows/ci.yml
```diff
@@ -229,14 +229,56 @@ jobs:
...
+      - name: Setup base virtual environment (no cache hit)
```
ryannikolaidis, line 250: err, why are we calling this "Setup base virtual environment (no cache hit)"?? seems like this is "Required cache check" ...or something to that extent

.github/workflows/ci.yml
```diff
@@ -229,14 +229,56 @@ jobs:
...
+        if: steps.base-virtualenv-cache.outputs.cache-hit != 'true'
+        uses: actions/github-script@v3
+        with:
+          script: |
+            core.setFailed('base cached environment couldn't be found')
```
ryannikolaidis, line 255: escaping single quotes required here. or wrapping in double quotes

ryannikolaidis:
@rbiseck3 @tabossert I realize this already merged, but dropped some notes per current issue hitting folks in ci

.github/workflows/ingest-test-fixtures-update-pr.yml
```diff
@@ -36,26 +36,68 @@ jobs:
           [ ! -d "$NLTK_DATA" ] && mkdir "$NLTK_DATA"
           make install-ci
 
-  update-fixtures-and-pr:
-    runs-on: ubuntu-latest-m
+  setup_ingest:
+    strategy:
+      matrix:
+        python-version: [ "3.8","3.9","3.10","3.11" ]
```
ryannikolaidis, line 42: no need for python matrix here

PR #1800, opened 2023-10-19
local connector metadata and deserialization fix

rbiseck3:
### Description
* Priority of this was to fix deserialization of ingest docs. Currently the source metadata wasn't being persisted
* To help debug this, source metadata was added to the local ingest doc as well. 
* Unit test added to make sure the metadata itself was persisted.
* As part of serialization, it was forcing docs to fetch source metadata if it hadn't already to add to the generated dict/json. This shouldn't have happened if the underlying variable `_source_metadata` was `None`. This way the doc can be serialized without any calls being made. 
* Serialization was moved to the `to_dict` method to make it more universal.


ryannikolaidis:
cc @potter-potter for visibility

ryannikolaidis (approved):
this looks great! thanks for the unit test and for jumping on this in general.

ryannikolaidis:
actually, running this locally to validate

ryannikolaidis (approved):
okay, re-approved with latest round. 
tested locally by adding a log line in update_source_metadata and validating it was called the number of expected times when running the google drive ingest test

potter-potter:
This is working correctly when I test it.

potter-potter (approved):
This is working for me. source_metadata persists into partitioning.

unstructured/ingest/connector/local.py
```diff
@@ -48,6 +50,19 @@ def cleanup_file(self):
...
         """Not applicable to local file system"""
 
+    def update_source_metadata(self, **kwargs) -> None:
+        try:
+            out = os.lstat(self.path)
+            self._source_metadata = SourceMetadata(
+                exists=True,
+                date_created=str(datetime.fromtimestamp(out.st_ctime)),
+                date_modified=str(datetime.fromtimestamp(out.st_mtime)),
+                permissions_data=[{"mode": out.st_mode}],
```
ryannikolaidis, line 60: reasoning / value on this?

ryannikolaidis:
where did the +14k changes come from? 😳

unstructured/ingest/connector/local.py
rbiseck3: this is the file mode data which I would consider the permissions associated with the file.

rbiseck3:
@ryannikolaidis, all the ingest tests that run local connector has source metadata added as part of this PR.

Issue #1564, opened 2023-09-28
bug/OCR language parameters isn't passed to table model


yuming-long:
**Describe the bug**
OCR language parameters isn't passed to table model thus we can't extract texts from tables from different language

**To Reproduce**
use this Korean files with tables: [korean_table.pdf](https://github.com/Unstructured-IO/unstructured/files/12750626/korean_table.pdf)
```
from unstructured.partition.pdf import partition_pdf


elements = partition_pdf(filename="/Users/yumingl/Downloads/korean_table.pdf", infer_table_structure=True, ocr_languages="kor+eng", strategy="hi_res")
tables = [el for el in elements if el.category == "Table"]
if len(tables) >= 1:
    for table in tables:
        print(table.text)
        print(table.metadata.text_as_html)
 ```
see output like:
```
업종 적립 대상 할인점 이마트, 이마트 트 레 이 더 스 , 롯 데 마트, 홈 플러스 주유 모든 주유소 온 라 인 쇼핑몰 심 성 카 드 쇼핑, 6 마 켓 옥션, 11 번 가, 인터파크, 삼 쿠 팡 , 티 몬 , 위 메 프
<table><thead><th>as</th><th>Ae ches</th></thead><tr><td></td><td>O/OLE, ODE SAo|yA, SHOVE, SSAA</td></tr><tr><td rowspan="2"></td><td>ATES AW GOP, SM, THI}, QIE{T ES,</td></tr><tr><td></td><td>$2, EE, Path jz</td></tr></table>
```
And the `text_as_html` doesn't have Korean compared to the `text` field.

**Expected behavior**
I notice the tesseract in table model doesn't have lang parameters passed in and i manually modified it with:
* pip install unstructured-inference into unstructured env with `pip install -e .` (make install-project-local)
* add `lang="kor+eng",` after [this line](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/tables.py#L91)
* also add `lang="kor+eng",` after [this line](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/tables.py#L112)

and get expected result like:
```
[...]

PR #1801, opened 2023-10-19
Chore (refactor): support table extraction with pre-computed ocr data

yuming-long:
### Summary

Table OCR refactor, move the OCR part for table model in inference repo to unst repo. 
* Before this PR, table model extracts OCR tokens with texts and bounding box and fills the tokens to the table structure in inference repo. This means we need to do an additional OCR for tables.
* After this PR, we use the OCR data from entire page OCR and pass the OCR tokens to inference repo, which means we only do one OCR for the entire document.

**Tech details:**
* Combined env `ENTIRE_PAGE_OCR` and `TABLE_OCR` to `OCR_AGENT`, this means we use the same OCR agent for entire page and tables since we only do one OCR.
* Bump inference repo to `0.7.9`, which allow table model in inference to use pre-computed OCR data from unst repo. Please check in [PR](https://github.com/Unstructured-IO/unstructured-inference/pull/256).
* All notebooks lint are made by `make tidy`
* This PR also fixes [issue](https://github.com/Unstructured-IO/unstructured/issues/1564), I've added test for the issue in `test_pdf.py::test_partition_pdf_hi_table_extraction_with_languages`
* Add same scaling logic to image [similar to previous Table OCR](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/tables.py#L109C1-L113), but now scaling is applied to entire image

### Test
* Not much to manually testing expect table extraction still works
* But due to change on scaling and use pre-computed OCR data from entire page, there are some slight (better) changes on table output, here is an comparison on test outputs i found from the same test `test_partition_image_with_table_extraction`:

screen shot for table in `layout-parser-paper-with-table.jpg`:
expected
before refactor:
before
after refactor:
after

### TODO
(added as a ticket) Still have some clean up to do in inference repo since now unst repo have duplicate logic, but can keep them as a fall back plan. If we want to remove anything OCR related in inference, here are items that is deprecated and can be removed:
* [`get_tokens`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/tables.py#L77) (already noted in code)
* parameter `extract_tables` in inference
* [`interpret_table_block`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/layoutelement.py#L88)
* [`load_agent`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/tables.py#L197)
* env `TABLE_OCR` 

### Note
if we want to fallback for an additional table OCR (may need this for using paddle for table), we need to:
* pass `infer_table_structure` to inference with `extract_tables` parameter
* stop passing `infer_table_structure` to `ocr.py`


unstructured/partition/ocr.py
```diff
@@ -196,19 +212,158 @@ def supplement_page_layout_with_ocr(
...
+                    padded_element.bbox.y2,
+                ),
+            )
+            table_tokens = get_table_tokens_per_element(
+                padded_element,
+                ocr_layout,
+            )
+            table_agent = init_table_agent()
+            element.text_as_html = table_agent.predict(cropped_image, ocr_tokens=table_tokens)
+    return elements
```
christinestraub, line 274: What do you think about moving `table_agent = init_table_agent()` to the first line (out of the for `loop`) and then not defining `table_agent` as a global variable?

christinestraub: or adding `if table_agent is None:` to `init_table_agent()`?
```
def init_table_agent():
    """Initialize a table agent from unstructured_inference as
    a global variable to ensure that we only load it once."""
    from unstructured_inference.models import tables

    global table_agent

    if table_agent is None:
        table_agent = tables.UnstructuredTableTransformerModel()
        table_agent.initialize(model="microsoft/table-transformer-structure-recognition")

    return table_agent
```

christinestraub (changes requested):
Overall, it looks great! Just a few comments.

unstructured/partition/ocr.py
```diff
@@ -196,19 +212,158 @@ def supplement_page_layout_with_ocr(
...
+        outer_bbox.y1,
+        outer_bbox.x2,
+        outer_bbox.y2,
+    )
+    return (
+        inner_x1 >= outer_x1
+        and inner_y1 >= outer_y1
+        and inner_x2 <= outer_x2
+        and inner_y2 <= outer_y2
+    )
```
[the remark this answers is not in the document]
yuming-long: thanks! replaced `is_inside()`

unstructured/partition/ocr.py
yuming-long: good catch! i moved `init_table_agent` outside the loop and added `is None` check to it

christinestraub (approved):
LGTM!

test_unstructured_ingest/expected-structured-output/local-single-file-with-pdf-infer-table-structure/layout-parser-paper-with-table.jpg.json
```diff
@@ -1,162 +1,152 @@
...
     "metadata": {
       "data_source": {},
       "filetype": "image/jpeg",
       "page_number": 1
     },
-    "text": "layout data structures, which are optimized for efficiency and versatility. 3) When necessary, users can employ existing or customized OCR models via the unified API provided in the OCR module. 4) LayoutParser comes with a set of utility fanctions for the visualization and storage of the layout data. 5) LayoutParser is also highly customizable, via its integration with functions for layout data annotation and model training We now provide detailed descriptions for each component."
+    "text": "For each dataset, we train several models of different sizes for different needs (the trade-off between accuracy vs. computational cost). For “base model” and “large model”, we refer to using the ResNet 50 or ResNet 101 backbones [13], respectively. One can train models of different architectures, like Fuster R-CNN [28] (P) and Mask R-CNN [12] (M). For example, an F in the Large Model column indicates it has m Faster R-CNN model trained using the ResNet 101 backbone. The platform is maintained and a number of additions will be made to the model zoo in coming months."
```
yuming-long, line 50: scaling on entire image improved ingest output, small font texts are now detected

PR #1802, opened 2023-10-19
Move python setup above cache restore on ingest

tabossert:


rbiseck3:
The `ingest-test-fixtures-update-pr.yml` file will also need this update.

rbiseck3:
can we also run the `Ingest Test Fixtures Update PR` action off this branch to make sure that works without error. Don't need to care about the actual output.

rbiseck3 (approved):
Assuming the ingest update CI work, LGTM

PR #1808, opened 2023-10-20
feat: clean pdfminer elements inside tables

benjats07:
This PR introduces `clean_pdfminer_inner_elements` , which deletes pdfminer elements inside other detection origins such as YoloX or detectron.
This function returns the clean document.

Also, the ingest-test fixtures were updated to reflect the new standard output.

The best way to check that this function is working properly is check the new test `test_clean_pdfminer_inner_elements` in `test_unstructured/partition/utils/test_processing_elements.py`


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 ### Enhancements
 
 * **Duplicate CLI param check** Given that many of the options associated with the `Click` based cli ingest commands are added dynamically from a number of configs, a check was incorporated to make sure there were no duplicate entries to prevent new configs from overwriting already added options.
+* **Remove pdfminer from inside tables** Previously when using `hi_res` some elements where extracted using pdfminer too, so we removed pdfminer from the tables pipeline to avoid duplicated elements. Now
```
scanny, line 16: Did you mean to say more? "Now".. all by its lonesome at the end :)

CHANGELOG.md
```diff
@@ -29,6 +31,11 @@ ocr agent tesseract/paddle in environment variable `OCR_AGENT` for OCRing the en
 * **Fix chunks breaking on regex-metadata matches.** Fixes "over-chunking" when `regex_metadata` was used, where every element that contained a regex-match would start a new chunk.
 * **Fix regex-metadata match offsets not adjusted within chunk.** Fixes incorrect regex-metadata match start/stop offset in chunks where multiple elements are combined.
 * **Map source cli command configs when destination set** Due to how the source connector is dynamically called when the destination connector is set via the CLI, the configs were being set incorrectoy, causing the source connector to break. The configs were fixed and updated to take into account Fsspec-specific connectors.
+* **Fix metrics folder not discoverable** Fixes issue where unstructured/metrics folder is not discoverable on PyPI by adding
+an `__init__.py` file under the folder.
```
scanny, line 35: Should be on single line so markdown formats it properly as a paragraph.

unstructured/partition/auto.py
```diff
@@ -374,6 +374,7 @@ def partition(
             include_page_breaks=include_page_breaks,
             infer_table_structure=infer_table_structure,
             strategy=strategy,
+            generate_extra_info=False,
```
scanny, line 377: `generate_extra_info` could mean anything. What would be a more descriptive name? Is it generate OCR info? generate parsing info? Generate table metadata? The name should better direct the caller as to what to expect.

scanny (changes requested):
A few remarks. Otherwise seems okay to me, although I don't know the PDF world well enough to really understand what this is doing. Might be worth asking Christine to have a look :)

unstructured/partition/utils/processing_elements.py
```diff
@@ -0,0 +1,27 @@
...
+        tables = [e for e in page.elements if e.type == "Table"]
+        for i, element in enumerate(page.elements):
+            if element.source == Source.PDFMINER:
+                element_inside_table = [element.bbox.is_in(t.bbox) for t in tables]
+                if sum(element_inside_table) == 1:
+                    parent_table_index = element_inside_table.index(True)
+                    parent_table = tables[parent_table_index]
+                    # Note(Benjamin): is this a good way to guess the id?
+                    future_id = hashlib.sha256(parent_table.text.encode()).hexdigest()[:32]
+                    extra_info[future_id].append(element)
```
ajjimeno, line 23: Should we assume that the element already had an id? I am wondering if it can be reused. Otherwise, I am not sure what should be the best id.

With extra_info, do you mean the removed pdfminer elements that would now be linked to the table?

CHANGELOG.md
benjats07: Thanks, I deleted that last part.

unstructured/partition/auto.py
benjats07: I agree, I changed the name to `generate_inner_elements_info` and added some info in the doc strings. Do you think is better now?

unstructured/partition/utils/processing_elements.py
benjats07: In fact...it doesn't have the id at the moment we're processing this, so I took the code from here:
https://github.com/Unstructured-IO/unstructured/blob/01a0e003d9050a076894aaa4957bd56fe712ea44/unstructured/documents/elements.py#L475-L477
Not sure under which circumstances could fail. 
Probably could try to assign it before in the process 🤔

unstructured/partition/auto.py
scanny: Definitely an improvement, that at leasts gives something to look around for (inner-elements) suits me :)

scanny (approved):
Looks fine to me. Not sure if someone closer to this particular part of the code ought to take a look too though :)

unstructured/partition/utils/processing_elements.py
benjats07: I dig a little bit in the code:
In order for `clean_pdfminer_inner_elements` to work I need a way to store the elements to be possible to recover them in the future (that is why I created that element_id) but at the moment the Element is created by 

https://github.com/Unstructured-IO/unstructured/blob/a5d91e91b9c2961fc8329d63651b7033fe6845f9/unstructured/partition/pdf.py#L439C14-L439C14

the `source` property is deleted (and is the idea), then we have ID but no source to filter.
So, this element_id probably wouldn't be the definitive, but for the moment we don't have other alternatives.

cragwolfe (changes requested):
CC @christinestraub -- at a high level, not sure about the pattern of using "detection_origin." as we generate sets of elements, we should know where they came from before merging them in the first place.

christinestraub (changes requested):
@cragwolfe IMO, the purpose of using `detection_origin` is to indicate the detection mechanism that emitted this element, for debugging purposes so it should be one of the [`sources`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/constants.py#L9) in the `inference`  library and the [`sources`](https://github.com/Unstructured-IO/unstructured/blob/refactor/partition-pdf-ocr_only/unstructured/partition/utils/constants.py#L5) in the `unstructured`  library for `hi_res` (Added `sources` for `ocr_only` strategy in PR [#1811](https://github.com/Unstructured-IO/unstructured/pull/1811)).

@benjats07 If you remove the bad `detection_origin` passed to the [`document_to_element_list()`](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L439), the correct `detection_origin` will be set by the [`normalize_layout_element()`](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/common.py#L101C5-L101C29).

benjats07:
> A lot of duplicate `pdfminer` elements have been removed from inside tables, but it looks like a few still remain. PDF: [recalibrating-risk-report_4-4.pdf](https://github.com/Unstructured-IO/unstructured/files/13181748/recalibrating-risk-report_4-4.pdf)
> 
> recalibrating-risk-report_4-4_1_xy-cut

I added some padding, so that elements should not be present anymore

christinestraub (changes requested):
IMO, the purpose of using detection_origin is to indicate the detection mechanism that emitted this element, for debugging purposes so it should be one of the [sources](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/constants.py#L9) in the inference library and the [sources](https://github.com/Unstructured-IO/unstructured/blob/refactor/partition-pdf-ocr_only/unstructured/partition/utils/constants.py#L5) in the unstructured library for hi_res (Added sources for OCR in PR https://github.com/Unstructured-IO/unstructured/pull/1811).

If you remove the bad detection_origin passed to the [document_to_element_list()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L439), the correct detection_origin will be set by the [normalize_layout_element()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/common.py#L101C5-L101C29).

benjats07:
> If you remove the bad detection_origin passed to the [document_to_element_list()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L439), the correct detection_origin will be set by the [normalize_layout_element()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/common.py#L101C5-L101C29).

Thanks for that, completely forgot about that, is corrected now 😄

christinestraub (approved):
LGTM!

CHANGELOG.md
```diff
@@ -36,8 +42,6 @@
 ### Features
 
 * **Functionality to catch and classify overlapping/nested elements** Method to identify overlapping-bboxes cases within detected elements in a document. It returns two values: a boolean defining if there are overlapping elements present, and a list reporting them with relevant metadata. The output includes information about the `overlapping_elements`, `overlapping_case`, `overlapping_percentage`, `largest_ngram_percentage`, `overlap_percentage_total`, `max_area`, `min_area`, and `total_area`.
-* **Add Local connector source metadata** python's os module used to pull stats from local file when processing via the local connector and populates fields such as last modified time, created time.
-* **Add Local connector source metadata.** python's os module used to pull stats from local file when processing via the local connector and populates fields such as last modified time, created time.
```
cragwolfe, line 40: mistaken deletion?

CHANGELOG.md
```diff
@@ -70,12 +74,13 @@ ocr agent tesseract/paddle in environment variable `OCR_AGENT` for OCRing the en
...
 * **Fix metrics folder not discoverable** Fixes issue where unstructured/metrics folder is not discoverable on PyPI by adding an `__init__.py` file under the folder.
-* **Fix a bug when `parition_pdf` get `model_name=None`** In API usage the `model_name` value is `None` and the `cast` function in `partition_pdf` would return `None` and lead to attribution error. Now we use `str` function to explicit convert the content to string so it is garanteed to have `starts_with` and other string functions as attributes
-* **Fix html partition fail on tables without `tbody` tag** HTML tables may sometimes just contain headers without body (`tbody` tag)
-* **Fix out-of-order sequencing of split chunks.** Fixes behavior where "split" chunks were inserted at the beginning of the chunk sequence. This would produce a chunk sequence like [5a, 5b, 3a, 3b, 1, 2, 4] when sections 3 and 5 exceeded `max_characters`.
```
cragwolfe, line 75: mistaken deletion?

CHANGELOG.md
benjats07: Oh, not exactly. 
That line (that was duplicated indeed) was moved to 
https://github.com/Unstructured-IO/unstructured/blob/412c26067614454f759bb37a83f9f4348ded1ec8/CHANGELOG.md?plain=1#L12
The changelog got old during development, so it should be in most recent -dev version

CHANGELOG.md
benjats07: same as previous comment, these lines are in:
https://github.com/Unstructured-IO/unstructured/blob/412c26067614454f759bb37a83f9f4348ded1ec8/CHANGELOG.md?plain=1#L19

cragwolfe (approved):

PR #1810, opened 2023-10-20
chore: improve kwarg handling

qued:
Closes `unstructured-inference` issue [#265](https://github.com/Unstructured-IO/unstructured-inference/issues/265).

Cleaned up the kwarg handling, taking opportunities to turn instances of handling kwargs as dicts to just using them as normal in function signatures.

#### Testing:

Should just pass CI.


christinestraub (approved):
LGTM!

unstructured/partition/pdf.py
```diff
@@ -334,6 +329,9 @@ def _partition_pdf_or_image_local(
     ocr_mode: str = OCRMode.FULL_PAGE.value,
     model_name: Optional[str] = None,
     metadata_last_modified: Optional[str] = None,
+    extract_images_in_pdf: bool = False,
```
cragwolfe, line 332: shouldn't these be floated up to the `partition_pdf` function to make it a bit easier for end users to consume?

cragwolfe: (i'd go as far to say they should be floated up to partition, and the pdf-only kwargs popped for all other partition_ functions)

unstructured/partition/pdf.py
```diff
@@ -420,7 +408,7 @@ def _partition_pdf_or_image_local(
 
     # NOTE(alan): starting with v2, chipper sorts the elements itself.
     if model_name == "chipper":
-        kwargs["sort_mode"] = SORT_MODE_DONT
+        kwargs["sort_mode"] = SortMode.SORT_MODE_DONT
```
cragwolfe, line 411: not related to kwargs, though, right?
otherwise, a reason i'm not enthusiastic about this change is that end users (even using clients in other programming languages) will want to make a call to `partition()` and should just be able to pass a string which doesn't cleanly translate to the python builtin `enum`. as long as there is a ValueError in the right place for a string value... constants are fine IMO.

unstructured/partition/pdf.py
qued: Floated up in [d777dc2](https://github.com/Unstructured-IO/unstructured/pull/1810/commits/d777dc242ea51577b14b65134c0c68439315fb18) and [4c0142a](https://github.com/Unstructured-IO/unstructured/pull/1810/commits/4c0142ad4dae66b1dd4cbc8646ad1f2edf7af03f).

unstructured/partition/pdf.py
qued: Reverted sort mode enum in [1d1998c](https://github.com/Unstructured-IO/unstructured/pull/1810/commits/1d1998c48d896c9a5f2f1a612be305aa153b9abf).

cragwolfe (approved):

Issue #1176, opened 2023-08-22
feat/ update _partition_pdf_or_image_with_ocr to optionally use PaddleOCR


Coniferish:
Currently, `_partition_pdf_or_image_with_ocr` uses `pytesseract` to detect text in images. The goal of this issue is to enable ocr partitioning to optionally use PaddleOCR as an alternative text detection model. In addition to updating `_partition_pdf_or_image_with_ocr `, `add_pytesseract_bbox_to_elements` will also need to be updated as it expects the bounding box format from `pytesseract`.

https://pypi.org/project/paddleocr/

In paddle, `ocr.ocr(img)` returns similar data to `pytesseract.image_to_boxes(img)`

yuanjie-ai:
https://github.com/yuanjie-ai/ChatLLM/blob/master/chatllm/llmchain/document_loaders/image.py

```python
    from chatllm.llmchain.document_loaders import UnstructuredImageLoader

    loader = UnstructuredImageLoader(
        "/Users/betterme/PycharmProjects/AI/MeUtils/meutils/ai_cv/invoice.jpg",
        strategy='ocr_only'
    )
    data = loader.load()
```

yuanjie-ai:
> https://github.com/yuanjie-ai/ChatLLM/blob/master/chatllm/llmchain/document_loaders/image.py
> 
> ```python
>     from chatllm.llmchain.document_loaders import UnstructuredImageLoader
> 
>     loader = UnstructuredImageLoader(
>         "/Users/betterme/PycharmProjects/AI/MeUtils/meutils/ai_cv/invoice.jpg",
>         strategy='ocr_only'
>     )
>     data = loader.load()
> ```

add UnstructuredImageLoader-PaddleOCR

yuming-long:
This PR https://github.com/Unstructured-IO/unstructured/pull/1811 is doing a `ocr_only` workflow refactor. And after the PR is merged, we might just need to add tests to double check paddle is working in `ocr_only` for this issue.

PR #1811, opened 2023-10-20
refactor: `partition_pdf()` for `ocr_only` strategy

christinestraub:


unstructured/partition/ocr.py
```diff
@@ -383,84 +437,66 @@ def zoom_image(image: PILImage, zoom: float = 1) -> PILImage:
...
-            or text_height > env_config.TESSERACT_MAX_TEXT_HEIGHT
-        ):
-            # rounding avoids unnecessary precision and potential numerical issues assocaited
-            # with numbers very close to 1 inside cv2 image processing
-            zoom = np.round(env_config.TESSERACT_OPTIMUM_TEXT_HEIGHT / text_height, 1)
-            ocr_df = unstructured_pytesseract.image_to_data(
-                np.array(zoom_image(image, zoom)),
+        if output_type == OCROutputType.STRING:
+            output_data = unstructured_pytesseract.image_to_string(
+                np.array(image),
```
yuming-long, line 462: we might also want to apply the same scaling logic to `image_to_string` as `image_to_data` below, but this could be an additional PR

christinestraub: Seems like a good catch! We'll consider it in an additional PR. Thanks!

unstructured/partition/ocr.py
```diff
@@ -347,6 +354,53 @@ def init_table_agent():
...
+    """
+
+    ocr_layout = get_ocr_data_from_image(
+        image,
+        ocr_languages=ocr_languages,
+        ocr_agent=ocr_agent,
+        output_type=OCROutputType.TEXT_REGIONS,
+    )
+
+    if ocr_agent == OCR_AGENT_PADDLE:
```
yuming-long, line 374: i am not sure why we need to separate the logic for paddle and tesseract here

yuming-long (commented):
Overall looks good 🎉  added few comments

unstructured/partition/ocr.py
```diff
@@ -383,84 +437,66 @@ def zoom_image(image: PILImage, zoom: float = 1) -> PILImage:
     return PILImage.fromarray(new_image)
 
 
-def get_ocr_layout_from_image(
+def get_ocr_data_from_image(
```
yuming-long, line 440: could you add docstring for this function? thanks

christinestraub: Done!

unstructured/partition/ocr.py
christinestraub: For **tesseract**, the `ocr_text` returned by `unstructured_pytesseract.image_to_string()` doesn't contain bounding box data but is well grouped. Conversely, the `ocr_layout` returned by parsing `unstructured_pytesseract.image_to_data()` contains bounding box data but is not well grouped. Therefore, we need to first group the `ocr_layout` by `ocr_text` and then merge the text regions in each group to create a list of layout elements.
For **paddle**, there is no difference in `ocr_layout` and `ocr_text` in terms of grouping because we get `ocr_text` from `ocr_layout, so the first two grouping and merging steps are not necessary.

yuming-long: gotcha good point! could you note this in the function? Thanks

christinestraub: Those two methods are combined in `get_elements_from_ocr_regions()`, which is what I did in this PR.

yuming-long (commented):
Noted some ingest changes, just want to make sure this is what we want :) Also please feel free to address the negative ingest changes in additional PR.

unstructured/partition/ocr.py
```diff
@@ -385,84 +450,86 @@ def zoom_image(image: PILImage, zoom: float = 1) -> PILImage:
...
     image: PILImage,
     ocr_languages: str = "eng",
-    ocr_agent: str = "tesseract",
-) -> List[TextRegion]:
+    ocr_agent: str = OCR_AGENT_TESSERACT,
+    output_type: OCROutputType = OCROutputType.STRING,
+) -> Union[str, List[TextRegion]]:
```
qued, line 458: In my experience a function that takes an input that specifies what format the output takes is an antipattern. I'd recommend the following:
1. Create another function (maybe named `get_ocr_text_from_image`) that handles `PILImage -> str`
2. Within both functions, spin the agent dependent code out into their own functions, so that `get_ocr_data_from_image` and it's `str` sibling are just routing functions. This already seems to be mostly done with the `parse_ocr_data_{agent}` functions, but the Tesseract preprocessing code is IMO out of place here and should live in a Tesseract-specific function. Maybe directly in `parse_ocr_data_tesseract`?

unstructured/partition/pdf.py
```diff
@@ -22,6 +21,8 @@
 from pdfminer.pdfpage import PDFPage
 from pdfminer.pdftypes import PDFObjRef
 from pdfminer.utils import open_filename
+from PIL import Image as PILImage
+from unstructured_inference.inference.layout import DocumentLayout
```
qued, line 25: This needs to be imported from within the function it's used, or we're picking this up as a base level dependency, which we don't want to do.

unstructured/partition/pdf.py
```diff
@@ -797,91 +789,64 @@ def _add_pytesseract_bboxes_to_elements(
...
+        include_page_breaks=include_page_breaks,
+        last_modification_date=metadata_last_modified,
+        # NOTE(crag): do not attempt to derive ListItem's from a layout-recognized "List"
+        # block with NLP rules. Otherwise, the assumptions in
+        # unstructured.partition.common::layout_list_to_list_items often result in weird chunking.
+        infer_list_items=False,
+        detection_origin="image" if is_image else "pdf",
+        languages=languages,
+        **kwargs,
+    )
```
qued, line 848: This seems roundabout to me... we're putting it in `unstructured-inference` format only to immediately convert it back to `unstructured` format?

What's the motivation there?

qued (changes requested):
One change needed for sure (moving the `unstructured_inference` import) and a few suggestions and questions.

unstructured/partition/ocr.py
christinestraub: @qued I combined `get_ocr_layout_from_image()` and `get_ocr_text_from_image()` to `get_ocr_data_from_image()` in this PR, following the same pattern as the `unstructured_pytesseract.image_to_data()`.
```
ocr_df = unstructured_pytesseract.image_to_data(
    np.array(image),
    lang=ocr_languages,
    output_type=Output.DATAFRAME,
)
```
Do you recommend to revert it back?

test_unstructured_ingest/expected-structured-output/pdf-fast-reprocess/azure/IRS-form-1987.pdf.json
```diff
@@ -57,10 +37,10 @@
       ],
       "page_number": 1
     },
-    "text": "Internal Revenue Service"
+    "text": "Department of the Treasury Internal Revenue Service"
```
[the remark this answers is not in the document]
christinestraub: Yes, I combined those two lines because I thought it looked better (referring to the result JSON for the [`hi_res` strategy](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/expected-structured-output/azure/IRS-form-1987.pdf.json#L19)). What do you think?

unstructured/partition/pdf.py
christinestraub: The motivation is to have a pattern similar to the `hi_res` strategy.

test_unstructured_ingest/expected-structured-output/pdf-fast-reprocess/azure/IRS-form-1987.pdf.json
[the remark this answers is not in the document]
yuming-long: Yeah combine the two makes sense here. Also I see the ingest combine words together to get one line like `Sections B-2 and B-3`, which is definitely better! Only one place I am not sure is for `"Specific Instructions Section A"` in the changes below, I feel like separate them into `"Specific Instructions"` and  `"Section A"` would be better.

yuming-long (approved):
LGTM once the test coverage is passed!

unstructured/partition/ocr.py
```diff
@@ -385,84 +430,97 @@ def zoom_image(image: PILImage, zoom: float = 1) -> PILImage:
...
+    return ocr_regions
+
+
+def get_ocr_regions_tesseract(
+    image: PILImage,
+    ocr_languages: str = "eng",
+) -> List[TextRegion]:
+    """Get the OCR regions from image as a list of text regions with tesseract."""
+
+    logger.info("Processing OCR with tesseract...")
```
yuming-long, line 475:
```suggestion
    logger.info("Processing entire page OCR with tesseract...")
```
to be consistent with paddle

unstructured/partition/ocr.py
```diff
@@ -385,84 +430,97 @@ def zoom_image(image: PILImage, zoom: float = 1) -> PILImage:
...
+
+    ocr_regions = parse_ocr_data_tesseract(ocr_df, zoom=zoom)
+
+    return ocr_regions
+
+
+def get_ocr_regions_paddle(image: PILImage) -> List[TextRegion]:
+    """Get the OCR regions from image as a list of text regions with paddle."""
+
+    logger.info("Processing entrie page OCR with paddle...")
```
yuming-long, line 514:
```suggestion
    logger.info("Processing entire page OCR with paddle...")
```

christinestraub:
> One change needed for sure (moving the `unstructured_inference` import) and a few suggestions and questions.

@qued I've updated the logic and no longer require the `unstructured_inference` import. Can you please review it again?

yuming-long:
This PR will also fix issue: https://github.com/Unstructured-IO/unstructured/issues/1176

cragwolfe (approved):

PR #1814, opened 2023-10-20
update ingest pipeline to share ingest docs via multiprocessing.manager.dict

rbiseck3:
### Description
* If the contents of a doc were updated by the process of reading/downloading it, this was not being persisted. To fix this, the data being passed around was updated to use a multiprocessing safe dict rather than the json string. Now that dict is updated after the `get_file` method is called. 
* Wikipedia connector was updated to use a static filename rather than one requiring a call to fetch data. 
* The read config param `re_download` was not being leveraged by the source node, this was fixed. 
* Added fix: chunking and embedding order reversed so chunking runs before embeddings


unstructured/ingest/connector/wikipedia.py
```diff
@@ -22,22 +22,30 @@
...
 class SimpleWikipediaConfig(BaseConnectorConfig):
     title: str
-    auto_suggest: bool
+    auto_suggest: bool = False
 
 
 @dataclass
 class WikipediaIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
     connector_config: SimpleWikipediaConfig = field(repr=False)
+    revision_id: t.Optional[int] = None
```
ryannikolaidis, line 31: actually, now that I'm looking at this, revision_id is already assigned to the version via source metadata. do we need to store as revision_id on the instance directly as well?

unstructured/ingest/pipeline/utils.py
```diff
@@ -1,8 +1,7 @@
 import hashlib
-import json
 
 
-def get_ingest_doc_hash(doc: str) -> str:
-    json_as_dict = json.loads(doc)
-    hashed = hashlib.sha256(json_as_dict.get("filename").encode()).hexdigest()[:32]
+def get_ingest_doc_hash(json_as_dict: dict) -> str:
+    print(json_as_dict.keys())
```
ryannikolaidis, line 5: do we want this print here? or was this for debugging?

unstructured/ingest/pipeline/source.py
```diff
@@ -1,21 +1,28 @@
 import typing as t
 from dataclasses import dataclass
 
-from unstructured.ingest.connector.registry import create_ingest_doc_from_json
+from unstructured.ingest.connector.registry import create_ingest_doc_from_dict
 from unstructured.ingest.interfaces import BaseSessionHandle, IngestDocSessionHandleMixin
 from unstructured.ingest.logger import logger
 from unstructured.ingest.pipeline.interfaces import SourceNode
 
+pass
```
ryannikolaidis, line 9: ?

unstructured/ingest/pipeline/source.py
```diff
@@ -1,21 +1,28 @@
...
 @dataclass
 class Reader(SourceNode):
-    def run(self, ingest_doc_json: str) -> t.Optional[str]:
+    def run(self, ingest_doc_dict: dict) -> t.Optional[str]:
         try:
             global session_handle
-            doc = create_ingest_doc_from_json(ingest_doc_json)
+            doc = create_ingest_doc_from_dict(ingest_doc_dict)
+            filename = doc.filename
+            print(f"##### Checking if {filename} exists:")
```
ryannikolaidis, line 22: debugging?

unstructured/ingest/pipeline/source.py
rbiseck3: linting mishap

unstructured/ingest/pipeline/utils.py
rbiseck3: nope, this was from a previous update I was testing. Not needed anymore.

unstructured/ingest/interfaces.py
```diff
@@ -13,17 +13,19 @@
...
 import requests
 from dataclasses_json import DataClassJsonMixin
+from dataclasses_json.core import Json, _asdict, _decode_dataclass
 
 from unstructured.chunking.title import chunk_by_title
 from unstructured.documents.elements import DataSourceMetadata
 from unstructured.embed.interfaces import BaseEmbeddingEncoder, Element
 from unstructured.embed.openai import OpenAIEmbeddingEncoder
 from unstructured.ingest.error import PartitionError, SourceConnectionError
-from unstructured.ingest.ingest_doc_json_mixin import IngestDocJsonMixin
```
ryannikolaidis, line 22: ...do we still need unstructured.ingest.ingest_doc_json_mixin in this case? or okay to remove it?

rbiseck3: This is waiting for the other PR to be merged in. I just copied it over for now.

ryannikolaidis (changes requested):
err, changes look good, but need changelog bump with description. also update the pr title as a conventional commit (and maybe a bit more descript)

ryannikolaidis (approved):
lgtm!

ahmetmeleq:
Suggested updates that might be complementary to this PR:

updating `embedding` to `chunking`
https://github.com/Unstructured-IO/unstructured/blob/40a265d0273bcdd18cb688b122e81c588c3c5172/unstructured/ingest/pipeline/reformat/chunking.py#L48 

updating `embeddings` to `chunking`
https://github.com/Unstructured-IO/unstructured/blob/40a265d0273bcdd18cb688b122e81c588c3c5172/unstructured/ingest/pipeline/reformat/chunking.py#L54

PR #1817, opened 2023-10-20
fix: partition html fail with table without tbody

### LINKED ISSUE #1807: bug/partition_html fails when parsing html with tables with no tbody
**Describe the bug**
partition_html fails when parsing tables with a thead but no tbody.

Inside `_process_leaf_table_item(tag_elem)` the following code fails with
AttributeError: 'NoneType' object has no attribute 'findall'

```
 if tag_elem.tag in TABLE_TAGS:
    nested_table = tag_elem.findall("table")
    if not nested_table:
        rows = tag_elem.findall("tr")
        if not rows:
             body = tag_elem.find("tbody")
             rows = body.findall("tr")

```
      
This happens because the table does not have a tbody. The table has a thead that contains th cells with values.

This happens in SEC Form 3,4 and 5 where the thead contains the headers for optional data that is not present in the table.
The table header is still present to show the structure of the table

[SEC Form 3](https://www.sec.gov/Archives/edgar/data/1004702/000100470223000152/xslF345X02/wk-form3_1697651740.xml)

**To Reproduce**

```
table_html = """
<table>
  <thead>
    <tr><th>Header 1</th><th>Header 2</th></tr>
    </thead>
</table>
    """
partitions = partition_html(text=table_html)
```

**Expected behavior**
partitionl_html returns a list of partitions. The table with the missing table body should still be returned as a HTMLTable

**Screenshots**

[Table with no table body](https://github.com/dgunning/edgartools/blob/main/images/tables_no_tbody.png)

**Environment Info**
OS version:  macOS-13.3.1-x86_64-i386-64bit
Python version:  3.10.10
unstructured version:  None
unstructured-inference version:  0.7.7
pytesseract version:  0.3.10
Torch version:  2.1.0
Detectron2 is not installed

**Additional context**
I am the author of [edgartools](https://github.com/dgunning/edgartools) which uses unstructured to partition SEC filing html

badGarnet:
This PR resolves #1807 
- fix a bug where when a table tagged content does not contain `tbody` tag but `thead` tag for the rows the code fails
- now when there is no `tbody` in a table section we try to look for `thead` isntead
- when both are not found return empty table

### DIFF
--- a/unstructured/documents/html.py
+++ b/unstructured/documents/html.py
@@ -39,7 +39,7 @@ TEXT_TAGS: Final[List[str]] = ["p", "a", "td", "span", "font"]
 LIST_ITEM_TAGS: Final[List[str]] = ["li", "dd"]
 LIST_TAGS: Final[List[str]] = ["ul", "ol", "dl"]
 HEADING_TAGS: Final[List[str]] = ["h1", "h2", "h3", "h4", "h5", "h6"]
-TABLE_TAGS: Final[List[str]] = ["table", "tbody", "td", "tr"]
+TABLE_TAGS: Final[List[str]] = ["table", "tbody", "td", "tr", "thead"]
 TEXTBREAK_TAGS: Final[List[str]] = ["br"]
 PAGEBREAK_TAGS: Final[List[str]] = ["hr"]
 EMPTY_TAGS: Final[List[str]] = PAGEBREAK_TAGS + TEXTBREAK_TAGS
@@ -494,8 +494,8 @@ def _process_leaf_table_item(
         if not nested_table:
             rows = tag_elem.findall("tr")
             if not rows:
-                body = tag_elem.find("tbody")
-                rows = body.findall("tr")
+                body = tag_elem.find("tbody") or tag_elem.find("thead")
+                rows = body.findall("tr") if body else []
             if len(rows) > 0:
                 table_data = [list(row.itertext()) for row in rows]
                 html_table = tabulate(table_data, tablefmt="html")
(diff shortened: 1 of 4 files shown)


dgunning (commented):
I would treat a table with no body as an empty table. I don't think the header content should be extracted as table content.

What happens on SEC Form 3 is that either Table 1 or Table 2 or both can contain data. If a table is empty using the head content makes it hard to compare against a table with data. 
It should either have content or no content.
So I would leave out the `or tag_elem.find('thead')`.

badGarnet:
> I would treat a table with no body as an empty table. I don't think the header content should be extracted as table content.
> 
> What happens on SEC Form 3 is that either Table 1 or Table 2 or both can contain data. If a table is empty using the head content makes it hard to compare against a table with data. It should either have content or no content. So I would leave out the `or tag_elem.find('thead')`.
> 
> 

that's a fair point but wouldn't it be better if partition doesn't drop contents. We could put this element as a different category or in table as html properly mark this is just headers. Downsteam application can then filter based on the need.

dgunning:
> > I would treat a table with no body as an empty table. I don't think the header content should be extracted as table content.
> > What happens on SEC Form 3 is that either Table 1 or Table 2 or both can contain data. If a table is empty using the head content makes it hard to compare against a table with data. It should either have content or no content. So I would leave out the `or tag_elem.find('thead')`.
> > 
> 
> that's a fair point but wouldn't it be better if partition doesn't drop contents. We could put this element as a different category or in table as html properly mark this is just headers. Downsteam application can then filter based on the need.

On the topic of dropping content, don't you skip the thead in the case where the tbody is present? I don't think HTMLTable gets the headers in the happy flow

amanda103 (approved):
lgtm 👍🏻

unstructured/documents/html.py
dgunning, line 1: Looks good

dgunning (approved):
I think these changes work. Thanks

PR #1819, opened 2023-10-20
fix: ModuleNotFoundError for unstructured metrics folder

shreyanid:
### Summary
Fix `ModuleNotFoundError: No module named 'unstructured.metrics'` where the unstructured/metrics folder is not discoverable from a package import since its missing an __init__.py file.


yuming-long (approved):
LGTM!

PR #1822, opened 2023-10-20
fix: model_name being None raises attribution error

### LINKED ISSUE #1754: Partition params sent as None hit AttributeError
**Describe the bug**
When model_name is passed as None, this error occurs:

```
from unstructured.partition.auto import partition_pdf
partition_pdf(filename="example-docs/layout-parser-paper.pdf", model_name=None)

File ~/unstructured/file_utils/filetype.py:565, in add_metadata_with_filetype.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
    560     params["filename"] = params.get("metadata_filename")
    562 metadata_kwargs = {
    563     kwarg: params.get(kwarg) for kwarg in ("filename", "url", "text_as_html")
    564 }
--> 565 if not cast(str, kwargs.get("model_name", "")).startswith("chipper"):
    566     # NOTE(alan): Skip hierarchy if using chipper, as it should take care of that
    567     elements = set_element_hierarchy(elements)
    569 for element in elements:
    570     # NOTE(robinson) - Attached files have already run through this logic
    571     # in their own partitioning function

AttributeError: 'NoneType' object has no attribute 'startswith'
```

Because the key is present in kwargs, the kwargs.get will return None and break. The api uses `model_name=None` as a default.

`unstructured` 0.10.22

badGarnet:
This PR resolves #1754 
- function wrapper tries to use `cast` to convert kwargs into `str` but when a value is `None` `cast(str, None)` still returns `None`
- fix replaces the conversion to simply using `str()` function call

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.25-dev6
+## 0.10.25-dev7
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


awalker4 (approved):
LGTM!

PR #1825, opened 2023-10-20
fix: partition docx from teams output

### LINKED ISSUE #1816: bug/Microsft Teams chat archive docx produces no elements
**Summary**
* Microsoft teams chat allows chat transcripts to be saved in DOCX format.
* These files use the absolute minimum DOCX structure. In particular, they omit any _sections_; paragraphs appear directly in the `<w:body>` element with no `<w:sectPr>` element to specify margins, headers, footers, etc.
* Because `partition_docx()` partitions by iterating through the document sections (because sections provide headers, footers, certain page-breaks, etc.), and there is no first section, no paragraphs or tables are found and so no elements are generated.

**To Reproduce**
[teams_chat.docx](https://github.com/Unstructured-IO/unstructured/files/13057024/teams_chat.docx)
```python
>>> partition_docx("teams_chat.docx")
[]
```

**Expected behavior**
```python
>>> partition_docx("teams_chat.docx")
[
    Text('0:0:0.0 --> 0:0:1.510\nSome Body\nOK. Yeah.')
    Text('0:0:3.270 --> 0:0:4.250\nJames Bond\nUmm.')
]
```

Related issue: #1798

badGarnet:
This PR resolves #1816 
- current docx partition assumes all contents are in sections
- this is not true for MS Teams chat transcript exported to docx
- now the code checks if there are sections or not; if not then iterate through the paragraphs and partition contents in the paragraphs


unstructured/partition/docx.py
```diff
@@ -287,6 +287,10 @@ def _iter_document_elements(self) -> Iterator[Element]:
         # -- functions like `._iter_paragraph_elements()` where the "just return when done"
         # -- characteristic of a generator avoids repeated code to form interim results into lists.
 
+        if not self._document.sections:
+            for paragraph in self._document.paragraphs:
+                yield from self._iter_paragraph_elements(paragraph)
```
scanny, line 292: This will work in this case, but not the general case, because a document can contain both paragrahs and tables and tables are of course skipped by `Document.paragraphs`.

If we're looking for a quick-fix for the Microsoft Teams chat transcript problem then I'd say this is an entirely valid approach. But longer term we probably want a solution that includes the (real) possibility of tables being in a "sectionless" document.

Offhand I'd expect tables are unlikely in a Teams chat transcript, but I've never used Teams or its chat and I suppose even if a user can't enter a table in chat, certain chat events (like I don't know, maybe multi-person joins or leaves or something) could conceivably be _rendered_ as a table when the chat is exported as DOCX.

I can sketch out a more robust solution for longer-term use and put that in a follow-up PR if we want to let this one go to get us around the immediate problem.

scanny (commented):
Hi Yao. tl;dr version is this is okay for a quick-fix but doesn't solve the general case. I'm happy to approve if we want the quick-fix and I can follow-up with a PR with a general-case fix if you want, just let me know :)

unstructured/partition/docx.py
badGarnet: hmm, do we have a way to know what possible iterables are in a docx object? Cause the impulse here is then "ok, lets add `self._document.other_iterables` to the loop". But then I guess the questions is are there documented way to access a "root" iterable? Or maybe reverse it and put everything inside a section upon loading a docx?

scanny: Yeah, I have a sketch almost done I can show you. Essentially this is the gist:

A Word document (the main document, not including headers footers, other bits like notes, footnotes, comments, etc.) can contain _block-items_, which are either paragraphs or tables.

For historical reasons, `python-docx` has `Document.paragraphs` and `Document.tables`. If we use both of those we get everything. Problem is the sequence is lost, like all paragraphs appear before all tables or vice-versa.

The way to go is to iterate through the `<w:p>` and `<w:tbl>` elements in the document body and then create a `Paragraph` or `Table` object as appropriate. This produces them in document order.

That's what I'm sketching out now.

scanny (approved):
I chatted with @cragwolfe on this and he wants the quick-fix. This LGTM for that. I'll work out the general-case along the lines of the sketch I mentioned here and put that up later this week :)

PR #1828, opened 2023-10-20
feat: add accuracy as wrapper for edit distance score

Klaijan:
Add `calculate_accuracy` function that is a wrapper of `calculate_edit_distance` that returns the "score".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Features
 
-* **Adds accurary function** The accuracy scoring was originally an option under `calculate_edit_distance`. For easy function call, it is now a wrapper around the original function that calls edit_distance and return as "score".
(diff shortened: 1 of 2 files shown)


shreyanid (approved):
great wrapper, each return type makes sense now :) thanks!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Features
 
+* **Adds accurary function** The accuracy scoring was originally an option under `calculate_edit_distance`. For easy function call, it is now a wrapper around the original function that calls edit_distance and return as "score".
```
shreyanid, line 9:
```suggestion
* **Adds accuracy function** The accuracy scoring was originally an option under `calculate_edit_distance`. For easy function call, it is now a wrapper around the original function that calls edit_distance and return as "score".
```
and bump version

PR #1835, opened 2023-10-21
build(deps): add typing extensions dep

### LINKED ISSUE #1330: TypedDict not available in python < 3.8
**Describe the bug**

**To Reproduce**
try parsing documents in python 3.7 versions as setup.py file's python requirement is >=3.7 and pip install didn't throw any errors.

**Expected behavior**

Import smoothly and parse documents.

**Screenshots**

from unstructured.documents.elements import (
  File "/home/airflow/.local/lib/python3.7/site-packages/unstructured/documents/elements.py", line 14, in <module>
    from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict, Union, cast
ImportError: cannot import name 'TypedDict' from 'typing' (/usr/local/lib/python3.7/typing.py)

**Environment Info**
This will help us understand more about the environment in which the bug occurred.

**Additional context**

Either update setup.py which will not install the library on the first hand, which is not that helpful as one have to install whole new python or add a condition which would import same from typing_extensions

qued:
Closes #1330.

Added `typing-extensions` as an explicit dependency (it was previously an implicit dependency via `dataclasses-json`).

This dependency should be explicit, since we import from it directly in `unstructured.documents.elements`. This has the added benefit that `TypedDict` will be available for Python 3.7 users.

Other changes:
* Ran `pip-compile`
* Fixed a bug in `version-sync.sh` that caused an error when using the sync functionality when syncing to a dev version from a release version.

#### Testing:

To test the Python 3.7 functionality, in a Python 3.7 environment install the base requirements and run
```python

```
This also works on `main` as `typing_extensions` is a requirement. However if you `pip uninstall typing-extensions`, and run the above code, it should fail. So this update makes sure `typing-extensions` doesn't get lost if the other dependencies move around.

To reproduce the `version-sync.sh` bug that was fixed, in `main`, increment the most recent version in `CHANGELOG.md` while leaving the version in `__version__.py`. Then add the following lines to `version-sync.sh` to simulate a particular set of circumstances, starting on line 114:

```
MAIN_IS_RELEASE=true
CURRENT_BRANCH="something-not-main"
```

Then run `make version-sync`.

The expected behavior is that the version in `__version__.py` is changed to the new version to match `CHANGELOG.md`, but instead it exits with an error.

The fix was to only do the version incrementation check when the script is running in `-c` or "check" mode.


scripts/version-sync.sh
```diff
@@ -111,6 +111,7 @@ MAIN_VERSION=$(git show origin/main:unstructured/__version__.py | grep -o -m 1 -
 MAIN_IS_RELEASE=false
 [[ $MAIN_VERSION != *"-dev"* ]] && MAIN_IS_RELEASE=true
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+MAIN_IS_RELEASE=true
```
shreyanid, line 114: this testing artifact should be removed, other than that version-sync changes have been tested and LGTM

qued: Good catch, removed in [2a92e96](https://github.com/Unstructured-IO/unstructured/pull/1835/commits/2a92e96182ac7ce9ef63d9c4cf5e4b1639eaf47a).

shreyanid (approved):
worked in 3.7 for me!

PR #1844, opened 2023-10-23
fix: chunk_by_title() interface is rude

scanny:
### `chunk_by_title()` interface is "rude"

**Executive Summary.** Perhaps the most commonly specified option for `chunk_by_title()` is `max_characters` (default: 500), which specifies the chunk window size.

When a user specifies this value, they get an error message:
```python
  >>> chunks = chunk_by_title(elements, max_characters=100)
  ValueError: Invalid values for combine_text_under_n_chars, new_after_n_chars, and/or max_characters.
```
A few of the things that might reasonably pass through a user's mind at such a moment are:
* "Is `110` not a valid value for `max_characters`? Why would that be?"
* "I didn't specify a value for `combine_text_under_n_chars` or `new_after_n_chars`, in fact I don't know what they are because I haven't studied the documentation and would prefer not to; I just want smaller chunks! How could I supply an invalid value when I haven't supplied any value at all for these?"
* "Which of these values is the problem? Why are you making me figure that out for myself? I'm sure the code knows which one is not valid, why doesn't it share that information with me? I'm busy here!"

In this particular case, the problem is that `combine_text_under_n_chars` (defaults to 500) is greater than `max_characters`, which means it would never take effect (which is actually not a problem in itself).

To fix this, once figuring out that was the problem, probably after opening an issue and maybe reading the source code, the user would need to specify:
  ```python
  >>> chunks = chunk_by_title(
  ...     elements, max_characters=100, combine_text_under_n_chars=100
  ... )
  ```

This and other stressful user scenarios can be remedied by:
* Using "active" defaults for the `combine_text_under_n_chars` and `new_after_n_chars` options.
* Providing a specific error message for each way a constraint may be violated, such that direction to remedy the problem is immediately clear to the user.

An *active default* is for example:
* Make the default for `combine_text_under_n_chars: int | None = None` such that the code can detect when it has not been specified.
* When not specified, set its value to `max_characters`, the same as its current (static) default.
This particular change would avoid the behavior in the motivating example above.

Another alternative for this argument is simply:
```python
combine_text_under_n_chars = min(max_characters, combine_text_under_n_chars)
```

### Fix

1. Add constraint-specific error messages.
2. Use "active" defaults for `combine_text_under_n_ chars` and `new_after_n_chars`.
3. Improve docstring to describe active defaults, and explain other argument behaviors, in particular identifying suppression options like `combine_text_under_n_chars = 0` to disable chunk combining.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Features
 
-* **Add Local connector source metadata** python's os module used to pull stats from local file when processing via the local connector and populates fields such as last modified time, created time.
+* **Add Local connector source metadata.** python's os module used to pull stats from local file when processing via the local connector and populates fields such as last modified time, created time.
 
 ### Fixes
 
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -159,7 +159,13 @@ def test_split_elements_by_title_and_table():
         Text("It is storming outside."),
         CheckBox(),
     ]
-    sections = _split_elements_by_title_and_table(elements, combine_text_under_n_chars=0)
+    sections = _split_elements_by_title_and_table(
+        elements,
+        multipage_sections=True,
+        combine_text_under_n_chars=0,
+        new_after_n_chars=500,
+        max_characters=500,
+    )
 
     assert sections == [
         [
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -22,6 +22,110 @@ from unstructured.documents.elements import (
 )
 from unstructured.partition.html import partition_html
 
+# == chunk_by_title() validation behaviors =======================================================
+
+
+@pytest.mark.parametrize("max_characters", [0, -1, -42])
+def test_it_rejects_max_characters_not_greater_than_zero(max_characters: int):
+    elements: List[Element] = [Text("Lorem ipsum dolor.")]
+
+    with pytest.raises(
+        ValueError, match=f"'max_characters' argument must be > 0, got {max_characters}"
+    ):
+        chunk_by_title(elements, max_characters=max_characters)
+
+
+def test_it_does_not_complain_when_specifying_max_characters_by_itself():
+    """Caller can specify `max_characters` arg without specifying any others.
+
+    In particular, When `combine_text_under_n_chars` is not specified it defaults to the value of
+    `max_characters`; it has no fixed default value that can be greater than `max_characters` and
+    trigger an exception.
+    """
+    elements: List[Element] = [Text("Lorem ipsum dolor.")]
+
+    try:
+        chunk_by_title(elements, max_characters=50)
+    except ValueError:
+        pytest.fail("did not accept `max_characters` as option by itself")
+
+
+@pytest.mark.parametrize("n_chars", [-1, -42])
+def test_it_rejects_combine_text_under_n_chars_for_n_less_than_zero(n_chars: int):
+    elements: List[Element] = [Text("Lorem ipsum dolor.")]
+
+    with pytest.raises(
+        ValueError, match=f"'combine_text_under_n_chars' argument must be >= 0, got {n_chars}"
+    ):
+        chunk_by_title(elements, combine_text_under_n_chars=n_chars)
+
+
+def test_it_accepts_0_for_combine_text_under_n_chars_to_disable_chunk_combining():
+    """Specifying `combine_text_under_n_chars=0` is how a caller disables chunk-combining."""
+    elements: List[Element] = [Text("Lorem ipsum dolor.")]
+
+    chunks = chunk_by_title(elements, max_characters=50, combine_text_under_n_chars=0)
+
+    assert chunks == [CompositeElement("Lorem ipsum dolor.")]
+
+
+def test_it_does_not_complain_when_specifying_combine_text_under_n_chars_by_itself():
+    """Caller can specify `combine_text_under_n_chars` arg without specifying any other options."""
+    elements: List[Element] = [Text("Lorem ipsum dolor.")]
+
+    try:
+        chunk_by_title(elements, combine_text_under_n_chars=50)
+    except ValueError:
+        pytest.fail("did not accept `combine_text_under_n_chars` as option by itself")
+
+
+@pytest.mark.parametrize("n_chars", [-1, -42])
+def test_it_rejects_new_after_n_chars_for_n_less_than_zero(n_chars: int):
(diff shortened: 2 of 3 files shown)


unstructured/chunking/title.py
```diff
@@ -73,27 +73,53 @@ def chunk_by_title(
...
-    ):
+    # -- validation and arg pre-processing ---------------------------
+
+    # -- chunking window must have positive length --
+    if max_characters <= 0:
+        raise ValueError(f"'max_characters' argument must be > 0, got {max_characters}")
+
+    # -- `combine_text_under_n_chars` defaults to `max_characters` when not specified and is
+    # -- capped at max-chars
+    if combine_text_under_n_chars is None or combine_text_under_n_chars > max_characters:
```
shreyanid, line 100: It makes sense to set `combine_text_under_n_chars` to `max_characters` when not specified, but when it is specified but is greater than `max_characters`, we are just capping it at max_character without a warning that the user input was capped. Perhaps I need a bit more context on this, but does it make sense to cap the user-inputted value without notifying the user of the change?

shreyanid:
Looking at the [chunking documentation](https://unstructured-io.github.io/unstructured/bricks/chunking.html#chunking) and the docstring of chunk_by_title in this PR, I'm trying to understand the relationship between these 3 parameters combine_text_under_n_chars, new_after_n_chars, and max_characters

- new_after_n_chars is after how many characters a section should be capped and a new section should start. Can this value in theory be greater than max_characters?
- combine_text_under_n_chars is if a section has under this many characters, it will be combined with another(?) section
- max_characters (not mentioned in the documentation) is to specify a "chunk size"? for elements text specifically? - im not sure how this hard max and the soft max new_after_n_chars act differently.

I am very new to chunking/RAG but I can review given a little context/crash course on how each of these are used :)

unstructured/chunking/title.py
```diff
@@ -73,27 +73,53 @@ def chunk_by_title(
...
+        raise ValueError(f"'new_after_n_chars' argument must be >= 0, got {new_after_n_chars}")
+
+    # -- `new_after_n_chars` takes precendence on conflict with `combine_text_under_n_chars` --
+    if combine_text_under_n_chars > new_after_n_chars:
+        combine_text_under_n_chars = new_after_n_chars
+
+    # ----------------------------------------------------------------
+
     chunked_elements: List[Element] = []
     sections = _split_elements_by_title_and_table(
```
shreyanid, line 124: the defaults to this (private) function are still 500, please remove

unstructured/chunking/title.py
shreyanid: discussion update: the behavior when `combine_text_under_n_chars` is equal to or greater than `max_characters` is the same, so the capping can be done without notification

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Features
```
shreyanid, line 8: add changelog message (and bump dev version before merging)

scanny: Hi @shreyanid, Crag said don't add a changelog message for a change that doesn't add a new feature or fix a bug. That's why I made the trivial change to the CHANGELOG to fool the checker (since that's the only way it will approve without a new entry).

test_unstructured/chunking/test_title.py
```diff
@@ -22,6 +22,100 @@
...
+def test_it_rejects_new_after_n_chars_for_n_less_than_zero(n_chars: int):
+    elements: List[Element] = [Text("Lorem ipsum dolor.")]
+
+    with pytest.raises(
+        ValueError, match=f"'new_after_n_chars' argument must be >= 0, got {n_chars}"
+    ):
+        chunk_by_title(elements, new_after_n_chars=n_chars)
+
+
+def test_it_does_not_complain_when_specifying_new_after_n_chars_by_itself():
```
shreyanid, line 82: Should `test_it_does_not_complain_when_specifying_combine_text_under_n_chars_by_itself` also exist (I see it exists here and for max_characters)? If not, what other param is required with it, and can we test that it warns/errs as needed?

CHANGELOG.md
shreyanid: Right, that works then!

test_unstructured/chunking/test_title.py
```diff
@@ -395,43 +495,6 @@ def test_add_chunking_strategy_on_partition_html_respects_multipage():
...
 
 
-@pytest.mark.parametrize(
-    ("combine_text_under_n_chars", "new_after_n_chars", "max_characters"),
-    [
-        (-1, -1, -1),  # invalid chunk size
-        (0, 0, 0),  # invalid max_characters
-        (-5666, -6777, -8999),  # invalid chunk size
-        (-5, 40, 50),  # invalid chunk size
-        (50, 70, 20),  # max_characters needs to be greater than new_after_n_chars
```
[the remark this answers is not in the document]
scanny: Sure, added a test for that.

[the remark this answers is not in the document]
scanny: All the cases from the old test were already covered.

shreyanid (approved):

unstructured/chunking/title.py
```diff
@@ -56,8 +56,8 @@ def chunk_table_element(element: Table, max_characters: int = 500) -> List[Table
 def chunk_by_title(
     elements: List[Element],
     multipage_sections: bool = True,
-    combine_text_under_n_chars: int = 500,
-    new_after_n_chars: int = 500,
+    combine_text_under_n_chars: int | None = None,
```
qued, line 59: As far as I can tell, this syntax wasn't introduced until Python 3.10. On the other hand, CI isn't complaining (maybe because typing isn't really enforced?). Everywhere else we're using the `Optional[int]` syntax, I'm inclined to stick with that here.

scanny: This syntax is accepted in Python 3.7+ (and maybe further back) when `from __future__ import annotations` appears in the import section.

This "pipe == Union" syntax is recommended as Union will be deprecated, so it's the "way forward" so to speak, which is why I started using it. Let me see if I can find a reference for that ... here's one: https://mail.python.org/archives/list/typing-sig@python.org/message/624BEVFNZXFEQP6ZWIGB4YKSPMCBE2M6/

next response in that thread is Guido approving.

They `from __future__ import annotations` also gives some other very nice benefits like not requiring forward references to be quoted and a few other things.

scanny: Ok, fix is in for this @Alan :)

scanny: Ok, fix is in for this @qued :)

PR #1846, opened 2023-10-23
refactor ingest CLI for better code reuse

rbiseck3:
### Description
Much of the current CLI code is copy-paste across subcommands. To alleviate this, most of the duplicate code was moved into base classes for src and destination connector commands. This also allows for code reuse when a destination command is called and it no longer has to jump through hoops to dynamically recreate what _would_ have been called by a source command.

The reason everything can't live in a single BaseCmd class is due to the need for a dynamic map to the source command. This runs into a circular dependency issue if it was all in one class. By splitting it into a `BaseSrcCmd` and a `BaseDestCmd` class, this helps avoid that issue.


ryannikolaidis (approved):
makes sense to me! totally agree on the reuse of code here.

PR #1850, opened 2023-10-23
Chore: stop passing extract_tables to inference and note table regression on entire doc OCR

yuming-long:
### Summary

A follow up ticket on https://github.com/Unstructured-IO/unstructured/pull/1801, I forgot to remove the lines that pass extract_tables to inference, and noted the table regression if we only do one OCR for entire doc

**Tech details:**
* stop passing `extract_tables` parameter to inference
* added table extraction ingest test for image, which was skipped before, and the "text_as_html" field contains the OCR output from the table OCR refactor PR
* replaced `assert_called_once_with` with `call_args` so that the unit tests don't need to test additional parameters
* added `error_margin` as ENV when comparing bounding boxes of`ocr_region` with `table_element`
* added more tests for tables and noted the table regression in test for partition pdf

### Test
* for stop passing `extract_tables` parameter to inference, run test `test_partition_pdf_hi_res_ocr_mode_with_table_extraction` before this branch and you will see warning like `Table OCR from get_tokens method will be deprecated....`, which means it called the table OCR in inference repo. This branch removed the warning.


christinestraub (approved):
For low-resolution documents, setting the appropriate `error_margin` seems to improve table OCR results. LGTM!

PR #1855, opened 2023-10-23
Reorganized the Examples section in Documentation & add Databricks example

ron-unstructured:
To test:
> cd docs && make html

Change log:
- Previously, the three examples were on one page, ie. the "Examples" page, ie. (i) Sentiment Analysis Labeling in LabelStudio, (ii) Extracting Metadata from Documents, and (iii) Exploring Source Documents. This PR improves the organization by splitting each example into its own page under the "Examples" parent page.
- Added a new example to work with Databricks Delta Table and Connector.


MthwRobinson:
The three original examples in there are all pretty old. "Sentiment Analysis Labeling in LabelStudio" uses the SEC filings repo, which we're not actively maintaining and "Extracting Metadata from Documents" and "Exploring Source Documents" use sections of the library that, while still functional, aren't really core user workflows. Can you drop these examples and pull in two more recent workflow examples? Those could be based on notebooks you've already worked up.

MthwRobinson:
@ron-unstructured - Is this one ready for review?

MthwRobinson (approved):

PR #1858, opened 2023-10-24
fix: sectioner does not consider separator length

scanny:
### sectioner-does-not-consider-separator-length

**Executive Summary.** A primary responsibility of the sectioner is to minimize the number of chunks that need to be split mid-text. It does this by computing text-length of the section being formed and "finishing" the section when adding another element would extend its text beyond the window size.

When element-text is consolidated into a chunk, the text of each element is joined, separated by a "blank-line" (`"\n\n"`). The sectioner does not currently consider the added length of separators (2-chars each) and so forms sections that need to be split mid-text when chunked.

Chunk-splitting should only be necessary when the text of a single element is longer than the chunking window.

**Example**

  ```python
    elements: List[Element] = [
        Title("Chunking Priorities"),  # 19 chars
        ListItem("Divide text into manageable chunks"),  # 34 chars
        ListItem("Preserve semantic boundaries"),  # 28 chars
        ListItem("Minimize mid-text chunk-splitting"),  # 33 chars
    ]  # 114 chars total but 120 chars with separators

    chunks = chunk_by_title(elements, max_characters=115)
  ```

  Want:

  ```python
[
    CompositeElement(
        "Chunking Priorities"
        "\n\nDivide text into manageable chunks"
        "\n\nPreserve semantic boundaries"
    ),
    CompositeElement("Minimize mid-text chunk-splitting"),
]
  ```

  Got:

  ```python
[
    CompositeElement(
        "Chunking Priorities"
        "\n\nDivide text into manageable chunks"
        "\n\nPreserve semantic boundaries"
        "\n\nMinimize mid-text chunk-spli"),
    )
    CompositeElement("tting")
  ```

### Technical Summary

Because the sectioner does not consider separator (`"\n\n"`) length when it computes the space remaining in the section, it over-populates the section and when the chunker concatenates the element text (each separated by the separator) the text exceeds the window length and the chunk must be split mid-text, even though there was an even element boundary it could have been split on.

### Fix

Consider separator length in the space-remaining computation.

The solution here extracts both the `section.text_length` and `section.space_remaining` computations to a `_TextSectionBuilder` object which removes the need for the sectioner (`_split_elements_by_title_and_table()`) to deal with primitives (List[Element], running text length, separator length, etc.) and allows it to focus on the rules of when to start a new section.

This solution may seem like overkill at the moment and indeed it would be except it forms the foundation for adding section-level chunk combination (fix: dissociated title elements) in the next PR. The objects introduced here will gain several additional responsibilities in the next few chunking PRs in the pipeline and will earn their place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.10.28-dev0
+
+### Enhancements
+
+### Features
+
+### Fixes
+
+* **Fix unnecessary mid-text chunk-splitting.** The "pre-chunker" did not consider separator blank-line ("\n\n") length when grouping elements for a single chunk. As a result, sections were frequently over-populated producing a over-sized chunk that required mid-text splitting.
+
 ## 0.10.27
 
 ### Enhancements
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -5,7 +5,11 @@ from typing import List
 import pytest
 
 from unstructured.chunking.title import (
+    _NonTextSection,
     _split_elements_by_title_and_table,
+    _TableSection,
+    _TextSection,
+    _TextSectionBuilder,
     chunk_by_title,
 )
 from unstructured.documents.coordinates import CoordinateSystem
@@ -15,6 +19,7 @@ from unstructured.documents.elements import (
     CoordinatesMetadata,
     Element,
     ElementMetadata,
+    ListItem,
     RegexMetadata,
     Table,
     Text,
@@ -190,6 +195,7 @@ def test_split_elements_by_title_and_table():
         Text("It is storming outside."),
         CheckBox(),
     ]
+
     sections = _split_elements_by_title_and_table(
         elements,
         multipage_sections=True,
@@ -198,29 +204,40 @@ def test_split_elements_by_title_and_table():
         max_characters=500,
     )
 
-    assert sections == [
-        [
-            Title("A Great Day"),
-            Text("Today is a great day."),
-            Text("It is sunny outside."),
-        ],
-        [
-            Table("<table></table>"),
-        ],
-        [
-            Title("An Okay Day"),
-            Text("Today is an okay day."),
-            Text("It is rainy outside."),
-        ],
-        [
-            Title("A Bad Day"),
-            Text("Today is a bad day."),
-            Text("It is storming outside."),
-        ],
-        [
-            CheckBox(),
-        ],
+    section = next(sections)
+    assert isinstance(section, _TextSection)
+    assert section.elements == [
+        Title("A Great Day"),
+        Text("Today is a great day."),
+        Text("It is sunny outside."),
     ]
+    # --
+    section = next(sections)
+    assert isinstance(section, _TableSection)
+    assert section.table == Table("<table></table>")
+    # ==
+    section = next(sections)
+    assert isinstance(section, _TextSection)
+    assert section.elements == [
+        Title("An Okay Day"),
+        Text("Today is an okay day."),
+        Text("It is rainy outside."),
+    ]
+    # --
+    section = next(sections)
+    assert isinstance(section, _TextSection)
+    assert section.elements == [
+        Title("A Bad Day"),
+        Text("Today is a bad day."),
+        Text("It is storming outside."),
+    ]
+    # --
+    section = next(sections)
+    assert isinstance(section, _NonTextSection)
+    assert section.element == CheckBox()
+    # --
+    with pytest.raises(StopIteration):
+        next(sections)
 
 
 def test_chunk_by_title():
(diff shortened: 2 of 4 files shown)


unstructured/chunking/title.py
```diff
@@ -121,31 +123,29 @@ def chunk_by_title(
     # ----------------------------------------------------------------
```
Coniferish, line 123: Commenting here since I'm unable to comment on the part I want to.

I'm not really understanding the difference between `new_after_n_chars` and `max_characters` from the docstring description. Why would users use `new_after_n_chars` when they can just use `max_characters`? Are they intended to be used together for some specific use case?

unstructured/chunking/title.py
```diff
@@ -198,59 +198,89 @@ def _split_elements_by_title_and_table(
...
+          into sections as big as possible without exceeding the chunk window size.
 
-        new_section = (
-            (section_length + len(str(element)) > max_characters)
-            or (isinstance(element, Title) and section_length > combine_text_under_n_chars)
-            or (metadata_differs or section_length > new_after_n_chars)
+        * **Minimize chunks that must be split mid-text.** Precompute the text length of each
+          section and only produce a section that exceeds the chunk window size when there is a
+          single element with text longer than that window.
+    """
```
Coniferish, line 222: What is the behavior when the last element is less than `combine_text_under_n_chars` (or if the only element is less than that)? We might want to add a test for that.

unstructured/chunking/title.py
```diff
@@ -198,59 +198,89 @@ def _split_elements_by_title_and_table(
...
+    metadata1 = preceding_element.metadata
+    metadata2 = element.metadata
     if metadata1.section != metadata2.section:
         return True
-    if include_pages and metadata1.page_number != metadata2.page_number:
-        return True
-    return False
+    if ignore_page_numbers:
+        return False
+    return metadata1.page_number != metadata2.page_number
```
Coniferish, line 285: Again, can't comment on the part I want to, but Line 303, there's a typo: "withoptional"

unstructured/chunking/title.py
```diff
@@ -198,59 +198,89 @@ def _split_elements_by_title_and_table(
...
 
 
 _P = ParamSpec("_P")
```
Coniferish, line 286: (repeat) A clearer distinction between the uses of `max_characters` and `new_after_n_chars` would be helpful in the `add_chunking_strategy` decorator.

test_unstructured/chunking/test_title.py
```diff
@@ -195,29 +201,40 @@ def test_split_elements_by_title_and_table():
...
+        Text("Today is a bad day."),
+        Text("It is storming outside."),
     ]
+    # --
+    section = next(sections)
+    assert isinstance(section, _NonTextSection)
+    assert section.element == CheckBox()
+    # --
+    with pytest.raises(StopIteration):
+        next(sections)
```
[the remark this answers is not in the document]
scanny: The behavior for `combine_text_under_n_chars` is uniform for values `>= max_characters`, that behavior being always combine up to the hard chunking window. Testing that it _doesn't_ do something at that threshold is like trying to prove a negative, so not something we can test effectively, however the tests added in the PR stacked on this one do a thorough job of exercising and verifying `_SectionCombiner` behavior, which is the only place this parameter is operative. So I'd be inclined to rely on those tests for unit-level verification: (link might break on rebase) https://github.com/Unstructured-IO/unstructured/pull/1861/files#diff-3b2d43d641f8967f696e3ad0e2c6fe2d726852023de07299fbecd5c4060cd729R808 if it does break just look for `test_title.py::Describe_SectionCombiner`.

unstructured/chunking/title.py
scanny: Well, I'm not sure i'm clear about the intended use case, but this one is plausible to me and the only one that's occurred to me so far:

* I (the library end-user) want chunks that are around 250 chars because that's currently optimal for my language model platform.
* However, I would rather exceed that window somewhat than split a chunk mid-text if I can do so within 300 chars.
* So I set `new_after_n_chars = 250`, which makes most of my chunks around that size, and I set `max_characters = 300`, which allows slightly larger elements to be chunked whole, but ensures I never get any chunks bigger than that even if it means splitting a chunk mid-text.
* I've found through experimentation that this gives me the best balance on my particular corpus of documents.

I'm not confident in this account, this is just the only plausible idea I've had, so I haven't presumed to update the docstring.

I could ask around on this though and consider updating the docstring in a subsequent PR if folks thought this was right. The documentation would have to be updated too I expect.

unstructured/chunking/title.py
scanny: Not sure what you're asking exactly, do you mean "last" like "preceding" or like "terminal element".

But elements don't really come into combination, only whole sections do. This is what the next stacked PR fixes, so I'd look at those tests and see if you think I've missed something.

So this PR doesn't try to change combination behavior or improve the tests for it, that is left to the next PR.

unstructured/chunking/title.py
scanny: Ah, yes, so there is :)

I've fixed that, must have been in there for a long time :)

unstructured/chunking/title.py
scanny: Yeah, I'll see what I can do there. Those are quite short and I'm not even sure where that docstring shows up. I'm not sure what I'll be able to do in the space constraints. Also it's kind of outside the scope of this PR to get into general clean-up. But I'll have a think and see what I can come up with, probably not in this PR though, I'd like to ship this one unless there's something substantial to be fixed, can probably slip this into the next PR.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.10.28dev0
```
qued, line 1:
```suggestion
## 0.10.28-dev0
```

Coniferish:
Doesn't look like I can approve, but LGTM!

Coniferish (approved):
LGTM!

PR #1862, opened 2023-10-24
build: element type frequency evaluation metrics workflow in ci

Klaijan:
**Executive Summary**
Measured element type frequency accuracy from the current version of code with the expected output. The performance is reported as tsv file under `metrics`.

**Technical Details**
- The evaluation measures element type frequencies from `structured-output-eval` against `expected-structured-output`
- `evaluation.py` has been edited to support function calling using `click.group()` and `command()`
- `evaluation-ingest-cp.sh` is now added to all the `test-ingest-xx.sh` scripts 

**Outputs**
2 tsv files is saved
image
image
9-0e05-41d4-b69f-841a2aa131ec)
and aggregated score is displayed.
image


unstructured/ingest/evaluate.py
```diff
@@ -123,6 +132,73 @@ def measure_edit_distance(
...
+    rows = []
+    accuracy_scores: List[float] = []
+
+    for doc in output_list:  # type: ignore
+        fn = (doc.split("/")[-1]).split(".json")[0]
+        connector = doc.split("/")[0]
+        if doc in source_list:  # type: ignore
+            output = get_element_type_frequency(_read_json(os.path.join(output_dir, doc)))
+            source = get_element_type_frequency(_read_json(os.path.join(source_dir, doc)))
+            accuracy = round(calculate_element_type_percent_match(output, source), 3)
```
shreyanid, line 180: should the result rounding be built into the `calculate_element_type_percent_match` function directly?

Klaijan: I think it is ok to leave it here in case we or the user wants more precision.

unstructured/ingest/evaluate.py
```diff
@@ -164,6 +240,8 @@ def _display(rows, headers):
 
 
 def _mean(scores: List[float], rounding: Optional[int] = 3):
+    if len(scores) < 1:
+        return None
```
shreyanid, line 244: are we returning None here because we want the result to be empty rather than 0 or something?

Klaijan: Yes, because if there's no value, we do not want to say the average is zero. But let me also add another condition for 0 real quick.

unstructured/ingest/evaluate.py
```diff
@@ -123,6 +132,73 @@ def measure_edit_distance(
...
+
+    rows = []
+    accuracy_scores: List[float] = []
+
+    for doc in output_list:  # type: ignore
+        fn = (doc.split("/")[-1]).split(".json")[0]
+        connector = doc.split("/")[0]
+        if doc in source_list:  # type: ignore
+            output = get_element_type_frequency(_read_json(os.path.join(output_dir, doc)))
+            source = get_element_type_frequency(_read_json(os.path.join(source_dir, doc)))
```
shreyanid, line 179: what are we currently using for gold standard element output, the expected_structured_output? From my understanding, the expected_structured_output represents the current state of the unstructured output for a given document, which might not necessarily be the gold standard human element division/classification. 

We will definitely need to start an odetta task to get these gold standard outputs, but in the meantime if it's only a few small documents maybe it should be done manually. By using expected_structured_output, the element accuracy result will always be 100% because by construction the expected_structured_output has to match the unstructured output exactly

unstructured/ingest/evaluate.py
shreyanid: hm, that's ok too, but I was imagining it could be an input ("rounding") to that calculate_element_type_percent_match function with a default value of say 3, so it can always be changed or removed for more precision there if desired. maybe we can make a different task / ticket to move this later

yuming-long (approved):
LGTM! 🎉

unstructured/ingest/evaluate.py
```diff
@@ -164,9 +240,16 @@ def _display(rows, headers):
...
+    elif len(scores) == 1:
+        mean = scores[0]
+    else:
+        mean = statistics.mean(scores)
     if not rounding:
-        return statistics.mean(scores)
-    return round(statistics.mean(scores), rounding)
+        return mean
+    else:
+        return round(mean, rounding)
```
badGarnet, line 252:
```suggestion
    return round(mean, rounding)
```

Issue #1859, opened 2023-10-24
bug/google drive connector don't return metadata for PNG


benjats07:
**Describe the bug**
Google drive connector don't return metadata when extracting information from PNG

**Expected behavior**
Metadata should be returned

PR #1863, opened 2023-10-25
fix: filename missing from image metadata

qued:
Closes [#1859](https://github.com/Unstructured-IO/unstructured/issues/1859).

* **Fixes elements partitioned from an image file missing certain metadata** Metadata for image files, like file type, was being handled differently from other file types. This caused a bug where other metadata, like the file name, was being missed. This change brought metadata handling for image files to be more in line with the handling for other file types so that file name and other metadata fields are being captured.

Additionally:
* Added test to verify filename is being captured in metadata
* Cleaned up `CHANGELOG.md` formatting

#### Testing:
The following produces output `None` on `main`, but outputs the filename `layout-parser-paper-fast.jpg` on this branch:
```python
elements = partition("example-docs/layout-parser-paper-fast.jpg")
print(elements[0].metadata.filename)

```


ryannikolaidis (approved):
perfect! test in the description worked exactly as suggested. thanks for jumping on this!

PR #1864, opened 2023-10-25
Fix ingest test CI job

rbiseck3:
Fixed some syntax errors in the update ingest CI job causing it to always fail

### DIFF
--- a/.github/workflows/ingest-test-fixtures-update-pr.yml
+++ b/.github/workflows/ingest-test-fixtures-update-pr.yml
@@ -6,6 +6,7 @@ on:
 
 env:
   GHA_CACHE_KEY_VERSION: "v1"
+  PYTHON_VERSION: "3.10"
 
 jobs:
   setup:
@@ -15,7 +16,6 @@ jobs:
       (github.event_name == 'push' && contains(github.event.head_commit.message, 'ingest-test-fixtures-update'))
     env:
       NLTK_DATA: ${{ github.workspace }}/nltk_data
-      PYTHON_VERSION: "3.10"
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ env.PYTHON_VERSION }}
@@ -37,9 +37,6 @@ jobs:
           make install-ci
 
   setup_ingest:
-    strategy:
-      matrix:
-        python-version: [ "3.8","3.9","3.10","3.11" ]
     runs-on: ubuntu-latest
     env:
       NLTK_DATA: ${{ github.workspace }}/nltk_data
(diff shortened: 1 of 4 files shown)


.github/workflows/ingest-test-fixtures-update-pr.yml
```diff
@@ -37,9 +37,6 @@ jobs:
           make install-ci
 
   setup_ingest:
-    strategy:
-      matrix:
-        python-version: [ "3.8","3.9","3.10","3.11" ]
```
ryannikolaidis, line 42: this would put everything on 3.10. there are still [a number of ingest tests ](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest.sh#L49-L58)we DO run through the entire matrix

rbiseck3: But what would that mean for the update CI job? A different PR for each version? Since they run all tests if the python version if `3.10`, I thought this made sense.

ryannikolaidis: update CI job only runs in 3.10. the fixtures themselves hopefully don't change based on python version, but we do want to test select flows across all versions to make sure the code/flow doesn't break across versions

rbiseck3: Isn't that what the CI job is for? This should only be to generate new outputs if they've changed.

ryannikolaidis: ...well..that's how it did run, until it was updated recently

ryannikolaidis: ohh sorry, I thought this was CI.yaml, my bad
yes, let's DEFINITELY undo this change..

ryannikolaidis: ...I never realized that the matrix was introduced to the ingest script which is what I think confused me

rbiseck3: Sorry, still confused, what reason should the update ingest test job be running through the matrix? Only `3.10` runs through all the ingest tests anyways.

ryannikolaidis (approved):
sorry, realized you were undoing the changes in ingest-test-fixture from [this pr](https://github.com/Unstructured-IO/unstructured/pull/1799/files#diff-dd400be1d149e0275e7f8c7dddc054b97d0d829ebad86a12af349deeb26cf2c9), which yes needs to be done.

changes here on the quotes make sense. 

investiation on:
> I can also look into github syntax and see if we can create a reusable step which gets run for the setup as well as for on the fly if no cache found. 

can happen in another pr if that's easier

.github/workflows/ingest-test-fixtures-update-pr.yml
rbiseck3: Or do you mean lets DEFINITELY get rid of the matrix?

rbiseck3:
To get others unblocked, keeping this PR as small as possible to only include what's needed to get this CI job running successfully again.

.github/workflows/ingest-test-fixtures-update-pr.yml
ryannikolaidis: yes, get rid of it in the ingest test fixture workflow..so return it to how it was

ryannikolaidis: ...which is what you're doing here

ryannikolaidis:
> To get others unblocked, keeping this PR as small as possible to only include what's needed to get this CI job running successfully again.

yep, thank you! totally appreciate that.

Issue #1521, opened 2023-09-25
feat/Decouple Partitioning User API and Implementation


newelh:
## Description
The user API is tightly bound to the internal implementation of the partitioners. As a result, changes in keyword arguments or function behavior immediately impact the user API. Users have direct access to the implementation functions, as indicated in the current documentation. This tight coupling restricts our flexibility to refactor or enhance the partitioners without introducing breaking changes.

The proposed changes are intended to increase the speed of development and eventually provide the ability to introduce new user facing APIs in the future.

## Proposed Solution
Introduce an abstraction layer between the user API and the partitioner implementations. This layer would expose only the necessary functionalities and delegate to the appropriate partitioners. This decoupling allows for more flexibility in internal changes without affecting the user API. A phased approach would be needed, where this issue would be the first step of the following:

1. Move the contents of the partition function for each partitioner into a private function `_partition_{type}`. For example, the email partitioner. While this seems trivial, it will enable us to further develop the partitioners without obstructing APIs that users may have in production. It will enable us to simplify and consolidate functionality in the partitioners so that we can eventually support a simpler interface for users in the future. 
```
@process_metadata()
@add_metadata_with_filetype(FileType.EML)
@add_chunking_strategy()
def partition_email(
    filename: Optional[str] = None,
    file: Optional[Union[IO[bytes], SpooledTemporaryFile]] = None,
    text: Optional[str] = None,
    content_source: str = "text/html",
    encoding: Optional[str] = None,
    include_headers: bool = False,
    max_partition: Optional[int] = 1500,
    include_metadata: bool = True,
    metadata_filename: Optional[str] = None,
    metadata_last_modified: Optional[str] = None,
    process_attachments: bool = False,
    attachment_partitioner: Optional[Callable] = None,
    min_partition: Optional[int] = 0,
    chunking_strategy: Optional[str] = None,
    **kwargs,
) -> List[Element]:
    """ Wrap the _partition_email function to separate user facing API from internal API"""
    return _partition_email(
        filename=filename,
        file=file,
        text=text,
        content_source=content_source,
        encoding=encoding,
        include_headers=include_headers,
        max_partition=max_partition,
        include_metadata=include_metadata,
        metadata_filename=metadata_filename,
        metadata_last_modified=metadata_last_modified,
        process_attachments=process_attachments,
        attachment_partitioner=attachment_partitioner,
        min_partition=min_partition,
        chunking_strategy=chunking_strategy,
        **kwargs,
    )

[...]

PR #1872, opened 2023-10-25
refactor text.py

### LINKED ISSUE #1520: refactor text.py
Refactor text.py so it passes strict type checking, uses "_private" methods, etc. 
Similar to https://github.com/Unstructured-IO/unstructured/pull/1476

Coniferish:
### Summary
Closes #1520 
Partial solution to #1521 

- Adds an abstraction layer between the user API and the partitioner implementation
- Adds comments explaining paragraph chunking
- Makes edits to pass strict type-checking for both text.py and test_text.py


test_unstructured/chunking/test_title.py
```diff
@@ -30,7 +30,8 @@ def test_it_rejects_max_characters_not_greater_than_zero(max_characters: int):
     elements: List[Element] = [Text("Lorem ipsum dolor.")]
 
     with pytest.raises(
-        ValueError, match=f"'max_characters' argument must be > 0, got {max_characters}"
+        ValueError,
+        match=f"'max_characters' argument must be > 0, got {max_characters}",
```
scanny, line 34: @Coniferish I think your local Black settings might be off because these changes are not required by our Black configuration and in fact not desired. If parameters fit on a single line they are left that way (without a trailing comma), which conserves vertical space (how much will fit in a single screenful of code).

unstructured/chunking/title.py
```diff
@@ -104,7 +104,7 @@ def chunk_by_title(
     # -- but a negative value is not
     if combine_text_under_n_chars < 0:
         raise ValueError(
-            f"'combine_text_under_n_chars' argument must be >= 0, got {combine_text_under_n_chars}"
```
scanny, line 107: Trailing comma is not necessary here, is it? Why would we want that?

scanny (approved):
LGTM aside from a few formatting changes that look unnecessary to me. Approving in advance :)

unstructured/chunking/title.py
Coniferish: These edits are coming up automatically when I try to commit--I believe it comes from the `[pre-commit](https://github.com/Unstructured-IO/unstructured#installation-instructions-for-local-development)` configuration.

This is what happens when I try to commit:
```
(other logging info above...)
ruff.....................................................................Failed
- hook id: ruff
- files were modified by this hook
```
Then I have unstaged changes (the added commas), so I add them and can then commit and push.

scanny: Hmm, I see. The operation of that pre-commit hook would explain why it's not happening on my machine (and why CI didn't complain about it when I added those lines).

I think that hook is missing something ... here it is, `COM812` in the ignore line for `ruff`, you can see it in the `pyproject.toml` here, which is what CI uses https://github.com/Unstructured-IO/unstructured/blob/main/pyproject.toml#L21

That needs to be added on line 31 of `.pre-commit-config.yaml`, might need to uninstall and reinstall the pre-commit hook for that to take effect.

Coniferish: 🙏 Sounds good. I'll create s PR and do that.

Coniferish:
@qued

PR #1877, opened 2023-10-25
chore: add metric helper for table structure eval

badGarnet:
- add helper to run inference over an image or pdf of table and compare it against a ground truth csv file
- this metric generates a similarity score between 1 and 0, where 1 is perfect match and 0 is no match at all
- add example docs for testing
- NOTE: this metric is only relevant to table structure detection. Therefore the input should be just the table area in an image/pdf file; we are not evaluating table element detection in this metric


ajjimeno (approved):
For the intended evaluation LGTM.

On the other hand, I am wondering if in the future we should consider the colspan values, e.g. you can have two rows that look the same in the csv format but a column might span across several cells in one case and another column in another but still look the same in the csv file.

badGarnet:
> For the intended evaluation LGTM.
> 
> On the other hand, I am wondering if in the future we should consider the colspan values, e.g. you can have two rows that look the same in the csv format but a column might span across several cells in one case and another column in another but still look the same in the csv file.

in this scheme we have cells span multiple columns/rows have one cell filled then the other cell being left empty. So if prediction split those cells or combined cells where it shouldn't we'd see differences. See here for more example on the score (without normalizing to 0-1): https://www.notion.so/Table-token-ration-metric-5a3222b430aa4c0a95da17c3fff6b011

PR #1878, opened 2023-10-25
build(deps): remove ebooklib

### LINKED ISSUE #1871: fix/remove ebooklib from unstructured
Due to licensing incompatibility with AGPL, remove ebooklib from unstructured requirements.

qued:
Closes #1871.

* **Removed `ebooklib` as a dependency** `ebooklib` is licensed under AGPL3, which is incompatible with the Apache 2.0 license. Thus it is being removed.

- Also fixed some issues with the `pip-compile` script.

#### Testing
The main thing is to remove `ebooklib` and still have a working `partition_epub`.

Run `pip uninstall ebooklib` and then ensure the following runs successfully (need `pandoc` installed):

```python
from unstructured.partition.epub import partition_epub
partition_epub("example-docs/winter-sports.epub")

```


test_unstructured/partition/epub/test_epub.py
```diff
@@ -222,4 +218,4 @@ def test_partition_epub_respects_detect_language_per_element():
     filename = "example-docs/language-docs/eng_spa_mult.epub"
     elements = partition_epub(filename=filename, detect_language_per_element=True)
     langs = [element.metadata.languages for element in elements]
-    assert langs == [["eng"], ["eng"], ["spa", "eng"], ["eng"], ["eng"], ["spa"]]
+    assert langs == [["eng"], ["spa", "eng"], ["eng"], ["eng"], ["spa"]]
```
Coniferish, line 221: Do you know why there's one fewer elements created for this document?

qued: I haven't looked closely, but if I had to guess maybe a section change happened in the middle of one of the current elements.

Coniferish: That would make sense, but I will take a look to confirm.

Coniferish: @qued for some reason I'm having trouble checking out to this branch locally, so can you check?

test_unstructured/partition/epub/test_epub.py
```diff
@@ -8,77 +8,61 @@
...
-    "CHAPTER III TEES AND CRAMPITS",
-    "CHAPTER IV TOBOGGANING",
-    # not included in expected sections because TOC doesn't perfectly match with EpubHtml items
-    # 'CHAPTER V ICE-HOCKEY',
-    "CHAPTER VI SKI-ING",
-    "CHAPTER VII NOTES ON WINTER RESORTS",
-    "CHAPTER VIII FOR PARENTS AND GUARDIANS",
-    "THE FULL PROJECT GUTENBERG LICENSE",
-}
+EXAMPLE_DOCS_PATH = os.path.join(DIRECTORY, "..", "..", "..", "example-docs")
```
badGarnet, line 11: seen this and other places assume where the root path is in tests so I think we should create a ticket to create a test level fixture called root_path or example_doc_path. https://unstructured-ai.atlassian.net/browse/CORE-2378

qued: Good idea.

badGarnet (commented):
on M2 mac testing it seems you have to run `brew install pandoc` to be able to run `pandoc`. Maybe that should be part of the install process? Right now the script states it is for CI and it seem to just aim for linux os: https://github.com/Unstructured-IO/unstructured/blob/845b349d2d7a437bf1e6824edfe122c92cc32671/scripts/install-pandoc.sh

test_unstructured/partition/epub/test_epub.py
Coniferish: There's a potential issue with a universal EXAMPLE_DOCS_PATH, though, because unlike the `partition` directory, most test files are nested in an additional directory based on their doc type (ppt and pptx are in their own "pptx" folder for example) while others are not. They used to all be in the same directory like the partition files, but I never looked into why this change occurred.

scripts/pip-compile.sh
```diff
@@ -8,8 +8,22 @@ if ! python -c "import sys; assert sys.version_info.major == $major and sys.vers
   exit 1
 fi
 
+# NOTE(alan): Order matters in compiling. Start with base and test. Then do the rest in any order.
```
badGarnet, line 11: this PR (https://github.com/Unstructured-IO/unstructured/pull/1295) might help with this situation

badGarnet (approved):

PR #1888, opened 2023-10-26
support passing credentials from memory for google connectors

rbiseck3:
### Google Drive
The existing service account parameter was expanded to support either a file path or a json value to generate the credentials when instantiating the google drive client.

### GCS
Google Cloud Storage already supports the value being passed in, from their docstring:
> - you may supply a token generated by the
      [gcloud](https://cloud.google.com/sdk/docs/)
      utility; this is either a python dictionary, the name of a file
      containing the JSON returned by logging in with the gcloud CLI tool,
      or a Credentials object.

I tested this locally:

```python
from gcsfs import GCSFileSystem
import json

with open("/Users/romanisecke/.ssh/google-cloud-unstructured-ingest-test-d4fc30286d9d.json") as json_file:
    json_data = json.load(json_file)
    print(json_data)

    fs = GCSFileSystem(token=json_data)
    print(fs.ls(path="gs://utic-test-ingest-fixtures/"))
```
`['utic-test-ingest-fixtures/ideas-page.html', 'utic-test-ingest-fixtures/nested-1', 'utic-test-ingest-fixtures/nested-2']`

### DIFF
--- a/unstructured/ingest/cli/cmds/gcs.py
+++ b/unstructured/ingest/cli/cmds/gcs.py
@@ -13,13 +13,13 @@ CMD_NAME = "gcs"
 
 @dataclass
 class GcsCliConfig(BaseConfig, CliMixin):
-    token: t.Optional[t.Union[dict, Path]] = None
+    service_account_key: t.Optional[t.Union[dict, Path]] = None
 
     @staticmethod
     def get_cli_options() -> t.List[click.Option]:
         options = [
             click.Option(
-                ["--token"],
+                ["--service-account-key"],
                 default=None,
                 type=FileOrJson(),
                 help="Either the file path of the credentials file to use or a json string of "
--- a/unstructured/ingest/cli/cmds/gcs.py
+++ b/unstructured/ingest/cli/cmds/gcs.py
@@ -1,12 +1,11 @@
 import typing as t
 from dataclasses import dataclass
+from pathlib import Path
 
 import click
 
 from unstructured.ingest.cli.base.src import BaseSrcCmd
-from unstructured.ingest.cli.interfaces import (
-    CliMixin,
-)
+from unstructured.ingest.cli.interfaces import CliMixin, FileOrJson
 from unstructured.ingest.interfaces import BaseConfig
 
 CMD_NAME = "gcs"
@@ -14,7 +13,7 @@ CMD_NAME = "gcs"
 
 @dataclass
 class GcsCliConfig(BaseConfig, CliMixin):
-    token: t.Optional[str] = None
+    token: t.Optional[t.Union[dict, Path]] = None
 
     @staticmethod
     def get_cli_options() -> t.List[click.Option]:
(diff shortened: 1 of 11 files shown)


unstructured/ingest/connector/google_drive.py
```diff
@@ -52,12 +52,21 @@ def create_service_account_object(key_path, id=None):
...
+        if isinstance(key_path, dict):
+            creds = service_account.Credentials.from_service_account_info(key_path)
+        elif os.path.isfile(key_path):
+            os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = key_path
+            creds, _ = default()
+        else:
+            raise ValueError(
+                f"key path not recognized as a dictionary or a file path: "
+                f"[{type(key_path)}] {key_path}",
+            )
```
ryannikolaidis, line 69: for clarify maybe just have a separate argument for the path versus the value rather than trying to infer. I think given naming folks are already confused on which we expec

ryannikolaidis:
going to include GCS as well?

ryannikolaidis:
See note on different field for path vs value. Maybe stick to the suggestion proposed in the issue ticket?

unstructured/ingest/cli/cmds/google_drive.py
```diff
@@ -1,20 +1,18 @@
...
-)
+from unstructured.ingest.cli.interfaces import CliMixin, CliRecursiveConfig, FileOrJson
 from unstructured.ingest.interfaces import BaseConfig
 
 
 @dataclass
 class GoogleDriveCliConfig(BaseConfig, CliMixin):
     drive_id: str
-    service_account_key: str
+    token: t.Union[dict, Path]
```
ryannikolaidis, line 15: I think in either case we should keep the old key and flag as deprecated.
let's defer to service_account_key as the one we make common to both, since this is more specific and seems to be more wildly referenced in their documents.

rbiseck3: What do we use if both are provided? Raise an error?

ryannikolaidis: actually, I think google drive is the one folks are really using. so I think fine to not deprecate, but let's just make service_account_key the common one.

unstructured/ingest/cli/cmds/gcs.py
```diff
@@ -1,28 +1,27 @@
...
-    CliMixin,
-)
+from unstructured.ingest.cli.interfaces import CliMixin, FileOrJson
 from unstructured.ingest.interfaces import BaseConfig
 
 
 @dataclass
 class GcsCliConfig(BaseConfig, CliMixin):
-    token: t.Optional[str] = None
+    service_account_key: t.Optional[t.Union[dict, Path]] = None
```
ryannikolaidis, line 14: should also bump the gcs docs page?

ryannikolaidis (commented):
tested both by passing the value as the argument, and both worked. 

after docs bump this is gtg!

ryannikolaidis (approved):
...per discussion, looks like docs use anonymous, so bumping shouldn't be required for this.

lgtm!

PR #1896, opened 2023-10-26
test: segregate long running integration tests

qued:
|                      | Time |
| --------------------- | ----- |
| Before changes: | 13:15 |
| After changes:  | 10:18 |
| Percent decrease: | 22% |


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.10.28-dev0
+
+### Enhancements
+
+* **Separated integration tests to reduce test time** Used `pytest` markers to mark long-running integration tests, to facilitate speedier local testing and to run separately in CI.
```
cragwolfe, line 5: i'm not sure "integration" is the right adjective here

cragwolfe: why not just call it "chipper," like it is?
also makes sense to move chipper tests to their own file IMO

qued: Not all the tests I've marked are chipper.

qued:
Closing for now as we are not currently going in this direction.

PR #1898, opened 2023-10-26
Update fsspecs-specific source connector docs

rbiseck3:
### Description
Add in the fsspec configs needed for the fsspec-based connectors

To match the behavior of the original CLI, the default used by the click option was mirrored in the base config for the api endpoint.


rbiseck3:
@ryannikolaidis, all docs ran through manually to validate.

ryannikolaidis (approved):
thanks for the update and testing these!

PR #1909, opened 2023-10-27
build: streamline ci

qued:
Updated CI to shave time off in some conditions with no real downside.

- When the base cache already exists, we don't download it during setup, and we skip all other steps as well.
- During ingest setup, we check if the ingest cache exists before downloading the base cache, and if the ingest cache already exists, we skip everything else.
- `check-deps` doesn't have to wait on `setup` or download a cache, as the dependencies aren't needed, only `pip`.

| CI version | Setup (on hit) | Ingest Setup (on hit) | Check deps (on hit) |
| ---------- | --------- | ------------ | -------- |
| Old     | 2:06-2:31  | 3:07-3:31  | 3:19-6:26 |
| New   | 0:08-0:09  |  0:07-0:09 | 2:05-2:23 |

#### Testing
- Results of cache hits (best case scenario) can be seen [here](https://github.com/Unstructured-IO/unstructured/actions/runs/6663042342/job/18108335916?pr=1909).
- Results of old cache hits can be seen [here](https://github.com/Unstructured-IO/unstructured/actions/runs/6662563238).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,12 @@ jobs:
         key: unstructured-${{ runner.os }}-${{ matrix.python-version }}-${{ env.GHA_CACHE_KEY_VERSION }}-${{ hashFiles('requirements/*.txt') }}
         lookup-only: true
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


cragwolfe (approved):
tricky

PR #1917, opened 2023-10-27
feat: support table conversion for tabular destination connectors

rbiseck3:
### Description
* A full schema was introduced to map the type of all output content from the json partition output and mapped to a flattened table structure to leverage table-based destination connectors. The delta table destination connector was updated at the moment to take advantage of this.
* Existing method to convert to a dataframe was updated because it had a bug in it. Object content in the metadata would have the key name changed when flattened but then this would be omitted since it didn't exist in the `_get_metadata_table_fieldnames` response. 
* Unit test was added to make sure we handle all values possible in an Element when converting to a table
* Delta table ingest test was split into a source and destination test (looking ahead to split these up in CI)


unstructured/staging/base.py
```diff
@@ -228,17 +233,78 @@ def convert_to_csv(elements: List[Element]) -> str:
 
 
 @requires_dependencies(["pandas"])
-def convert_to_dataframe(elements: List[Element], drop_empty_cols: bool = True) -> "pd.DataFrame":
+def get_default_dtypes() -> dict:
```
ryannikolaidis, line 236:
```suggestion
def get_pandas_default_dtypes() -> dict:
```
maybe? had no idea what a dtype was until I looked inside this function

ryannikolaidis:
knowing metadata is a moving target: how resilient is this to change of metadata? ...mostly: how does it break when things change or are added? wonder if we need a "adding metadata checklist".

ryannikolaidis:
> knowing metadata is a moving target: how resilient is this to change of metadata? ...mostly: how does it break when things change or are added? wonder if we need a "adding metadata checklist".

ahh, looks like you do protect against it by check if it exists. so worst case will be if something is renamed then it just won't be added the table

test_unstructured_ingest/python/test-ingest-delta-table-output.py
```diff
@@ -10,7 +10,13 @@ def run_check(table_uri):
...
     )
 
-    assert len(delta_table.to_pandas()) == 10
+    df = delta_table.to_pandas()
+    expected_rows = 5
+    expected_columns = 18
+    print(f"Number of rows in table vs expected: {len(df)}/{expected_rows}")
+    print(f"Number of columns in table vs expected: {len(df.columns)}/{expected_columns}")
+    assert len(df) == 5
+    assert len(df.columns) == 18
```
ryannikolaidis, line 19: on failure, would be nice if these printed the diff. also probably if this happens something about metadata went wrong? might be good to include some leading tips in that case

unstructured/ingest/utils/table.py
```diff
@@ -0,0 +1,24 @@
+import typing as t
+
+import pandas as pd
+
+from unstructured.staging.base import flatten_dict, get_default_dtypes
+
+
+def convert_to_df(
```
ryannikolaidis, line 8:
```suggestion
def convert_to_dataframe(
```
for readability? ..then I know at a glance exactly what this is

test_unstructured/staging/test_base_staging.py
```diff
@@ -102,6 +107,71 @@ def test_convert_to_dataframe_maintains_fields(
     assert "regex_metadata_punc" in df.columns
 
 
+def test_dataframe_dtypes():
```
ryannikolaidis, line 110: quick docstring on what you are testing..also maybe more specific test name? here it looks like you're specifically testing that for element there is a corollary default dtype?

ryannikolaidis (approved):
lgtm!

ryannikolaidis:
bumping pr title to conventional commit message

PR #1918, opened 2023-10-27
Add test for the embedding module

ahmetmeleq:
Adds an actual test for the embedding module (Huggingface). 

Additional info:
- Found `langchain.embeddings.HuggingFaceEmbeddings` to be deterministic even when there's no seed set
- Took 6.84s to pass the test (without cache, including model download)
- `langchain.embeddings.HuggingFaceEmbeddings` runs in local, making it zero cost

For all these reasons, testing embedding modules with the Huggingface model seems to be making sense

### DIFF
--- a/test_unstructured/embed/test_embed_huggingface.py
+++ b/test_unstructured/embed/test_embed_huggingface.py
@@ -1,23 +1,69 @@
+import numpy as np
+
 from unstructured.documents.elements import Text
 from unstructured.embed.huggingface import HuggingFaceEmbeddingEncoder
 
 
 def test_embed_documents_does_not_break_element_to_dict(mocker):
-    # Mocked client with the desired behavior for embed_documents
-    mock_client = mocker.MagicMock()
-    mock_client.embed_documents.return_value = [1, 2]
-
-    # Mock get_openai_client to return our mock_client
-    mocker.patch.object(
-        HuggingFaceEmbeddingEncoder,
-        "get_huggingface_client",
-        return_value=mock_client,
-    )
 
     encoder = HuggingFaceEmbeddingEncoder()
-    elements = encoder.embed_documents(
+
+    doc_embed_result = encoder.embed_documents(
         elements=[Text("This is sentence 1"), Text("This is sentence 2")],
     )
-    assert len(elements) == 2
-    assert elements[0].to_dict()["text"] == "This is sentence 1"
-    assert elements[1].to_dict()["text"] == "This is sentence 2"
+    query_embed_result = encoder.embed_query(Text("This is sentence 1"))
+
+    # Keep the number of elements same
+    assert len(doc_embed_result) == 2
+
+    # Have consistent embedding dimensionality
+    assert (
+        len(doc_embed_result[0].embeddings)
+        == len(doc_embed_result[1].embeddings)
+        == len(query_embed_result)
+        == 384
+    )
+
+    # Keep element text the same
+    assert doc_embed_result[0].to_dict()["text"] == "This is sentence 1"
+    assert doc_embed_result[1].to_dict()["text"] == "This is sentence 2"
+
+    # Generate reproducable / deterministic embeddings
+    assert all(
+        np.isclose(
+            doc_embed_result[0].embeddings[:5],
+            [
+                0.0356232188642025,
+                0.06595201045274734,
+                0.06317725777626038,
+                0.04989251494407654,
+                0.0425567664206028,
+            ],
+        ),
+    )
+
+    assert all(
+        np.isclose(
+            doc_embed_result[1].embeddings[:5],
+            [
+                0.0622941255569458,
+                0.0793377012014389,
+                0.05761214718222618,
+                0.0711262971162796,
+                0.05167127028107643,
+            ],
+        ),
+    )
+
+    assert all(
+        np.isclose(
+            query_embed_result[:5],
+            [
+                0.03562326729297638,
+                0.06595207005739212,
+                0.0631771832704544,
+                0.04989247024059296,
+                0.04255683720111847,
+            ],
+        ),
+    )


ahmetmeleq:
Awaiting for a final check on if the deterministic behavior will persist in CI.  (if the test passes, the answer is yes)

ryannikolaidis (changes requested):
how about instead bumping our [embed ingest test](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-embed.sh) to use huggingface module instead? reason being, even though it's a small amount of download time, having a unit test that has to reach out externally doesn't feel great / isolated; whereas hitting live services is an expectation of the ingest tests.

ahmetmeleq:
> how about bumping our [embed ingest test](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-embed.sh) to use huggingface module instead? reason being, even though it's a small amount of download time, having a unit test that has to reach out externally doesn't feel great / isolated; whereas hitting live services is an expectation of the ingest tests.


This will need incorporating the recently added embedders to ingest. Pausing this one until incorporation work is finished (and starting on incorporation now)

ahmetmeleq:
Updating test fixtures: https://github.com/Unstructured-IO/unstructured/actions/runs/6720600405/job/18264902583

ahmetmeleq:
Todo: update ingest examples and  that uses embeddings to specify a `provider` parameter

unstructured/ingest/interfaces.py
```diff
@@ -38,6 +40,13 @@
 ]
 
 
+EMBEDDING_PROVIDER_TO_CLASS_MAP = {
```
rbiseck3, line 43: We might want to add this map to the `unstructured.embed` directory so that code outside of ingest could leverage it as well.

unstructured/ingest/cli/interfaces.py
```diff
@@ -368,12 +368,22 @@ class CliEmbeddingConfig(EmbeddingConfig, CliMixin):
     @staticmethod
     def get_cli_options() -> t.List[click.Option]:
         options = [
+            click.Option(
+                ["--embedding-provider"],
+                help="Type of the embedding class to be used. Can be: "
+                "langchain-huggingface, langchain-openai, or langchain-aws-bedrock.",
```
rbiseck3, line 374: Unless we're setting this as an `ENUM`, lets avoid giving all the options available so that if more get added, the string doesn't become out of date. Or generate that list from the keys in the embedding map we're using and set this as an `ENUM` based on that.

unstructured/ingest/interfaces.py
```diff
@@ -169,17 +178,19 @@ class ReadConfig(BaseConfig):
...
-            "api_key": self.api_key,
-        }
+        kwargs = {}
+        if self.api_key:
+            kwargs["api_key"] = self.api_key
         if self.model_name:
             kwargs["model_name"] = self.model_name
-        return OpenAIEmbeddingEncoder(**kwargs)
+
+        cls = EMBEDDING_PROVIDER_TO_CLASS_MAP.get(self.provider, None)
```
rbiseck3, line 192: Rather than default this to `None` if not found, this should raise an error.

ahmetmeleq:
Started seeing indeterministic behavior after ~ the 7 decimal

Todo: check seed options

unstructured/ingest/interfaces.py
ahmetmeleq: Address with [move mapping object to embed module](https://github.com/Unstructured-IO/unstructured/pull/1918/commits/921dd56b83c8c35d4b05ecd988bff120d02bafa9)

Why would we need this outside of ingest, won't python users import the classes directly? Do we assume there'll be other tools which'll need a mapping?

unstructured/ingest/cli/interfaces.py
ahmetmeleq: I'm dynamically loading the list from the class map keys now. Let me know if you insist on enum inheritance:
[load help message contents dynamically](https://github.com/Unstructured-IO/unstructured/pull/1918/commits/dcb8400e3cd72cc18c106753d60d174f8a589abd)

unstructured/ingest/interfaces.py
ahmetmeleq: Agreed, addressed with [raise error when wrong provider name is provided](https://github.com/Unstructured-IO/unstructured/pull/1918/commits/9cac06ab1f059a071e1c190393a5a444105596b3)

unstructured/ingest/cli/interfaces.py
```diff
@@ -397,13 +397,25 @@ def get_cli_options() -> t.List[click.Option]:
 class CliEmbeddingConfig(EmbeddingConfig, CliMixin):
     @staticmethod
     def get_cli_options() -> t.List[click.Option]:
+        from unstructured.embed import EMBEDDING_PROVIDER_TO_CLASS_MAP
+
         options = [
+            click.Option(
+                ["--embedding-provider"],
+                help="Type of the embedding class to be used. Can be one of: "
+                f"{list(EMBEDDING_PROVIDER_TO_CLASS_MAP)}",
```
rbiseck3, line 406: leverage the [choice](https://click.palletsprojects.com/en/8.1.x/options/#choice-options) click option to allow click to throw an appropriate error if they don't pass in something that is allowed. 
```python
type=click.Choice(list(EMBEDDING_PROVIDER_TO_CLASS_MAP))
```

ahmetmeleq: This is cool.

Offloaded raising to Click, and removed the raise logic from EmbeddingConfig: 
[check provider error in cli via Click, instead of in EmbeddingConfig](https://github.com/Unstructured-IO/unstructured/pull/1918/commits/a8b95fdf84f9428f6595e27c42475ceb2c44d75b)

Let me know if we still would like to keep the try / except within EmbeddingConfig

rbiseck3 (approved):

ahmetmeleq:
Running a test fixtures update: https://github.com/Unstructured-IO/unstructured/actions/runs/6747950881

ryannikolaidis (approved):

PR #1919, opened 2023-10-27
feat: use yolox as default to table extraction for pdf/image

badGarnet:
- yolox has better recall than yolox_quantized, the current default model, for table detection
- update logic so that when `infer_table_structure=True` the default model is `yolox` instead of `yolox_quantized`
- user can still override the default by passing in a `model_name` or set the env variable `UNSTRUCTURED_HI_RES_MODEL_NAME`

## Test:

Partition the attached file with 

```python

yolox_elements = partition_pdf(filename, strategy="hi_re", infer_table_structure=True)
yolox_quantized_elements = partition_pdf(filename, strategy="hi_re", infer_table_structure=True, model_name="yolox_quantized")
```

Compare the table elements between those two and yolox (default) elements should have more complete table.

[AK_AK-PERS_CAFR_2008_3.pdf](https://github.com/Unstructured-IO/unstructured/files/13191198/AK_AK-PERS_CAFR_2008_3.pdf)


benjats07:
Probably we can just use this version of YoloX(non-quantized) for reprocessing if we already know that the page contains a table. For now, It's ok.

benjats07 (approved):

PR #1923, opened 2023-10-27
feat: support generic inputs for partition kwargs from ingest CLI

rbiseck3:
### Description
To always support the latest changed to the partition method and the possible kwargs it supports, the ingest CLI has been refactored to take in a valid json string to represent those values to allow a user more flexibility with controlling the partition method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.29-dev9
+## 0.10.29-dev10
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


ryannikolaidis:
per async conversation: let's not drop support for individual flags for partition params. it's very nice to have these exposed via --help and nice to not have to specify via dictionary json and lookup in some other place what options are available and specifically considered partition args. that said, could see having a backdoor via something like this to specify we haven't yet explicitly supported

ryannikolaidis:
looks good! maybe we update a test to set a parameter with this? I guess could just swap an existing parameter but set it via this?

ryannikolaidis (approved):
lgtm, thanks!

ryannikolaidis:
updated title to convention commit, since this will be used by merge queue as the commit.

PR #1939, opened 2023-10-30
tests: separate chipper tests

qued:
Separates chipper tests to speed up testing and CI.


cragwolfe (approved):
PR is in draft mode but looks solid to me.

PR #1947, opened 2023-10-30
fix: flaky chunk metadata

scanny:
**Executive Summary.**  When the elements in a _section_ are combined into a _chunk_, the metadata in each of the elements is _consolidated_ into a single `ElementMetadata` instance. There are two main problems with the current implementation:

1. The current algorithm simply uses the metadata of the first element as the metadata for the chunk. This produces:
	- **empty chunk metadata** when the first element has no metadata, such as a `PageBreak("")`
	- **missing chunk metadata** when the first element contains only partial metadata such as a `Header()` or `Footer()`
	- **misleading  metadata** when the first element contains values applicable only to that element, such as `category_depth`, `coordinates` (bounding-box), `header_footer_type`, or `parent_id`
2. Second, list metadata such as `emphasized_text_content`, `emphasized_text_tags`, `link_texts` and `link_urls` is only combined when it is unique within the combined list. These lists are "unzipped" pairs. For example, the first `link_texts` corresponds to the first `link_urls` value. When an item is removed from one (because it matches a prior entry) and not the other (say same text "here" but different URL) the positional correspondence is broken and downstream processing will at best be wrong, at worst raise an exception.

### Technical Discussion
Element metadata cannot be determined in the general case simply by sampling that of the first element. At the same time, a simple union of all values is also not sufficient. To effectively consolidate the current variety of metadata fields we need four distinct strategies, selecting which to apply to each field based on that fields provenance and other characteristics.

The four strategies are:
- `FIRST` - Select the first non-`None` value across all the elements. Several fields are determined by the document source (`filename`, `file_directory`, etc.) and will not change within the output of a single partitioning run. They might not appear in every element, but they will be the same whenever they do appear. This strategy takes the first one that appears, if any, as proxy for the value for the entire chunk.
- `LIST` - Consolidate the four list fields like `emphasized_text_content` and `link_urls` by concatenating them in element order (no set semantics apply). All values from `elements[n]` appear before those from `elements[n+1]` and existing order is preserved.
- `LIST_UNIQUE` - Combine only unique elements across the (list) values of the elements, preserving order in which a unique item first appeared.
- `REGEX` - Regex metadata has its own rules, including adjusting the `start` and `end` offset of each match based its new position in the concatenated text.
- `DROP` - Not all metadata can or should appear in a chunk. For example, a chunk cannot be guaranteed to have a single `category_depth` or `parent_id`.

Other strategies such as `COORDINATES` could be added to consolidate the bounding box of the chunk from the coordinates of its elements, roughly `min(lefts)`, `max(rights)`, etc. Others could be `LAST`, `MAJORITY`, or `SUM` depending on how metadata evolves.

The proposed strategy assignments are these:

- `attached_to_filename`: FIRST,
- `category_depth`: DROP,
- `coordinates`: DROP,
- `data_source`: FIRST,
- `detection_class_prob`: DROP,  # -- ? confirm --
- `detection_origin`: DROP,      # -- ? confirm --
- `emphasized_text_contents`: LIST,
- `emphasized_text_tags`: LIST,
- `file_directory`: FIRST,
- `filename`: FIRST,
- `filetype`: FIRST,
- `header_footer_type`: DROP,
- `image_path`: DROP,
- `is_continuation`: DROP,  # -- not expected, added by chunking, not before --
- `languages`: LIST_UNIQUE,
- `last_modified`: FIRST,
- `link_texts`: LIST,
- `link_urls`: LIST,
- `links`: DROP,            # -- deprecated field --
- `max_characters`: DROP,   # -- unused in code, probably remove from ElementMetadata --
- `page_name`: FIRST,
- `page_number`: FIRST,
- `parent_id`: DROP,
- `regex_metadata`: REGEX,
- `section`: FIRST,  # -- section unconditionally breaks on new section --
- `sent_from`: FIRST,
- `sent_to`: FIRST,
- `subject`: FIRST,
- `text_as_html`: DROP,     # -- not expected, only occurs in TableSection --
- `url`: FIRST,

**Assumptions:**
- each .eml file is partitioned->chunked separately (not in batches), therefore
  sent-from, sent-to, and subject will not change within a section.

### Implementation
Implementation of this behavior requires two steps:
1. **Collect** all non-`None` values from all elements, each in a sequence by field-name. Fields not populated in any of the elements do not appear in the collection.
```python
all_meta = {
    "filename": ["memo.docx", "memo.docx"]
    "link_texts": [["here", "here"], ["and here"]]
    "parent_id": ["f273a7cb", "808b4ced"]
}
```
2. **Apply** the specified strategy to each item in the overall collection to produce the consolidated chunk meta (see implementation).

### Factoring
For the following reasons, the implementation of metadata consolidation is extracted from its current location in `chunk_by_title()` to a handful of collaborating methods in `_TextSection`.
- The current implementation of metadata consolidation "inline" in `chunk_by_title()` already has too many moving pieces to be understood without extended study. Adding strategies to that would make it worse.
- `_TextSection` is the only section type where metadata is consolidated (the other two types always have exactly one element so already exactly one metadata.)
- `_TextSection` is already the expert on all the information required to consolidate metadata, in particular the elements that make up the section and their text.

Some other problems were also fixed in that transition, such as mutation of elements during the consolidation process.

### Technical Risk: adding new `ElementMetadata` field breaks metadata

If each metadata field requires a strategy assignment to be consolidated and a developer adds a new `ElementMetadata` field without adding a corresponding strategy mapping, metadata consolidation could break or produce incorrect results.

This risk can be mitigated multiple ways:
1. Add a test that verifies a strategy is defined for each (Recommended).
2. Define a default strategy, either `DROP` or `FIRST` for scalar types, `LIST` for list types.
3. Raise an exception when an unknown metadata field is encountered.

This PR implements option 1 such that a developer will be notified before merge if they add a new metadata field but do not define a strategy for it.

### Other Considerations
- If end-users can in-future add arbitrary metadata fields _before_ chunking, then we'll need to define metadata-consolidation behavior for such fields. Depending on how we implement user-defined metadata fields we might:
	- Require explicit definition of a new metadata field before use, perhaps with a method like `ElementMetadata.add_custom_field()` which requires a consolidation strategy to be defined (and/or has a default value).
	- Have a default strategy, perhaps `DROP` or `FIRST`, or `LIST` if the field is type `list`.

### Further Context
Metadata is only consolidated for `TextSection` because the other two section types (`TableSection` and `NonTextSection`) can only contain a single element.

---

### document-static
These fields are very likely to be the same for all elements in a single document:

- `attached_to_filename`
- `data_source`
- `file_directory`
- `filename`
- `filetype`
- `last_modified`
- `sent_from`
- `sent_to`
- `subject`
- `url`

*Consolidation strategy:* `FIRST` - use first one found, if any.

### section-static
These fields are very likely to be the same for all elements in a single section, which is the scope we really care about for metadata consolidation:

- `section` - an EPUB document-section unconditionally starts new section.

*Consolidation strategy:* `FIRST` - use first one found, if any.

### consolidated list-items
These `List` fields are consolidated by concatenating the lists from each element that has one:

- `emphasized_text_contents`
- `emphasized_text_tags`
- `link_texts`
- `link_urls`
- `regex_metadata` - special case, this one gets indexes adjusted too.

*Consolidation strategy:* `LIST` - concatenate lists across elements.

### dynamic
These fields are likely to hold unique data for each element:

- `category_depth`
- `coordinates`
- `image_path`
- `parent_id`

*Consolidation strategy:*
- `DROP` as likely misleading.
- `COORDINATES` strategy could be added to compute the bounding box from all bounding boxes.
- Consider allowing if they are all the same, perhaps an `ALL` strategy.

### slow-changing
These fields are somewhere in-between, likely to be common between multiple elements but varied within a document:

- `header_footer_type` - *strategy:* drop as not-consolidatable
- `languages` - *strategy:* take first occurence
- `page_name` - *strategy:* take first occurence
- `page_number` - *strategy:* take first occurence, will all be the same when `multipage_sections` is `False`. Worst-case semantics are "this chunk began on this page".

### N/A
These field types do not figure in metadata-consolidation:

- `detection_class_prob` - I'm thinking this is for debug and should not appear in chunks, but need confirmation.
- `detection_origin` - for debug only
- `is_continuation` - is _produced_ by chunking, never by partitioning (not in our code anyway).
- `links` (deprecated, probably should be dropped)
- `max_characters` - is unused as far as I can tell, is unreferenced in source code. Should be removed from `ElementMetadata` as far as I can tell.
- `text_as_html` - only appears in a `Table` element, each of which appears in its own section so needs no consolidation. Never appears in `TextSection`.

*Consolidation strategy:* `DROP` any that appear (several never will)

### DIFF
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -22,6 +22,7 @@ from unstructured.documents.elements import (
     Element,
     ElementMetadata,
     ListItem,
+    PageBreak,
     RegexMetadata,
     Table,
     Text,
@@ -207,7 +208,7 @@ def test_split_elements_by_title_and_table():
 
     section = next(sections)
     assert isinstance(section, _TextSection)
-    assert section.elements == [
+    assert section._elements == [
         Title("A Great Day"),
         Text("Today is a great day."),
         Text("It is sunny outside."),
@@ -219,7 +220,7 @@ def test_split_elements_by_title_and_table():
     # ==
     section = next(sections)
     assert isinstance(section, _TextSection)
-    assert section.elements == [
+    assert section._elements == [
         Title("An Okay Day"),
         Text("Today is an okay day."),
         Text("It is rainy outside."),
@@ -227,7 +228,7 @@ def test_split_elements_by_title_and_table():
     # --
     section = next(sections)
     assert isinstance(section, _TextSection)
-    assert section.elements == [
+    assert section._elements == [
         Title("A Bad Day"),
         Text("Today is a bad day."),
         Text("It is storming outside."),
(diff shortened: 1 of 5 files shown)


test_unstructured/chunking/test_title.py
```diff
@@ -722,16 +725,210 @@ def it_provides_access_to_its_table(self):
...
 
-    def it_provides_access_to_its_elements(self):
-        elements: List[Element] = [
-            Title("Introduction"),
-            Text(
-                "Lorem ipsum dolor sit amet consectetur adipiscing elit. In rhoncus ipsum sed"
-                "lectus porta volutpat.",
-            ),
-        ]
+    def it_can_combine_itself_with_another_TextSection_instance(self):
```
scanny, line 728: This test should have been added last PR, adding it here since I'm in the neighborhood.

test_unstructured/chunking/test_title.py
```diff
@@ -722,16 +725,210 @@ def it_provides_access_to_its_table(self):
 class Describe_TextSection:
     """Unit-test suite for `unstructured.chunking.title._TextSection objects."""
 
-    def it_provides_access_to_its_elements(self):
```
scanny, line 725: `.elements` need no longer be public now, this property has been removed, which better encapsulates the section and would be dead code anyway.

unstructured/documents/elements.py
```diff
@@ -265,6 +269,70 @@ def get_last_modified(self) -> Optional[datetime.datetime]:
...
 
+class ConsolidationStrategy(enum.Enum):
+    """Methods by which a metadata field can be consolidated across a collection of elements.
+
+    These are assigned to `ElementMetadata` field-names immediately below. Metadata consolidation is
+    part of the chunking process and may arise elsewhere as well.
+    """
+
+    DROP = "drop"
+    """Do not include this field in the consolidated metadata object."""
```
scanny, line 280: These enum-member docstrings show up in VisualStudio and other IDE "Intellisense" completion, so seem a better choice than comments for documenting these.

christinestraub:
Confirmed that the following fields are for debug only so the `DROP` strategy could be applied to those fields.
```
detection_class_prob: DROP, # -- ? confirm --
detection_origin: DROP, # -- ? confirm --
```
Questions:
- What field types should we apply the `list_unique` strategy to?
- Would the _Consolidation strategy_: `COORDINATES` be added in the next PR? I think the `coordinates` field is so important for the elements which contain coordinate data.

scanny:
> What field types should we apply the `LIST_UNIQUE` strategy to?

Hi Christine, currently the only metadata field to use `LIST_UNIQUE` is `.languages`. So a set of element metadata for that field like: `[["eng"], ["eng", "spa"], ["spa"], ["spa", "eng"]]` would consolidate to `["eng", "spa"]` (each item appears only once and order encountered is preserved). There could be more later of course but that's the only one so far and the reason we added that strategy.

> Would the _Consolidation strategy_: `COORDINATES` be added in the next PR? I think the `coordinates` field is so important for the elements which contain coordinate data.

Crag's view on that was the following:
> coordinates can also be dropped --  a chunk could span multiple pages, so there is no solution given the current 1-page assumption for coordinates

Which seems right to me. I hadn't initially thought about the possibility of spanning pages.

There is however a `multipage_sections` argument to chunking (default `True`). In cases where that is set to `False` we're guaranteed that all elements in a section come from the same page, so that's a possible partial solution, like consolidate coordinates when that value is False and drop otherwise. We would also have to _have_ page-numbers for that to work, otherwise the sectioner has no way of telling where one page ends and the next one starts. But still, something we can continue to noodle on. I'd love to learn more about how PDF handles page-numbers. I kind of expect that the presence of page numbers is "guaranteed" in that format since it's very page-oriented, isn't it? Like an actual print-out?

christinestraub (approved):
LGTM! The _Consolidation strategy_ for the `languages` field seems to be incorrect in the PR description.

PR #1951, opened 2023-10-31
test: separate ingest tests for source and destination connectors

rbiseck3:
### Description
This splits the source ingest tests from the destination ingest tests since they share a different pattern:
* src tests pull data from a source and compare the partitioned content to the expected results
* destingation tests leverage the local connector to produce results to push to a destination and leverages overhead to create temporary locations at those destinations to write to and delete when done. 

Only the src tests create partitioned content that needs to be checked so the update ingest test CI job only needs to run these.


.github/workflows/ci.yml
```diff
@@ -370,7 +370,109 @@ jobs:
...
+        UNS_API_KEY: ${{ secrets.UNS_API_KEY }}
+        NOTION_API_KEY: ${{ secrets.NOTION_API_KEY }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AZURE_SEARCH_ENDPOINT: ${{ secrets.AZURE_SEARCH_ENDPOINT }}
+        AZURE_SEARCH_API_KEY: ${{ secrets.AZURE_SEARCH_API_KEY }}
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        TABLE_OCR: "tesseract"
+        OCR_AGENT: "tesseract"
+        CI: "true"
```
ryannikolaidis, line 465: should probably only bring in secrets we need here

.github/workflows/ci.yml
```diff
@@ -370,7 +370,109 @@ jobs:
...
+    - name: 'Az CLI login'
+      uses: azure/login@v1
+      with:
+        client-id: ${{ secrets.AZURE_CLIENT_ID }}
+        tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
```
ryannikolaidis, line 406: can we remove these steps in the source job?

.github/workflows/ci.yml
```diff
@@ -370,7 +370,109 @@ jobs:
...
         tesseract --version
-        ./test_unstructured_ingest/test-ingest.sh
+        ./test_unstructured_ingest/test-ingest-src.sh
+
+
+  test_ingest_dest:
+    environment: ci
+    strategy:
+      matrix:
+        python-version: ["3.8","3.9","3.10","3.11"]
```
ryannikolaidis, line 380: maybe we only run on 3.10 until we have a reason to have a matrix dest job? ..this will save some ci time for those other versions spinning up only to skip all destination tests

ryannikolaidis: actually, I guess maybe there are some we do want to run against. hmm, one sec

rbiseck3: Since there are only so many of these, I have them all being run for every version right now.

ryannikolaidis: let's run: azure, gcs, and s3 on all version. let's only run the others on 3.10

ryannikolaidis (approved):
lgtm

ryannikolaidis (commented):
ping me when tests are passing and ready for review

ryannikolaidis (approved):
lgtm
note: in order to merge we're going to need to change branch protection policy.

PR #1952, opened 2023-10-31
Support nested tables in docx files

bborn:
Recursively parse cells that contain tables inside `docx` files. Uses tabulate's `unsafehtml` option to avoid escaping the nested tables


scanny:
Hi Bruno, I assume this PR means you've encountered nested tables in practice :) I'd be very interested to hear more about your experience with them, where and how often you find them occurring and the kind of outputs that would be most useful for the cases you've seen.

I like the recursive approach you've implemented here; I think that's the right idea.

A couple things we'll need to consider.

1. The current implementation of `convert_ms_office_table_to_text()` is used by both the DOCX and PPTX partitioner. However, only DOCX can have nested tables, so these changes will break the PPTX partitioner (a `python-pptx` Cell has no `.tables` property).

	I think the solution to this is easy enough, and that's to create a different implementation for each type and probably move them out of `partition.common` and into their respective partitioner modules since they will only be used there.

2. A DOCX cell is a _block-item container_ meaning it can contain both paragraphs and tables (the two kinds of block-item) and in any order and number. So a general-case solution will need to accommodate that and include _all_ items found in the cell.

	One thing that's on my near-term list to add to `python-docx` is `BlockItemContainer.iter_inner_content()` which would iterate these paragraphs and tables in document order. Using that (which we need for another PR anyway) we could place not only the first table, but any paragraphs before and/or after and any other tables that were nested. I'm not sure how much this actually happens but it wouldn't do to have something that broke except in the simplest case.

I think we might end up with something like this:

```python
def convert_docx_table_to_html(table: DocxTable) -> str:

    def iter_cell_block_items(cell) -> Iterator[str]:
        for item in cell.iter_inner_content():
            if isinstance(item, Paragraph):
                yield f"<p>{item.text}</p>"
            elif isinstance(item, DocxTable):
                yield convert_docx_table_to_html(item)

    def iter_cells(row) -> Iterator[str]:
        return (
            "\n".join(iter_cell_block_items(cell)) for cell in row.cells
        )

    rows = [list(iter_cells(row)) for row in table.rows]

    return tabulate(rows, headers="firstrow", tablefmt="unsafehtml")
```

We might want to rethink the assumption that the first row is column headers. That's less likely to be the case for a nested table and probably isn't our call anyway. We could add a parameter for that and only apply it to the "outside" table, that would let the caller decide what way they wanted it.

I can see when I might have time to add the upstream `python-docx` feature and separate the table converters for DOCX and PPTX. I should have a better idea tomorrow as to timing on those and will leave an update here.

bborn:
@scanny updated - please take a look. 

> I'd be very interested to hear more about your experience with them, where and how often you find them occurring and the kind of outputs that would be most useful for the cases you've seen.

TBH - not very much. So far just one user who tried to ingest a document with nested tables and had it break, so I can't say much about use cases. I think converting them to HTML (as I've done) makes sense, although on the other hand maybe people don't care about the table markup that much, and really just want to make sure they get all the text (e.g. vector indexing a document), in which case just converting the nested table content to plain text would make sense.

scanny:
@bborn Ok, I got the okay to work on the upstream `BlockItemContainer.iter_inner_content()` so I'll be picking that up this week.

A couple other things that occurred to me after a sleep:
1. With the simple algorithm for iterating over table cells for DOCX, when a cell is merged, its text will be duplicated. This will happen once for each cell in the merge, so might typically be two or three times but could be quite a few if a merged cell is large. What would be your point-of-view (POV) on how that should best be handled? Horizontal merges are relatively easily de-duplicated, vertical ones less so, but that produces a different number of "columns" for different rows and perhaps makes columnar groupings impossible to clearly distinguish.

    Example:

    ```txt
    +-----+-----+-----+-----+-----+
    | fiz | foo       | bar | baz |
    +-----+     +     +-----+-----+
    | bah |           | dah | bng |
    +-----+-----+-----+-----+-----+
    ```
    will enumerate as:
    ```txt
    +-----+-----+-----+-----+-----+
    | fiz | foo | foo | bar | baz |
    +-----+-----+-----+-----+-----+
    | bah | foo | foo | dah | bng |
    +-----+-----+-----+-----+-----+
    ```

    horz deduplicated would be:
    
    ```txt
    +-----+-----+-----+-----+
    | fiz | foo | bar | baz |
    +-----+-----+-----+-----+
    | bah | foo | dah | bng |
    +-----+-----+-----+-----+
    ```

    both horz and vert deduplicated would be:
    
    ```txt
    +-----+-----+-----+-----+
    | fiz | foo | bar | baz |
    +-----+-----+-----+-----+
    | bah | dah | bng |
    +-----+-----+-----+
    ```
2. Did you have a POV on unconditionally considering the first row to be headings? I'm a little inclined to not take a position on that at all, since the first column could just as easily be row-headings and there be no column headings at all. I suppose it's a question of the downstream consequences of making the wrong guess and whether we should be guessing at all.

scanny:
Hi @bborn, okay, another thing now that I'm digging into the code around this. What are we thinking should be the behavior for "plain" text mode nested tables? `partition_docx()` provides both formats so we'll need a plan for how to deal with this one too. I'll dig into what the current behavior is there in a bit but if you have a POV on this that would be very welcome :)

scanny:
Also @bborn can you say more about why we need the `unsafehtml` option and what we might be losing (or gaining) by using that instead of the regular "html" option?

bborn:
To be honest - for the use cases we're dealing with (vector indexing documents), it would be fine to just dump the nested tables as plain text. The markup and structure doesn't seem too important. I would probably de-dupe horizontal and vertical, and forget about trying to preserve the table structure.

`unsafehtml` was just because when I first tried it, the nested table was getting escaped, so using unsafe was the only way I could find to preserve the HTML elements. Maybe with the new content iteration approach it's no longer needed.

scanny:
@bborn this is incorporated in PR #2020 if you want to take a look. I've credited you with the commit although I took a bit different approach to certain aspects.

bborn:
> @bborn this is incorporated in PR #2020 if you want to take a look. I've credited you with the commit although I took a bit different approach to certain aspects.

Sweet! Looks awesome. Thanks - closing this one.

PR #1953, opened 2023-10-31
build: don't save cache for cache existence check

qued:
Using `actions/cache@v3` instead of `actions/cache/restore@v3` for the cache lookup in `setup_ingest` is causing CI to save the cache twice when there's a cache miss. This is unnecessary, but I'm also a little concerned it's causing some sort of race condition since I've seen instances of CI failing to save due to the cache already existing (which shouldn't be the case on a cache miss).

This PR switches the lookup to a `restore` action to avoid duplicate ingest cache saving.

#### Testing:

There should only be one "Post Run actions/cache@v3" step in each `setup_ingest` job when there's a cache miss.

[Here](https://github.com/Unstructured-IO/unstructured/actions/runs/6707740917/job/18227300507) is an example of a cache miss running with this PR.

For comparison, [here](https://github.com/Unstructured-IO/unstructured/actions/runs/6695668886/job/18195998114) is what it looks like when it saves twice.


ryannikolaidis (approved):
Agree, this is the correct way to do that

PR #1957, opened 2023-10-31
Add CliMix class to wrap both BaseConfig and CliMixin

rbiseck3:
### Description
Add new class to wrap base config and cli mixin to help with typing:
```python
class CliConfig(BaseConfig, CliMixin):
    pass
```


ryannikolaidis:
per convention maybe we just call the new thing CliConfig?

jasonbot (approved):

PR #1968, opened 2023-11-01
fix languages 500 error with empty string for ocr_languages

### LINKED ISSUE #1870: bug/Only one of languages and ocr_languages should be specified
Some API users are getting a 500 due to this ValueError. I think our client defaults to sending an empty string for `ocr_languages`, like the below. I think if we see both params, we should default to using `languages` and log a warning for the library users.
```
curl -X POST 'http://localhost:8000/general/v0/general' --form files="@$file" --form languages="eng" --form ocr_languages=""
```

Coniferish:
Closes #1870 
Defining both `languages` and `ocr_languages` raises a ValueError, but the api defaults to `ocr_languages` being an empty string, so if users define `languages` they are automatically hitting the ValueError.

This fix checks if `ocr_languages` is an empty string and converts it to `None` to avoid this.

### Testing
On the main branch, the following will raise the ValueError, but it will correctly partition on this branch
```
filename = "example-docs/category-level.docx"
elements = partition(filename,languages=['spa'],ocr_languages="")

elements[0].metadata.languages
```


awalker4 (approved):
LGTM!

shreyanid:
Seeing the version conflict with the previous PR in the merge queue. Confused on how to resolve it: setting it to dev5 like the previous wouldn't correctly increment the version for this change, but setting it to dev6 would still have a conflict. Perhaps that's why this was waiting to merge.

PR #1976, opened 2023-11-01
chore: remove copy line from non-matrix connectors

Klaijan:


test_unstructured_ingest/metrics/all-docs-cct.tsv
```diff
@@ -1,4 +1,4 @@
 filename	connector	cct-accuracy	cct-%missing
 handbook-1p.docx	box	0.974	0.03
 example-10k.html	local	0.686	0.04
-IRS-form-1987.pdf	azure	0.783	0.13
+science-exploration-1p.pptx	dropbox	0.861	0.09
```
ryannikolaidis, line 4: why is dropbox reported here?

test_unstructured_ingest/metrics/all-docs-cct.tsv
```diff
@@ -1,4 +1,4 @@
 filename	connector	cct-accuracy	cct-%missing
 handbook-1p.docx	box	0.974	0.03
 example-10k.html	local	0.686	0.04
-IRS-form-1987.pdf	azure	0.783	0.13
```
ryannikolaidis, line 4: probably safe to keep azure if it's helpful

test_unstructured_ingest/metrics/all-docs-cct.tsv
Klaijan: That's probably the version lingered in my local. Let me try to resolve that with the new src/dest setting.

ryannikolaidis (approved):
lgtm!

test_unstructured_ingest/metrics/metrics-json-manifest.txt
```diff
@@ -1,3 +1,4 @@
+handbook-1p.docx.json
 example-10k.html.json
 IRS-form-1987.pdf.json
 science-exploration-1p.pptx.json
```
ryannikolaidis, line 4: ohh, curious, why do we list all of these in the manifest, but only actually expect example-10k and the IRS form? also -- why adding handbook-1p if it's not in the all-docs-cct.tsv?

Klaijan: We are now combining all manifest for both cct and element type. I am working on creating gold standard for element type and should be in soon on another pr!

PR #1977, opened 2023-11-01
chore: add doctype to ingest evaluation functions

shreyanid:
### Summary
To combine ingest and holistic metrics efforts, add the `doctype` field to the results from the functions in evaluate.py for use in subsequent aggregation functions.

### Test
Run `sh ./test_unstructured_ingest/evaluation-metrics.sh text-extraction` and there will be a new doctype column with the file's doctype extension.
Screenshot 2023-11-01 at 2 23 11 PM

### DIFF
--- a/test_unstructured_ingest/metrics/all-docs-cct.tsv
+++ b/test_unstructured_ingest/metrics/all-docs-cct.tsv
@@ -1,4 +1,0 @@
-filename	doctype	connector	cct-accuracy	cct-%missing
-handbook-1p.docx	docx	box	0.974	0.03
-example-10k.html	html	local	0.686	0.04
-science-exploration-1p.pptx	pptx	dropbox	0.861	0.09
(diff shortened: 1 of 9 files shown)


test_unstructured_ingest/metrics/all-docs-cct.tsv
```diff
@@ -1,4 +1,4 @@
-filename	connector	cct-accuracy	cct-%missing
-handbook-1p.docx	box	0.974	0.03
-example-10k.html	local	0.686	0.04
-IRS-form-1987.pdf	azure	0.783	0.13
+filename	doctype	connector	cct-accuracy	cct-%missing
+handbook-1p.docx	docx	box	0.974	0.03
+example-10k.html	html	local	0.686	0.04
+science-exploration-1p.pptx	pptx	dropbox	0.861	0.09
```
shreyanid, line 4: If these numbers are all wrong due to my local run, please let me know the best way to update these metrics sheets (ingest text fixtures update?) with the new column as well

cragwolfe: yes, i'm pretty sure running the ingest test fixtures update (GH workflow) is the proper way to update these

Klaijan: Yes, you can run fixture update to correct these. You can also delete those files on local to avoid saving and pushing anything and just update fresh from the fixture update.

Klaijan (approved):
Lgtm!

PR #1980, opened 2023-11-01
refactor: separate click wrappers from core evaluation functionality

shreyanid:
### Summary
Click decorated functions cannot (properly) be called outside of the click interface. This makes it difficult to reuse the setup functionality in measure_text_edit_distance or measure_element_type_accuracy. This PR removes the click decoration and separates it into a wrapper function purely to execute the command.

### Technical Details
- Changed as suggested in [this StackOverflow post](https://stackoverflow.com/questions/40091347/call-another-click-command-from-a-click-command) response
    - We should also consider separating the locations of these now distinct functions: the `_command` click-decorated functions can stay in ingest/evaluate.py, and the core functions measure_text_edit_distance and measure_element_type_accuracy can move somewhere into the unstructured/metrics/ folder (which is a more logical location for them in my opinion). If agreed I will implement that in this PR

### Test
`sh ./test_unstructured_ingest/evaluation-metrics.sh text-extraction` functionality is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.29-dev6
+## 0.10.29-dev7
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.29-dev5
+## 0.10.29-dev6
 
 ### Enhancements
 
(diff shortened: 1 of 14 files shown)


shreyanid:
Closing as duplicate of https://github.com/Unstructured-IO/unstructured/pull/1981

PR #1981, opened 2023-11-01
refactor: separate click wrappers from core evaluation functionality

shreyanid:
### Summary
Click decorated functions cannot (properly) be called outside of the click interface. This makes it difficult to reuse the setup functionality in measure_text_edit_distance or measure_element_type_accuracy. This PR removes the click decoration and separates it into a wrapper function purely to execute the command.

### Technical Details
- Changed as suggested in [this StackOverflow post](https://stackoverflow.com/questions/40091347/call-another-click-command-from-a-click-command) response
- The locations of these now distinct functions are separate: the `_command` click-decorated functions stay in ingest/evaluate.py, and the core functions measure_text_edit_distance and measure_element_type_accuracy are moved into the unstructured/metrics/ folder (which is a more logical location for them).
- Initial test added for measure_text_edit_distance

### Test
`sh ./test_unstructured_ingest/evaluation-metrics.sh text-extraction` functionality is unchanged.


unstructured/ingest/evaluate.py
```diff
@@ -165,80 +81,16 @@ def measure_text_edit_distance(
...
-    if dir and not os.path.exists(dir):
-        os.makedirs(dir)
-    with open(os.path.join(os.path.join(dir, filename)), mode, newline="") as tsv:
-        writer = csv.writer(tsv, delimiter="\t")
-        if mode == "w":
-            writer.writerow(headers)
-        writer.writerows(rows)
 
 
 def _display(rows, headers):
```
shreyanid, line 96: I left _display here on the ingest side since it uses click.echo, but I get a vscode warning that the function is unused (in this file). Happy to move it to /metrics along with everything else but thought it would be cleaner to have all click-related functions in one file. Let me know!

Klaijan: I agree on the formatting. Can we ignore the error? Mot sure which command to use but can try to ##ignore and something like unused?

shreyanid: i don't believe its actually an error (but we'll see what ingest says), it's just greyed out on vscode. might not need to mark anything

example-docs/test_evaluate_files/gold_standard_cct/Bank Good Credit Loan.pptx.txt
Klaijan, line 1: I am uncertain whether if we want to push the docs with the commit or not. I normally upload to s3. But if you heard it differently would you mind loop me in on that as well?

Klaijan: Oh ok I suppose this is for the `test_unstructured`? Then this should be resolved!

test_unstructured/metrics/test_evaluate.py
```diff
@@ -0,0 +1,32 @@
...
+def test_text_extraction_takes_list():
+    output_dir = os.path.join(TESTING_FILE_DIR, UNSTRUCTURED_OUTPUT_DIRNAME)
+    output_list = ["currency.csv.json"]
+    source_dir = os.path.join(TESTING_FILE_DIR, GOLD_CCT_DIRNAME)
+    export_dir = os.path.join(TESTING_FILE_DIR, "test_evaluate_results_cct")
+    measure_text_edit_distance(
+        output_dir=output_dir,
+        output_list=output_list,
+        source_dir=source_dir,
+        source_list=None,
```
Klaijan, line 26: I think you can just omit the `source_list` arg altogether?

shreyanid: added defaults and omitted the arg. need to reshuffle the argument order for non default values to be listed first

unstructured/ingest/evaluate.py
```diff
@@ -69,88 +44,29 @@ def main():
     help="A tuple of weights to the Levenshtein distance calculation. \
         See text_extraction.py/calculate_edit_distance for more details.",
 )
-def measure_text_edit_distance(
+def measure_text_edit_distance_command(
     output_dir: str,
-    output_list: Optional[List[str]],
     source_dir: str,
+    output_list: Optional[List[str]],
     source_list: Optional[List[str]],
```
Klaijan, line 51:
```suggestion
    output_list: Optional[List[str]] = None,
    source_list: Optional[List[str]] = None,
```

unstructured/ingest/evaluate.py
```diff
@@ -165,132 +81,16 @@ def measure_text_edit_distance(
     help="Directory to save the output evaluation metrics to. Default to \
         your/working/dir/metrics/",
 )
-def measure_element_type_accuracy(
+def measure_element_type_accuracy_command(
     output_dir: str,
-    output_list: Optional[List[str]],
     source_dir: str,
+    output_list: Optional[List[str]],
     source_list: Optional[List[str]],
```
Klaijan, line 88:
```suggestion
    output_list: Optional[List[str]] = None,
    source_list: Optional[List[str]] = None,
```

unstructured/ingest/evaluate.py
```diff
@@ -39,14 +14,14 @@ def main():
 
 @main.command()
 @click.option("--output_dir", type=str, help="Directory to structured output.")
+@click.option("--source_dir", type=str, help="Directory to source.")
```
Klaijan, line 17: If this reordering works with the `generate_args` function in `evaluation-metrics.sh` line 41 then it works for me too.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
...
-* **Incorporate multiple embedding model options into ingest, add diff test embeddings** Problem: Ingest pipeline already supported embedding functionality, however users might want to use different types of embedding providers. Enhancement: Extend ingest pipeline so that users can specify and embed via a particular embedding provider from a range of options. Also adds a diff test to compare output from an embedding module with the expected output
+* **Incorporate multiple embedding model options into ingest, add diff test embeddings** Problem: Ingest pipeline already supported embedding functionality, however users might want to use different types of embedding providers. Enhancement: Extend ingest pipeline so that users can specify and embed via a particular embedding provider from a range of options. Also adds a diff test to compare output from an embedding module with the expected output.
```
Klaijan, line 10: Is this just adding the period for other Enhancement?

shreyanid: just adding the period to clear the changelog check, since this isn't a user facing change

unstructured/ingest/evaluate.py
shreyanid: hm i didn't add defaults for the _command functions because they're already built into the click parameters. This change isn't necessary for functionality. Do you still have a reason for me to change it?

unstructured/ingest/evaluate.py
shreyanid: Same as other function, defaults are handled by click. Let me know if I should change it anyway.

unstructured/ingest/evaluate.py
shreyanid: Yep! The args can still be formed and passed in any order from this script because they're named arguments

Klaijan (approved):

PR #1982, opened 2023-11-02
feat: rework aggregate metrics by doctype calculation

shreyanid:
### Summary
Previously, the holistic evaluation script was a copy of the ingest evaluation function with some modifications to aggregate the data by doctype. This refactor instead takes the result of the `measure_text_edit_distance` function (used by ingest) and aggregates the results by doctype. This pattern can also be followed by future aggregations we may want to perform.

### Test
Confirm the doctype aggregation functionality of the `aggregate_cct_data_by_doctype` function by calling it on the ingest metrics result sheet:
(from the top level unstructured folder)
```
python -c 'from unstructured.metrics.doctype_aggregation import *; aggregate_cct_data_by_doctype("./test_unstructured_ingest/metrics")'
```
The aggregated result will be written to the same metrics folder.
Screenshot 2023-11-03 at 2 56 20 PM


unstructured/metrics/doctype_aggregation.py
```diff
@@ -0,0 +1,77 @@
...
+
+@main.command()
+@click.option("--output_dir", type=str, help="Directory to structured output.")
+@click.option(
+    "--output_list",
+    type=str,
+    multiple=True,
+    help="Optional: list of selected structured output file names under the \
+        directory to be evaluate. If none, all files under directory will be use.",
+)
```
Klaijan, line 37: Do you want to reorder this to make it align with the `evaluation.py` one?

Klaijan (approved):
lgtm

PR #1996, opened 2023-11-02
Chore: don't pass empty language code to tesseract CLI

yuming-long:
Summary:

Close: https://github.com/Unstructured-IO/unstructured/issues/1920

* stop passing in empty string from `languages` to tesseract, which will result in passing empty string to language config `-l` for the tesseract CLI
* also stop passing in duplicate language code from `languages` to tesseract OCR
* if we failed to convert any iso languages from the `languages` parameter, proceed OCR with `eng` as default
  
### Test
* First confirm the tesseract error `Estimating resolution as X` before this: 
  * on the `unstructured-api` repo with main branch, run `make run-web-app`
  * curl to test error from empty string, or just any wrong input like `-F 'languages="eng,de"'`:
 ```
curl -X 'POST'  'http://0.0.0.0:8000/general/v0/general' \
  -H 'accept: application/json'   \
 -F 'files=@sample-docs/layout-parser-paper-with-table.jpg' \
-F 'languages=""'  \
-F 'strategy=hi_res'  \
-F 'pdf_infer_table_structure=True' \
 | jq -C . | less -R
``` 

* after this change:
   * in your unstructured API env, cd to unstructured repo and install it locally with `pip install -e .`
   * check out to this branch
   * run `make run-web-app` again in api repo
   * the curl command return output and see warning in log

### CODE REVIEW

**Comment by Coniferish** [comment]
@yuming-long, do you use pyenv for env management? I'm having trouble installing the local changes to `unstructured` in my api environment. Will keep working on it, but figured I'd ask.

**Comment by yuming-long** [comment]
Hi @Coniferish, yes I use pyenv.
I am on api repo, and then `pyenv activate unstructured-api`, `cd ../unstructured`, and `pip intall -e .` works for me. Let me know if you need help on the error :)

**Comment by Coniferish** [comment]
Sounds good. It's probably because I set 'local' envs for the different repos and need to clear out the config or something

File: unstructured/partition/lang.py [hunk 1]
```diff
@@ -146,8 +146,17 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
...
     converted_languages = list(
-        filter(None, [convert_language_to_tesseract(lang) for lang in languages]),
+        filter(
+            lambda x: x is not None and x != "",
+            [convert_language_to_tesseract(lang) for lang in languages],
+        ),
     )
+    # Remove duplicates from the list but keep the original order
+    converted_languages = list(dict.fromkeys(converted_languages))
+    if len(converted_languages) == 0:
```
shreyanid, line 156: If no languages are found, is there any way to proceed with the tesseract call anyway? Could we set a default of English since that's what tesseract uses as default (iirc), instead of raising an error? I would rather the user not be hung up at this step due to language conversion/detection and still be able to process through to tesseract

yuming-long: good call :) I think i will change this to a logger warning info and use language detection if no languages are converted, and worse case we use eng default

Coniferish:
@yuming-long, do you use pyenv for env management? I'm having trouble installing the local changes to `unstructured` in my api environment. Will keep working on it, but figured I'd ask.

> Hi @Coniferish, yes I use pyenv. I am on api repo, and then `pyenv activate unstructured-api`, `cd ../unstructured`, and `pip install -e .` works for me. Let me know if you need help on the error :)

Figured out I needed to first `pip uninstall unstructured` before going to the directory to install. Thank you @yuming-long!

Coniferish:
Not specific to this issue, but I'm noticing that I'm not actually seeing the `languages` field in the return value from the curl requests I'm making. Any ideas, @awalker4, @shreyanid, @yuming-long?

```
curl -X 'POST'  'http://0.0.0.0:8000/general/v0/general' \
-F 'files=@sample-docs/layout-parser-paper-with-table.jpg' \
-F 'languages=eng' \
-F 'strategy=hi_res' \
-F 'pdf_infer_table_structure=True' | jq -C . | less -R

[
  {
    "type": "Title",
    "element_id": "8e5839c2fb9b4d6b78cd2b1c1f5bed02",
    "metadata": {
      "filename": "layout-parser-paper-with-table.jpg",
      "filetype": "image/jpeg",
      "page_number": 1
    },
    "text": "LayoutParser: A Unified Toolkit for DL-Based DIA 5"
  },
```

unstructured/partition/lang.py
```diff
@@ -146,8 +146,20 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
     if languages is None:
         raise ValueError("`languages` can not be `None`")
     converted_languages = list(
-        filter(None, [convert_language_to_tesseract(lang) for lang in languages]),
+        filter(
```
Coniferish, line 149: Just a thought, but would we want to log a warning for any changes made to the languages users pass in (so the removal of duplicates is also logged in the warning message)?

yuming-long: hmm i don't think thats necessary. In my opinion its similar to convert languages parameter to standard language code and we don't need to tell them the converted language code.

yuming-long: cc@shreyanid ^^

Coniferish (approved):
LGTM

PR #1997, opened 2023-11-02
chore: update CHANGELOG.md

cragwolfe:
Remove bullets not related to end-user consumption of the unstructured library.


Klaijan (commented):
I have similar bullet on 0.10.26 that then should be removed too.

shreyanid (approved):
These removals make sense 👍 Klaijan on the 0.10.26 item, since it's already released we probably won't change it retroactively, but if decided otherwise that can be a separate PR

cragwolfe:
> These removals make sense 👍 Klaijan on the 0.10.26 item, since it's already released we probably won't change it retroactively, but if decided otherwise that can be a separate PR

Agreed, @Klaijan , we can leave the CHANGELOG for older releases intact.

PR #2003, opened 2023-11-03
ci(test): remove environment identifier

tabossert:
Moved Azure OIDC to use only Pull Request subject, this gets rid of the noise it creates in the PR’s

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -313,7 +313,6 @@ jobs:
 
 
   test_ingest_src:
-    environment: ci
     strategy:
       matrix:
         python-version: ["3.8","3.9","3.10","3.11"]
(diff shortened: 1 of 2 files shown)


ryannikolaidis (approved):
assuming tests pass, this looks good to me!

ryannikolaidis:
oh, maybe we run the test fixtures flow just to validate it doesn't get angry with azure cli?

ryannikolaidis (changes requested):
okay, worth double checking the ingest fixture. [looks like](https://github.com/Unstructured-IO/unstructured/actions/runs/6749558936/job/18350165090) it hit the issue we were worried about.

ryannikolaidis (approved):
sweet! validated the fixture flow passed

PR #2010, opened 2023-11-06
fix: update slack link to link shortener

qued:
Per @tabossert we're now using a link shortener behind which we can rotate the link to keep it current. That way we (🤞 ) never have to update this here again.

#### Testing:

Links should work. No more links should exist in the documentation except this one.


MthwRobinson (approved):
LGTM - thanks!

PR #2017, opened 2023-11-06
Klaijan/add ingest-test ci file

Klaijan:


shreyanid:
since the file was uploaded to s3 but the new .json files (added in this PR) are not available in expected_structured_output for it yet, ingest tests on mine and other PRs are failing. I've deleted the `IRS-form-1987.pdf` file from the small-pdf-set s3 bucket for now since this PR is still a draft, so hopefully you can just add it back when this is closer to ready to merge.

PR #2020, opened 2023-11-06
fix: #1952 support nested docx tables

scanny:
In DOCX, like HTML, a table cell can itself contain a table. This is not uncommon and is typically used for formatting purposes.

When a DOCX table is nested, create nested HTML tables to reflect that structure and create a plain-text table with captures all the text in nested tables, formatting it as a reasonable facsimile of a table.

This implements the solution described and spiked in PR #1952.


CHANGELOG.md
```diff
@@ -10,6 +18,7 @@
...
 * **Map full output schema for table-based destination connectors** A full schema was introduced to map the type of all output content from the json partition output and mapped to a flattened table structure to leverage table-based destination connectors. The delta table destination connector was updated at the moment to take advantage of this.
 * **Incorporate multiple embedding model options into ingest, add diff test embeddings** Problem: Ingest pipeline already supported embedding functionality, however users might want to use different types of embedding providers. Enhancement: Extend ingest pipeline so that users can specify and embed via a particular embedding provider from a range of options. Also adds a diff test to compare output from an embedding module with the expected output
+* **Support nested DOCX tables.** In DOCX, like HTML, a table cell can itself contain a table. In this case, create nested HTML tables to reflect that structure and create a plain-text table with captures all the text in nested tables, formatting it as a reasonable facsimile of a table.
```
qued, line 21: Looks like this was added twice.

qued (approved):
Looks like the changelog update was duplicated, but other than that LGTM!

CHANGELOG.md
scanny: k, fixed, thanks Alan! :)

PR #2023, opened 2023-11-07
fix: pass partition arguments to api when partitioning with unstructured-ingest and --partition-by-api

### LINKED ISSUE #1064: bug/unstructured ingest: partition parameters are not passed into request with `partition-by-api`
**Describe the bug**
while making a request.post with command partition-by-api, the partition parameters are not passed in: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/interfaces.py#L229, should pass all the partition parameters in format partition_XX with data field of request like data=partition_kwargs, example: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/api.py#L64

**Definition of Done**
* add partition-by-api ingest tests with existing partition parameters (four as i remember)

ryannikolaidis:
Closes #1064 

When using the `--partition-by-api` flag via unstructured-ingest, none of the partition arguments are forwarded, meaning that these options are disregarded. With this change, we now pass through all of the relevant partition arguments to the api. 

## Changes

* parse and pass relevant partition arguments to the api in unstructured-ingest
* bonus: leverage an existing `partition.api` function to call out to the api rather than including duplicative request logic in unstructured ingest
* bonus: --pdf-infer-table-structure is now a flag not an arg (it defaults false anyways, this is more succinct and consistent with similar parameters)
* bonus: adds `hi_res_model_name` so a user can specify the model to leverage when using a hi_res strategy.

## Testing

* update against_api.sh source test script to specify a partition argument and validates that the response from the api respected the argument
* manually ran a request and validated that it was processed with chipper as specified (not sure if we want to bake a chipper request into the ci tests) (validated that the response leveraged the chipper model):

```
    local \
    --output-dir /tmp/ingest-requests/chipper \
    --verbose \
    --reprocess \
    --strategy hi_res \
    --partition-by-api \
    --hi-res-model-name chipper \
    --api-key "$API_KEY" \
    --input-path 'example-docs/layout-parser-paper-with-table.pdf'
```


unstructured/ingest/interfaces.py
```diff
@@ -453,21 +453,17 @@ def partition_file(
...
+                k: str(v) for k, v in partition_kwargs.items() if v is not None
+            }
+            elements = partition_via_api(
+                filename=str(self.filename),
+                api_key=partition_config.api_key,
+                api_url=endpoint,
+                **passthrough_partition_kwargs,
+            )
+            # TODO: add m_data_source_metadata to unstructured-api pipeline_api and then
+            # pass the stringified json here
```
ahmetmeleq, line 466: Is this to be addressed before the PR gets merged?

ahmetmeleq:
1 comment, otherwise lgtm

unstructured/ingest/interfaces.py
ryannikolaidis: nope, that was an existing todo that we don't want to drop (see in the red changes)

potter-potter (approved):
looks solid

PR #2026, opened 2023-11-07
Update paddle_ocr.py

l4b4r4b4b4:
Use latest paddleocr lib & OCR model

### DIFF
--- a/unstructured/partition/utils/ocr_models/paddle_ocr.py
+++ b/unstructured/partition/utils/ocr_models/paddle_ocr.py
@@ -6,7 +6,7 @@ from unstructured.logger import logger
 @functools.lru_cache(maxsize=None)
 def load_agent(language: str = "en"):
     import paddle
-    from unstructured_paddleocr import PaddleOCR
+    from paddleocr import PaddleOCR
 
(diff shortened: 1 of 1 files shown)


qued:
Hi @l4b4r4b4b4, we can't use `paddleocr` directly as they are dependent on a package (PyMuPDF) that has a non-compatible license.

l4b4r4b4b4:
@qued so that means it would still be possible to either
1. Bump unstructured_paddleocr without PyMuPDF
2. Or even just pass the latest model as ocr_version param 
To use latest PaddleOCR model in unstructured partition@?

qued:
For 1. yes, that's the default way forward. 2. would require a code change but as long as it didn't introduce a non-compatible dependency, that would work.

l4b4r4b4b4:
@qued ok will try that out with an env variable.

l4b4r4b4b4:
New [PR](https://github.com/Unstructured-IO/unstructured/pull/2044) adding paddle ocr version as env variable to docker file with default value `PP-OCRv4` and passing it to PaddleOCR function calls.

PR #2037, opened 2023-11-08
Jj/2011 missing languages metadata

### LINKED ISSUE #2011: bug: missing `languages` metadata in curl request response
As mentioned [here](https://github.com/Unstructured-IO/unstructured/pull/1996#issuecomment-1792804910), the `languages` metadata isn't showing up in the response from curl requests from the command line. This needs to be investigated and resolved.

```
curl -X 'POST'  'http://0.0.0.0:8000/general/v0/general' \
-F 'files=@sample-docs/layout-parser-paper-with-table.jpg' \
-F 'languages=eng' \
-F 'strategy=hi_res' \
-F 'pdf_infer_table_structure=True' | jq -C . | less -R

[
  {
    "type": "Title",
    "element_id": "8e5839c2fb9b4d6b78cd2b1c1f5bed02",
    "metadata": {
      "filename": "layout-parser-paper-with-table.jpg",
      "filetype": "image/jpeg",
      "page_number": 1
    },
    "text": "LayoutParser: A Unified Toolkit for DL-Based DIA 5"
  },
```

Coniferish:
### Summary
Closes #2011 

`languages` was missing from the metadata when partitioning pdfs via `hi_res` and `fast` strategies and missing from image partitions via `hi_res`. This PR adds `languages` to the relevant function calls so it is included in the resulting elements.

### Testing
On the main branch, `partition_image` will include `languages` when `strategy='ocr_only'`, but not when `strategy='hi_res'`:
```
filename = "example-docs/english-and-korean.png"

elements = partition_image(filename, strategy="ocr_only", languages=['eng', 'kor'])
elements[0].metadata.languages

elements = partition_image(filename, strategy="hi_res", languages=['eng', 'kor'])
elements[0].metadata.languages
```

For `partition_pdf`, `'ocr_only'` will include `languages` in the metadata, but `'fast'` and `'hi_res'` will not.
```
filename = "example-docs/korean-text-with-tables.pdf"

elements = partition_pdf(filename, strategy="ocr_only", languages=['kor'])
elements[0].metadata.languages


elements = partition_pdf(filename, strategy="fast", languages=['kor'])
elements[0].metadata.languages


elements = partition_pdf(filename, strategy="hi_res", languages=['kor'])
elements[0].metadata.languages
```

On this branch, `languages` is included in the metadata regardless of strategy


cragwolfe:
@Coniferish , nit on the language codes in the description. I believe we are trying to standardize on two letter character codes per: https://github.com/Unstructured-IO/unstructured/blob/ad14321/unstructured/partition/lang.py#L187 , even though the old ones still work.

Coniferish:
@shreyanid, can you confirm that we're standardizing on two-character language codes? Everything we've done up to this point (that I've seen) has been three-character. I think I'll need to change a number of the language checks if we are wanting to standardize on 3-char.

shreyanid:
We are standardizing on 3 letter codes following the ISO 639-3 standard, mentioned in the conversion function [here](https://github.com/Unstructured-IO/unstructured/blob/ad14321016533dc03c1782f6ebea00bc9c804846/unstructured/partition/lang.py#L227). If you have seen standardization to 2-character anywhere let me know, only langdetect outputs 2 letter codes but they should get converted to 3 letter codes before addition to metadata and such.

test_unstructured/partition/pdf_image/test_image.py
```diff
@@ -195,6 +195,31 @@ def test_partition_image_raises_with_invalid_language(
...
+        DIRECTORY,
+        "..",
+        "..",
+        "..",
+        "example-docs",
+        "english-and-korean.png",
+    )
+    elements = image.partition_image(
+        filename=filename,
+        ocr_languages="eng+kor",
```
shreyanid, line 216: Should we use the `languages` parameter for this test instead? I'm ok with either since this old format can get converted

Coniferish: Yah, doesn't make a difference, but I'll change it since that's not what this test is supposed to be testing. Thanks for catching

unstructured/partition/pdf.py
```diff
@@ -553,6 +524,7 @@ def _process_pdfminer_pages(
     fp: BinaryIO,
     filename: str = "",
     include_page_breaks: bool = False,
+    languages: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
Coniferish: I tried removing the default, but then I was getting an linting error about having a non-default argument coming after a default argument. I was unsure about changing the order of the arguments to address that since I have been told to not change the order of the arguments in case users are using positional arguments. Thinking about it, though, that shouldn't be the case for private functions, so I will change it if all of that sounds correct to you.

PR #2043, opened 2023-11-09
Dynamic ElementMetadata implementation

scanny:
### Executive Summary
The structure of element metadata is currently static, meaning only predefined fields can appear in the metadata. We would like the flexibility for end-users, at their own discretion, to define and use additional metadata fields that make sense for their particular use-case.

### Concepts
A key concept for dynamic metadata is _known field_. A known-field is one of those explicitly defined on `ElementMetadata`. Each of these has a type and can be specified when _constructing_ a new `ElementMetadata` instance. This is in contrast to an _end-user defined_ (or _ad-hoc_) metadata field, one not known at "compile" time and added at the discretion of an end-user to suit the purposes of their application. 

An ad-hoc field can only be added by _assignment_ on an already constructed instance.

### End-user ad-hoc metadata field behaviors

An ad-hoc field can be added to an `ElementMetadata` instance by assignment:
```python
>>> metadata = ElementMetadata()
>>> metadata.coefficient = 0.536
```
A field added in this way can be accessed by name:
```python
>>> metadata.coefficient
0.536
```
and that field will appear in the JSON/dict for that instance:
```python
>>> metadata = ElementMetadata()
>>> metadata.coefficient = 0.536
>>> metadata.to_dict()
{"coefficient": 0.536}
```
However, accessing a "user-defined" value that has _not_ been assigned on that instance raises `AttributeError`:
```python
>>> metadata.coeffcient  # -- misspelled "coefficient" --
AttributeError: 'ElementMetadata' object has no attribute 'coeffcient'
```

This makes "tagging" a metadata item with a value very convenient, but entails the proviso that if an end-user wants to add a metadata field to _some_ elements and not others (sparse population), AND they want to access that field by name on ANY element and receive `None` where it has not been assigned, they will need to use an expression like this:
```python
coefficient = metadata.coefficient if hasattr(metadata, "coefficient") else None
``` 

### Implementation Notes

- **ad-hoc metadata fields** are discarded during consolidation (for chunking) because we don't have a consolidation strategy defined for those. We could consider using a default consolidation strategy like `FIRST` or possibly allow a user to register a strategy (although that gets hairy in non-private and multiple-memory-space situations.)
- ad-hoc metadata fields **cannot start with an underscore**.
- We have no way to distinguish an ad-hoc field from any "noise" fields that might appear in a JSON/dict loaded using `.from_dict()`, so unlike the original (which only loaded known-fields), we'll rehydrate anything that we find there.
- No real type-safety is possible on ad-hoc fields but the type-checker does not complain because the type of all ad-hoc fields is `Any` (which is the best available behavior in my view).
- We may want to consider whether end-users should be able to add ad-hoc fields to "sub" metadata objects too, like `DataSourceMetadata` and conceivably `CoordinatesMetadata` (although I'm not immediately seeing a use-case for the second one).


cragwolfe:
it looks like before, metadata elements were sorted by key name when serialized to json. is there a reason that cannot still be the case?

scanny:
@crag Yeah, I thought that too, but I found counter-examples so gave up on that idea.

However, now that you mention it and I think about it again, there would be benefits to doing that in that the metadata would be canonicalized and then any incremental later changes would produce small diffs and random changes in insertion-order or whatever would not produce diffs. As long as we're resetting the ingest expectations we might as well improve things :)

I'll add that ... okay, that's in. Would probably be more reliable to do that when serializing to JSON (or both places) but we can take that up another day.

cragwolfe:
Steve, I'm not sure of creating the new dict object there as we may not always be getting the dict for the sake of json serialization. could the order of the keys instead be enforced in `elements_to_json()` in https://github.com/Unstructured-IO/unstructured/blob/22aedc4d6f9b97c0192073a194c22781e73e556a/unstructured/staging/base.py#L115C28-L115C28 ?

scanny:
Hmm, we _could_ do it there with `json.dumps(element_dict_list, sort_keys=True)`.

I'm not seeing how we could miss `ElementMetadata.to_dict()` in the process because `elements_to_json()` calls `convert_to_dict()` calls `element.to_dict()` which calls `ElementMetadata.to_dict()`.

However, it couldn't hurt and if there is indeed some way to escape `ElementMetadata.to_dict()` that would handle that case. Probably create a bigger ingest diff, but if we're resetting that anyway we could be assured all those expectations were canonicalized going forward.

I'll give it a shot and see what it does ... :)

scanny:
@cragwolfe hmm, looks like the ingest tests use something other than `elements_to_json()` because the results here reflect _none_ of the keys being sorted.

We can certainly do _both_ if you think that's a good idea, for now that's what I'll do, but interesting datapoint on the ingest tests ... hmm, yeah, it looks like ingest tests use `convert_to_dict()` but not `elements_to_json()`.

cragwolfe (approved):

PR #2044, opened 2023-11-09
Use latest PaddleOCR model version with environment variable

l4b4r4b4b4:
Added paddle OCR version as env variable to docker file with default value `PP-OCRv4` and passed it to paddle ocr function calls.


l4b4r4b4b4:
@qued  do you see a reason not to approve, to check if the tests execute successfully?

qued:
Hi @l4b4r4b4b4 . The immediate problem is that `os` needs to be imported, that's easily fixed.

A more complex problem is that we use our own fork of PaddleOCR to remove a dependency that conflicts with our open source license. In our fork as far as I can tell version `PP-OCRv4` is not supported (https://github.com/Unstructured-IO/unstructured.PaddleOCR/blob/main/unstructured_paddleocr/paddleocr.py#L65). 

There have been explorations on updating the fork, which might solve the issue, I'll have to check on the status of that.

MthwRobinson:
Closing since this PR hasn't been active for a while, but this PR seems reasonable and seems to be an `import os` statement away from being complete. Feel free to reopen if this would still be helpful.

PR #2055, opened 2023-11-10
Fix: fast strategy fallback to ocr only

### LINKED ISSUE #2038: bug: using "fast" strategy to partition a PDF can still fall back to OCR only
**Describe the bug**
Calling:

    elements = partition(filename=filename, strategy="fast")

shows the following output:

    PDF text is not extractable. Cannot use the fast partitioning strategy. Falling back to partitioning with the ocr_only strategy.

for [9493801-p17.pdf](https://github.com/Unstructured-IO/unstructured/files/13292884/9493801-p17.pdf) , as of unstructured==0.10.29.

**Expected behavior**

An empty list of elements are returned practically instantly. (rather than falling back to "ocr_only" and taking significant time and compute). Note that the "auto" strategy should return "hi_res" output.

christinestraub:
Closes #2038.
### Summary
The `fast` strategy should not fall back to a more expensive strategy.

### Testing
For [9493801-p17.pdf](https://github.com/Unstructured-IO/unstructured/files/13292884/9493801-p17.pdf), the following code should return an empty list.

```
elements = partition(filename=filename, strategy="fast")
```


cragwolfe (approved):
Test instructions work as expected, and "auto" correctly resolves to "hi_res" for this file.
LGTM once tests pass.

cragwolfe (approved):

Issue #2056, opened 2023-11-11
bug/VSCode Reopen in Dev Container fails


IAINATTWATER:
**Describe the bug**
Having pulled the container per the instructions, created a  devcontainer.json in the .devcontainer folder. Chose reopen in container and it fails with

> Failed to cache user env: Shell server terminated (code: 2, signal: null)

**To Reproduce**
devcontainer.json:
```
{
    "name":"unstructured",
    "image":"downloads.unstructured.io/unstructured-io/unstructured:latest",
    "runArgs": ["--gpus=all","--name=unstructured",
             "--net=host","-it","--privileged=true",
             "-v","/home/****/.cache/huggingface/hub/:/workspaces/hub",
             "-v","/tmp/.X11-unix:/tmp/.X11-unix/",
             "-v","/var/run/docker.sock:/var/run/docker.sock",
             "--env-file", ".devcontainer/devcontainer.env"
            ]
}
```

**Expected behavior**
Folder structure be reopened in unstructured container and allow development as normal.

**Log**

> [2023-11-11T15:40:34.999Z] Failed to cache user env: Shell server terminated (code: 2, signal: null)

Thank you

IAINATTWATER:
Thanks - that's fixed it for me!

PR #2057, opened 2023-11-11
Remove ssh-keyscan and files

tabossert:
This was legacy and is no longer needed.  It also has the effect of incorrect owner for known_hosts of notebook-user

Relates to: #2056

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.10.31-dev1
+
+### Enhancements
+
+### Features
+
+### Fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.10.31-dev1
+
+### Enhancements
+
+### Features
+
+### Fixes
+* **Remove default user ./ssh folder** The default notebook user during image build would create the known_hosts file with incorrect ownership, this is legacy and no longer needed so it was removed.
```
cragwolfe, line 8: Let's remove the bullet, now only including items that impact end consumers of the library

cragwolfe (approved):

PR #2068, opened 2023-11-14
add mongo db destination connector

rbiseck3:
### Description
This adds the basic implementation of pushing the generated json output of partition to mongodb. None of this code provisions the mondo db instance so things like adding a search index around the embedding content must be done by the user. Any sort of schema validation would also have to take place via user-specific configuration on the database. This update makes no assumptions about the configuration of the database itself.


test_unstructured_ingest/python/test-ingest-mongodb.py
```diff
@@ -0,0 +1,520 @@
+#!/usr/bin/env python
+import time
+
+import click
+from pymongo.mongo_client import MongoClient
+from pymongo.operations import SearchIndexModel
+
+exact_embedding = [
```
ryannikolaidis, line 8: what was the thinking around hardcoding this? will this be brittle to any changes in partitioning (and no easy way to update like with fixtures)?

test_unstructured_ingest/python/test-ingest-mongodb.py
```diff
@@ -0,0 +1,520 @@
...
+    print(f"checking the count in the db ({count}) matches what's expected: {expected_records}")
+    assert (
+        count == expected_records
+    ), f"expected count ({expected_records}) does not match how many records were found: {count}"
+    print("successfully checked that the expected number of records was found in the db!")
+
+
+@cli.command()
+@click.pass_context
+def check_vector(ctx):
```
ryannikolaidis, line 478: quick docstring? something to roughly capture what you're doing. wait for status ready and then search against exact match and validate against perfect match score

test_unstructured_ingest/python/test-ingest-mongodb.py
rbiseck3: Open to ideas, tried to come up with a test specific to the vector search index functionality. The result should be deterministic given that we're using huggingface now but this does leave the possibility that the text may change. Since the embedding is based on the text in an element, I don't see many reasons this should change, unlike the metadata.

docs/source/ingest/destination_connectors/mongodb.rst
```diff
@@ -0,0 +1,68 @@
+MongoDB
+======================
+
+Batch process all your records using ``unstructured-ingest`` to store structured outputs locally on your filesystem and upload those local files to an MongoDB collection.
+
+First you'll need to install the azure cognitive search dependencies as shown here.
```
ryannikolaidis, line 6:
```suggestion
First you'll need to install the MongoDB dependencies as shown here.
```

test_unstructured_ingest/python/test-ingest-mongodb.py
ryannikolaidis: yea, only thing I could think is that we grab the embeddings from an element in the partitioned results file and then use that as the query?

ryannikolaidis: totally agree that this is likely not going to change often if at all, but thinking around "how bad is it" the one time it does.
if you're not feeling the suggestion I just mentioned, I guess alternatively would at least leave some documentation in case it ever breaks and code snippet for how to regenerate this.

rbiseck3: ya that's probably the best approach for this. Updated accordingly.

ryannikolaidis (approved):
looks great!

test_unstructured_ingest/python/test-ingest-mongodb.py
ryannikolaidis: thank you!

PR #2071, opened 2023-11-14
feat: update all ingest tests to use huggingface for embeddings

rbiseck3:
### Description
Update any use of OpenAI for generating embeddings in the ingest tests to use Huggingface

**Bonus Changes:**
* Remove duplicate delta table test
* Delete delta table destination directory at the beginning of the test to make sure it doesn't exist and prevent the test from breaking.


ryannikolaidis:
otherwise looks good

ryannikolaidis (approved):

PR #2074, opened 2023-11-14
Refactor: partition pdf

christinestraub:
### Summary
- add constants for strategies
- add `_process_uncategorized_text_elements()` to remove code block duplication
### Testing
CI should pass.


test_unstructured/partition/pdf_image/test_chipper.py
```diff
@@ -1,13 +1,14 @@
 import pytest
 
 from unstructured.partition import pdf
+from unstructured.partition.utils.constants import PartitionStrategy
```
badGarnet, line 4: do we want a changelog item to advertise that we got this class for the constants now?

badGarnet (approved):

test_unstructured/partition/pdf_image/test_chipper.py
christinestraub: Sounds good. Thanks!

cragwolfe (approved):

PR #2076, opened 2023-11-15
fix(chunk): #1985 mis-splits of Table chunks

### LINKED ISSUE #1985: XLSX Chunk by TItle HTML not working properly
I have noticed that the 'text_as_html' attribute in the chunk metadata contains the HTML for the entire sheet, as opposed to just the chunk (example code below). I am iterating the chunks and using the HTML if the category is Table. So the value in 'chunk.text' is correct, but chunk.metadata.text_as_html contains the same value as "elements". This seems to be caused by the below line of code, as the attribute is not located in the area it's looking.

```
from io import BytesIO
from unstructured.partition.xlsx import partition_xlsx

filepath = 'C:\\Path\\To\\file.xlsx'

with open(filepath, "rb") as fh:
    bytes_io = BytesIO(fh.read())

elements = partition_xlsx(file=bytes_io)

chunks = chunk_by_title(elements, multipage_sections=True)

for i, chunk in enumerate(chunks):
    if chunk.metadata.page_number is None:
        page_list = [1]
    else:
        page_list = [chunk.metadata.page_number]
    # substitute html if text is a table
    if chunk.category == 'Table':
        chunk_text = chunk.metadata.text_as_html
    else:
        chunk_text = chunk.text
```

https://github.com/Unstructured-IO/unstructured/blob/2f553333bd6fe7462b7f52a04f1c9f7fdf619e90/unstructured/chunking/title.py#L33

scanny:
Closes #1985 

**Summary.** Due to an interaction of coding errors, HTML text in `TableChunk` splits of a `Table` element were repeating the entire HTML for the table in each chunk.

**Technical Summary.** This behavior was fixed but not published in the last chunking PR of a series. Finish up that PR and submit it all here. This PR extracts chunking to the particular Section type (each has their own distinct chunking behavior).


test_unstructured/chunking/test_title.py
```diff
@@ -189,7 +190,7 @@ def test_split_elements_by_title_and_table():
         Title("A Great Day"),
         Text("Today is a great day."),
         Text("It is sunny outside."),
-        Table("<table></table>"),
+        Table("Heading\nCell text"),
```
scanny, line 193: This global change is purely psychological, but `"<table></table>" would never appear as `Table.text`, rather it would appear as `Table.metadata.text_as_html`. This apparent misconception could be related to the underlying mistake in the code leading to this so I made these more realistic so the reader's intuition about these elements is not led astray.

test_unstructured/chunking/test_title.py
```diff
@@ -705,19 +706,130 @@ def test_it_considers_separator_length_when_sectioning():
...
+            "</thead>\n"
+            "<tbody>\n"
+            "<tr><td>Lorem ipsum  </td><td>adipiscing   </td></tr>\n"
+            "</tbody>\n"
+            "</table>"
+        )
+        with pytest.raises(StopIteration):
+            next(chunk_iter)
+
+    def but_it_splits_its_table_into_TableChunks_when_the_table_text_exceeds_the_window(self):
```
qued, line 758: Confirmed this reproduces the bug that previously existed where `text_as_html` was sought in the table element rather than the table element metadata.

CHANGELOG.md
```diff
@@ -12,6 +13,7 @@
 * **Include `languages` in metadata when partitioning strategy='hi_res' or 'fast'** User defined `languages` was previously used for text detection, but not included in the resulting element metadata for some strategies. `languages` will now be included in the metadata regardless of partition strategy for pdfs and images.
 * **Handle a case where Paddle returns a list item in ocr_data as None** In partition, while parsing PaddleOCR data, it was assumed that PaddleOCR does not return None for any list item in ocr_data. Removed the assumption by skipping the text region whenever this happens.
 * **Fix some pdfs returning `KeyError: 'N'`** Certain pdfs were throwing this error when being opened by pdfminer. Added a wrapper function for pdfminer that allows these documents to be partitioned.
+* **Fix mis-splits on Table chunks.** Closes #1985. Remedies repeated appearance of full .text_as_html on metadata of each TableChunk split from a Table element too large to fit in the chunking window.
```
qued, line 16: Changed formatting to render references to code in monospace. Also removed reference to github issue as we typically don't include those in the changelog.
```suggestion
* **Fix mis-splits on `Table` chunks** Remedies repeated appearance of full `.text_as_html` on metadata of each `TableChunk` split from a `Table` element too large to fit in the chunking window.
```

qued (approved):
Made a suggestion on the formatting for the changelog, but overall LGTM. I don't think it's quite right how the text and HTML are chunked as if they are in isolation, but aligning them sounds really hard and out of scope here.

PR #2079, opened 2023-11-15
chore: for ingest-test metrics, also check subdirs

Klaijan:
- Copy script only went through one layer of subdirectory so it did not found the match between manifest file and structured output. Now edited to search all subdirectories.
- `set -e` causes the script to exit at any exit rather than `exit 0`, fix all scripts that needs to run the copy script to be `set +e` right before the check diff, then back to `set -e` after
- Edit the default evaluation metrics output from `metrics` to `metrics-tmp` to account for diff check
- Add a script that checks the differences between old eval metric output (metrics) and new eval metrics output (metrics-tmp)


test_unstructured_ingest/src/azure.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set +e
```
cragwolfe, line 3: the `set -e` means that the script will exit on failure if any lines in the bash script fail, as it should. so, need to find another way to have the script gracefully complete without doing this. in the worst case, could put a `set +e` before the line causing the issue that is OK to fail, immediately followed by `set -e`, but ideally the line in question does not need that in the first place.

Klaijan: Having `set -e` causes the script to exit at check diff, and therefore never reaches the copy line of the script. That is why the cct for s3 did not run.

cragwolfe: i see, so we do want the script to fail in normal conditions there if there is a diff. we can do that by preserving the exit code, and then exiting with that. something like:

```
set +e
"$SCRIPT_DIR"/check-diff-expected-output.sh $OUTPUT_FOLDER_NAME
EXIT_CODE=$?
set -e

"$SCRIPT_DIR"/evaluation-ingest-cp.sh "$OUTPUT_DIR" "$OUTPUT_FOLDER_NAME"
exit $EXIT_CODE
```

cragwolfe (approved):

PR #2080, opened 2023-11-15
fix(ingest): canonicalize ingest JSON

scanny:
Canonicalize JSON produced for ingest tests such that incidental changes is _form_ of the JSON objects (keys moving around) that does not change the _content_ of that JSON object does not trigger an ingest-test failure.


cragwolfe:
Might as well change it everywhere?

    cd unstructured; find -name '*py' | xargs grep json.dumps | grep -v sort_keys

Issue #2028, opened 2023-11-07
bug/'UnstructuredTableTransformerModel' object has no attribute 'model'


awalker4:
API users are hitting this error. It appears to be only UnstructuredTableTransformerModel.

```
File "unstructured/partition/pdf.py", line 416, in _partition_pdf_or_image_local
    final_layout = process_data_with_ocr(
  File "unstructured/partition/ocr.py", line 81, in process_data_with_ocr
    merged_layouts = process_file_with_ocr(
  File "unstructured/partition/ocr.py", line 167, in process_file_with_ocr
    raise e
  File "unstructured/partition/ocr.py", line 156, in process_file_with_ocr
    merged_page_layout = supplement_page_layout_with_ocr(
  File "unstructured/partition/ocr.py", line 240, in supplement_page_layout_with_ocr
    page_layout.elements[:] = supplement_element_with_table_extraction(
  File "unstructured/partition/ocr.py", line 276, in supplement_element_with_table_extraction
    element.text_as_html = table_agent.predict(cropped_image, ocr_tokens=table_tokens)
  File "unstructured_inference/models/tables.py", line 53, in predict
    super().predict(x)
  File "unstructured_inference/models/unstructuredmodel.py", line 32, in predict
    if self.model is None:
```

yuming-long:
link to [sentry error](https://unstructuredio.sentry.io/issues/4579740245/events/55499300c641490b835488623ad56d4a/?project=4505909127086080)

yuming-long:
Didn't find a way to reproduce error on API level, but i think using table_agent from inference in https://github.com/Unstructured-IO/unstructured/pull/2087 could fix this

PR #2087, opened 2023-11-15
Chore: Import tables_agent from inference

yuming-long:
Related to https://github.com/Unstructured-IO/unstructured/issues/2028

Import `tables_agent` from inference and don't init in unst again

copy the same logic from [`interpret_table_block`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/layoutelement.py#L89) (currently deprecated).


unstructured/partition/ocr.py
```diff
@@ -315,7 +315,8 @@ def init_table_agent():
 
     global table_agent
 
-    if table_agent is None:
+    if not hasattr(table_agent, "model"):
+        logger.info("Loading the Table Transformer Model...")
```
badGarnet, line 319: maybe the problem is usage of global here? Not sure why this is necessarily safer. 
we could convert table agent into a singleton class https://python-patterns.guide/gang-of-four/singleton/ 
the example is : https://python-patterns.guide/gang-of-four/singleton/#a-more-pythonic-implementation

yuming-long: yeah make sense, i think i was trying to make sure the `table_agent` always has attribute `model` here and i am not sure if a singleton class can help?

cragwolfe (approved):

PR #2088, opened 2023-11-15
drop cloud cli deps

rbiseck3:
### Description
To not require additional dependencies on cloud-related CLIs (i.e. gcloud and az), using python and the existing dependencies already used to run out code to interact with those providers for overhead work associated with destination ingest tests.


rbiseck3:
@tabossert, not sure how this might impact things that you set up for the CI...

tabossert:
Shouldn't have any real effects, if this passes CI then I don't see any issue with this.

tabossert (approved):

PR #2094, opened 2023-11-16
chore: ingest-script cleanup, better skip condition

cragwolfe:
When testing ingest tests, one often wants to keep the .json output or generated metrics files around for inspection after the fact. This updates the bash condition to actually honor the comment that mentions

    # export UNSTRUCTURED_CLEANUP_DEV_FIXTURES=1
    
** Test Instructions **

Run:

    export UNSTRUCTURED_CLEANUP_DEV_FIXTURES=1
    ./test_unstructured_ingest/src/s3.sh
    ./test_unstructured_ingest/evaluation-metrics.sh text-extraction
    
and witness test directories/files do not get cleaned up. E.g., `test_unstructured_ingest/metrics-tmp/`. One can also add a `set -x` at the top of test_unstructured_ingest/cleanup.sh to see what is getting skipped (it's a lot!).

### DIFF
--- a/test_unstructured_ingest/cleanup.sh
+++ b/test_unstructured_ingest/cleanup.sh
@@ -4,7 +4,9 @@
 function cleanup_dir() {
   # NOTE(crag): for developers that want to always clean up .json outputs, etc., set
   # UNSTRUCTURED_CLEANUP_DEV_FIXTURES=1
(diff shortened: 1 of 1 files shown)


Klaijan (approved):
lgtm

PR #2096, opened 2023-11-16
Chore: move hi res initialization `initialize.py` file out of ingest

yuming-long:
Move Hi_res model initialization file out of ingest to `partition` dir


rbiseck3:
I don't see the actual code that does the initialization moved as part of this PR...

yuming-long:
its a filename change `File renamed without changes.` so it doesn't show up in the changes i guess

cragwolfe (approved):

PR #2099, opened 2023-11-16
feat: create CI cache if one not already found at each step

rbiseck3:
### Description
To avoid the entire pipeline breaking if something happens to the cache it depends on between steps, recreate it at each one if there is no cache hit. To avoid a lot of duplicate code, both cache setups were abstracted away into custom github actions. 

**Additional Changes:**
* Downgrade runner instance for setup job
* Remove any python setup from docker file test


.github/actions/create-cache/action.yml
```diff
@@ -0,0 +1,34 @@
+
+name: 'Base Cache Build'
+description: 'Build the base python cache for CI to use'
+inputs:
+  python-version:
+    description: 'which python version to user'
```
ryannikolaidis, line 6:
```suggestion
    description: 'python version associated with the cache'
```

.github/actions/create-ingest-cache/action.yml
```diff
@@ -0,0 +1,46 @@
+
+name: 'Ingest Cache Build'
+description: 'Build ingest specific cache on top of base cache'
+inputs:
+  python-version:
+    description: 'which python version to user'
```
ryannikolaidis, line 6:
```suggestion
    description: 'python version associated with the cache'
```

.github/actions/create-ingest-cache/action.yml
```diff
@@ -0,0 +1,46 @@
...
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/cache/restore@v3
+      id: virtualenv-cache
+      with:
+        path: |
+          nltk_data
+          .venv
+        key: unstructured-${{ runner.os }}-${{ inputs.python-version }}-${{ env.GHA_CACHE_KEY_VERSION }}-${{ hashFiles('requirements/*.txt') }}
```
ryannikolaidis, line 20: same as note aboe on env. GHA_CACHE_KEY_VERSION

.github/actions/create-ingest-cache/action.yml
```diff
@@ -0,0 +1,46 @@
...
+  python-version:
+    description: 'which python version to user'
+    required: true
+  force-recreate:
+    description: 'regardless of cache existing, recreate it'
+    default: "false"
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/cache/restore@v3
```
ryannikolaidis, line 14: give this a name that makes it clear we're trying to restore base cache here?

.github/actions/create-ingest-cache/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        path: |
+          nltk_data
+          .venv
+        key: unstructured-${{ runner.os }}-${{ inputs.python-version }}-${{ env.GHA_CACHE_KEY_VERSION }}-${{ hashFiles('requirements/*.txt') }}
+    - name: Setup virtual environment (no cache hit)
+      if: steps.virtualenv-cache.outputs.cache-hit != 'true'
+      uses: ./.github/actions/create-cache
+      with:
+        python-version: ${{ inputs.python-version }}
+        force-recreate: 'true'
```
ryannikolaidis, line 26: nice

.github/actions/create-ingest-cache/action.yml
```diff
@@ -0,0 +1,46 @@
...
+    - uses: actions/cache@v3
+      id: ingest-virtualenv-cache
+      with:
+        path: |
+          .venv
+          nltk_data
```
ryannikolaidis, line 32: realize this would have been true before, but does the ingest-specific cache end up with only the additional ingest-specific content? or is the ingest cache actually the base cache + the ingest content?

.github/workflows/ci.yml
```diff
@@ -339,6 +283,12 @@ jobs:
         key: unstructured-ingest-${{ runner.os }}-${{ matrix.python-version }}-${{ env.GHA_CACHE_KEY_VERSION }}-${{ hashFiles('requirements/ingest/*.txt') }}
         # Due to the dependency on setup_ingest, the cache should exist before this is ran. Set failed if it wasn't found.
         fail-on-cache-miss: true
+    - name: Setup virtual environment (no cache hit)
+      if: steps.virtualenv-cache.outputs.cache-hit != 'true'
+      uses: ./.github/actions/create-ingest-cache
+      with:
+        python-version: ${{ matrix.python-version }}
+        force-recreate: 'true'
```
ryannikolaidis, line 291: if we fail on cache miss I would think that means we never get to the next step. is that not true??

ryannikolaidis: ...at which point checking cache-hit seems moot

tabossert:
I took a look and I don't have anything other than what @ryannikolaidis is suggesting.

rbiseck3:
@ryannikolaidis, slight refactor since some of your comments. The use of `force_recreate` should never have been required but it got around some weird issues I was seeing in the CI. I believe the most recent changes addressed all of that.

.github/workflows/ci.yml
```diff
@@ -532,23 +426,19 @@ jobs:
   test_dockerfile:
     runs-on: ubuntu-latest-m
     needs: [ setup, lint ]
+    strategy:
+      matrix:
+        python-version: [ "3.8","3.9","3.10","3.11" ]
```
ryannikolaidis, line 431: why are you adding this?

ryannikolaidis: note this is the dockerfile test...

rbiseck3: So we not want it being tested against all python versions we support?

ryannikolaidis: it's in a docker container, I don't think the python version would affect it? unless I'm missing something

rbiseck3: Forgot we’re using our own base image, I had thought we were dynamically building it off a base python image in which case this would make sense. Will revert.

.github/workflows/ci.yml
```diff
@@ -152,24 +116,14 @@ jobs:
...
-      if: steps.virtualenv-cache.outputs.cache-hit != 'true'
-      run: |
-        python${{ matrix.python-version}} -m venv .venv
-        source .venv/bin/activate
-        mkdir "$NLTK_DATA"
-        make install-ci
+    - name: Setup virtual environment
+      uses: ./.github/actions/base-cache
+      with:
+        python-version: ${{ matrix.python-version }}
```
ryannikolaidis, line 126: do we still need the python setup step at this point? that's already embedded in base-cache action, right?

.github/workflows/ci.yml
```diff
@@ -265,51 +209,15 @@ jobs:
     strategy:
       matrix:
         python-version: [ "3.8","3.9","3.10","3.11" ]
-    runs-on: ubuntu-latest-m
+    runs-on: ubuntu-latest
```
ryannikolaidis, line 212: do we know why it was switched to the larger size and why it's okay to switch it back now?

rbiseck3: It shouldn't need the large size for setup since it's not using multiple processes to pull in the dependencies. This may have been done in error (by me) when the instance size was being increased for ingest related CI steps.

.github/workflows/ci.yml
rbiseck3: Yes because python setup is only being run in the cache action if the cache wasn't found to create it.

ryannikolaidis: ohh I see, got it!

.github/workflows/ingest-test-fixtures-update-pr.yml
```diff
@@ -43,36 +28,10 @@ jobs:
...
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ env.python-version }}
-      - name: Setup virtual environment (no cache hit)
-        run: |
-          python${{ env.python-version }} -m venv .venv
-          source .venv/bin/activate
-          make install-all-ingest
+          python-version: ${{ env.PYTHON_VERSION }}
+          check-only: 'true'
```
ryannikolaidis, line 34: I'm confused..if we check-only and it DOES exist, it looks like we don't actually restore it. I'm assuming I'm not following some piece of this...

ryannikolaidis (approved):
lgtm! thanks for hopping on this!

PR #2105, opened 2023-11-17
feat: support both single and batch ingest docs

rbiseck3:
### Description
There are some source ingest connectors that would be more efficient to read the content in batches rather than use an entire process per document. For example, reading from ElasticSearch. Given an index with possible hundreds of documents, reading each one individually is not as optimal as reading in batches. To try and maintain as much of the ingest doc paradigm already being supported, a new class `BaseIngestDocBatch` was added to handle reading in batches. It produces a list of `BaseSingleIngestDoc` which is what all current implementations were renamed to.  This list is generated after it runs its `get_files` method. Past the source node, all other steps in the pipeline should not be affected, this is just an optimization for the read step. 

**Additional Changes:**
* Removed use of jq and instead converted this into a fields filter on the content to let the database handle the filtering and limit the amount of data being pulled in.


unstructured/ingest/pipeline/utils.py
```diff
@@ -2,5 +2,5 @@
 
 
 def get_ingest_doc_hash(json_as_dict: dict) -> str:
-    hashed = hashlib.sha256(json_as_dict["filename"].encode()).hexdigest()[:32]
+    hashed = hashlib.sha256(json_as_dict["unique_id"].encode()).hexdigest()[:32]
```
potter-potter, line 5: Whats the reason for this? Is it because a batch is not a filename?

unstructured/ingest/connector/elasticsearch.py
```diff
@@ -63,104 +61,37 @@ class ElasticsearchIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
...
-            return
-        self.source_metadata = SourceMetadata(
-            version=document["_version"],
-            exists=document["found"],
-        )
-
     @SourceConnectionError.wrap
-    @requires_dependencies(["jq"], extras="elasticsearch")
-    @BaseIngestDoc.skip_if_file_exists
+    @requires_dependencies(["elasticsearch"], extras="elasticsearch")
```
potter-potter, line 91: I like removing jq. I don't think it pip installs easily.

test_unstructured_ingest/expected-structured-output/elasticsearch/movies/0-57554198.json
```diff
@@ -7,8 +7,7 @@
           "document_id": "0",
           "index_name": "movies",
           "url": "http://localhost:9200"
-        },
-        "version": "1"
```
potter-potter, line 11: What was this version 1? and why did we remove it. Is this how we were going to incrementally ingest docs?

ryannikolaidis:
adding @ahmetmeleq as a reviewer since he built the ES connector

test_unstructured_ingest/expected-structured-output/elasticsearch/movies/0-57554198.json
rbiseck3: This was a ramification of getting a specific record by id rather than a general scan for a lot of ids at once in a batch call. This value doesn't exist on the response from that call.

rbiseck3: Actually, that might be a configuration I can set to get that back. Let me take a look.

rbiseck3: Added it back in.

unstructured/ingest/pipeline/pipeline.py
```diff
@@ -40,6 +42,21 @@ def get_nodes_str(self):
...
         return " -> ".join([node.__class__.__name__ for node in nodes])
 
+    def expand_batch_docs(self, dict_docs: t.List[dict]) -> t.List[dict]:
+        expanded_docs = []
+        for d in dict_docs:
+            doc = create_ingest_doc_from_dict(d)
+            if isinstance(doc, BaseSingleIngestDoc):
+                expanded_docs.append(doc.to_dict())
+            elif isinstance(doc, BaseIngestDocBatch):
+                expanded_docs.extend([single_doc.to_dict() for single_doc in doc.ingest_docs])
```
potter-potter, line 52: This is nice. I was expecting it to have to loop through the batch and then loop through single. But this does it all at once.

unstructured/ingest/pipeline/source.py
```diff
@@ -12,7 +17,34 @@
...
 @dataclass
 class Reader(SourceNode):
-    def run(self, ingest_doc_dict: dict) -> t.Optional[str]:
+    def get_single(self, doc: BaseSingleIngestDoc, ingest_doc_dict: dict) -> str:
+        if (
+            not self.read_config.re_download
+            and doc.filename.is_file()
+            and doc.filename.stat().st_size
+        ):
+            logger.info(f"File exists: {doc.filename}, skipping download")
```
potter-potter, line 26: I guess with batch there is no real way to "skip download" since at that time we won't know whats in the batch compared to what already exists.

How is batch going to work with incremental ingestion? I guess that part comes before the actual deciding what to download?

rbiseck3: We don't support that at the moment so my thoughts are right now is lets optimize for what we do support and figure that out when we cross the bridge concerning incremental workflows. We'll need some sort of state when we introduce incremental workflows so we can play around with the data there. We have no guarantee of last modified dates existing on the ES content anyways so we'd most likely have to reingest the entire thing each time.

potter-potter:
This is all working as expected when I test it with elasticsearch and some of the other connectors.

potter-potter (approved):
Looks good. Works well.

PR #2093, opened 2023-11-16
ci: reorganize eval output folders and add azure to matrix test

Klaijan:
**Summary**
The CI workflow for evaluation previously saved the metric outputs to the `metrics/` folder. Currently structured in subfolders e.g. `metrics/text-extraction` `metrics/element-type` for the folder clean up purpose. 

Additionally, Azure connector is also added to `full_python_matrix_tests` in this PR.

PR #2108, opened 2023-11-17
refactor: measure_text_edit_distance function for aggregation

Klaijan:
- Refactor `metrics/evaluation.py` to accepts `grouping` as parameter. 
- Switch to `DataFrame` for easier analysis and aggregation.


cragwolfe (commented):
i think should wait on merging this until https://github.com/Unstructured-IO/unstructured/pull/2093 merges, since this could potentially impact metrics files that are updated by the ingest tests

cragwolfe (approved):

PR #2113, opened 2023-11-18
fix: assorted partition_html() bugs

### LINKED ISSUE #1928: bug/Empty table content should be a valid html str in metadata.text_as_html
**Describe the bug**
When we fixed [partition_html fails when parsing html with tables with no tbody](https://github.com/Unstructured-IO/unstructured/issues/1807) we set the element.metadata.text_as_html to be ''.

However the empty string cannot be parsed down stream, which mean we have a dead table element among the other elements, requiring special error handling.

Basically, if unstructured parses a html element into a TableElement it should deliver a TableElement with a valid html text .. otherwise downstream client will have to do more exception handling

Potential solution: Instead of using '' for an empty table

use 
```<table><tr><td> </td></tr></table>```

This is a valid html table, and pandas read_html() correctly parses it as an empty list of tables.
`pd.read_html(table_html) == []`

 However note that this works when there is a space between the td tags <td> </td>
 
 
 Additionally there is a deprecation warning introduced by the fix for Bug 1807
 unstructured\documents\html.py:498: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.

**To Reproduce**

```
    table_html = """
    <table>
      <thead>
        <tr><th>Header 1</th><th>Header 2</th></tr>
        </thead>
    </table>
    """
    elements = partition_html(text=table_html)
    tables = pd.read_html(elements[0].metadata.text_as_html)
 ```   
    This produces lxml.etree.ParserError: Document is empty

**Expected behavior**
Code to parse metadata.text_as_html should not fail

**Screenshots**

**Environment Info**
Environment Info
OS version: macOS-13.3.1-x86_64-i386-64bit
Python version: 3.10.10
unstructured version: None
unstructured-inference version: 0.7.7
pytesseract version: 0.3.10
Torch version: 2.1.0
Detectron2 is not installed

**Additional context**

scanny:
Addresses a cluster of HTML-related bugs:
- empty table is identified as bulleted-table
- `partition_html()` emits empty (no text) tables (#1928)
- `.text_as_html` contains inappropriate `<br>` elements in invalid locations.
- cells enclosed in `<thead>` and `<tfoot>` elements are dropped (#1928)
- `.text_as_html` contains whitespace padding

Each of these is addressed in a separate commit below.

Fixes #1928.


test_unstructured_ingest/expected-structured-output/Sharepoint-with-permissions/Shared Documents/ideas-page.html.json
```diff
@@ -106,7 +106,7 @@
         "eng"
       ],
       "page_number": 1,
-      "text_as_html": "<table><br><tbody><br><tr><td>January 2023</td><td>(</td><td>Someone</td><td>fed my essays into GPT to make something that could answer<br>questions based on them, then asked it where good ideas come from.  The<br>answer was ok, but not what I would have said. This is what I would have said.)</td><td>The way to get new ideas is to notice anomalies: what seems strange,<br>or missing, or broken? You can see anomalies in everyday life (much<br>of standup comedy is based on this), but the best place to look for<br>them is at the frontiers of knowledge.</td><td>Knowledge grows fractally.<br>From a distance its edges look smooth, but when you learn enough<br>to get close to one, you&#x27;ll notice it&#x27;s full of gaps. These gaps<br>will seem obvious; it will seem inexplicable that no one has tried<br>x or wondered about y. In the best case, exploring such gaps yields<br>whole new fractal buds.</td></tr><br></tbody><br></table>"
```
scanny, line 109: This is an example of the extra `<br>` elements that were being inserted. Basically everywhere `tabulate` added a newline for human readability it was replaced with a `<br>` element. Note these are not legal positions for a `<br>` element and the resulting HTML only parses when the parser used has that specific permissiveness. Also of course it takes up unnecessary space.

test_unstructured_ingest/expected-structured-output/confluence-diff/MFS/1540126.json
```diff
@@ -17,7 +17,7 @@
...
+      "text_as_html": "<table><tr><td>Driver</td></tr><tr><td>Approver</td></tr><tr><td>Contributors</td></tr><tr><td>Informed</td></tr><tr><td>Objective</td></tr><tr><td>Due date</td></tr><tr><td>Key outcomes</td></tr><tr><td>Status</td><td>NOT STARTED</td><td>/</td><td>IN PROGRESS</td><td>/</td><td>COMPLETE</td></tr></table>"
```
scanny, line 20: This is an interesting case because cells previously present no longer appear. I believe this is due to the `NOT STARTED / IN PROGRESS / COMPLETE` content being in a nested table. The current algorithm for that is to get each of those as a separate "cell", then `tabulate` would pad out the per-row cell-count to equal the longest row. That's why are the removed cells are empty.

A better rendering might be to place all content in a nested table into a single cell (that's what we do for DOCX). I thought I'd address that in a following PR rather than extend this one.

test_unstructured_ingest/expected-structured-output/confluence-diff/MFS/1540126.json
```diff
@@ -327,7 +327,7 @@
         "eng"
       ],
       "page_number": 1,
-      "text_as_html": "<table><br><tbody><br><tr><td>Milestone</td><td>Owner</td><td>Deadline</td><td>Status</td></tr><br><tr><td>         </td><td>     </td><td>        </td><td>      </td></tr><br><tr><td>         </td><td>     </td><td>        </td><td>      </td></tr><br><tr><td>         </td><td>     </td><td>        </td><td>      </td></tr><br></tbody><br></table>"
+      "text_as_html": "<table><tr><td>Milestone</td><td>Owner</td><td>Deadline</td><td>Status</td></tr><tr></tr><tr></tr><tr></tr></table>"
```
scanny, line 330: This one is also interesting because it produces empty rows (rows with no cells). Those rows also appeared in the prior case but with empty cells.

We should probably suppress emission of rows with no cells. I'll add that in.

test_unstructured_ingest/expected-structured-output/confluence-diff/MFS/1540126.json
```diff
@@ -84,7 +84,7 @@
         "eng"
       ],
       "page_number": 1,
-      "text_as_html": "<table><br><tbody><br><tr><td>Must have:   </td></tr><br><tr><td>Nice to have:</td></tr><br><tr><td>Not in scope:</td></tr><br></tbody><br></table>"
+      "text_as_html": "<table><tr><td>Must have:</td></tr><tr><td>Nice to have:</td></tr><tr><td>Not in scope:</td></tr></table>"
```
scanny, line 87: This one demonstrates the new `.strip()` behavior of cell contents. Leading and trailing space in an HTML cell is not meaningful and not rendered so another opportunity to compact the already longer-than-.text `.text_as_html` value. This length mismatch is especially important when it comes to chunking.

test_unstructured/documents/test_html.py
```diff
@@ -196,7 +336,6 @@ def test_parse_nothing():
 def test_read_with_existing_pages():
     page = Page(number=0)
     html_document = HTMLDocument.from_pages([page])
-    html_document._read()
```
scanny, line 199: These calls are unnecessary and do not affect the outcome of the test. This method is called in due course simply by accessing `.pages` in the next line. Remove these as misleading the reader to think this call somehow changes state and is somehow required for test effectiveness.

cragwolfe (approved):

PR #2116, opened 2023-11-18
fix: local connector with input path to single file

ryannikolaidis:
When passed an absolute file path for the input document path, the local connector incorrectly writes the output file to the wrong directory. Also, in the single file input path cases we are currently including parent path as part of the destination writing, instead when a single file is specified as input the output file should be located directly in the specified outputs directory. Note: this change meant that we needed to bump the file path of some expected results. This fixes such that the output in this case is written to `output-dir/input-filename.json`. 

## Changes
- Fix for incorrect output path of files partitioned via the local connector when the input path is a file path (rather than directory)
- Updated single-local-file test to validate the flow where we specify an absolute file path (since this was particularly broken)

## Testing
Note: running the updated `local-single-file` test without the changes to the local connector will result in a final output copy of:

```
Copying /Users/ryannikolaidis/Development/unstructured/unstructured/test_unstructured_ingest/workdir/local-single-file/partitioned/a48c2abec07a9a31860429f94e5a6ade.json -> /Users/ryannikolaidis/Development/unstructured/unstructured/test_unstructured_ingest/../example-docs/language-docs/UDHR_first_article_all.txt.json
```

where the output path is the input path and not the expected `output-dir/input-filename.json`

Running with this change we can now expect the file at that directory.


unstructured/ingest/connector/local.py
```diff
@@ -41,10 +41,12 @@ class LocalIngestDoc(BaseIngestDoc):
 
     @property
     def base_filename(self) -> t.Optional[str]:
-        download_path = str(Path(self.connector_config.input_path).resolve())
-        full_path = str(self.filename)
-        base_path = full_path.replace(download_path, "")
-        return base_path
+        download_path = Path(self.connector_config.input_path).resolve()
+        full_path = Path(self.filename).resolve()
+        if download_path.is_file():
```
potter-potter, line 46: This is cool. It actually "attempts" to open the file. It's not just depending on an extension.

potter-potter (approved):
Solution looks good. I ran it both old way and with this pr and it appears to fix it nicely.

PR #2135, opened 2023-11-21
fix(docx): tables in header/footer dropped

### LINKED ISSUE #2126: bug/docx does not parse tables in header
**Describe the bug**

Tables in headers are not parsed in `partition_docx`. The underyling reason is that [docx._DocxPartitioner._iter_section_headers()](https://github.com/Unstructured-IO/unstructured/blob/0.10.30/unstructured/partition/docx.py#L566) only iterates over the paragraphs of the header, as shown in the following code: `text = "\n".join([p.text for p in header.paragraphs])`. 

**To Reproduce**

[test_document.docx](https://github.com/Unstructured-IO/unstructured/files/13412860/test_document.docx)

```python
elements = partition_docx(filename="/path/to/test_document.docx")
for e in elements:
    print(e)
```

Actual output:
```
First line of the header
Third line of the header
Main document
```

**Expected behavior**

Expected output:
```
First line of the header
Table cell1  Table cell2
Third line of the header
Main document
```

**Screenshots**
Screenshot of the test word document
image

**Environment Info**
OS version:  macOS-14.1-arm64-arm-64bit
Python version:  3.11.6
unstructured version:  0.10.30

scanny:
A DOCX header or footer is a so-called "story part" meaning like the document body (which is also a story part) it can contain both paragraphs and tables. The implementation of `Header.text` and `Footer.text` gather only the paragraphs.

Add a new method to extract all content from a header or footer, including table content, suitable for use as the `.text` attribute of that element.

Fixes #2126.


test_unstructured/partition/docx/test_docx.py
```diff
@@ -258,18 +267,6 @@ def test_partition_docx_from_file(mock_document_file_path: str, expected_element
...
-@pytest.mark.parametrize("infer_table_structure", [True, False])
-def test_partition_docx_infer_table_structure(infer_table_structure: bool):
-    elements = partition_docx(
-        example_doc_path("fake_table.docx"), infer_table_structure=infer_table_structure
-    )
-    table_element_has_text_as_html_field = (
-        hasattr(elements[0].metadata, "text_as_html")
-        and elements[0].metadata.text_as_html is not None
-    )
-    assert table_element_has_text_as_html_field == infer_table_structure
```
scanny, line 270: No changes to this test, just moved to group related tests.

test_unstructured/partition/docx/test_docx.py
```diff
@@ -197,20 +197,29 @@ def str_repr(e: Element) -> str:
...
-    assert [e.text for e in elements] == [
-        "0:0:0.0 --> 0:0:1.510\nSome Body\nOK. Yeah.",
-        "0:0:3.270 --> 0:0:4.250\nJames Bond\nUmm.",
-        "saved-by Dennis Forsythe",
-    ]
-    assert [e.category for e in elements] == [
-        ElementType.UNCATEGORIZED_TEXT,
-        ElementType.UNCATEGORIZED_TEXT,
-        ElementType.TABLE,
-    ]
```
scanny, line 213: No changes to this test, just moved to group related tests.

cragwolfe (approved):

Issue #1922, opened 2023-10-27
bug/TypeError: 'PDFObjRef' object is not subscriptable


awalker4:
API users are hitting this error. Seems related to #1921, a PDFObjRef is passed where it isn't expected. The stack at this time looks like:

annotation | <PDFObjRef:845>
annotation_dict | <PDFObjRef:845>
annotation_list | []
annots | [<PDFObjRef:845>]
coordinate_system | <unstructured.documents.coordinates.PixelSpace object at 0x7fc73cd5eb30>
height | 595.276
page_number | 20

```
TypeError: 'PDFObjRef' object is not subscriptable
  File "unstructured/partition/pdf.py", line 157, in partition_pdf
    return partition_pdf_or_image(
  File "unstructured/partition/pdf.py", line 261, in partition_pdf_or_image
    extracted_elements = extractable_elements(
  File "unstructured/partition/pdf.py", line 180, in extractable_elements
    return _partition_pdf_with_pdfminer(
  File "unstructured/utils.py", line 178, in wrapper
    return func(*args, **kwargs)
  File "unstructured/partition/pdf.py", line 496, in _partition_pdf_with_pdfminer
    elements = _process_pdfminer_pages(
  File "unstructured/partition/pdf.py", line 556, in _process_pdfminer_pages
    annotation_list = get_uris(page.annots, height, coordinate_system, i + 1)
  File "unstructured/partition/pdf.py", line 952, in get_uris
    return get_uris_from_annots(annots.resolve(), height, coordinate_system, page_number)
  File "unstructured/partition/pdf.py", line 979, in get_uris_from_annots
    if str(annotation_dict["Subtype"]) != "/'Link'" or "A" not in annotation_dict:
```

Issue #1921, opened 2023-10-27
bug/TypeError: cannot unpack non-iterable PDFObjRef object


awalker4:
**Describe the bug**
API users are hitting this error. The issue is that the value of rect is `<PDFObjRef:606>` which cannot be unpacked. We don't have a reproducer yet, so we need to:

* Reproduce the issue or otherwise understand why it happens
* Account for the argument to `rect_to_bbox` not being a tuple

```
  File "unstructured/partition/pdf.py", line 157, in partition_pdf
    return partition_pdf_or_image(
  File "unstructured/partition/pdf.py", line 261, in partition_pdf_or_image
    extracted_elements = extractable_elements(
  File "unstructured/partition/pdf.py", line 180, in extractable_elements
    return _partition_pdf_with_pdfminer(
  File "unstructured/utils.py", line 178, in wrapper
    return func(*args, **kwargs)
  File "unstructured/partition/pdf.py", line 496, in _partition_pdf_with_pdfminer
    elements = _process_pdfminer_pages(
  File "unstructured/partition/pdf.py", line 556, in _process_pdfminer_pages
    annotation_list = get_uris(page.annots, height, coordinate_system, i + 1)
  File "unstructured/partition/pdf.py", line 952, in get_uris
    return get_uris_from_annots(annots.resolve(), height, coordinate_system, page_number)
  File "unstructured/partition/pdf.py", line 981, in get_uris_from_annots
    x1, y1, x2, y2 = rect_to_bbox(annotation_dict["Rect"], height)
  File "unstructured/partition/pdf.py", line 1041, in rect_to_bbox
    x1, y2, x2, y1 = rect
```

PR #2137, opened 2023-11-21
Chore: Repair invalid PDF structure for PDFminer when PSSyntaxError

yuming-long:
### Summary

Add a procedure to repair PDF when the PDF structure is invalid for `PDFminer` to process.

This PR handles two cases of `PSSyntaxError  Invalid dictionary construct: ...`:
* PDFminer open entire document and create pages generator on `PDFPage.get_pages(fp)`: [sentry log example](https://unstructuredio.sentry.io/issues/4655715023/?alert_rule_id=14681339&alert_type=issue&notification_uuid=d8db4cf4-686f-4504-8a22-74a79a8e966f&project=4505909127086080&referrer=slack)
* PDFminer's interpreter process a single page on `interpreter.process_page(page)`: [sentry log example](https://unstructuredio.sentry.io/issues/4655898781/?referrer=slack&notification_uuid=0d929d48-f490-4db8-8dad-5d431c8460bc&alert_rule_id=14681339&alert_type=issue)

**Additional tech details:**
* Add new dependency `pikepdf` in `requirements/extra-pdf-image.in`, which is used for repairing PDF.
* Add new denpendenct `pypdf` in `requirements/extra-pdf-image.in`,  which is used to find the error page from entire document by reading the PDF file again (can't find a way to split pdf in PDFminer).
* Refactor the `is null` check for `get_uris_from_annots`, since the root cause is that `get_uris` passed a None `annots` to `get_uris_from_annots`, so the Null check should happen in `get_uris`.
* Add more type protection in `get_uris_from_annots` when using any `PDFObjRef.resolve()` as `dict` (it could still be a `PDFObjRef`). This should fix :
   * https://github.com/Unstructured-IO/unstructured/issues/1922 where `annotation_dict` is a `PDFObjRef`
   * https://github.com/Unstructured-IO/unstructured/issues/1921 where `rect` is a `PDFObjRef`

### Test
Added three test files (both are larger than 500 KB) for unittests to test:
* Repair entire doc
* Repair one page
* Reprocess failure after repairing one page (just return the elements before error page in this case). 
   * Also seems like splitting the document into smaller pages could fix this problem, but not sure why. For example, I saw error from reprocess in the whole [cancer.pdf](https://github.com/Unstructured-IO/unstructured/files/13461616/cancer.pdf) doc, but no error when i split the pdf by error page....
   * tested if i can repair the entire doc again in this case, saw other error which means repairing is not helping imo
   * PDFminer can process the whole doc after pikepdf only repaired the entire doc in the first place, but we can't repair by pages in this way


unstructured/partition/pdf.py
```diff
@@ -557,10 +561,40 @@ def _process_pdfminer_pages(
...
+        try:
+            # Detect invalid dictionary construct for one page
+            interpreter.process_page(page)
+            page_layout = device.get_result()
+        except PSSyntaxError:
+            logger.info("Detected invalid dictionary construct for PDFminer")
+            logger.info("Repairing the PDF page...")
+            fp.seek(0)
+            # find the error page from binary data fp
+            pdf_reader = pypdf.PdfReader(fp)
```
awalker4, line 585: Nit: Can we make helper functions for the pdf rewrites? You can reassign `page` (also for `pages` above) with the return value of the helper.

unstructured/partition/pdf.py
```diff
@@ -557,10 +572,39 @@ def _process_pdfminer_pages(
     laparams = LAParams()
     device = PDFPageAggregator(rsrcmgr, laparams=laparams)
     interpreter = PDFPageInterpreter(rsrcmgr, device)
-
-    for i, page in enumerate(PDFPage.get_pages(fp)):  # type: ignore
-        interpreter.process_page(page)
-        page_layout = device.get_result()
+    try:
+        # Detect invalid dictionary construct for entire PDF
+        pages = list(PDFPage.get_pages(fp))
```
badGarnet, line 577: Are we calling `list` conversion here to ensure all pages gets processed so we can detect error? but this also means we are holding more data in memory compared to before (when pages is a generator). 
Maybe a cleaner solution is to create a helper like this suggestions here: https://stackoverflow.com/a/27924011

yuming-long: > Are we calling list conversion here to ensure all pages gets processed so we can detect error?

Yes, converting to list is needed for error detection, otherwise the error will show up in `for page in PDFPage.get_pages(fp)`

unstructured/partition/pdf.py
```diff
@@ -556,56 +556,69 @@ def get_page_data(fp: BinaryIO, page_number: int):
...
+                    try:
+                        interpreter.process_page(page)
+                        page_layout = device.get_result()
+                    except Exception:
+                        logger.warning(
+                            f"PDFMiner failed to process PDF page {i+1} after repairing it."
+                        )
+                        break
+            yield page, page_layout
     except PSSyntaxError:
```
Klaijan, line 595: I'm having trouble following why aren't all these under `except` be captured in the `try` above that we need to process them again here?

yuming-long: the first `try` is for a raw PDF page and the second `try` is for a repaired PDF page, we could still see error after `pikepdf` repaired the page :( and i am trying to see if we have a better way to handle this

so the first `try` catch `PSSyntaxError` -> repair PDF page in first `except` -> reprocess the repaired PDF page in the second `try` -> saw error other than `PSSyntaxError` in second `except`-> return the elements before error page

Klaijan: Gotcha!

yuming-long: > Is it because we 1) loop then get_pages vs 2) get_pages then loop? What would happen if we try to get_pages first in the first try?

I am not sure i follow this, which line you are referring to?

yuming-long: Also we have another option, which is to repair the entire doc regardlessly as:
```
try:
        i = 0
        pages = PDFPage.get_pages(fp)
        for page in pages:
               interpreter.process_page(page)
               page_layout = device.get_result()
except PSSyntaxError:
 ```
This also works since `PSSyntaxError` for `get_pages` and `process_page` will both be caught in this `try... except` block, but this way we can't repair doc by pages, and might be unnecessary if only need to repair one page of the doc, what do you think?

Comment (awalker4) [reply unstructured/partition/pdf.py:597]: I agree that per page will probably be more efficient.

Comment (Klaijan) [reply unstructured/partition/pdf.py:597]: I second the per page too.

So to summarize, the outermost `except` kicks in when line 570 fails
```
pages = PDFPage.get_pages(fp)
```

And the inner `except` is when the generator is successfully created, but fails at per page process.

In the first case, we repair the whole document then process the repaired document page by page. And in the second case, we process the original document pages when possible, and repair then process the repaired pages when needed?

yuming-long: yes

Klaijan: Then if we do not have a more well-rounded way to approach this, I think this looks good to me. I'm gonna wait a little more in case others come up with something if that's alright with you.

yuming-long: for sure! thanks

Klaijan (approved):
Just resolve all the conflicts and this looks good to me!

PR #2144, opened 2023-11-22
ci: fix broken API unit test for using unsupported `fast` strategy for images

yuming-long:
### Summary

This should fix the broken unit test on main CI

* change the strategy in `test_partition_multiple_via_api_valid_request_data_kwargs` from `fast` to `auto`, since the test was using `fast` for images, and we don't support it.

Note: need to catch this 500 error in API and return to user in issue: https://github.com/Unstructured-IO/unstructured-api/issues/314


test_unstructured/partition/test_api.py
```diff
@@ -406,7 +402,6 @@ def get_api_key():
 
 
 @pytest.mark.skipif(skip_outside_ci, reason="Skipping test run outside of CI")
-@pytest.mark.skipif(skip_not_on_main, reason="Skipping test run outside of main branch")
```
cragwolfe, line 409: we still want to skip API tests for CI in PR's, so please revert these skips

cragwolfe (changes requested):
let's continue to skip API tests if they are not on commits running in CI on the main branch. otherwise the "fast" -> "auto" change LGTM!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+* **Fix broken API unit test** where the test was using `fast` stragy for images, which is not supported.
```
cragwolfe, line 9: can remove this since it does not impact the library. we can force merge if there is a changelog complaint.

cragwolfe (approved):

PR #2146, opened 2023-11-22
bug: fsspec output filepath including base directory

rbiseck3:
### Description
When passing in a remote path for fsspec-based source connectors, the base directory was always being included in the output path itself. This was updated to exclude the base directory any only include any child directories relative to the base one.

### DIFF
--- a/unstructured/ingest/connector/fsspec.py
+++ b/unstructured/ingest/connector/fsspec.py
@@ -69,10 +69,10 @@ class FsspecIngestDoc(IngestDocCleanupMixin, BaseIngestDoc):
 
     @property
     def _output_filename(self):
-        return (
-            Path(self.processor_config.output_dir)
-            / f"{self.remote_file_path.replace(f'{self.connector_config.dir_path}/', '')}.json"
+        filename = (
+            f"{self.remote_file_path.replace(self.connector_config.path_without_protocol, '')}.json"
         )
+        return Path(self.processor_config.output_dir) / filename
 
     def _create_full_tmp_dir_path(self):
         """Includes "directories" in the object path"""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.1-dev1
+## 0.11.1-dev2
 
 ### Enhancements
 
@@ -8,6 +8,7 @@
 
 * **Do not extract text of `<style>` tags in HTML.** `<style>` tags containing CSS in invalid positions previously contributed to element text. Do not consider text node of a `<style>` element as textual content.
 * **Fix DOCX merged table cell repeats cell text.** Only include text for a merged cell, not for each underlying cell spanned by the merge.
+* **Fix output filepath for fsspec-based source connectors** Previously the base directory was being included in the output filepath unnecessarily.
 
 ## 0.11.0
 
--- a/unstructured/__version__.py
+++ b/unstructured/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.11.1-dev1"  # pragma: no cover
+__version__ = "0.11.1-dev2"  # pragma: no cover


ryannikolaidis:
awesome! should require bump to test fixtures, right? since this updates the output path.

ryannikolaidis:
changes look good though!

ryannikolaidis (approved):
lgtm!

unstructured/ingest/connector/fsspec.py
```diff
@@ -69,10 +69,10 @@ def _tmp_download_file(self):
 
     @property
     def _output_filename(self):
-        return (
-            Path(self.processor_config.output_dir)
-            / f"{self.remote_file_path.replace(f'{self.connector_config.dir_path}/', '')}.json"
+        filename = (
+            f"{self.remote_file_path.replace(self.connector_config.path_without_protocol, '')}.json"
```
potter-potter, line 73: Are you sure this is going to work for nested folders? Or maybe we not nesting anymore? (It's been a long time since I've thought about the where the partitioned files go.) 

As I test S3 with this, there is no `path_without_protocol` even though there is a nested folder path. So if 2 files have the same name or are in different folders maybe this will be an issue.

rbiseck3: When I tested it locally it maintained the nested path. Can you post exactly what you ran where the sub folders weren’t in the output path?

potter-potter: Ok. Let me look into it a little deeper. I was just comparing the result of the new _output_filename to the old _output_filename for S3. I'll have to run the full test to check.

Just curious, how far down the pipeline do nested paths go?

potter-potter: I got this with Box:
`INFO     Copying /Users/davidpotter/Documents/Unstructured/sessions/unstructured/test_unstructured_ingest/workdir/box/partitioned/f354b481320e7405c87cbcd32bde8fce.json -> /handbook-1p.docx.json`
It tried to write the file to my root directory.

potter-potter: You've probably seen how the Path module barfs when it gets "/filename.txt" after the ` / ` I wish it was smarter.

ryannikolaidis: let's add a test case for this ^^

ryannikolaidis (changes requested):
add a test case with a nested case if we don't have one?

unstructured/ingest/connector/fsspec.py
ryannikolaidis: but this is working in CI and the box test has nested folders?

potter-potter: @ryannikolaidis Where did it pass CI? Last CI I see is with `update unit tests` `a43e58c` and it did't get to run box.sh (which is probably in 3.10. (maybe it ran CI since then and I don't see it. I would like to learn where to find it if thats the case)

rbiseck3:
I've tested this manually with the s3 connector using:
* A specific file as the input path
* An input path with and without the leading slash
* Recursive input with subfolders

ryannikolaidis (approved):

potter-potter:
I just ran with box. The nesting doesn't seem to be working right. I'll have to run against the old version again, but nested-2 should be in nested-1 but now nested-2 is next to nested-1.

potter-potter:
Well. Looks like the directory and files in Box got shifted around sometime around 9/10. So it's not nested anyway and there is a partitioned file in the Box folder. So I can't make a conclusion with Box. I'll try to fix the files in Box.com if I can later.

I will try dropbox to see if the nesting works. :)

potter-potter:
Dropbox is the same. I can tell from expected-structured-output. No double nesting. Same with GCS. No double nesting.

potter-potter (approved):
Works as expected.

Ok. I fixed the files in Box.com and it all works good. You will need to `export OVERWRITE_FIXTURES=true` for Box.

ryannikolaidis:
> Works as expected.
> 
> Ok. I fixed the files in Box.com and it all works good. You will need to `export OVERWRITE_FIXTURES=true` for Box.

hmm, what do you mean by that?

> Dropbox is the same. I can tell from expected-structured-output. No double nesting. Same with GCS. No double nesting.

were we every doubly nesting these? I don't remember us explicitly doing that? but also if we were, I don't remember us undoing that at any point?

ryannikolaidis:
> Works as expected.
> 
> Ok. I fixed the files in Box.com and it all works good. You will need to `export OVERWRITE_FIXTURES=true` for Box.

if you make changes to any source test files, please create a quick PR to explicitly fix the expected values. in this case, I think we're still ignoring Box tests so slightly less urgent, but otherwise this could be breaking tests for other active PRs.
I'll go ahead and kick one off now.

also, what changes did you make to the source? we should probably discuss per the "double nested" conversation, since these file storage services are all treated similarly may be worth making sure we're testing consistently.

potter-potter:
> > Works as expected.
> > Ok. I fixed the files in Box.com and it all works good. You will need to `export OVERWRITE_FIXTURES=true` for Box.
> 
> @potter-potter if you make changes to any source test files, please create a quick PR to explicitly fix the expected values. in this case, I think we're still ignoring Box tests so slightly less urgent, but otherwise this could be breaking tests for other active PRs. I'll go ahead and kick one off now.
> 
> also, what changes did you make to the source? we should probably discuss per the "double nested" conversation, since these file storage services are all treated similarly may be worth making sure we're testing consistently.

@ryannikolaidis Shoot! You’re right! Will be more cautious next time.

potter-potter:
@rbiseck3 I had them double nested when I built the dropbox and the box connectors. I wanted to have at least 2 levels ov nesting. (there's not much value in having two folders next to each other called nested1 and nested2) I am very curious when/why they got un-nested. 
Also there was a partitioned .json file in box that was probably put in when someone was working on the Box destination connector. In our test we were partitioning a partitioned file...
That being said I should not have made those changes in Box.com without consulting with the team. :)

PR #2163, opened 2023-11-27
feat: Add sftp source connector

potter-potter:
Adds source connector for SFTP which uses fsspec and paramiko via fsspec. Paramiko is the standard sftp package for python used in pysftp etc...

```
--username foo \
--password bar \
--remote-url sftp://localhost:47474/upload/
```

Will only download a specifically requested file if it has an extension. (i.e. `--remote-url sftp://localhost:47474/upload/bob.zip`) It will treat any other remote_url as a folder path. This is intentional.


unstructured/ingest/connector/sftp.py
```diff
@@ -0,0 +1,56 @@
...
+    port: str = ""
+
+    def __post_init__(self):
+        super().__post_init__()
+
+        _, ext = os.path.splitext(self.remote_url)
+        if ext:
+            # We only want the file_path if it has an extension
+            self.file_path = Path(self.remote_url).name
+            self.dir_path = Path(urlparse(self.remote_url).path).parent.as_posix().lstrip("/")
```
rbiseck3, line 28: small nit, since we're using `urlparse(self.remote_url)` so often, can we create a variable and use that everywhere instead? 
```python
parsed_url = urlparse(self.remote_url)
```

unstructured/ingest/cli/cmds/sftp.py
```diff
@@ -0,0 +1,40 @@
...
+CMD_NAME = "sftp"
+
+
+@dataclass
+class SftpCliConfig(CliConfig):
+    username: str
+    password: str
+
+    @staticmethod
+    def get_cli_options() -> t.List[click.Option]:
```
rbiseck3, line 20: Would we want to support a user passing in the host/port explicitly as a cli argument?

rbiseck3:
reminder to change PR title to match convention

unstructured/ingest/connector/sftp.py
potter-potter: Definitely.

unstructured/ingest/cli/cmds/sftp.py
potter-potter: @rbiseck3 Would they have the choice of either as cli or in the --remote-url? I know in the url is pretty standard for sftp

potter-potter: seems like once you use `urlparse` with sftp protocol, it is really expecting a hostname right after the `//` (port is optional) so maybe its natural to leave it in the --remote_url

unstructured/ingest/connector/sftp.py
```diff
@@ -0,0 +1,75 @@
...
+class SftpIngestDoc(FsspecIngestDoc):
+    connector_config: SimpleSftpConfig
+    registry_name: str = "sftp"
+
+    @SourceConnectionError.wrap
+    @requires_dependencies(["paramiko", "fsspec"], extras="sftp")
+    def get_file(self):
+        super().get_file()
+
+
```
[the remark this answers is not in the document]
rbiseck3: Actually something I've been trying to move towards is avoid setting that annotation at the class level. It would be better to add this: 
```python
@requires_dependencies(["paramiko", "fsspec"], extras="sftp")
def initialize(self):
    super().initialize()
```

PR #2166, opened 2023-11-28
feat: add weighted metrics

qued:
Added versions of accuracy and missing % weighted by character count and word count respectively to the aggregate metrics. The reason for doing this is to avoid the practice of averaging averages... the accuracy metric is essentially 1 minus the average Levenshtein distance per character (clipped to lie in the interval [0, 1]), while the missing % is the average missing words per word.


MthwRobinson:
@qued - Anything in here you'd still like to work on?

PR #2173, opened 2023-11-28
chore: improve logging sensitive fields in ingest CLI

rbiseck3:
### Description
Create a more robust way to hide sensitive data when logging fields:
* If any of the following strings exist in the key name, omit the value: `["key", "cred", "token"]`
* If the key equals any of the following, omit: `["account_name", "client_id"]`
* If the value itself is a dictionary, run it through the method as well to omit any nested sensitive fields
* If the value is a string that can be converted into json, to the same as above and return the value as a string again

### DIFF
--- a/unstructured/ingest/cli/common.py
+++ b/unstructured/ingest/cli/common.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import Optional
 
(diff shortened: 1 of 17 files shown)


potter-potter:
Thanks for working on this. I was working on a similar but this is good.
It also needs to address these log statements:
https://github.com/Unstructured-IO/unstructured/blob/2bb463d00682cf96720476b58e8fda3658897249/unstructured/ingest/pipeline/interfaces.py#L113

https://github.com/Unstructured-IO/unstructured/blob/2bb463d00682cf96720476b58e8fda3658897249/unstructured/ingest/connector/fsspec.py#L91C33-L91C33

potter-potter:
Do you think we should mask None values? Before we didn't. Is it a better developer experience if it doesn't mask None?

potter-potter:
Also, I think masking 'cred' may be too general. It will mask things like 'credit'. Unless this is just a very specific function for this particular log line. (In that case it is fine)

I was working on a more universal function that we could use in those other spots as well. But I guess it still needs to be pretty specific to what is being logged.

rbiseck3:
@potter-potter is there any option that has `credit`? I'd rather be overly conservative on hiding sensitive information and accommodating fields that get set in the future that this is covering when it shouldn't.

potter-potter:
I agree on being overly conservative. And since this is only masking options, I think you're good.

potter-potter:
This is awesome. I'm checking if the log lines here need it. There is an api_key
https://github.com/Unstructured-IO/unstructured/blob/2bb463d00682cf96720476b58e8fda3658897249/unstructured/ingest/pipeline/interfaces.py#L171
And maybe the openai key for embedding?

PR #2174, opened 2023-11-28
feat: add logging to ingest CLI for tests being skipped at the end

rbiseck3:
### Description
Often times there are tests being skipped either due to missing env vars or explicitly defined in the base script but these get lost in the logs. This PR updates the scripts to leverage a custom error code if being skipped due to missing env vars and this custom error code is being caught by the base script and logs all files being skipped to a file. At the end of the script, this file gets logged in the CI output.

### DIFF
--- a/test_unstructured_ingest/test-ingest-src.sh
+++ b/test_unstructured_ingest/test-ingest-src.sh
@@ -1,51 +1,57 @@
 #!/usr/bin/env bash
 
-set -eu -o pipefail
+set -u -o pipefail
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+SKIPPED_FILES_LOG=$SCRIPT_DIR/skipped-files.txt
+# If the file already exists, reset it
+if [ -f "$SKIPPED_FILES_LOG" ]; then
+  rm "$SKIPPED_FILES_LOG"
+  touch "$SKIPPED_FILES_LOG"
+fi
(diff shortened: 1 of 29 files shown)


test_unstructured_ingest/test-ingest-src.sh
```diff
@@ -1,50 +1,56 @@
 #!/usr/bin/env bash
 
-set -eu -o pipefail
+set -u -o pipefail
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+SKIPPED_FILES_LOG=$SCRIPT_DIR/skipped-files.txt
```
potter-potter, line 6: I assume test-ingest-src and test-ingest-dest do not run parallel in CI, or the skipped-files.txt might be overwritten or deleted while the other is running? Or does the CI actually make completely separate environments for the different tests?

rbiseck3:
@potter-potter, this was because I forgot to leverage the `-a` flag when using the `tee` command so it wasn't appending if the file already existed. Should be fixed now.

test_unstructured_ingest/test-ingest-src.sh
rbiseck3: They are run in parallel but on different machines so this shouldn't cause any conflict.

rbiseck3: And that file isn't persisted between tests so it should only live long enough for the script to print it in the logs.

potter-potter (approved):
This works well for me.

PR #2175, opened 2023-11-28
feat: leverage logger to hide sensitive data in ingest logs

rbiseck3:
### Description
Modify the logger being used by ingest to leverage a new class implemented inheriting `logging.Formatter` which adds in some middleware to update the message being logged to omit any sensitive content. It does this by dynamically pulled out any valid json from the string being logged and runs that through a `hide_sensitive_fields` method which updated any values that are considered sensitive. Replaces the original json strings with the `json.dumps` version of the new dictionary.


potter-potter:
Very nice. I noticed it left null for the Unstructured api_key when it was missing and then masked it when it was there. Thats great.

potter-potter:
This is working well and masks all the things that I was finding.

unstructured/ingest/logger.py
```diff
@@ -1,12 +1,81 @@
+import ast
+import json
 import logging
 
 logger = logging.getLogger("unstructured.ingest")
 
 
+def hide_sensitive_fields(data: dict) -> dict:
```
potter-potter, line 8: Can you make a simple unit test for this function and extract_json. Would help others to see exactly whats going on a bit quicker.

potter-potter (approved):
Works great.

PR #2178, opened 2023-11-28
feat: add visualize param to command and add test

Klaijan:
- Add `visualize` parameter to the click command -- now callable using `--visualize` flag to show the progress bar.
- Refactor the name.


test_unstructured/metrics/test_evaluate.py
```diff
@@ -17,6 +18,35 @@
...
+def test_text_extraction():
+    output_dir = os.path.join(TESTING_FILE_DIR, UNSTRUCTURED_OUTPUT_DIRNAME)
+    source_dir = os.path.join(TESTING_FILE_DIR, GOLD_CCT_DIRNAME)
+    export_dir = os.path.join(TESTING_FILE_DIR, "test_evaluate_results_cct")
+    measure_text_extraction_accuracy(
+        output_dir=output_dir, source_dir=source_dir, export_dir=export_dir
+    )
+    df = pd.read_csv(os.path.join(export_dir, "all-docs-cct.tsv"), sep="\t")
+    assert len(df) == 3
+    assert len(df.columns) == 5
```
shreyanid, line 34: will this columns value always be 5? I know we have a lot of changes slated for this output file. What exactly is this test testing, and can we update the test name to reflect that?

unstructured/metrics/doctype_aggregation.py
```diff
@@ -66,7 +66,7 @@ def measure_holistic_eval_cct(
     weights: Tuple[int, int, int],
 ) -> None:
     export_dir = "result_doctype_aggregate"
-    measure_text_edit_distance(
+    measure_text_extraction_accuracy(
```
shreyanid, line 69: when will this function be passed `visualize` as input (here or in the test, for example)? I see that the default value is false, so is it getting used yet?

test_unstructured/metrics/test_evaluate.py
Klaijan: As long as we don't make any changes to `measure_text_extraction_accuracy` function, the column will always be 5 ("filename", "doctype", "connector", "cct-accuracy", "cct-%missing"). This is just testing the `measure_text_extraction_accuracy` in general if it behave and save as expected. I can add another assertion to check if the file existed before all the file opening if that makes more sense.

unstructured/metrics/doctype_aggregation.py
Klaijan: Referring to this click option
```
@click.option(
    "--visualize",
    is_flag=True,
    show_default=True,
    default=False,
    help="Add the flag to show progress bar.",
)
```
The `--visualize` param will be passed in as a flag. Meaning you can use by adding `--visualize` to the command line that calls the script. It is not being used yet. The main purpose of this additional param/flag is to track the progress of the `evaluate.py` when the list of files to evaluate against are large so we don't get stuck staring at blank terminal for like an hour. I find that quite frustrating when I experienced. So to answer your question, no it is not getting used yet anywhere in this repo or script since the evaluating list is relatively small that it won't get stuck at processing for long.

unstructured/metrics/doctype_aggregation.py
shreyanid, line 1: why was doctype_aggregation deleted? where was this functionality moved to?

Klaijan: That file was accidentally added to the first refactor pr. The grouping functionality is now integrated with the `measure_text_extraction_accuracy` (see line 42 in `evaluate.py`)

Klaijan: For your reference: https://github.com/Unstructured-IO/unstructured/pull/2108

shreyanid (approved):

PR #2179, opened 2023-11-28
feat: introduce dataclass with support for sensitive fields

rbiseck3:
### Description
Given the use of dataclasses everywhere, it would help if we can designate explicitly which fields should be ommited when being logged (vs being serialized). This PR extends the already used dataclass and json mixin libraries to support a sensitive field, and redact the text when converting to dict/json. 

This only introduces the new dataclass without incirporating it into all the connectors. Simply opening up this PR for discussion at the moment. 

Example:
```python

from unstructured.ingest.enhanced_dataclass.dataclasses import enhanced_field
from unstructured.ingest.enhanced_dataclass.json_mixin import EnhancedDataClassJsonMixin


@dataclass
class AuthData(EnhancedDataClassJsonMixin):
    username: str
    password: str = enhanced_field(sensitive=True)


a = AuthData(username="myname", password="pw")
print(a.to_json(indent=2, preserve_sensitive=False))
print(a.to_json(preserve_sensitive=False, redacted_text="************"))
print(a.to_dict())
```

```shell
{
  "username": "myname",
  "password": "***REDACTED***"
}
{"username": "myname", "password": "************"}
{'username': 'myname', 'password': 'pw'}
```


potter-potter:
Super cool. I wonder if there is a better variable name than `preserve_sensitive`. I would want something that is brainless to understand. Maybe like `keep_hidden` but better (actually the opposite of that :) ). 

I guess we probably won't be printing it often anyway...

And then the idea is that the logger will pick up on this automatically, and always hide it? Or hide it by default unless specifically exposed?

And we could potentially use this to decide what data to encrypt/decrypt in the future?

rbiseck3:
Once this is merged in, will begin updating connector configs to leverage this.

potter-potter:
Tried this out on `output_dir`
https://github.com/Unstructured-IO/unstructured/blob/341f0f428c6760d9a1063557fce9ecfa676fc76d/unstructured/ingest/interfaces.py#L98
And got:
```
File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/dataclasses.py", line 504, in _init_fn
    raise TypeError(f'non-default argument {f.name!r} '
TypeError: non-default argument 'output_dir' follows default argument
```

rbiseck3:
@potter-potter, regarding your comment, you have to set the default still:
```python
output_dir: str = enhanced_field(default="structured-output", sensitive=True)
``` 

Dataclass expects a particular order in the fields that are defined, with required ones with a default being before optional ones with a default set.

potter-potter:
Running python 3.9
f"Config: {self.partition_config.to_json(redact_sensitive=True, redacted_text="***** BOB ******")}, "
Strings nested within an f-string cannot use the same quote character as the f-string prior to Python 3.12

Guess this is just really a Python issue since we are in an f-string. And nothing to do with this PR. 

So far it is working well.

test_unstructured_ingest/unit/test_enhanced_dataclass.py
```diff
@@ -0,0 +1,36 @@
...
+from dataclasses import Field, dataclass, fields
+
+from unstructured.ingest.enhanced_dataclass import EnhancedDataClassJsonMixin, enhanced_field
+from unstructured.ingest.enhanced_dataclass.dataclasses import EnhancedField
+
+
+@dataclass
+class AuthData(EnhancedDataClassJsonMixin):
+    username: str
+    password: str = enhanced_field(sensitive=True)
```
potter-potter, line 11: Can you add number, list and dict as well, just for peace of mind?

rbiseck3: What test case are you trying to validate?

potter-potter: Sorry I deleted the request. All good.

PR #2180, opened 2023-11-29
Add coco staging brick to unstructured base

pravin-unstructured:


unstructured/staging/base.py
```diff
@@ -337,3 +338,92 @@ def filter_element_types(
         return filtered_elements
 
     return elements
+
+
+def convert_to_coco(
+    elements: List[Element],
+    dataset_description: str = f"Unstructured COCO Dataset {datetime.now().strftime('%Y-%m-%d')}",
```
qued, line 345: If we set the default dynamically in the signature like this it will be set once at loadtime, but I think the intent is to set it whenever the function is called. If you want to timestamp it to when the function is called, make the type `Optional[str]`, set the default to `None`, and then in the body of the function, when `dataset_description is None` set it to the timestamped value.

unstructured/staging/base.py
```diff
@@ -337,3 +338,92 @@ def filter_element_types(
...
+    dataset_version: str = "1.0",
+    contributors: List = ["Unstructured Developers"],
```
qued, line 347: It might be safer to make this a `Tuple[str]` instead. Mutable default values can result in some weird behavior, see [here](https://unstructuredai.slack.com/archives/C0416NK4BLM/p1687997373994349). Since the list doesn't get changed in the function the weird behavior won't occur here, but making it a `Tuple` doesn't leave a landmine.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Features
 
+* **Staging Brick for Coco Format** Staging brick which converts a list of Elements into Coco Format.
```
qued, line 9: Also we should bump the version to a release version.

qued (changes requested):

unstructured/staging/base.py
```diff
@@ -337,3 +338,96 @@ def filter_element_types(
         return filtered_elements
 
     return elements
+
+
+def convert_to_coco(
+    elements: List[Element],
+    dataset_description: Optional[str] = None,
+    dataset_version: str = "1.0",
+    contributors: Tuple[str] = ("Unstructured Developers"),
```
qued, line 347: Add a comma in the parentheses to make it a tuple, otherwise Python thinks these are just order-of-operations parentheses, not tuple-parentheses.
```suggestion
    contributors: Tuple[str] = ("Unstructured Developers",),
```

PR #2187, opened 2023-11-29
feat: Add basic encryption support for cached content in ingest CLI

rbiseck3:
### Description
Given the recent conversations around security and logging sensitive data, realized that this same data is current;y being saved in plain text when being cached during the ingest pipeline. This may be fine for some use cases, but we should support basic encryption if a user doesn't want this data being saved in plaintext if it's potentially on a shared resource. 

Add two abstractions to support encryption:
* `Encryption` : basic interface with an `encrypt()` and `decrypt()` methods
  * `Fernet`: To leverage the built in crypto library, the Fernet implementation was added for basic encryption of cached content if a key file is passed in.
* `File Handler`: dataclass interface with an encryption field. Uses this at all times to encrypt content being written and decrypt it when being read. 
  * `Local File Handler`: To refactor for this approach but maintain what was currently being done, a `LocalFileHandler` implementation was created to read/write from the local file system
  * This can also be expanded using the Fsspec library to write/read to any other file-based system


unstructured/ingest/pipeline/encrypt/fernet.py
```diff
@@ -0,0 +1,20 @@
...
+
+from cryptography.fernet import Fernet
+
+from unstructured.ingest.pipeline.encrypt.interfaces import Encryption
+
+
+@dataclass
+class FernetEncryption(Encryption):
+    key: str
+    encoding: str = "UTF-8"
```
potter-potter, line 11: Is this supposed to be used in the `decode()` below?
Also can you add a comment on why we encode then decode. I'm assuming self.fernet.encrypt encodes into something that is not UTF-8? Like bytes or something?

potter-potter:
This is working. Pretty cool to see encrypted files in the `workdir`. But shouldn't it be decrypting in the Copy node into `structured-output`. I'm getting an encrypted .json in the `structured-output/delta-table-dest` folder.

I realize this is a work in progress and you haven't built it for use yet.

MthwRobinson:
@rbiseck3 - Closing this since last commits were in November, but feel free to reopen if it's still needed

PR #2190, opened 2023-11-29
chore: refactor ingest unit test to run in it's own CI step

rbiseck3:
### Description
Currently the ingest unit tests are running in both the source and destination ingest steps. This PR moved it out into it's own step that can run on a more basic runner since it doesn't require multiple cores/cpus for parallelization.

Further optimizations:
* Added the changelog step as a dependency for the ingest tests to avoid running them (fail fast) of the pipeline has already failed due to the changelog not being updated.
* The cache was never actually being saved if it needed to be recreated, a save job was added at the end of each custom action


.github/workflows/ci.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     env:
       NLTK_DATA: ${{ github.workspace }}/nltk_data
     runs-on: ubuntu-latest
-    needs: setup
+    needs: [setup, changelog]
```
ryannikolaidis, line 51: thinking just don't bother linting if we fail changelog?

ryannikolaidis (approved):
woops, good catch on the ci caching. changes lgtm.

PR #2194, opened 2023-11-30
feat: improve sensitive data handling by fsspec connectors

rbiseck3:
### Description
Building off of PR https://github.com/Unstructured-IO/unstructured/pull/2179, updating fsspec based connectors to use better authentication field handling. This PR adds in the following changes:

* Update the base classes to inherit from the enhanced json mixin
* Add in a new access config dataclass that should be used as a nest dataclass in the connector configs
* Update the code extracting configs out of the cli options dictionary to support the nested access config if it exists on the parent config
* Update all fsspec connectors with explicit access configs given what each one's SDKs support
* Update the json mixin and enhanced field to support a name override when serializing/deserializing from json/dicts. This allows a different name to be used for the CLI option than what the name of the field is on the dataclass. 
* Update all the writes to use class-based approach and share the same structure of the runner classes
* Above update allowed for better code to be used in the base source and destination CLI commands
* Add in utility code around paring a flat dictionary (coming from the click based options) into dataclass-based configs with potentially nested dataclasses. 

**Slightly unrelated changes:**
* session handle removed from pinecone connector as this was breaking the serialization of the write config and didn't have any benefit as a connection was never being shared, the index used simply makes a new http call each time it's invoked. 
* Dedicated write configs were created for all destination connectors to better support serialization
* Refactor of Elasticsearch connector included, with update to ingest test to use auth

**TODOs**
* Left a `#TODO` in the code but the way session handler is implemented right now, it breaks serialization since it adds a generic variable based on the library being used for a connector (i.e. `googleapiclient.discovery.Resource`) which is not serializable. This will need to be updated to omit that from serialization but still support the current workflow.

### DIFF
--- a/unstructured/ingest/cli/cmds/azure.py
+++ b/unstructured/ingest/cli/cmds/azure.py
@@ -26,31 +26,32 @@ class AzureCliConfig(AzureAccessConfig, CliConfig):
                 default=None,
                 help="The storage account key. This is used for shared key "
                 "authentication. If any of account key, sas token or "
-                "client_id is specified, anonymous access will be used.",
+                "client_id are not specified, anonymous access will be used.",
             ),
             click.Option(
                 ["--account-name"],
                 default=None,
                 help="The storage account name. This is used to authenticate "
-                "requests signed with an account key and to construct the "
-                "storage endpoint. It is required unless a connection string "
-                "is given, or if a custom domain is used with anonymous authentication.",
+                "requests signed with an account key and to construct "
+                "the storage endpoint. It is required unless a connection "
+                "string is given, or if a custom domain is used with "
+                "anonymous authentication.",
             ),
             click.Option(
                 ["--connection-string"],
                 default=None,
-                help="If specified, this will override all other parameters besides "
-                "request session. See http://azure.microsoft.com/en-us/"
-                "documentation/articles/storage-configure-connection-string/ "
+                help="If specified, this will override all other parameters. See "
+                "http://azure.microsoft.com/en-us/documentation/articles/storage-configure-connection-string/ "  # noqa: E501
                 "for the connection string format.",
             ),
             click.Option(
                 ["--sas_token"],
                 default=None,
-                help="A shared access signature token to use to authenticate requests "
-                "instead of the account key. If account key and sas token are both "
-                "specified, account key will be used to sign. If any of account key, "
-                "sas token or client_id are specified, anonymous access will be used.",
+                help="A shared access signature token to use to authenticate "
+                "requests instead of the account key. If account key and "
+                "sas token are both specified, account key will be used "
+                "to sign. If any of account key, sas token or client_id "
+                "are not specified, anonymous access will be used.",
             ),
         ]
         return options


unstructured/ingest/interfaces.py
```diff
@@ -248,7 +256,7 @@ class SourceMetadata(DataClassJsonMixin, ABC):
     permissions_data: t.Optional[t.List[t.Dict[str, t.Any]]] = None
 
 
-class IngestDocJsonMixin(DataClassJsonMixin):
+class IngestDocJsonMixin(EnhancedDataClassJsonMixin):
```
potter-potter, line 259: Should the isinstance on line 289 be EnhancedDataClassJsonMixin?

unstructured/ingest/cli/utils.py
```diff
@@ -23,35 +25,110 @@ def conform_click_options(options: dict):
...
+                f_type = dataclass_union_values[0]
+            origin = t.get_origin(f_type)
+            if origin:
+                f_type = origin
+            if is_subclass(f_type, BaseConfig):
+                dd[field.name] = conform_dict(inner_d=dd, inner_config=f_type)
+        return dd
+
+    adjusted_dict = conform_dict(inner_d=flat_data, inner_config=config)
+    print(adjusted_dict)
```
potter-potter, line 87: left over print statement

CHANGELOG.md
```diff
@@ -1,14 +1,14 @@
-## 0.11.4-dev2
+## 0.11.4-dev3
 
 ### Enhancements
 
 * **Refactor pdfminer code.** The pdfminer code is moved from `unstructured-inference` to `unstructured`.
+* **Improve handling of auth data for fsspec connectors** Leverage an extension od the dataclass paradigm to support a `sensitive` annotation for fields related to auth (i.e. passwords, tokens). Refactor all fsspec connectors to use explicit access configs rather than a generic dictionary.
```
potter-potter, line 6: Picky of me but can you fix the od to of. I stopped for a second to decypher od and then realized its of. Since it will probably be read by many.

unstructured/ingest/enhanced_dataclass/core.py
```diff
@@ -0,0 +1,197 @@
...
+    _isinstance_safe,
+    _support_extended_types,
+    _user_overrides_or_exts,
+    get_type_hints,
+    is_dataclass,
+)
+
+
+def _recursive_repr(user_function):
+    # Copied from dataclasses as this method isn't exposed for importing
```
ahmetmeleq, line 28:
```suggestion
    """Decorator to make a repr function return "..." for a recursive call. Useful in cases where calling __repr__ would cause recursion errors. """
    # Copied from dataclasses as this method isn't exposed for importing
```

test_unstructured_ingest/unit/test_utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_extract_config_union():
+    @dataclass
+    class C(BaseConfig):
+        c: bool
+        b: t.Optional[t.Union[B, int]] = None
+
+    c = extract_config(flat_data=flat_data, config=C)
+    expected_result = {"b": 4, "c": True}
```
ahmetmeleq, line 52: Why is extracting `b` as an `int` favored in this test? 

When I checked the line below, I thought b as a dataclass (B) would be favored over b as an int for Unions; but then why the expected result is `{"b": 4, "c": True}` here?
https://github.com/Unstructured-IO/unstructured/blob/de5224f0c24c941c487fdbd28be5901a8123d0fa/unstructured/ingest/cli/utils.py#L78

rbiseck3: it only favors b as a dataclass if it doesn't already exist in the dictionary. It always defaults to the base class first if possible because that introduces less complex logic such as paring a new dataclass.

ahmetmeleq: Sounds good 👍

unstructured/ingest/cli/cmds/elasticsearch.py
```diff
@@ -32,6 +31,45 @@ def get_cli_options() -> t.List[click.Option]:
                 help="If provided, will limit the fields returned by Elasticsearch "
                 "to this comma-delimited list",
             ),
+            click.Option(
```
potter-potter, line 34: Would you know why none of these new ones show up with:
`unstructured-ingest elasticsearch --help`

Does a process have to run before this info is accessable? --batch-size is there.

rbiseck3: I can see them:
```shell
PYTHONPATH=. python ./unstructured/ingest/main.py elasticsearch --help
Usage: main.py elasticsearch [OPTIONS] COMMAND [ARGS]...ticsearch --help             

Options:
  --index-name TEXT               Name for the Elasticsearch index to pull
                                  data from  [required]
  --hosts DELIMITED-STRING        List of the Elasticsearch hosts to connect
                                  to, e.g. "http://localhost:9200"
  --fields DELIMITED-STRING       If provided, will limit the fields returned
                                  by Elasticsearch to this comma-delimited
                                  list
  --username TEXT                 username when using basic auth
  --password TEXT                 password when using basic auth or connecting
                                  to a cloud instance
...
```

potter-potter: On same note, but not related to this pr. How do we get Weaviate in the --help destinations list?

unstructured/ingest/connector/gcs.py
```diff
@@ -1,21 +1,37 @@
...
 from unstructured.ingest.error import SourceConnectionError
+from unstructured.ingest.interfaces import AccessConfig
 from unstructured.utils import requires_dependencies
 
 
 @dataclass
-class SimpleGcsConfig(SimpleFsspecConfig):
+class GcsAccessConfig(AccessConfig):
+    token: t.Optional[str] = enhanced_field(
+        default=None, sensitive=True, overload_name="service_account_key"
```
potter-potter, line 21: overload_name is nice. so user sees `--service-account-key` but we pass it into fsspec as token?

unstructured/ingest/cli/cmds/elasticsearch.py
rbiseck3: Are you not seeing it listed? When I run the same command as above I see this list: 
```shell
Destinations:
  azure
  azure-cognitive-search
  box
  delta-table
  dropbox
  fsspec
  gcs
  mongodb
  pinecone
  s3
  weaviate
```

potter-potter: ok. interesting. is there something I have to run? `make doc` or something? I just have your branch loaded in git.

unstructured/ingest/cli/cmds/azure.py
```diff
@@ -7,43 +7,57 @@
...
     def get_cli_options() -> t.List[click.Option]:
         options = [
             click.Option(
                 ["--account-key"],
                 default=None,
-                help="Azure Blob Storage or DataLake account key (not required if "
-                "`azure_account_name` is public).",
+                help="The storage account key. This is used for shared key "
+                "authentication. If any of account key, sas token or "
+                "client_id is specified, anonymous access will be used.",
```
ahmetmeleq, line 29: Is this a typo, it says anonymous access will be used when credentials are provided

unstructured/ingest/cli/cmds/azure.py
```diff
@@ -7,43 +7,57 @@
...
+                "documentation/articles/storage-configure-connection-string/ "
+                "for the connection string format.",
+            ),
+            click.Option(
+                ["--sas_token"],
+                default=None,
+                help="A shared access signature token to use to authenticate requests "
+                "instead of the account key. If account key and sas token are both "
+                "specified, account key will be used to sign. If any of account key, "
+                "sas token or client_id are specified, anonymous access will be used.",
```
ahmetmeleq, line 53: Is this a typo, it says anonymous access will be used when credentials are provided

unstructured/ingest/runner/gcs.py
```diff
@@ -1,15 +1,21 @@
...
 from unstructured.ingest.runner.base_runner import FsspecBaseRunner
 from unstructured.ingest.runner.utils import update_download_dir_remote_url
 
+if t.TYPE_CHECKING:
+    from unstructured.ingest.connector.gcs import SimpleGcsConfig
 
+
+@dataclass
 class GCSRunner(FsspecBaseRunner):
+    fsspec_config: t.Optional["SimpleGcsConfig"] = None
```
ahmetmeleq, line 15: Given line 10,
```suggestion
    fsspec_config: t.Optional[SimpleGcsConfig] = None
```

unstructured/ingest/runner/dropbox.py
```diff
@@ -1,12 +1,19 @@
...
 from unstructured.ingest.runner.base_runner import FsspecBaseRunner
 from unstructured.ingest.runner.utils import update_download_dir_remote_url
 
+if t.TYPE_CHECKING:
+    from unstructured.ingest.connector.dropbox import SimpleDropboxConfig
 
+
+@dataclass
 class DropboxRunner(FsspecBaseRunner):
+    fsspec_config: t.Optional["SimpleDropboxConfig"] = None
```
ahmetmeleq, line 15: Given line 10,
```suggestion
    fsspec_config: t.Optional[SimpleDropboxConfig] = None
```

unstructured/ingest/cli/cmds/azure.py
rbiseck3: I actually copied these texts from the source library but I think they might have a typo themselves, will fix: [adlfs/spec.py#L127](https://github.com/fsspec/adlfs/blob/main/adlfs/spec.py#L127)

unstructured/ingest/runner/box.py
```diff
@@ -5,11 +5,15 @@
 from unstructured.ingest.runner.base_runner import FsspecBaseRunner
 from unstructured.ingest.runner.utils import update_download_dir_remote_url
 
+if t.TYPE_CHECKING:
+    from unstructured.ingest.connector.box import SimpleBoxConfig
+
 
 class BoxRunner(FsspecBaseRunner):
+    fsspec_config: t.Optional["SimpleBoxConfig"] = None
```
ahmetmeleq, line 13: Given line 9,
```suggestion
    fsspec_config: t.Optional[SimpleBoxConfig] = None
```

unstructured/ingest/runner/gcs.py
rbiseck3: When you use 
```
if t.TYPE_CHECKING:
    from unstructured.ingest.connector.gcs import SimpleGcsConfig
```
The dependency when used in the file needs to be in quotes. Otherwise it tried to import it.

rbiseck3: For reference: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

potter-potter (approved):
I've gone through everything. A lot to go through. But works well and I didn't see anything suspicious. Looks good. Really liking the access_configs instead of access_kwargs.

unstructured/ingest/runner/gcs.py
ahmetmeleq: Checked it now and sounds good 👍

PR #2195, opened 2023-11-30
chore: unstructured python client

### LINKED ISSUE #2033: Use the official python client when calling partition_by_api
We want to replace the requests call [here](https://github.com/Unstructured-IO/unstructured/blob/04fcdb91fe63ce31b18d1843f418e879bf3b9df7/unstructured/ingest/interfaces.py#L456C10-L456C10) with our [python client](https://github.com/Unstructured-IO/unstructured-python-client). This will give us better retry logic and error handling.

Coniferish:
### Summary
Closes #2033
Updates `partition_via_api` to use `UnstructuredClient` for api calls instead of `requests`.
Updates associated tests.

Note: This PR does **not** update `partition_multiple_via_api` as documentation in `unstructured-python-client` indicates it does not support multiple files. A new issue should be opened to add that functionality to `unstructured-python-client`.


awalker4 (approved):
Works locally! LGTM once tests pass (looks like those are transient ingest errors?).

PR #2204, opened 2023-12-04
feat: support glob filters for fsspec source connectors

rbiseck3:
### Description
The local source connector already supports filters on a list of glob filters optionally provided. This same code was leveraged in the base fsspec connector to filter on the full file paths returned from the underlying fs library.


potter-potter (approved):
This works well. Looks good.

PR #2215, opened 2023-12-05
chore: lint for print statements in ingest code

rbiseck3:
### Description
Given the filtering in the ingest logger, anything going to console should go through that. This adds a linter that only checks for `print()` statements in the ingest code and ignored it elsewhere for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.4-dev2
+## 0.11.4-dev3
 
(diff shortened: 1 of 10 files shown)


potter-potter:
This works. Looks good. Would you mind adding the Azure dest fix here since we still haven't merged the other one. And I'd like to see the CI pass.

potter-potter (approved):
Works well. Thanks.

PR #2217, opened 2023-12-05
chore: refactor ingest test script to use single script

rbiseck3:
### Description
This tackles 2 issues:
* Code reuse across the src and dest ingest test scripts
* Splitting up the src tests into core and non core to parallelize them a bit in the CI since the full run is now taking >45 min.


test_unstructured_ingest/test-ingest.sh
```diff
@@ -0,0 +1,101 @@
...
+  for test in $tests_to_run; do
+    CURRENT_TEST="$test"
+    echo "--------- RUNNING SCRIPT $test ---------"
+    echo "Running ./test_unstructured_ingest/$test_type/$test"
+    if [[ "${tests_to_omit[*]}" =~ $test ]]; then
+      echo "$test (skipped without running)" | tee -a "$SKIPPED_FILES_LOG"
+      continue
+    fi
+    ./test_unstructured_ingest/"$test_type"/"$test"
+    rc=$?
```
ahmetmeleq, line 79:
```suggestion
    # We get the exit code from the script in the previous line 
    rc=$?
```

test_unstructured_ingest/test-ingest.sh
```diff
@@ -0,0 +1,101 @@
...
+SKIPPED_FILES_LOG=$SCRIPT_DIR/skipped-files.txt
+# If the file already exists, reset it
+if [ -f "$SKIPPED_FILES_LOG" ]; then
+  rm "$SKIPPED_FILES_LOG"
+fi
+touch "$SKIPPED_FILES_LOG"
+cd "$SCRIPT_DIR"/.. || exit 1
+CURRENT_TEST="none"
+
+# Pull in variables specific to each
```
ahmetmeleq, line 33:
```suggestion
# Pull in variables specific to each test scenario
```

ahmetmeleq:
`test_unstructured_ingest/test-ingest-configs-src.sh`
`test_unstructured_ingest/test-ingest-configs-dest.sh`

The ones above sound like we're testing something about configs, I'd rather rename them as `configs-test-ingest-x.sh`.

test_unstructured_ingest/test-ingest.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2154
+
+set -u -o pipefail
```
ahmetmeleq, line 4:
```suggestion
# Make sure shell exits when there are empty variables
# Make sure the exit code for the last failed step (if any) in the pipeline will be pipeline's exit code
set -u -o pipefail
```

ahmetmeleq:
Love the parallelization. Looks good overall

test_unstructured_ingest/test-ingest.sh
ahmetmeleq: I don't think the line is self explanatory or a trivial command, could be good to add comments

test_unstructured_ingest/configs/dest.sh
```diff
@@ -0,0 +1,15 @@
...
+	'azure.sh'
+	'gcs.sh'
+	's3.sh'
+)
+tests_to_ignore=(
+	'notion.sh'
+	'dropbox.sh'
+	'sharepoint.sh'
+)
+tests_to_omit=()
```
ahmetmeleq, line 14: Checking [CI logs](https://github.com/Unstructured-IO/unstructured/actions/runs/7182374100/job/19559108932?pr=2217#step:7:1135), whole box test is commented out. Maybe we can omit it here instead

ahmetmeleq: Not setting credentials for `sharepoint-embed-cog-index.sh` [skips it](https://github.com/Unstructured-IO/unstructured/actions/runs/7182374100/job/19559108932?pr=2217#step:7:1556) implicitly, rather than explicitly omitting it. We could omit that one here as well

ahmetmeleq:
Looks good and ready to merge after tests pass

ahmetmeleq (approved):
LGTM

ryannikolaidis:
src and dest test script reuse makes sense to me. not sure I'm following on the benefits of the non-core as a separate job? each of the matrix python versions are run in parallel anyways, right? how does this save time? also, on net we actually add lines of code in this pr

ryannikolaidis:
> src and dest test script reuse makes sense to me. not sure I'm following on the benefits of the non-core as a separate job? each of the matrix python versions are run in parallel anyways, right? how does this save time? also, on net we actually add lines of code in this pr

ohh, nm, I see. because the 3.10 core tests run in parallel to 3.10 non-core tests. okay, makes sense to me.

.github/workflows/ci.yml
```diff
@@ -255,10 +255,88 @@ jobs:
...
+        UNS_API_KEY: ${{ secrets.UNS_API_KEY }}
+        NOTION_API_KEY: ${{ secrets.NOTION_API_KEY }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AZURE_SEARCH_ENDPOINT: ${{ secrets.AZURE_SEARCH_ENDPOINT }}
+        AZURE_SEARCH_API_KEY: ${{ secrets.AZURE_SEARCH_API_KEY }}
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        PINECONE_API_KEY: ${{secrets.PINECONE_API_KEY}}
+        TABLE_OCR: "tesseract"
+        OCR_AGENT: "tesseract"
```
ryannikolaidis, line 323: why are we setting ALL of these? shouldn't we just be setting keys needed for core tests? this is misleading and also means unnecessarily updating in multiple places

ryannikolaidis: and conversely should remove unneeded core-test secrets from the non-core job

.github/workflows/ci.yml
```diff
@@ -255,10 +255,88 @@ jobs:
...
+        CI: "true"
+      run: |
+        source .venv/bin/activate
+        sudo apt-get update
+        sudo apt-get install -y libmagic-dev poppler-utils libreoffice pandoc
+        sudo add-apt-repository -y ppa:alex-p/tesseract-ocr5
+        sudo apt-get install -y tesseract-ocr
+        sudo apt-get install -y tesseract-ocr-kor
+        sudo apt-get install diffstat
+        tesseract --version
```
ryannikolaidis, line 333: why isn't this in the venv already? issues with caching? this probably chews a bunch of time yea?

ryannikolaidis (changes requested):
couple quick comments and questions

MthwRobinson:
Closing since last commits were in December. Feel free to reopen if needed

Issue #2208, opened 2023-12-04
Enhancement: improve reading order


christinestraub:
**Is your feature request related to a problem? Please describe.**

PDF: [124_PDFsam_Basel III - Finalising post-crisis reforms.pdf](https://github.com/Unstructured-IO/unstructured/files/13550123/124_PDFsam_Basel.III.-.Finalising.post-crisis.reforms.pdf)

```
elements = partition_pdf(
    filename="124_PDFsam_Basel III - Finalising post-crisis reforms.pdf",
    strategy="hi_res",
    infer_table_structure=True,
)
```

There are 2 cases of formulas appear at the beginning of the element for some reason:
```
= . % kRW 2 25 . Risk weights for both risk-free yield curve and inflation rate are set at
...
= kRW RW ⋅ 6 , where Risk weights for both interest rate and inflation volatilities are set to σRW is set at 55%. σ
```

**Describe the solution you'd like**
Merge with XY-cut, with a preference of walking horizontally instead of vertically like the current default.

**Describe alternatives you've considered**
Pre-process bboxes before merge to alter size to Y-axis is aligned for tiny differences.

**Additional context**
- Extracted elements by `pdfminer`
![124_PDF_sam_Basel_1_extracted](https://github.com/Unstructured-IO/unstructured/assets/9475974/eb0f3ee8-2667-465a-9678-73600fdfcbcc)

PR #2219, opened 2023-12-05
Feat: improve reading order

christinestraub:


cragwolfe:
Confirmed that it improved order for the doc in https://github.com/Unstructured-IO/unstructured/issues/2208 .

Looking at element text joined by "\n":
```
Risk weights for both risk-free yield curve and inflation rate are set at
= . %
kRW 2 25 .
...
Risk weights for both interest rate and inflation volatilities are set to σRW is set at 55%. = kRW RW σ ⋅ 6 , where
```

PR #2220, opened 2023-12-05
update Run the library in a container section in README.md

PieterjanCriel:
from unstructured.partition.pdf import partition_pdf does not work anymore with the re structure.


Coniferish:
Thanks @PieterjanCriel! We've actually reorganized the structure again so pdf.py is in the partition directory, so that command is accurate again :)

PR #2152, opened 2023-11-23
feat: add elasticsearch destination connector

ahmetmeleq:
Closes https://github.com/Unstructured-IO/unstructured/issues/1842
Closes https://github.com/Unstructured-IO/unstructured/issues/2202
Closes https://github.com/Unstructured-IO/unstructured/issues/2203

This PR:
- Adds Elasticsearch destination connector to be able to ingest documents from any supported source, embed them and write the embeddings / documents into Elasticsearch.
- Defines an example unstructured elements schema for users to be able to setup their unstructured elasticsearch indexes easily.
- Includes parallelized upload and lazy processing for elasticsearch destination connector.
- Rearranges elasticsearch test helpers to source, destination, and common folders.
- Adds util functions to be able to batch iterables in a lazy way for uploads
- Fixes a bug where removing the optional parameter `--fields` broke the connector due to an integer processing error.
- Fixes a bug where using an [elasticsearch config](https://github.com/Unstructured-IO/unstructured/blob/8fa5cbf036c4b6a29a8e6c0cd81f22ef3ae84ed1/unstructured/ingest/connector/elasticsearch.py#L26-L35) for a destination connector resulted in a serialization issue when optional parameter `--fields` was not provided.

PR #2240, opened 2023-12-08
feat: add chroma destination connector

potter-potter:
Adds Chroma (also known as ChromaDB) as a vector destination.

Currently Chroma is an in-memory single-process oriented library with plans of a hosted and/or more production ready solution
-https://docs.trychroma.com/deployment

Though they now claim to support multiple Clients hitting the database at once, I found that it was inconsistent. Sometimes multiprocessing worked (maybe 1 out of 3 times) But the other times I would get different errors. So I kept it single process.


potter-potter:
I wonder why the chroma.txt requirements is so huge.

unstructured/ingest/connector/chroma.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @DestinationConnectionError.wrap
+    def check_connection(self):
+        _ = self.chroma_collection
+
+    @requires_dependencies(["chromadb"], extras="chroma")
+    def create_collection(self) -> "ChromaCollection":
+        import chromadb
+
+        chroma_client = chromadb.PersistentClient(path=self.connector_config.db_path)
```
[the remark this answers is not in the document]
rbiseck3: They also support an http client, would that make sense to expose as well? Could see someone using a remote instance more often than purely a local one.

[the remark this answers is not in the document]
potter-potter: Definitely. Thanks.

docs/source/ingest/destination_connectors/chroma.rst
```diff
@@ -0,0 +1,75 @@
...
+.. code:: shell
+
+  pip install "unstructured[chroma]"
+
+Run Locally
+-----------
+The upstream connector can be any of the ones supported, but for convenience here, showing a sample command using the
+upstream local connector. This will create new files on your local. And then upload those to Chroma.
+
+.. tabs::
```
ahmetmeleq, line 17: We seem to be having a new convention here where we render code from other folders:
https://github.com/Unstructured-IO/unstructured/blob/main/docs/source/ingest/destination_connectors/azure.rst

But it doesn't seem to be working / rendering for now so I'm not sure what to do here. I'd say this one is fine as well, just a note

unstructured/ingest/cli/cmds/weaviate.py
```diff
@@ -32,7 +32,8 @@ def get_cli_options() -> t.List[click.Option]:
                 help=(
                     "String representing a JSON-like dict with key,value containing "
                     "the required parameters to create an authentication object. "
-                    "The connector resolves th authentication object from the parameters. "
+                    'example: \'{"api_key":"123abc!"}\' '
```
ahmetmeleq, line 35: Thanks for this

unstructured/ingest/connector/chroma.py
```diff
@@ -0,0 +1,162 @@
...
+    WriteConfig,
+)
+from unstructured.ingest.logger import logger
+from unstructured.staging.base import flatten_dict
+from unstructured.utils import requires_dependencies
+
+if t.TYPE_CHECKING:
+    from chromadb import Collection as ChromaCollection
+
+    # from chromadb.config import Settings as ChromaSettings
```
ahmetmeleq, line 21: Is this still relevant?

unstructured/ingest/connector/chroma.py
```diff
@@ -0,0 +1,162 @@
...
+                ids=batch["ids"],
+                documents=batch["documents"],
+                embeddings=batch["embeddings"],
+                metadatas=batch["metadatas"],
+            )
+        except Exception as e:
+            raise ValueError(f"chroma error: {e}") from e
+
+    @staticmethod
+    def chunks(iterable, batch_size=100):
```
ahmetmeleq, line 108: This seems usable for other connectors, maybe we can move it to ingest utils. I also had [something similar](https://github.com/Unstructured-IO/unstructured/pull/2152/files#diff-f9e7c943e53c317768eadafb884c26862d85943655348d0a56e17ec7b76be445R191) for elasticsearch.

unstructured/ingest/connector/chroma.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def write(self, docs: t.List[BaseIngestDoc]) -> None:
+        dict_list: t.List[t.Dict[str, t.Any]] = []
+        for doc in docs:
+            local_path = doc._output_filename
+            with open(local_path) as json_file:
+                dict_content = json.load(json_file)
+
+                dict_content = [
+                    {
```
ahmetmeleq, line 150: Maybe we could unite this dict parsing part with the function `prepare_chroma_dict` above, putting all dict forming logic at one place

unstructured/staging/base.py
```diff
@@ -179,6 +179,9 @@ def elements_from_json(
 def flatten_dict(
     dictionary, parent_key="", separator="_", flatten_lists=False, keys_to_omit: List[str] = None
 ):
+    """Flattens a nested dictionary into a single level dictionary. keys_to_omit is a list of keys
```
ahmetmeleq, line 182: Thanks for this

unstructured/ingest/cli/cmds/chroma.py
```diff
@@ -0,0 +1,106 @@
...
+
+def get_base_dest_cmd():
+    from unstructured.ingest.cli.base.dest import BaseDestCmd
+
+    cmd_cls = BaseDestCmd(
+        cmd_name="chroma",
+        cli_config=ChromaCliConfig,
+        additional_cli_options=[ChromaCliWriteConfig],
+        write_config=ChromaWriteConfig,
+        # addition_configs={
```
ahmetmeleq, line 101: Are these still relevant?

scripts/chroma-test-helpers/create-and-check-chroma.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -e
+
+# $1 is first argument passed to shell. & runs process in background
```
ahmetmeleq, line 5: Better to instead comment on what we intend to do with $1 
```suggestion
# $1 is the path for chroma to write the contents to. The symbol "&" runs process in background
```

ahmetmeleq:
A few minor comments, otherwise looks good

docs/source/ingest/destination_connectors/chroma.rst
potter-potter: This will have to match the new doc format where we break out the .py and .sh Thanks for the reminder

unstructured/ingest/connector/chroma.py
potter-potter: I like the thought. Unfortunately then the chunks method and multiprocessing (which will work eventually) become much more complicated because you have a bunch of parallel lists instead of a list of dicts. (i changed prepare_chroma_dict to prepare_chroma_lists) So I'd like to keep them separate.

unstructured/ingest/connector/chroma.py
```diff
@@ -0,0 +1,162 @@
...
+
+if t.TYPE_CHECKING:
+    from chromadb import Collection as ChromaCollection
+
+    # from chromadb.config import Settings as ChromaSettings
+
+
+@dataclass
+class SimpleChromaConfig(BaseConnectorConfig):
+    collection_name: str
```
rbiseck3, line 26: How does this authenticate when communicating with something not on the user's local? Should contain that within an `AccessConfig`.

unstructured/ingest/connector/chroma.py
```diff
@@ -0,0 +1,162 @@
...
+    path: t.Optional[str] = None
+    settings: t.Optional[t.Dict[str, str]] = None
+    tenant: t.Optional[str] = None
+    database: t.Optional[str] = None
+    host: t.Optional[str] = None
+    port: t.Optional[int] = None
+    ssl: bool = False
+    headers: t.Optional[t.Dict[str, str]] = ({},)
```
rbiseck3, line 34: Should default to `None` if optional.
```suggestion
    headers: t.Optional[t.Dict[str, str]] = None
```

unstructured/ingest/cli/cmds/chroma.py
```diff
@@ -0,0 +1,106 @@
...
+                ["--ssl"],
+                required=False,
+                default=False,
+                type=bool,
+                help="Whether to use SSL to connect to the Chroma server.",
+            ),
+            click.Option(
+                ["--headers"],
+                required=False,
+                type=dict,
```
rbiseck3, line 65: I don't think `Click` knows how to receive a dict from the command line. Should use the custom `Dict` type in the cli interfaces file.

unstructured/ingest/cli/cmds/chroma.py
```diff
@@ -0,0 +1,106 @@
...
+            click.Option(
+                ["--path"],
+                required=False,
+                type=str,
+                help="Location where Chroma is persisted," "if not connecting via http.",
+            ),
+            click.Option(
+                ["--settings"],
+                required=False,
+                type=dict,
```
rbiseck3, line 26: Same comment as what I left for headers

unstructured/ingest/cli/cmds/chroma.py
```diff
@@ -0,0 +1,106 @@
...
+            click.Option(
+                ["--port"],
+                required=False,
+                type=int,
+                help="The port of the Chroma server.",
+            ),
+            click.Option(
+                ["--ssl"],
+                required=False,
+                default=False,
```
rbiseck3, line 58: For bool types, it's nice to leverage the `is_flag` option
```suggestion
                default=False,
                is_flag=True,
```

unstructured/ingest/connector/chroma.py
potter-potter: Agree. It should be "very" similar.
I think Chunks is too similar to Chunking though. We should rename it.

unstructured/ingest/connector/chroma.py
potter-potter: @rbiseck3 Was definitely looking into this. You think just `headers` and `settings` belong in AccessConfig or also `port, host, ssh`? I chose just `headers, settings` due to the `AccessConfig` docstring.

https://github.com/chroma-core/chroma/blob/main/chromadb/__init__.py

unstructured/ingest/connector/chroma.py
ahmetmeleq: Sorry, rethinking, what you do does make sense. Agreed with all

unstructured/ingest/connector/pinecone.py
```diff
@@ -87,15 +87,6 @@ def upsert_batch(self, batch):
             raise WriteError(f"http error: {api_error}") from api_error
         logger.debug(f"results: {response}")
 
-    @staticmethod
-    def chunks(iterable, batch_size=100):
```
ahmetmeleq, line 91: Thank you for this

test_unstructured_ingest/dest/chroma.sh
```diff
@@ -0,0 +1,57 @@
...
+	--chunk-elements \
+	--chunk-multipage-sections \
+	--embedding-provider "langchain-huggingface" \
+	chroma \
+	--host "localhost" \
+	--port 8000 \
+	--collection-name "$COLLECTION_NAME" \
+	--batch-size 80
+
+python "$SCRIPT_DIR"/python/test-ingest-chroma-output.py --collection-name "$COLLECTION_NAME"
```
ahmetmeleq, line 57: Smart

ahmetmeleq (approved):
LGTM!

PR #2243, opened 2023-12-11
chore: update sphinx ingest docs with new connectors

rbiseck3:
### Description
Given the latest refactor of the ingest connector code base, this PR update all python code to reflect the new approach. Part of this PR also refactors the way docs were handled were all code is now in a dedicated file the the appropriate file extension to get handled by the linters and referenced where needed in the actual doc.

PR #2245, opened 2023-12-11
chore: update sphinx ingest docs with new connectors

rbiseck3:
Replacing https://github.com/Unstructured-IO/unstructured/pull/2243


docs/source/ingest/destination_connectors/code/bash/pinecone.sh
```diff
@@ -0,0 +1,18 @@
...
+	--input-path example-docs/book-war-and-peace-1225p.txt \
+	--output-dir local-output-to-pinecone \
+	--strategy fast \
+	--chunk-elements \
+	--embedding-provider "$EMBEDDING_PROVIDER" \
+	--num-processes 2 \
+	--verbose \
+	pinecone \
+	--api-key "$PINECONE_API_KEY" \
+	--index-name "$PINECONE_INDEX_NAME" \
```
potter-potter, line 16: would be nice to add num-processes here since I think we added that. (maybe for just pinecone only? it doesn't work on chroma)

potter-potter (approved):
added nice to have for --num-processes for pinecone dest.

Other than that it looks good.

PR #2246, opened 2023-12-11
chore: add shfmt

rbiseck3:
### Description
Given all the shell files that now exist in the repo, would be nice to have linting/formatting around them (in addition to the existing shellcheck which doesn't do anything to format the shell code). This PR introduces `shfmt` to both check for changes and apply formatting when the associated make targets are called.


examples/ingest/pinecone/ingest.sh
```diff
@@ -5,26 +5,25 @@
...
-        --index-name "<Pinecone index name, ie: ingest-test>" \
-        --environment "<Pinecone index name, ie: ingest-test>" \
-        --batch-size "<Number of elements to be uploaded per batch, ie. 80>" \
-        --num-processes "<Number of processes to be used to upload, ie. 2>"
+	local \
+	--input-path example-docs/book-war-and-peace-1225p.txt \
+	--output-dir local-to-pinecone \
+	--strategy fast \
+	--chunk-elements \
+	--embedding-provider unstructured embedding provider, ie. langchain-huggingface \
```
potter-potter, line 20: This looks wrong

examples/ingest/pinecone/ingest.sh
```diff
@@ -5,26 +5,25 @@
...
+	2 \
```
potter-potter, line 21: And  here looks wrong too

scripts/docker-build.sh
```diff
@@ -5,16 +5,16 @@ DOCKER_REPOSITORY="${DOCKER_REPOSITORY:-quay.io/unstructured-io/unstructured}"
 PIP_VERSION="${PIP_VERSION:-23.1.2}"
 DOCKER_IMAGE="${DOCKER_IMAGE:-unstructured:dev}"
 
-DOCKER_BUILD_CMD=(docker buildx build --load -f Dockerfile \
-  --build-arg PIP_VERSION="$PIP_VERSION" \
-  --build-arg BUILDKIT_INLINE_CACHE=1 \
-  --progress plain \
-  --cache-from "$DOCKER_REPOSITORY":latest \
-  -t "$DOCKER_IMAGE" .)
+DOCKER_BUILD_CMD=(docker buildx build --load -f Dockerfile
```
potter-potter, line 8: Double check that backslashes aren't needed. :)

rbiseck3: Verified, this works when building the docker image

scripts/performance/benchmark.sh
```diff
@@ -20,49 +20,48 @@ S3_RESULTS_DIR="performance-test/results"
...
+		unstructured:perf-test \
+		bash /home/notebook-user/scripts/performance/benchmark-local.sh 2>&1 | tee >(while IFS= read -r line; do
+			read_benchmark_logs_for_results
+		done)
 else
-    NUM_ITERATIONS="$NUM_ITERATIONS" INSTANCE_TYPE="$INSTANCE_TYPE" GIT_HASH="$GIT_HASH" SLOW_FILES="${SLOW_FILES[*]}" HI_RES_STRATEGY_FILES="${HI_RES_STRATEGY_FILES[*]}" "$SCRIPT_DIR"/benchmark-local.sh 2>&1 | \
-    tee >(while IFS= read -r line; do
-        read_benchmark_logs_for_results
-    done)
+	NUM_ITERATIONS="$NUM_ITERATIONS" INSTANCE_TYPE="$INSTANCE_TYPE" GIT_HASH="$GIT_HASH" SLOW_FILES="${SLOW_FILES[*]}" HI_RES_STRATEGY_FILES="${HI_RES_STRATEGY_FILES[*]}" "$SCRIPT_DIR"/benchmark-local.sh 2>&1 |
```
potter-potter, line 52: I don't know shell well enough to tell if the backslash can be removed...

potter-potter:
This looks good. We are going to have some good looking shell scripts! Mostly it was lots of simple reformatting. I was a little hesitant when it removed some backslashes. But that was it.

scripts/performance/benchmark.sh
rbiseck3: verified this works

examples/ingest/pinecone/ingest.sh
potter-potter: This should still be --num-processes 2 \

examples/ingest/pinecone/ingest.sh
potter-potter: This should have quotations and then "<an blablabla>"

potter-potter (approved):
Looks great!

cragwolfe:
The massive indentation in some scripts feels less reader-friendly to me. Can some scripts go back to 2 spaces per level of indentation?

E.g., https://github.com/Unstructured-IO/unstructured/blob/main/scripts/user/unstructured-get-json.sh

PR #2252, opened 2023-12-12
fix: image may be scaled too large for tesseract

badGarnet:
This PR addresses [CORE-2965](https://unstructured-ai.atlassian.net/browse/CORE-2965) by limiting zoom factor so that the scaled image can still be processed by tesseract.

- tesseract has a 2^31 byte limit on image data
- occasionally an image may be scaled too much and larger than that size
- fix limits the scaling factor so that we never scale an image larger than what tesseract can handle

## test

A unit test is added in this PR to test a unlikely case where we'd scale an image a few thousand times and massively exceed the limit without the fix. 

Unstructured reviewers can also use the document in the ticket to test.

[CORE-2965]: https://unstructured-ai.atlassian.net/browse/CORE-2965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test_unstructured/partition/pdf_image/test_ocr.py
+++ b/test_unstructured/partition/pdf_image/test_ocr.py
@@ -468,3 +468,28 @@ def test_get_table_tokens(mock_ocr_layout):
         ]
 
         assert table_tokens == expected_tokens
+
+
+def test_auto_zoom_not_exceed_tesseract_limit(monkeypatch):
+    monkeypatch.setenv("TESSERACT_MIN_TEXT_HEIGHT", "1000")
+    monkeypatch.setenv("TESSERACT_OPTIMUM_TEXT_HEIGHT", "100000")
+    monkeypatch.setattr(
+        unstructured_pytesseract,
+        "image_to_data",
+        lambda *args, **kwargs: pd.DataFrame(
+            {
+                "left": [10, 20, 30, 0],
+                "top": [5, 15, 25, 0],
+                "width": [15, 25, 35, 0],
+                "height": [10, 20, 30, 0],
+                "text": ["Hello", "World", "!", ""],
+            },
+        ),
+    )
+
+    image = Image.new("RGB", (1000, 1000))
+    assert [region.text for region in ocr.get_ocr_layout_tesseract(image)] == [
+        "Hello",
+        "World",
+        "!",
+    ]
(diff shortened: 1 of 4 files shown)


test_unstructured/partition/pdf_image/test_ocr.py
```diff
@@ -468,3 +468,28 @@ def test_get_table_tokens(mock_ocr_layout):
...
+            },
+        ),
+    )
+
+    image = Image.new("RGB", (1000, 1000))
+    assert [region.text for region in ocr.get_ocr_layout_tesseract(image)] == [
+        "Hello",
+        "World",
+        "!",
+    ]
```
qued, line 495: What's being tested here, that Tesseract doesn't error out during this operation?

unstructured/partition/pdf_image/ocr.py
```diff
@@ -459,9 +460,13 @@ def get_ocr_layout_tesseract(
         text_height < env_config.TESSERACT_MIN_TEXT_HEIGHT
         or text_height > env_config.TESSERACT_MAX_TEXT_HEIGHT
     ):
+        max_zoom = max(0, np.round(np.sqrt(TESSERACT_MAX_SIZE / np.prod(image.size) / 32), 1))
```
qued, line 463: What does 32 represent here? It seems like it should be bytes per pixel but 32 seems too big to be that. Is it 32 bits per pixel and maybe it should be 4 bytes per pixel?

qued (commented):
A few questions.

test_unstructured/partition/pdf_image/test_ocr.py
badGarnet: right; and scaled image still returns expected results

unstructured/partition/pdf_image/ocr.py
badGarnet: color depth; ideally we compute that from the image property but conveniently we always pass in a 32 color image in by the PIL open as RGB mode.

awalker4 (approved):
LGTM! Verified that the file in the ticket works, as well as the one attached in [CORE-2696](https://unstructured-ai.atlassian.net/browse/CORE-2696).

qued (approved):
Confirmed the test fails on `main` and passes on this branch.

PR #2253, opened 2023-12-12
rfctr: skip CheckBox elements during chunking

scanny:
`CheckBox` elements get special treatment during chunking. `CheckBox` does not derive from `Text` and can contribute no text to a chunk. It is considered "non-combinable" and so is emitted as-is as a chunk of its own. A consequence of this is it breaks an otherwise contiguous chunk into two wherever it occurs.

This is problematic, but becomes much more so when overlap is introduced. Each chunk accepts a "tail" text fragment from its preceding element and contributes its own tail fragment to the next chunk. These tails represent the "overlap" between chunks. However, a non-text chunk can neither accept nor provide a tail-fragment and so interrupts the overlap. None of the possible solutions are terrific.

Give `Element` a `.text` attribute such that _all_ elements have a `.text` attribute, even though its value is the empty-string for element-types such as CheckBox and PageBreak which inherently have no text. As a consequence, several `cast()` wrappers are no longer required to satisfy strict type-checking.

This also allows a `CheckBox` element to be combined with `Text` subtypes during chunking, essentially the same way `PageBreak` is, contributing no text to the chunk.

Also, remove the `_NonTextSection` object which previously wrapped a `CheckBox` element during pre-chunking as it is no longer required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.4-dev9
+## 0.11.4-dev10
 
 ### Enhancements
 
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -5,7 +5,6 @@ from typing import List
 import pytest
 
 from unstructured.chunking.title import (
-    _NonTextSection,
     _SectionCombiner,
     _split_elements_by_title_and_table,
     _TableSection,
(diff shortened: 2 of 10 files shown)


christinestraub (changes requested):
What do you think about adding "text" key to `to_dict()` method of the `Element` class?

christinestraub (approved):
LGTM!

scanny:
Thanks @christinestraub! :)

PR #2261, opened 2023-12-13
rfctr: rename pre chunk

scanny:
The original naming for the pre-cursor to a chunk in `chunk_by_title()` was conflated with the idea of how these element subsequences were bounded (by document-section) for that strategy. I mistakenly picked that up as a universal concept but in fact no notion of section arises in the `by_character` or other chunking strategies.

Fix this misconception by using the name `pre-chunk` for this concept throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.4-dev10
+## 0.11.4-dev11
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


christinestraub (approved):
LGTM!

scanny:
Thanks very much @christinestraub! :)

PR #2265, opened 2023-12-13
rfctr(chunking): move add_chunking_strategy() decorator up

scanny:
The chunking subpackage `unstructured.chunking` currently contains only the `title` module and the `@add_chunking_strategy()` decorator is located in that module even though it has no special relationship to the `by_title` chunking strategy.

Move it to the `__init__.py` module such that it is exported from `unstructured.chunking`. Adjust all references, pretty much one per partitioner, to import it from there.

This prepares the way for further separation of the chunking package into modules, including a new `character` module for the `by_character` chunking strategy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.4-dev13
+## 0.11.4-dev14
 
 ### Enhancements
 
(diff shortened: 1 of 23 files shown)


christinestraub (approved):
LGTM!

scanny:
Thanks @christinestraub! :)

PR #2266, opened 2023-12-14
rfctr: extract ChunkingOptions

scanny:
Chunking options for things like chunk-size are largely independent of chunking strategy. Further, validating the args and applying defaults based on call arguments is sophisticated to make its use easy for the caller. These details distract from what the chunker is actually doing and would need to be repeated for every chunking strategy if left where they are.

Extract these settings and the rules governing chunking behavior based on options into its own immutable object that can be passed to any component that is subject to optional behavior (pretty much all of them).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.11.5-dev0
+
+### Enhancements
+
+### Features
+
+### Fixes
+
 ## 0.11.4
 
 ### Enhancements
--- a/test_unstructured/chunking/test_base.py
+++ b/test_unstructured/chunking/test_base.py
@@ -1,0 +1,113 @@
+"""Unit-test suite for the `unstructured.chunking.base` module."""
+
+from __future__ import annotations
+
+import pytest
+
+from unstructured.chunking.base import ChunkingOptions
+
+
+class DescribeChunkingOptions:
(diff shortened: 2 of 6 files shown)


christinestraub:
```
def _validate(self) -> None:
    ...
    if new_after_n_chars is not None and new_after_n_chars < 0:
        raise ValueError(
            f"'new_after_n_chars' argument must be >= 0," f" got {new_after_n_chars}"
        )
    ...
```

```
def soft_max(self) -> int:
    ...
    return (
        max_chars
        if (new_after_n_chars is None or new_after_n_chars < 0 or new_after_n_chars > max_chars)
        else new_after_n_chars
    )
```

@scanny if `new_after_n_chars < 0`, would a ValueError be raised, or would it be set to `max_chars`?

scanny:
@christinestraub it will raise `ValueError` on construction. This behavior is tested here: https://github.com/Unstructured-IO/unstructured/pull/2266/files#diff-195456ab10930a23a2630125985194c82a8fae098c6f5d4436c3468bd0ee201aR68-R74

The clause on `.soft_max` is just being thorough by mentioning that. Otherwise it would be depending on a distant behavior that could change later.

christinestraub (approved):
LGTM!

scanny:
Thanks very much @christinestraub! :)

PR #2268, opened 2023-12-14
Fix: MongoDB connector URI password redaction, basic unit tests for Git connector

jakub-sandomierz-deepsense-ai:
MongoDB connector:
Issue:
[MongoDB documentation](https://www.mongodb.com/docs/manual/reference/connection-string/) states that characters `$ : / ? # [ ] @` must be percent encoded. URI with password containing such special character will not be redacted.

Fix:
This fix removes usage of `unquote_plus` on password which allows detected password to match with one inside URI and successfully replace it.

Git connector:
Added very basic unit tests for repository filtering methods. Their impact is rather minimal but showcases current limitation in `is_file_type_supported` method.


rbiseck3:
@jakub-sandomierz-deepsense-ai, can you add a detailed PR description with what the issue was, how this fixed it. The impact of the unit tests added were.

rbiseck3 (approved):
Just need to rebase off main but otherwise, great find.

PR #2281, opened 2023-12-15
rfctr(chunking): extract general-purpose objects to base

scanny:
Many of the classes defined in `unstructured.chunking.title` are applicable to any chunking strategy and will shortly be used for the "by-character" chunking strategy as well.

Move these and their tests to `unstructured.chunking.base`.

Along the way, rename `TextPreChunkBuilder` to `PreChunkBuilder` because it will be generalized in a subsequent PR to also take `Table` elements such that inter-pre-chunk overlap can be implemented.

Otherwise, no logic changes, just moves.


christinestraub (approved):
LGTM!

scanny:
Thanks @christinestraub! :)

PR #2283, opened 2023-12-15
rfctr(chunking): generalize PreChunkBuilder

scanny:
To implement inter-pre-chunk overlap, we need a context that sees every pre-chunk both before and after it is accumulated (from elements).

- We need access to the pre-chunk when it is completed so we can extract the "tail" overlap to be applied to the next chunk.
- We need access to the as-yet-unpopulated pre-chunk so we can add the prior tail to it as a prefix.

This "visibility" is split between `PreChunkBuilder` and the pre-chunker itself, which handles `TablePreChunk`s without the builder.

Move `Table` element and TablePreChunk` formation into `PreChunkBuilder` such that _all_ element types (adding `Table` elements in particular) pass through it. Then `PreChunkBuilder` becomes the context we require.

The actual overlap harvesting and application will come in a subsequent commit.

### DIFF
--- a/test_unstructured/chunking/test_base.py
+++ b/test_unstructured/chunking/test_base.py
@@ -18,6 +18,7 @@ from unstructured.chunking.base import (
 )
 from unstructured.documents.elements import (
     CompositeElement,
+    Element,
     ElementMetadata,
     PageBreak,
     RegexMetadata,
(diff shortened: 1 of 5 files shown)


test_unstructured/chunking/test_base.py
```diff
@@ -572,24 +573,83 @@ class DescribePreChunkBuilder:
     def it_is_empty_on_construction(self):
         builder = PreChunkBuilder(opts=ChunkingOptions.new(max_characters=50))
 
-        assert builder.text_length == 0
-        assert builder.remaining_space == 50
+        assert builder._text_length == 0
+        assert builder._remaining_space == 50
```
scanny, line 577: Interface-wise, these two attributes are replaced by `.will_fit(element)` and so become private.

christinestraub (approved):
LGTM!

scanny:
Thanks @christinestraub! :)

PR #2287, opened 2023-12-17
chore: shell scripts default indent of 2 instead of 4

cragwolfe:
Given the tendency for shell scripts to easily enter into a few levels of indentation and long line lengths, update the default to 2 spaces.


christinestraub (approved):
LGTM!

PR #2289, opened 2023-12-18
chore: add hi_res_model_name kwarg

### LINKED ISSUE #2160: Add `hi_res_model_name` as an alias kwarg
The api param `hi_res_model_name` does not match with `model_name` in `unstructured`. Rather than changing the param in unstructured-api, add it as an alias kwarg in this library. See discussion [here](https://github.com/Unstructured-IO/unstructured-api/issues/315)

Coniferish:
Closes #2160 

Explicitly adds `hi_res_model_name` as kwarg to relevant functions and notes that `model_name` is to be deprecated.

Testing:
```
filename = "example-docs/DA-1p.pdf"
elements = partition(filename, strategy="hi_res", hi_res_model_name="yolox")
```


awalker4 (approved):
LGTM! I'd just suggest a quick note in the changelog

PR #2294, opened 2023-12-19
rfctr(chunking): extract BasePreChunker

scanny:
The `_split_elements_by_title_and_table()` function fulfills the pre-chunker role for `chunk_by_title()`, but most of its operation is not strategy-specific and can be reused by other chunking strategies.

Extract `BasePreChunker` and use it as the base class for `_ByTitlePreChunker` which now only needs to provide the boundary predicates specific to that strategy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.6-dev3
+## 0.11.6-dev4
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


christinestraub (approved):
The base class `BasePreChunker` seems useful and reusable when implementing a specific chunking strategy :)

scanny:
Thanks @christinestraub! :)

PR #2297, opened 2023-12-19
rfctr(chunking): split oversized chunks on word boundary

scanny:
The text of an oversized chunk is split on an arbitrary character boundary (mid-word). The `chunk_by_character()` strategy introduces the idea of allowing the user to specify a separator to use for chunk-splitting. For `langchain` this is typically "\n\n", "\n", or " "; blank-line, newline, or word boundaries respectively.

Even if the user is allowed to specify a separator, we must provide fall-back for when a chunk contains no such character. This can be done incrementally, like blank-line is preferable to newline, newline is preferable to word, and word is preferable to arbitrary character.

Further, there is nothing particular to `chunk_by_character()` in providing such a fall-back text-splitting strategy. It would be
preferable for all strategies to split oversized chunks on even-word boundaries for example.

Note that while a "blank-line" ("\n\n") may be common in plain text, it is unlikely to appear in the text of an element because it would have been interpreted as an element boundary during partitioning.

Add _TextSplitter with basic separator preferences and fall-back and apply it to chunk-splitting for all strategies. The `by_character` chunking strategy may enhance this behavior by adding the option for a user to specify a particular separator suited to their use case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.11.7-dev0
+
+### Enhancements
+
+### Features
+
+### Fixes
+
 ## 0.11.6
 
 ### Enhancements
--- a/test_unstructured/chunking/test_base.py
+++ b/test_unstructured/chunking/test_base.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from typing import List
+from typing import List, Sequence
 
 import pytest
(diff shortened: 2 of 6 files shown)


christinestraub (approved):
LGTM!

scanny:
Thanks @christinestraub! :)

PR #2299, opened 2023-12-19
feat: update cct eval for text dir

Klaijan:
The code makes edit to the `measure_text_extraction_accuracy` function to allows dir of txt as well as json. The function also takes input `output_type` to be either "json" or "txt" only, and checks if the files under given directory/list contains only specified file type or not.

To test this feature, run the following code:

```PYTHONPATH=. python unstructured/ingest/evaluate.py measure-text-extraction-accuracy-command --output_dir <clean-text-path> --source_dir <cct-label-path> --output_type txt```

### CODE REVIEW

File: unstructured/metrics/evaluate.py [hunk 1]
```diff
@@ -59,13 +60,23 @@ def measure_text_extraction_accuracy(
...
+    if output_type not in ["json", "txt"]:
+        raise ValueError(
+            f"Specified file type under `output_dir` or `output_list` should be one of \
+                'json' or 'txt'. The given file type is {output_type}, exiting."
+        )
+    if not all(_.endswith(output_type) for _ in output_list):
+        print(
+            "The directory contains file type inconsistent with the given input. \
+                Please note that some files will be skipped."
+        )
```
badGarnet, line 72: can we change all those instances of `print` to `logger.info` or `logger.warning` depends on the message please?

unstructured/metrics/evaluate.py
```diff
@@ -59,13 +60,23 @@ def measure_text_extraction_accuracy(
...
 
     rows = []
 
     # assumption: output file name convention is name-of-file.doc.json
     # NOTE(klaijan) - disable=True means to not show, disable=False means to show the progress bar
     for doc in tqdm(output_list, leave=False, disable=not visualize):  # type: ignore
-        filename = (doc.split("/")[-1]).split(".json")[0]
+        filename = (doc.split("/")[-1]).split(f".{output_type}")[0]
```
badGarnet, line 79: a small trick: here using split to just drop the file extension actually lead to creation of a list (i.e., expensive)
here we really need is the part of the string before the extension so we can do something like 
```
ext_index = -len(output_type)
for doc in tqdm(...):
    filename = os.path.basename(doc)[:ext_index]
```
also here using split on "/" is potentially problematic as it assumes the os separator is "/"

unstructured/metrics/evaluate.py
```diff
@@ -77,11 +88,13 @@ def measure_text_extraction_accuracy(
             fn_txt = fn + ".txt"
 
         if fn_txt in source_list:  # type: ignore
-            output_cct = elements_to_text(elements_from_json(os.path.join(output_dir, doc)))
-            source_cct = _read_text(os.path.join(source_dir, fn_txt))
+            try:
+                output_cct = _prepare_output_cct(os.path.join(output_dir, doc), output_type)
+                source_cct = _read_text(os.path.join(source_dir, fn_txt))
+            except Exception:
+                pass
```
badGarnet, line 95: this might be too generous; we at least should bring the exception message up so the log have a record on if any doc has issues with processing

unstructured/metrics/evaluate.py
```diff
@@ -212,9 +235,20 @@ def _write_to_file(dir: str, filename: str, df: pd.DataFrame, mode: str = "w"):
...
+        filename = _uniquity_file(dir, filename)
     df.to_csv(os.path.join(dir, filename), sep="\t", mode=mode, index=False, header=(mode == "w"))
 
 
+def _uniquity_file(dir, filename):
+    counter = 1
+    original_filename, extension = filename.rsplit(".", 1)
+    while os.path.exists(os.path.join(dir, filename)):
+        filename = original_filename + " (" + str(counter) + ")." + extension
+        counter += 1
```
badGarnet, line 248: I would suggest to use `os.listdir` and filtering to find all the files match the pattern and get the counter values and get max -> increment by 1
Reason:
- repeatedly calling os.path.join, path.exists is more expensive than just calling once listdir
- a while loop is a dangerous trap for infinite loop and I'd suggest we avoid it as much as we can

unstructured/metrics/utils.py
```diff
@@ -0,0 +1,165 @@
...
+
+import click
+import pandas as pd
+
+from unstructured.staging.base import elements_from_json, elements_to_text
+
+logger = logging.getLogger("unstructured.eval")
+
+
+def _prepare_output_cct(docpath: str, output_type: str):
```
cragwolfe, line 15: these could all use docstrings (even though they are internal functions)

cragwolfe:
> To test this feature, run measure_text_edit_distance as usual, and specify --output_type txt.

please provide a test snippet

awalker4:
In order to run that test command - can you add instructions for populating the source dir and output dir?

Klaijan:
> In order to run that test command - can you add instructions for populating the source dir and output dir?

For `txt`:

```PYTHONPATH=. python unstructured/ingest/evaluate.py measure-text-extraction-accuracy-command --output_dir unstructured-output-cct --source_dir gold-standard-cct --output_type txt```

For `json`:

```PYTHONPATH=. python unstructured/ingest/evaluate.py measure-text-extraction-accuracy-command --output_dir unstructured-output --source_dir gold-standard-cct --output_type json```

awalker4 (approved):
Command works for me!

PR #2303, opened 2023-12-20
Refactor: rename image extraction kwargs

christinestraub:
Currently, we're using different kwarg names in partition() and partition_pdf(), which has implications for the API since it goes through partition().

### Summary
- rename `extract_element_types` -> `extract_image_block_types`
- rename `image_output_dir_path` to `extract_image_block_output_dir`
- rename `extract_to_payload` -> `extract_image_block_to_payload`
- rename `pdf_extract_images` -> `extract_images_in_pdf` in `partition.auto`
- add unit tests to test element extraction for `pdf/image` via `partition.auto`
### Testing
CI should pass.


Coniferish:
Looks good. Only question/concern is if `extract_images_in_pdf` needs to just be deprecated instead of replaced. I'd assume if it was recently added it'd be safe to do so, but will leave that to @qued 's discretion.

christinestraub:
> Looks good. Only question/concern is if `extract_images_in_pdf` needs to just be deprecated instead of replaced. I'd assume if it was recently added it'd be safe to do so, but will leave that to @qued 's discretion.

CC: @cragwolfe

unstructured/partition/pdf.py
```diff
@@ -172,16 +172,16 @@ def partition_pdf(
         with Tesseract, you'll first need to install the appropriate Tesseract language pack.
     metadata_last_modified
         The last modified date for the document.
-    extract_images_in_pdf
```
cragwolfe, line 175: let's keep the old kwarg around as a deprecated argument for awhile to avoid breaking changes.

Coniferish:
[re: comment here](https://github.com/Unstructured-IO/unstructured/pull/2310) I wish I had thought of this earlier, but with so many interdependent parameters, I wonder if it would made more sense to have `extract_to_payload` be a Tuple(str, list) instead of a list. Then the first index could be used to determine `extract_to_payload` or `image_output_dir_path` since those two parameters seem to be mutually exclusive.

christinestraub:
> [re: comment here](https://github.com/Unstructured-IO/unstructured/pull/2310) I wish I had thought of this earlier, but with so many interdependent parameters, I wonder if it would made more sense to have `extract_to_payload` be a Tuple(str, list) instead of a list. Then the first index could be used to determine `extract_to_payload` or `image_output_dir_path` since those two parameters seem to be mutually exclusive.

@Coniferish Are you referring to something like this?
CC: @cragwolfe
```
element_extraction=("payload", ["Image", "Table"])
or 
element_extraction = ("output_dir", ["Image", "Table"])
```

Coniferish:
> @Coniferish Are you referring to something like this?

Yah, that's what I was thinking @christinestraub

Coniferish (approved):
LGTM! I like these names a lot more!

Coniferish (changes requested):
Just noticing that this is removing the old params, but @cragwolfe mentioned keeping/deprecating them instead

christinestraub:
> Just noticing that this is removing the old params, but @cragwolfe mentioned keeping/deprecating them instead

For the `extract_images_in_pdf` param, I kept it in both `auto.partition()` and `partition.pdf()`
For the `image_output_dir_path` param, I renamed it to `extract_image_block_output_dir` since this param is optional (can be omitted), so this update will not bring the breaking change. CC: @cragwolfe

Coniferish (approved):

PR #2309, opened 2023-12-21
feat(chunking): add inter-chunk overlap

scanny:
Reviewer: This PR probably reviews faster commit-by-commit. Each of the commits is groomed and focuses on a separate clear aspect of this implementation.

This  PR adds inter-chunk overlap capability to chunking. It does not yet expose it via the API.

Inter-chunk overlap is overlap between whole pre-chunks, prior to any text-splitting required for oversized chunks. Contrast with intra-chunk overlap implemented in the prior PR which implements overlap on these latter text-splitting boundaries.

Inter-chunk overlap is disabled by default since a pre-chunk already has a "clean" semantic boundary (composed of whole elements) and adding overlap there introduces noise from the adjacent context. If the user wants inter-chunk overlap they must specify `overlap_all=True` in the options. Inter-chunk overlap uses the same `overlap` length value used by intra-chunk overlap and does not overlap when that value is 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.9-dev2
+## 0.11.9-dev3
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


christinestraub (approved):
LGTM

PR #2310, opened 2023-12-21
Feat: return base64 encoded images for PDF's

### LINKED ISSUE #2302: Enhancement: return base64 encoded images for PDF's
Rather than saving to file, saving as base64 encoded data of the image bytes in a new metadata field (if that is what the user specifies by some other param like `pdf_extract_to_payload`), and the mimetype for the image in yet another field. this would allow the API to have parity with the library.

Note: This issue is associated with https://github.com/Unstructured-IO/unstructured-api/issues/306

christinestraub:
Closes #2302.
### Summary
- add functionality to get a Base64 encoded string from a PIL image
- store base64 encoded image data in two metadata fields: `image_base64` and `image_mime_type`
- update the "image element filter" logic to keep all image elements in the output if a user specifies image extraction
### Testing
```

elements = partition_pdf(
    filename="example-docs/embedded-images-tables.pdf",
    strategy="hi_res",
    extract_element_types=["Image", "Table"],
    extract_to_payload=True,
)
```
or
```

elements = partition(
    filename="example-docs/embedded-images-tables.pdf",
    strategy="hi_res",
    pdf_extract_element_types=["Image", "Table"],
    pdf_extract_to_payload=True,
)
```


awalker4 (approved):
LGTM!

christinestraub:
Thanks @awalker4 :)

unstructured/partition/auto.py
```diff
@@ -193,11 +195,26 @@ def partition(
...
+        'extract_element_types' for broader extraction capabilities.
+    pdf_extract_element_types
+        Only applicable if `strategy=hi_res`.
+        Images of the element type(s) specified in this list (e.g., ["Image", "Table"]) will be
+        saved in the path specified by 'image_output_dir_path' or stored as base64 encoded data
+        within metadata fields.
+    pdf_extract_to_payload
+        Only applicable if `strategy=hi_res`.
+        If True, images of the element type(s) defined in 'extract_element_types' will be encoded
+        as base64 data and stored in two metadat fields: 'image_base64' and 'image_mime_type'.
```
Coniferish, line 211: small typo: "metadat" > "metadata"

Coniferish: LGTM otherwise!

Coniferish:
I wish I had thought of this earlier, but with so many interdependent parameters, I wonder if it would have made more sense to have `extract_to_payload` be a Tuple(str, list) instead of a list. Then the first index could be used to determine `extract_to_payload` or `image_output_dir_path` since those two parameters seem to be mutually exclusive.

Coniferish:
I know this isn't the intended use for these parameters, but `NarrativeText` isn't working as a element type for `extract_element_types`. It raises a warning: `The requested type (NarrativeText) doesn't match any available type`.

Coniferish:
Also noticing on this document that the table element isn't complete. It's element[5], but element[6] is categorized as a Title element and element[6].text is 'Corrosion rate (mm/year)', which is cut off in the image that's generated when using `image_output_dir_path`.

Coniferish:
An idea for a future enhancement, but would it make sense to have "all" be an option for `extract_element_types`?

PR #2313, opened 2023-12-21
feat: update encoders to be dataclasses

rbiseck3:
### Description
Convert all encoders to be based off dataclasses. Purpose: this will allow encoders to be used in a generic way amongst other  dataclasses. Otherwise, it'll break validation in those parent dataclasses.


ryannikolaidis (approved):
seems reasonable to me

PR #2318, opened 2023-12-26
feat: improve dataclass use for encoders

rbiseck3:
### Description
Leverage a similar pattern to what is used for connectors, where there is a nested config dataclass as a field, along with cached content for things like the client and sample embedding for each. This required an update on the embeddings config in ingest and I left a TODO in there because the current approach breaks on other encoders such as bedrock because the parameters in that config don't map to all encoders. But this keeps the existing functionality working.

This update makes sure all variables associated with the dataclass exist when it's instantiated rather than being added in the `__post_init__()` method or the `initialize()`, allowing other libraries like pydantic to appropriately generate schemas from it.  It also now follows the pattern of the connectors in that each class has a nested config class used to instantiate the client itself as well as a field/property approach used to cache the client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 * **Add intra-chunk overlap capability.** Implement overlap for split-chunks where text-splitting is used to divide an oversized chunk into two or more chunks that fit in the chunking window. Note this capability is not yet available from the API but will shortly be made accessible using a new `overlap` kwarg on partition functions.
+* **Update encoders to leverage dataclasses** All encoders now follow a class approach which get annotated with the dataclass decorator. Similar to the connectors, it uses a nested dataclass for the configs required to configure a client as well as a field/property approach to cache the client. This makes sure any variable associated with the class exists as a dataclass field.
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.7-dev1
+## 0.11.7-dev2
 
 ### Enhancements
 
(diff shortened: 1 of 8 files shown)


ryannikolaidis:
>  I left a TODO in there because the current approach breaks on other encoders such as bedrock because the parameters in that config don't map to all encoders. 

trying to follow, is the bedrock encoder broken in the existing version on main? if so, we should make sure we have an issue to track that and also should have a test that is failing / would have caught that?

ryannikolaidis:
> Leverage a similar pattern to what is used for connectors, where there is a nested config dataclass as a field

curious: just doing this for consistency with what we're doing with connectors? or is this actually solving something?

rbiseck3:
@ryannikolaidis, this was a change that was needed to fix an issue with how enterprise uses the encoders. But when looking into it, I realized that the current implementation actually only supports OpenAi and Huggingface given that the only parameters captured are model name and an api keuy. The bedrock encoder required aws credential that aren't covered by the current `EmbeddingConfig`: 
```
@dataclass
class EmbeddingConfig(BaseConfig):
    provider: str
    api_key: t.Optional[str] = enhanced_field(default=None, sensitive=True)
    model_name: t.Optional[str] = None
```

Regarding 
> Leverage a similar pattern to what is used for connectors, where there is a nested config dataclass as a field

This matches how we have the connectors set up with connector_config: 
```
@dataclass
class BaseSourceConnector(BaseConnector, ABC):
    connector_config: BaseConnectorConfig
```

rbiseck3:
Created an issue to track the bedrock encoder: https://github.com/Unstructured-IO/unstructured/issues/2319

ryannikolaidis:
> @ryannikolaidis, this was a change that was needed to fix an issue with how enterprise uses the encoders. But when looking into it, I realized that the current implementation actually only supports OpenAi and Huggingface given that the only parameters captured are model name and an api keuy. The bedrock encoder required aws credential that aren't covered by the current `EmbeddingConfig`:

yep, do you mind summarizing the issue it solves in a generic way so it's clear what this resolves for any application that similarly might leverage the code? and include that info in the changelog?

ryannikolaidis (approved):

PR #2344, opened 2024-01-03
Updated the example 'links' references to 'link_urls'. 

matthewaltenburg:
This change was necessary because the original 'links' element has been deprecated, resulting in the example not functioning properly.


MthwRobinson:
Thanks for the docs contribution! Closing this since docs have now moved to https://docs.unstructured.io

PR #2345, opened 2024-01-03
fix(html): unequal row lengths in HTMLTable.text_as_html

### LINKED ISSUE #2339: bug/html_table_extraction
**Describe the bug**
HTML partitioning performance appears to have degraded, specifically for tables.

In llama-index, we have tool to extract tables from HTML documents using unstructured. We remove any tables that don't have a consistent number of columns.

In v0.10.30, this works fine. But the same code using v0.11.X ends up filtering out all tables, since every table does not meet the filtering requirements above. To me, this feels like table extraction got much noisier?

You can see our filtering code here
https://github.com/run-llama/llama_index/blob/1736097a5902ef39fd289a645ad9765058f1c8b3/llama_index/node_parser/relational/unstructured_element.py#L99

**To Reproduce**
You can install llama-index and unstructured, and run this notebook. https://docs.llamaindex.ai/en/stable/examples/query_engine/sec_tables/tesla_10q_table.html#extract-elements

But more specifically, I've narrowed it down to loading the HTML and running `partition_html(text=text)` and seeing that the extracted tables are different.

**Expected behavior**
Uniform tables extracted.

scanny:
Fixes #2339

Fixes to HTML partitioning introduced with v0.11.0 removed the use of `tabulate` for forming the HTML placed in `HTMLTable.text_as_html`. This had several benefits, but part of `tabulate`'s behavior was to make row-length (cell-count) uniform across the rows of the table.

Lacking this prior uniformity produced a downstream problem reported in

On closer inspection, the method used to "harvest" cell-text was producing more text-nodes than there were cells and was sensitive to where whitespace was used to format the HTML. It also "moved" text to different columns in certain rows.

Refine the cell-text gathering mechanism to get exactly one text string for each row cell, eliminating whitespace formatting nodes and producing strict correspondence between the number of cells in the original HTML table row and that placed in HTML.text_as_html.

HTML tables that are uniform (every row has the same number of cells) will produce a uniform table in `.text_as_html`. Merged cells may still produce a non-uniform table in `.text_as_html` (because the source table is non-uniform).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.9-dev1
+## 0.11.9-dev2
 
 ### Enhancements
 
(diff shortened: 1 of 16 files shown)


logan-markewich:
I have tested this PR, seems to work well for my case (parsing SEC 10K files)

cragwolfe (approved):

PR #2352, opened 2024-01-04
fix: languages/ocr_languages

Coniferish:


test_unstructured/partition/pdf_image/test_pdf.py
```diff
@@ -928,6 +928,7 @@ def test_partition_hi_res_model_name_default_to_None():
         pytest.fail("partition_pdf() raised AttributeError unexpectedly!")
 
 
+# NOTE (jennings): check that the equivalent for this works for partitioing images
```
Coniferish, line 931: partition_pdf and partition_image do not have the same default value for `languages`, so I'd like to make sure this test also works for `partition_image`

PR #2357, opened 2024-01-05
feat: add Vectara destination connector 

potter-potter:
Thanks to Ofer at Vectara, we now have a Vectara destination connector.

- There are no dependencies since it is all REST calls to API
-


unstructured/ingest/cli/cmds/vectara.py
```diff
@@ -0,0 +1,59 @@
...
+                required=True,
+                type=str,
+                help="Vectara OAuth2 secret.",
+                envvar="VECTARA_OAUTH_SECRET",
+                show_envvar=True,
+            ),
+            click.Option(
+                ["--corpus-name"],
+                required=False,
+                default="vectara-unstructured",
```
rbiseck3, line 42: Rather than use an `unstructured` default, let's make this required and have no default.

unstructured/ingest/connector/vectara.py
```diff
@@ -0,0 +1,246 @@
...
+class SimpleVectaraConfig(BaseConnectorConfig):
+    access_config: VectaraAccessConfig
+    customer_id: t.AnyStr = None
+    corpus_name: t.AnyStr = "vectara-unstructured"
+    corpus_id: t.AnyStr = None
+
+
+@dataclass
+class VectaraWriteConfig(WriteConfig):
+    batch_size: int = 100
```
rbiseck3, line 39: Didn't see this exposed in the CLI options.

unstructured/ingest/connector/vectara.py
```diff
@@ -0,0 +1,246 @@
...
+@dataclass
+class VectaraAccessConfig(AccessConfig):
+    oauth_client_id: str = enhanced_field(sensitive=True)
+    oauth_secret: str = enhanced_field(sensitive=True)
+
+
+@dataclass
+class SimpleVectaraConfig(BaseConnectorConfig):
+    access_config: VectaraAccessConfig
+    customer_id: t.AnyStr = None
```
rbiseck3, line 32: If we're setting this to a default of `None`, the typing should be `t.Optional[]`. But should this be optional? Or are these required? Also why are we supporting `bytes`? Which is part of the `AnyStr` constraint.

unstructured/ingest/connector/vectara.py
```diff
@@ -0,0 +1,246 @@
...
+                data = {
+                    "corpus": {
+                        "name": self.connector_config.corpus_name,
+                    }
+                }
+                create_corpus_response = self._request(endpoint="create-corpus", data=data)
+                self.connector_config.corpus_id = create_corpus_response.get("corpusId")
+
+        except Exception as e:
+            return str(e) + "\n" + "".join(traceback.TracebackException.from_exception(e).format())
```
rbiseck3, line 87: Whys is this returning a string on error? This can reraise a new error based on the first one if needed, but otherwise won't this just be swallowed up? Also it's returning a string, but that string is never used since this gets called without an output in initialize:
```python
    def initialize(self):
        self.vectara()
```

unstructured/ingest/connector/vectara.py
```diff
@@ -0,0 +1,246 @@
...
+        response = requests.request(
+            method=http_method, url=url, headers=headers, params=params, data=json.dumps(data)
+        )
+        response.raise_for_status()
+        return response.json()
+
+    # get OAUth2 JWT token
+    def _get_jwt_token(self):
+        """Connect to the server and get a JWT token."""
+        token_endpoint = f"https://vectara-prod-{self.connector_config.customer_id}.auth.us-west-2.amazoncognito.com/oauth2/token"  # noqa: E501
```
rbiseck3, line 122: Will this endpoint always have this pattern? This looks like something we might want to expose as an option. The fact that the region is hard coded already makes me feel like it won't always be the same.

unstructured/ingest/connector/vectara.py
```diff
@@ -0,0 +1,246 @@
...
+        self,
+        endpoint: str,
+        http_method: str = "POST",
+        params: t.Mapping[str, t.Any] = None,
+        data: t.Mapping[str, t.Any] = None,
+    ):
+        url = f"{self.BASE_URL}/{endpoint}"
+
+        current_ts = datetime.datetime.now().timestamp()
+        if self.jwt_token_expires_ts - current_ts <= 60:
```
rbiseck3, line 102: Do we ever have to worry about token refresh since the connector is fairly short lived and it'll fetch a new jwt token every time it's instantiated. Also we should set a `_jwt_token` variable as part of the dataclass defaulted to None, and maybe have a `@property` that returns a jwt token, which can be responsible for fetching a new one if the instance variable is `None`. Similar to how we cache clients.

test_unstructured_ingest/dest/vectara.sh
```diff
@@ -0,0 +1,93 @@
...
+  -H 'Content-Type: application/json' \
+  -H 'Accept: application/json' \
+  -H "customer-id: $VECTARA_CUSTOMER_ID" \
+  -H "Authorization: Bearer $access_token" \
+  --data-raw "{
+  \"corpusId\": $corpus_id
+}")
+corpus_size=$(echo "$get_corpus_size" | jq -r '.size.size')
+
+if [ "$corpus_size" == "$EXPECTED_CORPUS_SIZE" ]; then
```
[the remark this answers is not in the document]
potter-potter: Great question. Its kind of a black box and only wants partitioned documents from my understanding. The rest of the process happens behind closed doors with no options. (at least that's my understanding) I will try to dive deeper and see if theres something else we can test for.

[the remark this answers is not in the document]
potter-potter: I thought list-documents would give me the document title that we uploaded. But it just gives the indexing uuid when it uploaded. Nothing constant.

unstructured/ingest/cli/cmds/vectara.py
potter-potter: yep. The default doesn't make any sense...

unstructured/ingest/connector/vectara.py
potter-potter: Ha. Yes. Leftover from elasticsearch. I'm going to ditch it.

unstructured/ingest/connector/vectara.py
potter-potter: customer_id should not be optional. corpus_name and corpus_id should be. I thought AnyStr was interesting. I saw that it included bytes. But I think we just need `str`

unstructured/ingest/connector/vectara.py
potter-potter: Yep. My mistake. Leftover. No need to be a string. Should just raise.

unstructured/ingest/connector/vectara.py
potter-potter: Option created with current as default.

unstructured/ingest/connector/vectara.py
potter-potter: Great idea. Done.

unstructured/ingest/connector/vectara.py
ofermend: Great!

ryannikolaidis (dismissed):
thanks for addressing all the feedback, this looks good to me!

potter-potter:
> checking, was there any resolution on validation of the stored and embedded results?

I can make a post request with a query. And then have it give the top example. We do get some metadata we can check for. But once the partitioned file hits Vectara, it is a black box. So no guarantee what the are doing as far as embeddings and search.

Example of response:
`{
    "responseSet": [
        {
            "response": [
                {
                    "text": "I confess all these festivities and fireworks are becoming wearisome. \"If they had known that you wished it, the entertainment would have been put off,\" said the prince, who, like a wound-up clock, by force of habit said things he did not even wish to be believed. \"Don't tease! Well, and what has been decided about Novosiltsev's dispatch? <b>You know everything.</b> \"What can one say about it?\" replied the prince in a cold, listless tone. \"What has been decided?",
                    "score": 0.75966287,
                    "metadata": [
                        {
                            "name": "url",
                            "value": "example-docs/book-war-and-peace-1p.txt"
                        },
                        {
                            "name": "filename",
                            "value": "book-war-and-peace-1p.txt"
                        },
                        {
                            "name": "filetype",
                            "value": "text/plain"
                        },
                        {
                            "name": "last_modified",
                            "value": "2024-01-18T09:39:46"
                        },
                        {
                            "name": "lang",
                            "value": "eng"
                        },
                        {
                            "name": "offset",
                            "value": "76"
                        },
                        {
                            "name": "len",
                            "value": "20"
                        }
                    ],
                    "documentIndex": 0,
                    "corpusKey": {
                        "customerId": 0,
                        "corpusId": 61,
                        "semantics": "DEFAULT",
                        "dim": [],
                        "metadataFilter": "",
                        "lexicalInterpolationConfig": null
                    },
                    "resultOffset": 344,
                    "resultLength": 20
                }
            ],
            "status": [],
            "document": [
                {
                    "id": "726c32ce-acf0-471e-88b2-a4a8b040d324",
                    "metadata": []
                }
            ],
            "summary": [],
            "futureId": 1
        }
    ],
    "status": [],
    "metrics": null
}`

ryannikolaidis:
> > checking, was there any resolution on validation of the stored and embedded results?
> 
> @ryannikolaidis I can make a post request with a text query. (but not an embedding) And then have it give the top example. We do get some metadata we can check for. But once the partitioned file hits Vectara, it is a black box. So no guarantee what the are doing as far as embeddings and search.
> 
> Example of response: `{ "responseSet": [ { "response": [ { "text": "I confess all these festivities and fireworks are becoming wearisome. \"If they had known that you wished it, the entertainment would have been put off,\" said the prince, who, like a wound-up clock, by force of habit said things he did not even wish to be believed. \"Don't tease! Well, and what has been decided about Novosiltsev's dispatch? <b>You know everything.</b> \"What can one say about it?\" replied the prince in a cold, listless tone. \"What has been decided?", "score": 0.75966287, "metadata": [ { "name": "url", "value": "example-docs/book-war-and-peace-1p.txt" }, { "name": "filename", "value": "book-war-and-peace-1p.txt" }, { "name": "filetype", "value": "text/plain" }, { "name": "last_modified", "value": "2024-01-18T09:39:46" }, { "name": "lang", "value": "eng" }, { "name": "offset", "value": "76" }, { "name": "len", "value": "20" } ], "documentIndex": 0, "corpusKey": { "customerId": 0, "corpusId": 61, "semantics": "DEFAULT", "dim": [], "metadataFilter": "", "lexicalInterpolationConfig": null }, "resultOffset": 344, "resultLength": 20 } ], "status": [], "document": [ { "id": "726c32ce-acf0-471e-88b2-a4a8b040d324", "metadata": [] } ], "summary": [], "futureId": 1 } ], "status": [], "metrics": null }`
> 
> I looked for some other endpoints that might give us embedding information. But there wasn't anything I found useful.

Ahh, I see. I think we _could_ but we would need to [build and hit the appropriate app in Vectara](https://console.vectara.com/console/build) that also leverages the data we've written.

ryannikolaidis (approved):
Okay, I think checking count is good enough for now. It would be nice as a future enhancement to actually test end to end with a test application we build in Vectara, but this should at least given us confidence that we're writing to it.

PR #2366, opened 2024-01-08
bug: omit session handler from serialization to avoid mp issues

rbiseck3:
### Description
The session handler variable can be anything, because it's specific to the SDK being used for the connector. This can break the serialization depending on what that is. To avoid this all together, the session handler itself is not serialized. Instead, it needs to be recreated if an object is serialized and then deserialized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.11.9-dev3
+## 0.11.9-dev4
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


potter-potter:
Makes sense. 

Looks like this does not include recreating the session handler. Is there a plan for that, or is it necessary at this point?

potter-potter (approved):
Looks good.

PR #2367, opened 2024-01-08
feature(chunking): add basic strategy and overlap

scanny:
This PR culminates the restructuring of chunking over my prior dozen-or-so commits by adding the new options to the API and documentation.

Separately I'll be adding a new ingest test to defend against regression, although the integration test included in this PR will do a pretty good job of that too.

### DIFF
--- a/docs/source/core/chunking.rst
+++ b/docs/source/core/chunking.rst
@@ -2,50 +2,171 @@
 Chunking
 ########
 
-Chunking functions in ``unstructured`` use metadata and document elements
-detected with ``partition`` functions to split a document into subsections
-for uses cases such as Retrieval Augmented Generation (RAG).
+Chunking functions in ``unstructured`` use metadata and document elements detected with
+``partition`` functions to split a document into smaller parts for uses cases such as Retrieval
+Augmented Generation (RAG).
+
+Chunking Basics
+---------------
+
+Chunking in ``unstructured`` differs from other chunking mechanisms you may be familiar with.
+Typical approaches start with the text extracted from the document and form chunks based on
+plain-text features, character sequences like ``"\n\n"`` or ``"\n"`` that might indicate a paragraph
+boundary or list-item boundary.
+
+Because ``unstructured`` uses specific knowledge about each document format to partition the
+document into semantic units (document elements), we only need to resort to text-splitting when a
+single element exceeds the desired maximum chunk size. Except in that case, all chunks contain one
+or more whole elements, preserving the coherence of semantic units established during partitioning.
+
+A few concepts about chunking are worth introducing before discussing the details.
+
+- Chunking is performed on *document elements*. It is a separate step performed *after*
+  partitioning, on the elements produced by partitioning. (Although it can be combined with
+  partitioning in a single step.)
+
+- In general, chunking *combines* consecutive elements to form chunks as large as possible without
+  exceeding the maximum chunk size.
+
+- A single element that by itself exceeds the maximum chunk size is divided into two or more chunks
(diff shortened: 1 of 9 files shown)


unstructured/chunking/__init__.py
```diff
@@ -25,44 +26,70 @@ def add_chunking_strategy() -> Callable[[Callable[_P, List[Element]]], Callable[
...
+            call_args = get_call_args_applying_defaults()
+
+            if call_args.get("chunking_strategy") == "by_title":
+                return chunk_by_title(
                     elements,
-                    multipage_sections=params.get("multipage_sections", True),
-                    combine_text_under_n_chars=params.get("combine_text_under_n_chars", 500),
-                    new_after_n_chars=params.get("new_after_n_chars", 500),
-                    max_characters=params.get("max_characters", 500),
+                    combine_text_under_n_chars=call_args.get("combine_text_under_n_chars", 500),
```
badGarnet, line 76: the default parameters also appear in other spots in code -> should we consider making them constants to ensure consistency in all interfaces?

unstructured/chunking/basic.py
```diff
@@ -0,0 +1,80 @@
...
+
+from __future__ import annotations
+
+from typing import List, Optional, Sequence
+
+from unstructured.chunking.base import BasePreChunker, ChunkingOptions
+from unstructured.documents.elements import Element
+
+
+def chunk_elements(
```
MthwRobinson, line 24: ❤️ how simple this function is

MthwRobinson (approved):
Docs look good from my end and love how simple the basic chunking is after the refactor.

test_unstructured/chunking/test_basic.py
```diff
@@ -0,0 +1,108 @@
+"""Unit-test suite for the `unstructured.chunking.basic` module.
+
+That module implements the baseline chunking strategy. The baseline strategy has all behaviors
+shared by all chunking strategies and no extra rules like perserve section or page boundaries.
+"""
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
scanny: This line enables the so-called _postponed annotations_ behavior.  Postponed annotations removes the need to quote forward references in most cases and provides some other typing behavior improvements. This is described in [PEP 563](https://peps.python.org/pep-0563/).

Postponed annotations was originally going to be the default in Python 3.10 but was postponed. I can't find a hard commit by the Python team on "postponed to when", but I expect we can assume it would be at least Python 3.14 since the "New in Python 3.13" page makes no mention of it.

docs/source/core/chunking.rst
```diff
@@ -2,50 +2,171 @@
...
+Chunking in unstructured differs from other chunking mechanisms you may be familiar with. Typical
+approaches start with the text extracted from the document and form chunks based on plain-text
+features, character sequences like ``"\n\n"`` or ``"\n"`` that might indicate a paragraph boundary
+or list-item boundary.
 
-The ``chunk_by_title`` function combines elements into sections by looking
-for the presence of titles. When a title is detected, a new section is created.
-Tables and non-text elements (such as page breaks or images) are always their
-own section.
+Because unstructured uses specific knowledge about each document format to partition the document
```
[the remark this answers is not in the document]
scanny: Done.

unstructured/chunking/__init__.py
scanny: I'm going to introduce the constants in a separate PR. It looks like there are several spots outside this immediate area that will need to use them and there may also be ingest-test expection-file changes required too.

PR #2369, opened 2024-01-08
chore: refactor _convert_to_standard_langcode

Coniferish:
This PR is one in a series of PRs for refactoring and fixing the `languages` parameter so it can address incorrect input by users. #2293 

This PR adds a dictionary for helping map fully spelled out languages to tesseract language codes

PR #2370, opened 2024-01-08
refactor: extract language helper functions

Coniferish:
This PR is one in a series of PRs for refactoring and fixing the `languages` parameter so it can address incorrect input by users. #2293 

Refactor `_convert_language_code_to_pytesseract_lang_code` and extract `_get_iso639_language_object` to its own function

```
from unstructured.partition.lang import _convert_language_code_to_pytesseract_lang_code as convert
convert("English") # this will raise an error on both main and this branch
convert("en") # this will return "eng" on both branches
```


shreyanid:
Generally LGTM! Could you provide test instructions? 

Also, which order should this PR and https://github.com/Unstructured-IO/unstructured/pull/2369 be in, if dependent? (could help to mention in the description)

Coniferish:
> Generally LGTM! Could you provide test instructions?
> 
> Also, which order should this PR and #2369 be in, if dependent? (could help to mention in the description)

Thanks for asking, but I'm not sure which would be better to do first (if one would be better than the other)--the order shouldn't matter since they're both branching off main. I'll rebase and resolve any conflicts if there are any.

I'll create a test for `convert_language_to_tesseract`/`_convert_language_code_to_pytesseract_lang_code` since there wasn't one before.

unstructured/partition/lang.py
```diff
@@ -200,61 +200,70 @@ def convert_old_ocr_languages_to_languages(ocr_languages: str):
...
+            matched_langcodes = _get_all_tesseract_langcodes_with_prefix(lang_iso639.part2t)
+            return TESSERACT_LANGUAGES_SPLITTER.join(matched_langcodes)
 
-    else:
-        logger.warning(f"{lang} is not a language supported by Tesseract.")
-        return ""
+        else:
+            logger.warning(f"{lang} is not a language supported by Tesseract.")
+            return ""
+    return ""
```
shreyanid, line 239: If line 220 (`if lang_iso639:`) fails it comes to this return but doesn't provide a warning as to what happened. Can we include the warning from line 237 (`logger.warning(f"{lang} is not a language supported by Tesseract.")`) or something similar for this case as well?

shreyanid (approved):

Issue #2322, opened 2023-12-28
bug/gcs connector converting json content to str (breaking auth)


rbiseck3:
For the GCS connector, it can take in either a filepath or a json string as input which gets converted into a python-native dictionary. However, because the config dataclass has the token denoted as a `str` type, that dictionary gets converted into a string during serialization so by the time the pipeline fetched the content, this causes a `gcsfs.retry.HttpError: Invalid Credentials, 401`.

PR #2371, opened 2024-01-09
fix: Ingest GCS accepts JSON auth token (#2322)

jakub-sandomierz-deepsense-ai:
FSSpec serialization caused conversion of JSON token to string with single quotes. GCS requires JSON token in form of dict so this format is now assured. Other forms of auth are not modified but there is improved validation for all of the options.


unstructured/ingest/connector/fsspec/gcs.py
```diff
@@ -21,6 +23,30 @@ class GcsAccessConfig(AccessConfig):
...
+        ALLOWED_AUTH_VALUES = "google_default", "cache", "anon", "browser", "cloud"
+        # Case: null value
+        if not self.token:
+            return
+        # Case: one of auth constants
+        if self.token in ALLOWED_AUTH_VALUES:
+            return
+        # Case: token as json
+        try:
+            str_token = self.token.replace("'", '"')
```
rbiseck3, line 36: Why is this replace required? Just curious.

jakub-sandomierz-deepsense-ai: I have found that dict is converted to string with single quotes for string type of keys and values. This leads to inability to convert back to JSON.

rbiseck3 (approved):

PR #2372, opened 2024-01-09
drop python3.8

rbiseck3:
### Description
Remove all uses of python3.8


MthwRobinson (approved):
Good to go once the 3.9 build is working

PR #2378, opened 2024-01-10
bug: weaviate serialization broken

rbiseck3:
### Description
This PR handles two things:
* Fixes the serialization of the weaviate destination connector since the client content breaks serialization when present due to `TypeError: cannot pickle '_thread.lock' object`.
* Set finer auth control rather than generic dictionary on the CLI and access config.


ryannikolaidis (approved):
lgtm!

PR #2380, opened 2024-01-10
feat(ingest): add basic chunking to ingest

scanny:
The new "basic" chunking strategy and overlap options need to be available from the ingest CLI. An ingest test of those features is also welcome, both to verify the ingest feature and to defend against regressions in the chunking code.

Add a local ingest test exercising both the "basic" chunking strategy and intra-chunk overlap. Since there is no new source connector involved, use the local ingest source and destination. Update documentation to suit, filling in some details that hadn't made it into the docs yet.


ryannikolaidis (approved):
perfect, thanks for the quick change!

PR #2381, opened 2024-01-10
fix: ensure consistency in method signatures across destination connectors

rbiseck3:
### Description
* Make sure all destination connectors implement the base abstract methods using the same signatures. 
* Also leverage conform dict in the base methods to make sure it's called in a consistent fashion.
* Additional updates to move the common code into the base destination connector class

### DIFF
--- a/test_unstructured_ingest/dest/azure-cognitive-search.sh
+++ b/test_unstructured_ingest/dest/azure-cognitive-search.sh
@@ -11,7 +11,7 @@ OUTPUT_DIR=$OUTPUT_ROOT/structured-output/$OUTPUT_FOLDER_NAME
 WORK_DIR=$OUTPUT_ROOT/workdir/$OUTPUT_FOLDER_NAME
 max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
 
-AZURE_SEARCH_ENDPOINT=${AZURE_SEARCH_ENDPOINT:-"https://ingest-test-azure-cognitive-search.search.windows.net"}
+AZURE_SEARCH_ENDPOINT="https://ingest-test-azure-cognitive-search.search.windows.net"
 
 random_id=$(uuidgen)
 # index name must be all lowercase
--- a/test_unstructured_ingest/dest/azure-cognitive-search.sh
+++ b/test_unstructured_ingest/dest/azure-cognitive-search.sh
@@ -11,8 +11,6 @@ OUTPUT_DIR=$OUTPUT_ROOT/structured-output/$OUTPUT_FOLDER_NAME
 WORK_DIR=$OUTPUT_ROOT/workdir/$OUTPUT_FOLDER_NAME
 max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
 
-AZURE_SEARCH_ENDPOINT="https://ingest-test-azure-cognitive-search.search.windows.net"
-
 random_id=$(uuidgen)
 # index name must be all lowercase
 random_id=$(echo "$random_id" | tr '[:upper:]' '[:lower:]')
@@ -21,8 +19,8 @@ DESTINATION_INDEX="utic-test-ingest-fixtures-output-$random_id"
 # 2023-07-01-Preview, 2021-04-30-Preview, 2020-06-30-Preview
 API_VERSION=2023-07-01-Preview
 
-if [ -z "$AZURE_SEARCH_API_KEY" ]; then
-  echo "Skipping Azure Cognitive Search ingest test because AZURE_SEARCH_API_KEY env var is not set."
+if [ -z "$AZURE_SEARCH_API_KEY" ] || [ -z "$AZURE_SEARCH_ENDPOINT" ]; then
+  echo "Skipping Azure Cognitive Search ingest test because AZURE_SEARCH_API_KEY or AZURE_SEARCH_ENDPOINT env var is not set."
   exit 8
 fi
 # shellcheck disable=SC1091
--- a/test_unstructured_ingest/dest/azure-cognitive-search.sh
+++ b/test_unstructured_ingest/dest/azure-cognitive-search.sh
@@ -11,6 +11,8 @@ OUTPUT_DIR=$OUTPUT_ROOT/structured-output/$OUTPUT_FOLDER_NAME
 WORK_DIR=$OUTPUT_ROOT/workdir/$OUTPUT_FOLDER_NAME
 max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
 
+AZURE_SEARCH_ENDPOINT=${AZURE_SEARCH_ENDPOINT:-"https://ingest-test-azure-cognitive-search.search.windows.net"}
+
 random_id=$(uuidgen)
 # index name must be all lowercase
 random_id=$(echo "$random_id" | tr '[:upper:]' '[:lower:]')
--- a/test_unstructured_ingest/dest/azure-cognitive-search.sh
+++ b/test_unstructured_ingest/dest/azure-cognitive-search.sh
@@ -25,18 +25,21 @@ if [ -z "$AZURE_SEARCH_API_KEY" ] || [ -z "$AZURE_SEARCH_ENDPOINT" ]; then
   echo "Skipping Azure Cognitive Search ingest test because AZURE_SEARCH_API_KEY or AZURE_SEARCH_ENDPOINT env var is not set."
   exit 8
 fi
+
+endpoint="$AZURE_SEARCH_ENDPOINT/indexes/$DESTINATION_INDEX?api-version=$API_VERSION"
+echo "Connecting to endpoint: $endpoint"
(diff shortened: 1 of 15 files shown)


rbiseck3:
@ahmetmeleq I believe this should handle the issue you brought up with `conform_dict()`

ryannikolaidis:
I think you're going to hit issues with the unit tests leveraging assignment like [here](https://github.com/Unstructured-IO/unstructured/blob/e2f0de3c5049850be191e140cdc28af811ce907e/test_unstructured_ingest/unit/test_sql_conform_dict.py#L72). should be quick bump

ryannikolaidis:
bump PR title (for conventional) and description to include some more info about the what and why? also maybe include a note in the changelog?

ryannikolaidis:
not sure if this is in scope for this PR, but calling `write_elements()` rather than `write()` will still hit issues with some of these connectors, right? i.e. with Pinecone we [do formatting ](https://github.com/Unstructured-IO/unstructured/blob/e2f0de3c5049850be191e140cdc28af811ce907e/unstructured/ingest/connector/pinecone.py#L111-L139)directly in `write()`

unstructured/ingest/connector/azure_cognitive_search.py
```diff
@@ -150,8 +150,6 @@ def write(self, docs: t.List[BaseSingleIngestDoc]) -> None:
             local_path = doc._output_filename
             with open(local_path) as json_file:
                 json_content = json.load(json_file)
-                for content in json_content:
-                    self.conform_dict(data=content)
```
ahmetmeleq, line 154: How will ingest conform the dicts now since we remove it from write methods?

ahmetmeleq:
> not sure if this is in scope for this PR, but calling `write_elements()` rather than `write()` will still hit issues with some of these connectors, right? i.e. with Pinecone we [do formatting ](https://github.com/Unstructured-IO/unstructured/blob/e2f0de3c5049850be191e140cdc28af811ce907e/unstructured/ingest/connector/pinecone.py#L111-L139)directly in `write()`

I think this PR should: 
1. isolate all formatting logic from `write` methods
2. call the isolated formatting logic both in `write` , and `write_elements` to support both how ingest and other downstream applications

test_unstructured_ingest/dest/azure-cognitive-search.sh
```diff
@@ -21,8 +19,8 @@ DESTINATION_INDEX="utic-test-ingest-fixtures-output-$random_id"
 # 2023-07-01-Preview, 2021-04-30-Preview, 2020-06-30-Preview
 API_VERSION=2023-07-01-Preview
 
-if [ -z "$AZURE_SEARCH_API_KEY" ]; then
-  echo "Skipping Azure Cognitive Search ingest test because AZURE_SEARCH_API_KEY env var is not set."
+if [ -z "$AZURE_SEARCH_API_KEY" ] || [ -z "$AZURE_SEARCH_ENDPOINT" ]; then
```
ahmetmeleq, line 22: Why is this changed?

unstructured/ingest/connector/azure_cognitive_search.py
rbiseck3: This was moved to the base classes in the destination connector, meaning whenever you call `write()` or `write_elements` not, it'll call the `conform_dict` method on your behalf.

test_unstructured_ingest/dest/azure-cognitive-search.sh
rbiseck3: Not sure how this was passing in CI, but that endpoint was a configurable value. It broke for me locally without this change.

unstructured/ingest/interfaces.py
```diff
@@ -654,22 +654,34 @@ def __init__(self, write_config: WriteConfig, connector_config: BaseConnectorCon
...
+
+    @abstractmethod
+    def get_elements_dict(self, docs: t.List[BaseSingleIngestDoc]) -> t.List[t.Dict[str, t.Any]]:
         pass
 
     @abstractmethod
     def write_dict(self, *args, elements_dict: t.List[t.Dict[str, t.Any]], **kwargs) -> None:
         pass
 
+    def write_raw_dict(self, *args, elements_dict: t.List[t.Dict[str, t.Any]], **kwargs) -> None:
```
ahmetmeleq, line 677: I'd rather name this as `write_conformed_dict`, raw sounds as not processed / formatted

unstructured/ingest/connector/pinecone.py
```diff
@@ -87,26 +87,28 @@ def upsert_batch(self, batch):
...
-                pool.map(self.upsert_batch, list(chunk_generator(dict_list, pinecone_batch_size)))
+                pool.map(
+                    self.upsert_batch, list(chunk_generator(elements_dict, pinecone_batch_size))
+                )
 
-    def write(self, docs: t.List[BaseIngestDoc]) -> None:
+    def get_elements_dict(self, docs: t.List[BaseSingleIngestDoc]) -> t.List[t.Dict[str, t.Any]]:
         dict_list: t.List[t.Dict[str, t.Any]] = []
         for doc in docs:
             local_path = doc._output_filename
```
ahmetmeleq: Lines 119-138 are still not being called by `write_elements`, and they are essential while writing content. We need to isolate those parts as well. Both `write` and `write_elements` should work

rbiseck3:
@ahmetmeleq, look at the latest changes. I created two base classes in the `BaseDestinationConnector` class:
* `conform_dict`: Changes that are needed in place
* `normalize_dict`: map to a new dict to accommodate the input needed by the destination 

These are then called in the base `write_raw_dict` method which gets called by the `write` and `write_elements` methods.

test_unstructured_ingest/dest/azure-cognitive-search.sh
ahmetmeleq: https://github.com/Unstructured-IO/unstructured/blob/a8a103bc5c378900c3c17bffde26a3b41287ff89/test_unstructured_ingest/dest/azure-cognitive-search.sh#L14C1-L14C22

The variable was defined at the top (being removed here) and it is not a secret value. We've made a similar discussion in a past PR, we don't want to bloat CI values with non secret variables

ahmetmeleq: Also, the azure cognitive search credentials has recently been modified. If you have a local credential which is not updated, that might be the reason why it failed in local

ahmetmeleq:
The design looks good 👍 , waiting for the tests - also two comments on naming and azure cognitive search test script

test_unstructured_ingest/dest/azure-cognitive-search.sh
rbiseck3: I assumed because we had that value in LastPass, it was being used in CI.

ryannikolaidis: bump - can we remove? there should never be a case where this isn't defined given L14 right?

unstructured/ingest/interfaces.py
ryannikolaidis: @rbiseck3 I do see your point, but I think I agree with @ahmetmeleq here. at a glance I immediately know what to expect from `normalize_and_write_dict` whereas I might not have assumed `write_raw_dict` does this without looking at the code.

ryannikolaidis: ahh, looks like you just bumped

unstructured/ingest/interfaces.py
```diff
@@ -654,22 +655,54 @@ def __init__(self, write_config: WriteConfig, connector_config: BaseConnectorCon
...
+                dict_list.extend(dict_content)
+        return dict_list
 
     @abstractmethod
     def write_dict(self, *args, elements_dict: t.List[t.Dict[str, t.Any]], **kwargs) -> None:
         pass
 
+    def modify_and_write_dict(
+        self, *args, elements_dict: t.List[t.Dict[str, t.Any]], **kwargs
+    ) -> None:
```
ryannikolaidis, line 697: nit: maybe just a quick line docstring... modify=conform+normalize and why it's needed.

test_unstructured_ingest/dest/azure-cognitive-search.sh
rbiseck3: removed

ahmetmeleq: Agreed that it shouldn't exist in Lastpass 👍 I tried updating but having permission issues

image

ahmetmeleq: Updated now 👍

potter-potter:
This is great! `conform_dict` is something we've been wanting to address. I like the approach.

ahmetmeleq:
Looks all good now, waiting for tests to approve

ryannikolaidis (approved):
lgtm! thanks for the clean up here.

ahmetmeleq (approved):
Destination tests passed as well; LGTM, thanks for everything here.

PR #2382, opened 2024-01-10
feat: keep all image elements when using `hi_res` strategy.

christinestraub:
### Summary
The goal of this PR is to keep all image elements when using "hi_res" strategy. Previously, `Image` elements with small chunks of text were ignored unless the image block extraction parameters (`extract_images_in_pdf` or `extract_image_block_types`) were specified. Now, all image elements are kept regardless of whether the image block extraction parameters are specified.

### Testing
- on `main` branch,
```
elements = partition_pdf(
    filename="example-docs/embedded-images.pdf",
    strategy="hi_res",
)
image_elements = [el for el in elements if el.category == ElementType.IMAGE]
print("number of image elements: ", len(image_elements))
```
The above code will display `number of image elements: 0`. 

- on this `feature` branch,

The same code will display `number of image elements: 3`


Coniferish:
@christinestraub Looks like it's doing what's intended! 

Just curious about why we've decided to start keeping the "junk" text as noted by @cragwolfe in the code, though.
Also, wondering why we are keeping elements that have no text.

Also, maybe add the testing snippet to `test_pdf` with "example-docs/embedded-images.pdf" showing `len(image_elements) > 0`?

christinestraub:
> Just curious about why we've decided to start keeping the "junk" text as noted by @cragwolfe in the code, though. Also, wondering why we are keeping elements that have no text.

@Coniferish 
1. We now support extracting image blocks (`images` and `tables`).
2. "we should probably keep image elements. i think we still have the issue that tesseract try to come up with something and output garbage, but we should deal with that in a better way than just ignoring the image all together" mentioned by @cragwolfe

Coniferish (approved):
LGTM

christinestraub:
Also, maybe add the testing snippet to test_pdf with "example-docs/embedded-images.pdf" showing len(image_elements) > 0?

Added a testing snippet. Thanks @Coniferish :)

cragwolfe (approved):

PR #2386, opened 2024-01-11
Docs various updates

ron-unstructured:
To test:
> cd docs && make html 

Changelogs:
* Added verbiage about the cap limit and data usage for the Freemium AP
* Added deprecated warning on Staging bricks
* Added warning and code examples to use the SaaS API Endpoints using CLI-vs-SDKs
* Fixed example page formatting
* Added deprecation warning on ``model_name`` param in favor of ``hi_res_model_name``
* Added ``extract_images_in_pdf`` usage and code example in ``partition_pdf`` section
* Reorganized and improved the documentation Intro section


docs/source/api.rst
```diff
@@ -1,15 +1,35 @@
-Unstructured API
-#################
+Unstructured API Services
+#########################
 
-Try our hosted API! You can get your API key `here <https://unstructured.io/api-key>`__ now and start using it today.
+The Unstructured API presents an extensive selection of API services, encompassing free access, Software-as-a-Service (SaaS), and Azure & AWS Marketplace integrations. These API services offer a high-quality and scalable solution capable of ingesting and digesting files of various types and sizes.
```
MthwRobinson, line 4: `integrations` -> `applications`

docs/source/api.rst
```diff
@@ -1,15 +1,35 @@
...
 
 
+Changes to the Free API Service
+*******************************
+
+Implementation of a 1000-Page Cap
+---------------------------------
+Starting from January 10th, 2024, our free API service includes a cap of 1,000 pages. This limitation is introduced to ensure the sustainability of our free offerings and to enhance service quality.
+
+.. warning::
```
MthwRobinson, line 19: Also add that documents are never stored on the paid API, and that users can switch to that if they don't want documents stored.

MthwRobinson: I know you allude to that in the section below, but just want to make sure we're double tapping that

docs/source/api.rst
```diff
@@ -1,15 +1,35 @@
...
+    Documents submitted to the free API may be utilized for our proprietary model training and evaluation purposes. This is part of our continuous effort to improve the API's performance and capabilities.
+
+Premium API Services
+--------------------
+
+For advanced users requiring privacy and scalable API services, we provide several alternatives:
```
MthwRobinson, line 25: Drop `advanced`

docs/source/apis/saas_api.rst
```diff
@@ -58,13 +58,19 @@ Using the API Key and URL
...
 
-Example: Parsing a Document
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Example: Parsing a Document via SaaS API
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Please refer to the sample code and output provided below to begin processing documents with your SaaS API.
 
+.. warning::
+    When using the Command Line Interface (CLI), ensure that you copy the complete SaaS API URL provided in your email. This URL should **do include** the specific path ``/general/v0/general``.
```
MthwRobinson, line 67: `do include` -> `include`

docs/source/best_practices/table_extraction_pdf.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. code-block:: python
+
+    from unstructured.partition.auto import partition
+
+    filename = "example-docs/layout-parser-paper.pdf"
+
+    elements = partition(filename=filename,
+                         skip_infer_table_types=[],
+                         pdf_infer_table_structure=True,
```
MthwRobinson, line 51: Do you need `skip_infer_table_types` and `pdf_infer_table_structured`? I thought `skip_infer_table_types` superseded `pdf_infer_table_types`.

docs/source/best_practices/table_extraction_pdf.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. code-block:: bash
+
+     curl -X 'POST' \
+      'https://api.unstructured.io/general/v0/general' \
+      -H 'accept: application/json' \
+      -H 'Content-Type: multipart/form-data' \
+      -F 'files=@sample-docs/layout-parser-paper.pdf' \
+      -F 'strategy=hi_res' \
+      -F 'pdf_infer_table_structure=true' \
```
MthwRobinson, line 71: @awalker4 - Is `pdf_infer_table_types` or `skip_infer_table_types` the preferred way of doing this now?

MthwRobinson: I was thinking `skip_infer_table_types` had replaced `pdf_infer_table_types`

MthwRobinson (changes requested):
Nice doc updates! Have a few questions on parameters, once those are sorted this should be good to go.

docs/source/best_practices/table_extraction_pdf.rst
ron-unstructured: It seems both params must be defined when using `auto` partition to extract tables. This is because the `skip_infer_table_types` includes `pdf` by default. If the `pdf_infer_table_structured` is enabled, without removing `pdf` in `skip_infer_table_types`, it'll trigger a warning.
Ref: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py#L564

ron-unstructured: Clarification from @yuming-long 
- `skip_infer_table_types` won't supersede `pdf_infer_table_structure` (otherwise it is a bug)
- user should set `skip_infer_table_types` to a list without "pdf" for example ["jpg"] or []
- if the "pdf" in `skip_infer_table_types` is removed, and the user doesn't want to do table extraction for pdf, then do not set `pdf_infer_table_structure` to "True"; Otherwise, we will do table extraction since we still use both parameters to decide table extraction, so: (i)if the "pdf" in `skip_infer_table_types` is removed, `pdf_infer_table_structure` to "True". do table extraction, give a warning. (ii) if the "pdf" in `skip_infer_table_types` is removed, `pdf_infer_table_structure` to "False", no table extraction (and of no warning)
There is no need to define both, but just if there is a conflict between the two, give a warning.

MthwRobinson: We should have both of those parameters in the docs, that'll get confusing. Keep `skip_infer_table_types` and drop references to `pdf_infer_table_types`.

MthwRobinson: @ron-unstructured - I'll create a follow on issue to deprecate one of those parameters. As far as this docs PR goes, we're good to go here.

yuming-long: i tested this and got table output
```

elements = partition("layout-parser-paper.pdf", strategy='hi_res',
    skip_infer_table_types=[],
)

print([el.metadata.text_as_html for el in elements if el.metadata.text_as_html])
```
i think only set `skip_infer_table_types` should do the trick for pdf

MthwRobinson (approved):
LGTM!

PR #2390, opened 2024-01-12
feat: allows for reorientation of pdf pages when strategy=HI_RES

MillCheck:
Some of the PDFs, especially the scanner-produced ones, contain pages that are oriented wrong. To combat this, I'm introducing a very crude step of detecting the orientation and rotating the pages.
The solution I used (creating a TemporaryFile and re-running partition) is very hacky, but avoids a complete refactor, which would likely be the other option.

### DIFF
--- a/unstructured/partition/pdf.py
+++ b/unstructured/partition/pdf.py
@@ -4,7 +4,6 @@ import contextlib
 import io
 import os
 import re
-import tempfile
 import warnings
 from tempfile import SpooledTemporaryFile
 from typing import (
@@ -267,7 +266,6 @@ def _partition_pdf_or_image_local(
     extract_image_block_to_payload: bool = False,
     analysis: bool = False,
     analyzed_image_output_dir_path: Optional[str] = None,
-    hi_res_reorientation_strategy: str = "",
     **kwargs,
 ) -> List[Element]:
     """Partition using package installed locally"""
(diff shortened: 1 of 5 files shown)


MillCheck:
The current test fails on exactly one letter - it OCRs into '!' when reading the original file, it OCRs into "l" when reading the same file with the contents rotated 180 and the page itself rotated 180 (for a full 360 of the contents). All the other characters are the same.

Test passes on a local, non-Docker envornment. This is highly unusual, investingating further.

christinestraub:
@MillCheck Can you please add an example pdf document with rotated pages?

unstructured/partition/pdf.py
```diff
@@ -292,6 +285,36 @@ def _partition_pdf_or_image_local(
...
+                extract_images_in_pdf=extract_images_in_pdf,
+                extract_image_block_types=extract_image_block_types,
+                extract_image_block_output_dir=extract_image_block_output_dir,
+                extract_image_block_to_payload=extract_image_block_to_payload,
+                analysis=analysis,
+                analyzed_image_output_dir_path=analyzed_image_output_dir_path,
+                hi_res_reorientation_strategy="",  # already reoriented
+                **kwargs,
+            )
+
```
christinestraub, line 317: Can we do this within `partition_pdf_or_image()` after line [536](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L536)?

christinestraub (changes requested):

MillCheck:
> @MillCheck Can you please add an example pdf document with rotated pages?

I could, but the unit tests right now do it by generating one dynamically - I thought it sufficient, as I don't believe a rotated PDF is that useful otherwise.

unstructured/partition/pdf.py
MillCheck: I need clarification - line 536 might now refer to a different line than at the time of writing the above comment.

christinestraub: It now refers to line [528](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L528) , before calling the `_partition_pdf_or_image_local()` within `partition_pdf_or_image()`

MillCheck: Newest commit moves the logic to partition_pdf_or_image and removes the reorientation_strategy argument from _partition_pdf_or_image_local outright.

I'm not sure I like it more from an in-code perspective, but the data flow itself indeed feels more elegant now.

PR #2393, opened 2024-01-12
feat: add MongoDB source connector

potter-potter:
Adds MongoDB as a source (we already had it as a destination connector)


examples/ingest/mongodb/destination.sh
```diff
@@ -0,0 +1,35 @@
...
+  local \
+  --input-path example-docs/book-war-and-peace-1225p.txt \
+  --output-dir local-to-elasticsearch \
+  --strategy fast \
+  --chunk-elements \
+  --embedding-provider "<an unstructured embedding provider, ie. langchain-huggingface>" \
+  --num-processes 2 \
+  --verbose \
+  --work-dir "<directory for intermediate outputs to be saved>" \
+  elasticsearch \
```
ryannikolaidis, line 23: why is this example about ES destination and not MongoDB?

examples/ingest/mongodb/source.sh
```diff
@@ -0,0 +1,30 @@
...
+  --password "<Password to authenticate into the index>" \
+  --fields "<If provided, will limit the fields returned by Elasticsearch to this comma-delimited list" \
+  --batch-size "<How many records to read at a time per process>" \
+  --num-processes "<Number of processes to be used to upload, ie. 2>" \
+  --cloud-id "<Id used to connect to Elastic Cloud>" \
+  --es-api-key "<Api key used for authentication>" \
+  --api-key-id "<Id associated with api key used for authentication: https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html>" \
+  --bearer-auth "<Bearer token used for HTTP bearer authentication>" \
+  --ca-certs "<path/to/ca/certs>" \
+  --ssl-assert-fingerprint "<SHA256 fingerprint value>"
```
ryannikolaidis, line 30: why is this example about ES and not MongoDB?

ryannikolaidis:
I think you need to update[ list of sources in docs ](https://github.com/Unstructured-IO/unstructured/blob/f1b8ceff30a16e21d93403b1504571fea092583a/docs/source/ingest/source_connectors.rst#L1)to actually have the mongodb source doc available in the docs

unstructured/ingest/cli/cmds/mongodb.py
```diff
@@ -25,32 +27,36 @@ def get_cli_options() -> t.List[click.Option]:
...
             ),
             click.Option(
                 ["--collection"], required=True, type=str, help="collection name to connect to"
             ),
+            click.Option(
+                ["--batch-size"],
+                default=100,
+                type=click.IntRange(0),
+                help="how many records to read at a time per process",
+            ),
```
ryannikolaidis, line 41: this option only applies to the source connector, right? ...in which case this probably belongs somewhere other than MongoDBCliConfig otherwise both the source and destination connectors have this option?

unstructured/ingest/connector/salesforce.py
```diff
@@ -184,10 +185,12 @@ def _get_response(self):
...
         response = self._get_response()
-        logger.debug(f"response from salesforce record request: {response}")
+        logger.debug(f"response from salesforce record request: {json.dumps(response)}")
         records = response["records"]
         if not records:
-            raise ValueError(f"No record found with record id {self.record_id}: {response}")
+            raise ValueError(
+                f"No record found with record id {self.record_id}: {json.dumps(response)}"
+            )
         record_json = records[0]
```
ryannikolaidis, line 194: do we really want to be logging these records at all? ...that aside, they could have sensitive data in them. probably best to just log the record id?

unstructured/ingest/connector/registry.py
```diff
@@ -22,6 +22,7 @@
 from unstructured.ingest.connector.hubspot import HubSpotIngestDoc
 from unstructured.ingest.connector.jira import JiraIngestDoc
 from unstructured.ingest.connector.local import LocalIngestDoc
+from unstructured.ingest.connector.mongodb import MongoDBIngestDoc, MongoDBIngestDocBatch
```
ryannikolaidis, line 25: not a big deal, but do we actually need MongoDBIngestDoc in the registry? it's only ever used internally by MongoDBIngestDocBatch, right? ...i.e. it won't need to be serialized/deserialized?

unstructured/ingest/connector/salesforce.py
potter-potter: @ryannikolaidis I was thinking about that. But it is Debug. I will change to just record_id.

unstructured/ingest/connector/mongodb.py
```diff
@@ -78,39 +88,203 @@ def to_dict(
...
+            with open(filename, "w", encoding="utf8") as f:
+                f.write(concatenated_values)
+
+            self.ingest_docs.append(ingest_doc)
+
+
+@dataclass
+class MongoDBSourceConnector(SourceConnectorCleanupMixin, BaseSourceConnector):
+    connector_config: SimpleMongoDBConfig
     _client: t.Optional["MongoClient"] = field(init=False, default=None)
```
ryannikolaidis, line 230: please define a to_dict method and in it remove this _client since it won't be serializable. [Here](https://github.com/Unstructured-IO/unstructured/pull/2399)is a reference PR to where this has been done for another connector

unstructured/ingest/connector/mongodb.py
```diff
@@ -78,39 +88,203 @@ def to_dict(
...
+class MongoDBDestinationConnector(BaseDestinationConnector):
+    connector_config: SimpleMongoDBConfig
+    _client: t.Optional["MongoClient"] = field(init=False, default=None)
 
     @property
     def client(self) -> "MongoClient":
         if self._client is None:
-            self._client = self.generate_client()
+            self._client = self.connector_config.generate_client()
         return self._client
```
ryannikolaidis, line 289: while in here, please define a to_dict method and in it remove this _client since it won't be serializable. Hereis a reference PR to where this has been done for another connector

unstructured/ingest/cli/cmds/mongodb.py
potter-potter: Yep. moving it to a ReadConfig

unstructured/ingest/connector/registry.py
potter-potter: Seems as though it needs both for:

`create_ingest_doc_from_dict`

unstructured/ingest/connector/mongodb.py
potter-potter: Added the to_dict. Didn't remove _client.

ryannikolaidis: You need to remove the client in the to_dict method. See that pr for reference

potter-potter: @ryannikolaidis Sorry. Maybe I didn't communicate well. The elasticsearch version sets the `_client` attribute to None. I did exactly the same thing. Is that what you mean by remove? Or possibly I'm not understanding.

test_unstructured_ingest/src/mongodb.sh
```diff
@@ -0,0 +1,40 @@
...
+OUTPUT_FOLDER_NAME=mongodb
+OUTPUT_ROOT=${OUTPUT_ROOT:-$SCRIPT_DIR}
+OUTPUT_DIR=$OUTPUT_ROOT/structured-output/$OUTPUT_FOLDER_NAME
+WORK_DIR=$OUTPUT_ROOT/workdir/$OUTPUT_FOLDER_NAME
+DOWNLOAD_DIR=$SCRIPT_DIR/download/$OUTPUT_FOLDER_NAME
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+SOURCE_MONGO_COLLECTION="sample-mongodb-data"
+CI=${CI:-"false"}
+
+if [ -z "$MONGODB_URI" ] && [ -z "$MONGODB_DATABASE_NAME" ]; then
```
rbiseck3, line 18: Can this be tested via docker?

unstructured/ingest/connector/mongodb.py
```diff
@@ -78,39 +90,209 @@ def to_dict(
...
+                    date_created=doc.get("_id").generation_time.isoformat(),
+                ),
+                document=doc,
+            )
+            ingest_doc.update_source_metadata()
+            del doc["_id"]
+            filename = ingest_doc.filename
+            flattened_dict = flatten_dict(dictionary=doc)
+            str_values = [str(value) for value in flattened_dict.values()]
+            concatenated_values = "\n".join(str_values)
```
rbiseck3, line 213: Not sure if a doc was ever started but would be good to remembed to bucket this in with elasticsearch as ewe explore alternatives on how to represent the json data without loosing some of the structure.

unstructured/ingest/connector/mongodb.py
```diff
@@ -78,39 +90,209 @@ def to_dict(
...
+    ingest_docs: t.List[MongoDBIngestDoc] = field(default_factory=list)
+    list_of_ids: t.List[str] = field(default_factory=list)
+    registry_name: str = "mongodb_batch"
+
+    @property
+    def unique_id(self) -> str:
+        return ",".join(sorted(self.list_of_ids))
+
+    @requires_dependencies(["pymongo"], extras="mongodb")
+    def _get_docs(self):
```
rbiseck3, line 180: Can we add a return type here to make it explicit what sort of data this method is giving back.

unstructured/ingest/connector/mongodb.py
```diff
@@ -78,39 +90,209 @@ def to_dict(
...
+    def initialize(self):
+        _ = self.client
+
     @requires_dependencies(["pymongo"], extras="mongodb")
-    def generate_client(self) -> "MongoClient":
-        from pymongo import MongoClient
-        from pymongo.server_api import ServerApi
+    def _get_doc_ids(self):
+        """Fetches all document ids in a collection."""
+        collection = self.connector_config.get_collection(self.client)
```
rbiseck3, line 246: Just making sure, this only brings in the ids and not the whole record?

test_unstructured_ingest/src/mongodb.sh
potter-potter: @rbiseck3 Definitely. But the dest connector was with the hosted MongoDB. How do we decide which to use a Docker or the hosted?

unstructured/ingest/connector/mongodb.py
potter-potter: Yep. Just the ids. I'll add a return type.

unstructured/ingest/connector/mongodb.py
ryannikolaidis: @potter-potter yep, that's exactly what I meant. thanks!

unstructured/ingest/connector/registry.py
ryannikolaidis: ahh, forgot that we now serialize between each stage. yep, makes sense.

test_unstructured_ingest/src/mongodb.sh
ryannikolaidis: cc @rbiseck3 @potter-potter

ryannikolaidis (approved):
lgtm!

PR #2394, opened 2024-01-12
fix: pinecone serialization issue

ryannikolaidis:
This fixes the serialization of the Pinecone destination connector. Presence of the PineconeIndex object breaks serialization due to TypeError: cannot pickle '_thread.lock' object. This removes that object before serialization.

### DIFF
--- a/unstructured/ingest/connector/pinecone.py
+++ b/unstructured/ingest/connector/pinecone.py
@@ -1,3 +1,4 @@
+import copy
 import json
 import multiprocessing as mp
 import typing as t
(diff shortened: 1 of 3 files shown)


unstructured/ingest/connector/pinecone.py
```diff
@@ -47,6 +49,18 @@ class PineconeDestinationConnector(IngestDocSessionHandleMixin, BaseDestinationC
     connector_config: SimplePineconeConfig
     _index: t.Optional["PineconeIndex"] = None
 
+    def to_dict(self, **kwargs):
+        """
+        The _index variable in this dataclass breaks deepcopy due to:
```
jasonbot, line 54: Oof! This is why I hate pickle

jasonbot (approved):

unstructured/ingest/connector/pinecone.py
ryannikolaidis: Hard same!

PR #2396, opened 2024-01-13
Fix the table of contents warning message on transforming html into a…

cmscmadd:
There's a warning message on this page in the documentation: https://unstructured-io.github.io/unstructured/examples/chroma.html

This is the patch as advised in the chroma doc to remove the error message if you're inserting a TOC at the top.


MthwRobinson:
Thanks for the docs contribution! Closing since docs have moved to https://docs.unstructured.io

PR #2397, opened 2024-01-14
test: update test Elasticsearch mappings to validate embedding search

ryannikolaidis:
Currently in the Elasticsearch Destination ingest test we are writing the embeddings to a "float" type field. In order to leverage this field for similarity search it should be mapped as "dense_vector" with the respective dimensions assigned.

This PR updates that mapping and adds a test query to validate that this works as expected.

### DIFF
--- a/scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
+++ b/scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
@@ -10,6 +10,7 @@ from es_cluster_config import (
     PASSWORD,
     USER,
 )
+
 from unstructured.embed.huggingface import HuggingFaceEmbeddingConfig, HuggingFaceEmbeddingEncoder
 
 N_ELEMENTS = 1404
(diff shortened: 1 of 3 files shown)


scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
```diff
@@ -28,3 +56,16 @@
...
+    # Verify that the top 1 result matches the expected chunk by checking the start text
+    print("Testing query to the embedded index.")
+    query_text = (
+        "A gathering of Russian nobility and merchants in historic uniforms, "
+        "discussing the Emperor's manifesto with a mix of solemn anticipation "
+        "and everyday concerns, while Pierre, dressed in a tight nobleman's uniform, "
+        "ponders the French Revolution and social contracts amidst the crowd."
+    )
+    query_response = query(client, query_text)
+    assert query_response["hits"]["hits"][0]["_source"]["text"].startswith("CHAPTER XXII")
```
ryannikolaidis, line 70: query text is designed to semantically match with the beginning text of that chapter

jasonbot (approved):

PR #2398, opened 2024-01-14
chore: make Elasticsearch Destination connector write settings optional

ryannikolaidis:
* set required=False to all write config options
* update num_processes to default to 1 since that will always work


jasonbot (approved):
Don’t work on weekends!

PR #2399, opened 2024-01-14
fix: elasticsearch serialization issue 

ryannikolaidis:
This fixes the serialization of the Elasticsearch destination connector. Presence of the _client object breaks serialization due to TypeError: cannot pickle '_thread.lock' object. This removes that object before serialization.


jasonbot (approved):
Thank you for keeping this small

PR #2402, opened 2024-01-15
enhancement: normalize Salesforce artifact extensions

jakub-sandomierz-deepsense-ai:
Connectors use predictable result file naming convention so consumers of library can write code in abstraction of particular connector.

This change introduces compatibility with said naming convention.

`_output_filename` returns now filename with format.


ryannikolaidis (approved):
looks good!

PR #2403, opened 2024-01-16
fix documentation for chroma

potter-potter:
To test:

cd docs && make HTML

changelogs:

point main readme to the correct connector html page
point chroma docs to correct sample code


unstructured/ingest/connector/chroma.py
```diff
@@ -29,8 +29,8 @@ class SimpleChromaConfig(BaseConnectorConfig):
     access_config: ChromaAccessConfig
     collection_name: str
     path: t.Optional[str] = None
-    tenant: t.Optional[str] = None
-    database: t.Optional[str] = None
+    tenant: t.Optional[str] = "default_tenant"
+    database: t.Optional[str] = "default_database"
```
ryannikolaidis, line 33: what's going on here? why do we have these defaults??

ryannikolaidis: I mean, and likewise the defaults [here](https://github.com/Unstructured-IO/unstructured/blob/f1b8ceff30a16e21d93403b1504571fea092583a/unstructured/ingest/cli/cmds/chroma.py#L27-L41)?

ryannikolaidis: seems like these fields should just be required, right?

unstructured/ingest/cli/cmds/chroma.py
```diff
@@ -27,17 +27,15 @@ def get_cli_options() -> t.List[click.Option]:
...
+                help="The tenant to use for this client." "Chroma default is 'default_tenant'.",
             ),
             click.Option(
                 ["--database"],
-                required=False,
-                default="default_database",
+                required=True,
                 type=str,
-                help="The database to use for this client.",
+                help="The database to use for this client." "Chroma default is 'default_database'.",
```
ryannikolaidis, line 38: sorry, circling back from DMs. did not realize Chroma actually defaults to these values itself. in which case this is fine! ...maybe add a comment somewhere to that extent

CHANGELOG.md
```diff
@@ -790,6 +790,7 @@ allowing the document to be loaded. Fix: Change parent class for Formula to Text
 * make notion module discoverable
 * fix emails with `Content-Distribution: inline` and `Content-Distribution: attachment` with no filename
 * Fix email attachment filenames which had `=` in the filename itself
+* Fix documentation and sample code for Chroma
```
ryannikolaidis, line 793: please bump version. also probably should mention default/required change since that could actually break someone's workflow

potter-potter: reset it so that it shouldn't break someones workflow

ryannikolaidis (approved):
thanks!

PR #2404, opened 2024-01-16
feat: kafka src/dest connector

tullytim:
Adds a src and dest connector for Kafka


potter-potter:
Very cool! Is this something we can test with a Kafka docker container? Or does it need an account at a hosted service?

rbiseck3:
Looks like this is using the confluent kafka sdk. Do we know how this is any different from the generic kafka sdk? https://kafka-python.readthedocs.io.

tullytim:
Actually that's a gap right now - it only works with confluent's kafka in the cloud. Will submit a subsequent PR to make it work with plain Kafka as well, as the config is a subset.

tullytim:
> Very cool! Is this something we can test with a Kafka docker container? Or does it need an account at a hosted service?

It needs a hosted service to test the Confluent usage.  For the Kafka connection, it would need a container or something to spawn an instance.

rbiseck3:
For testing, can we at least leverage a kafka docker container for the non-confluent use case? There are sample scripts in the `/scripts` directory of how other services are stood up via docker/docker-compose for ingest testing (i.e. elasticsearch).

test_unstructured_ingest/dest/kafka.sh
```diff
@@ -0,0 +1,120 @@
...
+
+
+if [ -z "$KAFKA_CLUSTER_ID" ]; then
+	echo "Skipping Kafka ingest test because KAFKA_CLUSTER_ID env var is not set."
+	exit 0
+fi
+
+RANDOM_SUFFIX=$((RANDOM % 100000 + 1))
+
+LC_ALL=C
```
potter-potter, line 37: I'm not familiar with this. Is it necessary? What does it do?

unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,315 @@
...
+        producer.flush()
+        logger.info(f"Uploaded {num_uploaded} documents to Kafka")
+
+    def write(self, docs: t.List[BaseIngestDoc]) -> None:
+        content_list: t.List[t.Dict[str, t.Any]] = []
+        for doc in docs:
+            local_path = doc._output_filename
+            with open(local_path) as json_file:
+                dict_content = json.load(json_file)
+                for content in dict_content:
```
potter-potter, line 160: I notice that we are only writing `type, text, filename` into kafka. Why are we limiting ourself to this? Is this a limit of Kafka? 
Can we pass in embeddings and the other metadata? 

Maybe we should skip the chunking and embedding phase then in the destination test...
Please add some info in the PR description about the limitations or exact purpose of the connectors.

unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,315 @@
...
+    Note that messages have the format of:
+    <filename>: the name of the file (with correct file extension)
+    <content>: base64 encoded (whether was binary or not)
+    """
+
+    connector_config: SimpleKafkaConfig
+    _consumer: t.Optional["Consumer"] = None
+
+    def check_connection(self):
+        pass
```
potter-potter, line 233: Can we do something similar here to the destination connector? We will start using check_connection soon.

potter-potter:
@tullytim I'm going to be looking at this again in the upcoming weeks. It's super cool. What was your vision for a Kafka connector? Kafka has a message size limit of 1meg. So is that the document size that can be processed. Just a general description would help.
Just curious, are you using this connector yourself with unstructured?

PR #2413, opened 2024-01-17
enhancement: add _clean_ocr_languages_arg helper function

Coniferish:
This PR is one in a series of PRs for refactoring and fixing the languages parameter so it can address incorrect input by users. #2293

This PR adds _clean_ocr_languages_arg. There are no calls to this function yet, but it will be called in later PRs related to this series.


shreyanid:
I know you mentioned this will be used in future PR, but could you give me an idea of where this will be used? Why does this change only impact to the (soon deprecated) `ocr_languages` arg and not `languages` as well?

shreyanid (approved):
for whatever use `ocr_languages` still has, this change LGTM. let me know if you can give more context (see earlier question) so I can merge

Coniferish:
@shreyanid This is all stemming from users passing in incorrect arguments to the API. The that was opened is about accepting those common mistakes (even though `ocr_languages` will be deprecated). I don't know when we're planning to officially deprecate `ocr_languages`.

shreyanid:
ok, just revisited the linked issue. I also am not sure when we plan to officially deprecate it. I suppose handling messy `ocr_languages` is helpful anyway so that it becomes compatible with our conversion from ocr_languages to languages (if languages is not specified). sgtm thanks!

PR #2415, opened 2024-01-17
enhancement: allow setting image block crop padding parameter

### LINKED ISSUE #2320: bug: some extracted images have content clipped
**Describe the bug**
In some cases images extracted from a pdf are clipped (cutting off image content)

**To Reproduce**
Process this pdf with `extract_images_in_pdf=True`
[LM339-D.PDF](https://github.com/Unstructured-IO/unstructured/files/13781511/LM339-D.PDF)

**Expected behavior**
Extracted images should contain complete image content.

**Screenshots**
Some of the images in here will have clipped content. e.g.
figure-2-4

christinestraub:
Closes #2320 .

### Summary
In certain circumstances, adjusting the image block crop padding can improve image block extraction by preventing extracted image blocks from being clipped.

### Testing
- PDF: [LM339-D_2-2.pdf](https://github.com/Unstructured-IO/unstructured/files/13968952/LM339-D_2-2.pdf)
- Set two environment variables `EXTRACT_IMAGE_BLOCK_CROP_HORIZONTAL_PAD` and `EXTRACT_IMAGE_BLOCK_CROP_VERTICAL_PAD` 
(e.g. `EXTRACT_IMAGE_BLOCK_CROP_HORIZONTAL_PAD = 40`, `EXTRACT_IMAGE_BLOCK_CROP_VERTICAL_PAD = 20`

```
elements = partition_pdf(
    filename="LM339-D_2-2.pdf",
    extract_image_block_types=["image"],
)
```


Coniferish:
@Christine
This is what I'm using to test this, but I'm not seeing a difference in the image that's saved from the `main` branch and this one:
```
filename = "example-docs/LM339-D_2-2.pdf"
elements = partition_pdf(
    filename=filename,
    extract_image_block_types=["image"],
    strategy="hi_res",
    extract_image_block_output_dir="output/"
)
```

Do I need to define some additional param to increase the padding?

unstructured/partition/pdf_image/pdf_image_utils.py
```diff
@@ -75,6 +78,38 @@ def convert_pdf_to_image(
     return images
 
 
+def pad_element_bboxes(
+    element: "LayoutElement",
```
Coniferish, line 82: Not sure this needs to be changed since I'm not seeing an error from it, but why does the type hint have parentheses around it?

Coniferish:
Do the `.rst` files in `docs/source/` need to be updates with these changes?

christinestraub:
> @christine This is what I'm using to test this, but I'm not seeing a difference in the image that's saved from the `main` branch and this one:
> 
> ```
> from unstructured.partition.pdf import partition_pdf
> filename = "example-docs/LM339-D_2-2.pdf"
> elements = partition_pdf(
>     filename=filename,
>     extract_image_block_types=["image"],
>     strategy="hi_res",
>     extract_image_block_output_dir="output/"
> )
> ```
> 
> Do I need to define some additional param to increase the padding?

To see a difference in the image, you'll need to set two environment variables `EXTRACT_IMAGE_BLOCK_CROP_HORIZONTAL_PAD` and `EXTRACT_IMAGE_BLOCK_CROP_VERTICAL_PAD`. 
I updated testing instruction in description.

unstructured/partition/pdf_image/pdf_image_utils.py
christinestraub: No error is raised from it. I changed it as a refactoring. Do you mean double quotes("") by the "parentheses"?

Coniferish: Whoops. Yes, I meant quotes not parentheses.
Sounds good. Was mostly curious since I hadn't seen that before.

christinestraub: The double quotes are used to reference the LayoutElement type with the `TYPE_CHECKING` constant in a way that is compatible with both runtime execution (where LayoutElement is not actually imported) and static type checking (where LayoutElement is available).

Coniferish (approved):
LGTM once we get that docker test passing :)

Coniferish:
@christinestraub, one last question: I want to make sure that this should NOT update `element.text` that's detected even if the image now includes the full text (in this example "Output" is now visible, but it's still not a part of `element.text`)

christinestraub:
> @christinestraub, one last question: I want to make sure that this should NOT update the `element.text` that's detected even if the image now includes the full text (in this example "Output" is now visible, but it's still not a part of `element.text`)

I didn't expect to update the `element.txt`. @cragwolfe What do you think of this? Do we need to update `element.text` as well when extracting by adding padding to the detected image block?

cragwolfe:
> > @christinestraub, one last question: I want to make sure that this should NOT update the `element.text` that's detected even if the image now includes the full text (in this example "Output" is now visible, but it's still not a part of `element.text`)
> 
> I didn't expect to update the `element.txt`. @cragwolfe What do you think of this? Do we need to update `element.text` as well when extracting by adding padding to the detected image block?

I think the primary intent of this PR is with respect to the image artifacts saved, so do not need to update the .text in this PR. It's a good callout though and worth tracking in another github issue. But, there is a fair bit of nuance -- it seems find to extend OCR .text, but also want to be careful not to duplicate other text if the bbox overlaps another element. Or, don't allow a bbox overlap in the first place for OCR, only extend to min(EXTRACT_IMAGE_BLOCK_CROP_HORIZONTAL_PAD,distance_to_next_bbox).

cragwolfe (approved):
LGTM!

PR #2420, opened 2024-01-18
fix: Fix and use FSSpec dest check_connection

jakub-sandomierz-deepsense-ai:
FSSpec destination connectors did not use `check_connection`. There was an error when trying to `ls` destination directory - it may not exist at the moment of creation of connector.

Now `check_connection` calls `ls` on bucket root and this method is called on `initialize` of destination connector.


potter-potter:
Looks like this broke the dropbox.sh destination test. (was working in the main branch)
https://github.com/Unstructured-IO/unstructured/actions/runs/7571073407/job/20617893308#step:7:1366

jakub-sandomierz-deepsense-ai:
> Looks like this broke the dropbox.sh destination test. (was working in the main branch) https://github.com/Unstructured-IO/unstructured/actions/runs/7571073407/job/20617893308#step:7:1366 We should probably just give dropbox it's own version like azure.
> 
> Looks like azure (an fsspec connector) skips this. Which should be fine. S3 and GCS work with this change. At some point we should get the box destination connector working again.

Good point! I did not look there because the job has not failed (!!!). I saw customization for Dropbox path convention in other function/method, will try to fix it.

potter-potter (approved):
Looks good to me

PR #2423, opened 2024-01-18
fix: update flatten dict to support flattening tuples 

ahmetmeleq:
This PR updates flatten_dict function to support flattening tuples. 

This is necessary for objects like Coordinates, when the object is not written to the disk, therefore not being converted to a list before getting flattened.


potter-potter:
Could you add a few simple tests:
https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured/staging/test_base_staging.py

potter-potter (approved):
Looks like it works well.

PR #2425, opened 2024-01-18
fix: chroma destination connector serialization

ahmetmeleq:
This fixes the serialization of the ChromaDB destination connector. Presence of the _collection object breaks serialization due to TypeError: cannot pickle 'module' object. This removes that object before serialization.

### DIFF
--- a/unstructured/ingest/connector/chroma.py
+++ b/unstructured/ingest/connector/chroma.py
@@ -65,8 +65,8 @@ class ChromaDestinationConnector(BaseDestinationConnector):
     def to_dict(self, **kwargs):
         """
         The _collection variable in this dataclass breaks deepcopy due to:
-        TypeError: cannot pickle '_thread.lock' object
-        When serializing, remove it, meaning client data will need to be reinitialized
+        TypeError: cannot pickle 'module' object
+        When serializing, remove it, meaning collection data will need to be reinitialized
         when deserialized
         """
         self_cp = copy.copy(self)
--- a/unstructured/ingest/connector/chroma.py
+++ b/unstructured/ingest/connector/chroma.py
@@ -1,7 +1,9 @@
+import copy
 import typing as t
 import uuid
 from dataclasses import dataclass
 
+from unstructured.ingest.enhanced_dataclass.core import _asdict
 from unstructured.ingest.error import DestinationConnectionError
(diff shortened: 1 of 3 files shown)


unstructured/ingest/connector/chroma.py
```diff
@@ -60,6 +62,18 @@ def initialize(self):
     def check_connection(self):
         _ = self.chroma_collection
 
+    def to_dict(self, **kwargs):
+        """
+        The _collection variable in this dataclass breaks deepcopy due to:
+        TypeError: cannot pickle '_thread.lock' object
```
jasonbot, line 68: This comment looks copy pasted from the Postgres. Is it correct?

ahmetmeleq: It is false - fixing now

unstructured/ingest/connector/chroma.py
```diff
@@ -60,6 +62,18 @@ def initialize(self):
...
+        When serializing, remove it, meaning client data will need to be reinitialized
+        when deserialized
+        """
+        self_cp = copy.copy(self)
+        if hasattr(self_cp, "_collection"):
+            setattr(self_cp, "_collection", None)
```
jasonbot, line 74: Would it be safer to `delattr`? Though now that I think about it probably not.

unstructured/ingest/connector/chroma.py
ahmetmeleq: Addressed with [fix docstring](https://github.com/Unstructured-IO/unstructured/pull/2425/commits/d87d441bc27776005a44d83272c9eb2d4de56c2c), had missed those two lines

jasonbot (approved):
Looks fine as long as you’ve validated it. Test would be nice.

potter-potter (approved):
Looks good. Basically the same code we've had to add to most of the dest connectors.

unstructured/ingest/connector/chroma.py
ahmetmeleq: Since we have it in dataclass initialization as a `None` value, is `setattr` to `None` better?
https://github.com/Unstructured-IO/unstructured/blob/4a34765fdfc09ec47ebe84dbacf50e6257afa65e/unstructured/ingest/connector/chroma.py#L44-L48

PR #2427, opened 2024-01-19
fix: fsspec connectors returning data source version as integer

ryannikolaidis:
Connector data source versions should always be string values, however we were using the integer checksum value for the version for fsspec connectors. This casts that value to a string.

## Changes

* Cast the checksum value to a string when assigning the version value for fsspec connectors.
* Adds test to validate that these connectors will assign a string value when an integer checksum is fetched.

## Testing

Unit test added.


potter-potter (approved):
Looks solid. Thanks for the test.

Issue #711, opened 2023-06-09
Add indentation level as a metadata field


MthwRobinson:
The goal of this issue is to add indentation level as a metadata field. For documents such as PDFs, we can detect this based on the position of the bounding boxes. For MS office documents this may be available as an attribute with the document xml.

newelh:
This is in progress with #1433

orlandounstructured:
Older than 180 days old but keeping active due to recent reference in #2428

PR #2428, opened 2024-01-19
[WIP] Feat : Code partition

Plemeur:
Hello, 
Since I want to partition code file, I though I'd better contribute to the project.

Still a work in progress but I think it's better to share where I'm at right now.

The code partition is done using TreeSitter, it's basically the same concept as the one explained in this blog post with some tweaking that made sense to me.
https://docs.sweep.dev/blogs/chunking-2m-files#our-solution-

Most of the grunt work (FileTypes, Extension to str) is done, still need to figure out the mimetype for all the different files, but that's more or less details

For now all the elements are just `Code` elements, since the original solution I wanted to apply to differentiate between function/classes or whatnot is not working. If you have any suggestion on that I'm all ears.

I did experience some issues when testing with the auto partition, as some file mimetypes were not properly detected. It does not error out, just fall back to text partition. 
One fix I can think of is to try and use treesitter to parse the file with different language parser, and if one is correct just assuming the type. This might be unpractical because it either mean running the parser on all files that are detected as plaintext, or creating auto function specially for code file, with the corresponding API endpoint or request parameters

Now there is also the question of how to actually compile the parser and include the library, right now I'm doing it before running anything, but it might be done when loading the library for the first time, and cached in a folder like `~/.treesitter_build`


unstructured/file_utils/filetype.py
```diff
@@ -383,7 +404,7 @@ def detect_filetype(
         # NOTE(robinson) - we'll treat all code files as plain text for now.
         # we can update this logic and add filetypes for specific languages
         # later if needed.
-        return FileType.TXT
+        return STR_TO_FILETYPE.get(mime_type, FileType.TXT)
```
PublisherApplications, line 407: good job

unstructured/file_utils/filetype.py
```diff
@@ -383,7 +404,7 @@ def detect_filetype(
         # NOTE(robinson) - we'll treat all code files as plain text for now.
         # we can update this logic and add filetypes for specific languages
         # later if needed.
-        return FileType.TXT
```
PublisherApplications, line 386: bad job

PublisherApplications (approved):

unstructured/partition/auto.py
```diff
@@ -524,6 +541,13 @@ def partition(
             detect_language_per_element=detect_language_per_element,
             **kwargs,
         )
+    elif filetype in CODE_FILETYPES:
```
MthwRobinson, line 544: Could we update this logic to fallback to `partition_text` if `tree_sitter` doesn't exist? That way people who a processing these as text files can still do so if they don't have the code dependencies installed.

MthwRobinson (changes requested):
Thanks for the PR! Could you add the following:
- A changelog entry
- Bump the `unstructured/__version__` (should go to `0.12.1-dev19`)
- A test for `partition_code`

qued (commented):
Some initial comments on the draft. One main thing is that tests are needed, and I haven't attempted to test the code.

unstructured/partition/code.py
```diff
@@ -0,0 +1,217 @@
...
+        min_partition is not None
+        and max_partition is not None
+        and (min_partition > max_partition or min_partition < 0 or max_partition < 0)
+    ):
+        raise ValueError("Invalid values for min_partition and/or max_partition.")
+
+    # Verify that only one of the arguments was provided
+    exactly_one(filename=filename, file=file, text=text)
+    file_text = bytes()
+    language = None
```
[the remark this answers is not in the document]
Plemeur: Sure thing,
Also, for now I've put the programming language as the language metadata, but it might be confusing in the rest of the pipeline, any input on that ?

unstructured/partition/auto.py
Plemeur: Sure, 
We can either :
- define `partition_code = partition_text` when checking for the dependancy
- Add a global switch (or recheck for dependancy) and change the filetype from any code filetype to TXT
- Add a global switch (or recheck for dependancy) and add a nested if in the `if filetype in CODE_FILETYPES:`

I'm fine with any of those

Makefile
```diff
@@ -62,6 +62,12 @@ install-dev:
 install-build:
 	python3 -m pip install -r requirements/build.txt
 
+.PHONY: install-code
+install-code:
+	python3 -m pip install -r requirements/extra-code.txt
+	scripts/install_treesitter.sh
```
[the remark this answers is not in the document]
Plemeur: Hello
I remove that script in favor of dynamically building the treesitter library at runtime

However I have a small dilemma, treesitter is loading a shared library to be able to parse code, and there are only dirty method to actually unload libraries in python

Now, what I was thinking of is to build new libraries when new languages to parse show up, updating the main one as well for future use, and cleaning the extra built libraries whenever we leave the script/program that uses the parser. This works, but it means that there is a increase in memory usage for all new language that is greater than what it should actually be, which might be a problem if you spin up multiple API endpoint for load balancing or whatnot

We can partially fix it by setting a list of supported languages, and just build that the first time the module is imported. Or by using all the languages (even if the probability of people parsing things like Motorola 68000 Assembly is quite low)

In the end, to me anything is fine since I don't have infrastructure running for clients, so it's up to you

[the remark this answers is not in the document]
Plemeur: Went for the install on first load in the current commits

unstructured/partition/auto.py
Plemeur: Added tests, 
There are some gotcha that I will try to fix such as javascript files detected as cpp by libmagic, or go file detected as plain text

On top of that, I found that it's tricky to make the text actually fit both the minimum and the maximum size properly. I went with enforcing the minimum because it means less elements so potentially better search in the vector database

test_unstructured/partition/test_code.py
```diff
@@ -0,0 +1,81 @@
...
+
+@pytest.mark.parametrize("min_partition", [50, 100, 200])
+@pytest.mark.parametrize("language", TEST_FILE_URLS.keys())
+def test_partition_code_min_partition(min_partition: int, language: str, test_files):
+    elements = partition_code(filename=test_files[language], min_partition=min_partition)
+    assert len(elements) > 0
+
+    # No enforcement is done on the last element
+    for el in elements[:-1]:
+        assert len(el.text) >= min_partition
```
scanny, line 81: I'd like to see a couple tests that actually compare the generated JSON.

Especially in early stages like this, those JSON "expected output" files give a clear view to the partitioning behavior, like this input generates this output. Input files need to be shortish so you don't end up with huge JSON files of course but a few hundred lines of JSON is fine.

These will eventually provide the needful for ingest tests but the JSON "expectation" files can live in this test directory for a start.

scanny (commented):
A few remarks.

@MthwRobinson if you can weigh in on what the desired behavior is, there are a couple mentions for you on that in my comments.

MthwRobinson:
Hey everyone, tracking this one for review and hopefully will be able to give it a pass tomorrow.

unstructured/partition/code.py
```diff
@@ -0,0 +1,226 @@
...
+            filename=metadata_filename or filename,
+            last_modified=metadata_last_modified or last_modification_date,
+            languages=[programming_language],
+        )
+        metadata.detection_origin = detection_origin
+    else:
+        metadata = ElementMetadata()
+
+    # Note (Pierre) Some languages (i.e Python) use blank space and tabs for semantic
+    # Maybe removing the spaces is not a good idea
```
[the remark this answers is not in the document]
Plemeur: I agree to some extent.
In a language that enforces the use of brackets for inner blocks, we can then retrieve the actual context by adding the chunks before and after.
For Python (and maybe other languages but I don't know), there is no way to understand where a code block fits unless we have it's indentation (or check for errors)

We could re-format all code coming in and then find a more compact representation for the indentation, but it seems a bit too involved

test_unstructured/partition/test_code.py
```diff
@@ -0,0 +1,81 @@
...
+def test_files(tmp_path_factory):
+    test_files = {}
+    for language, file in TEST_FILE_URLS.items():
+        text = requests.get(file).content.decode("utf-8")
+        filename = file.split("/")[-1]
+        fn = tmp_path_factory.mktemp("data") / filename
+        fn.write_text(text)
+        test_files[language] = fn
+
+    return test_files
```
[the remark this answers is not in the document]
Plemeur: I'm open to suggestion for good tests files

test_unstructured/partition/test_code.py
Plemeur: I think I get your point,
Same as above, if you have some good test files idea please feel free to share.
I can generate the expected output with the current code, make sure that the metadata are fine, and it can be a reference point for future changes in the partitioning strategy

scanny: The ones you link to are a good starting point, just that they're local rather than need to be downloaded.

unstructured/partition/code.py
```diff
@@ -0,0 +1,226 @@
...
+        ) > min_chars:
+            # Child would make the current chunk too big, so start a new chunk
+            new_chunks.append(current_chunk)
+            current_chunk = text[last_end : child.end_byte]
+        else:
+            current_chunk += text[last_end : child.end_byte]
+        last_end = child.end_byte
+    if len(current_chunk) > 0:
+        new_chunks.append(current_chunk)
+    return new_chunks, last_end
```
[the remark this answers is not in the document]
Plemeur: This code is probably fine,
If the current code is merged (and even if it's not) please feel free to make a new pull request based on this idea, I personally find this code difficult to read and reason about

unstructured/partition/code.py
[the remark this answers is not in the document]
scanny: Hmm, @Plemeur I think I see the complication you indicate. When we _combine_ elements (concatenate their text), if we've gotten rid of the indentation at the TreeSitter (TS) node level, they could combine to produce invalid code, at least in Python.

So if we did decide to do this, it would have to be at the element level (encapsulated in `CodeElementBuilder`) such that "leading" indentation was removed consistently for all nodes in a single element.

Also, we would have to expect that elements were never combined during chunking.

The more I think about it the more the current approach seems to be partitioning and chunking welded together, which makes me uneasy about it fitting in consistently with the broader pipeline concepts across other document types.

At the same time it's not clear to me what an "atomic" element would be. Clearly a function would be too big, at least sometimes, and a statement would be the next level down; I don't think we could distinguish "passages" without some sort of inference.

@Plemeur does the division this produce turn out to be effective in your application? And would there be the prospect of chunking these elements with one of the unstructured chunkers?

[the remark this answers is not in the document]
Plemeur: I'm just starting out with this
I'm using this https://github.com/bionic-gpt/bionic-gpt and for now and my plan was to go up the chain from the chunking to the agent to understand how things work together

From what I understand from the chunkers code, it would not really be applicable to code broadly because of the different writing styles

I don't feel there is a valid abstraction that will satisfy 300 lines of parameters definition in a python dataclass, and at the same time 300 lines of JSX in the return of a react functional component

For the "atomic" element talk, when I tried to rewrite the parsing the first time, I was trying to find a good logic but came out to those conclusion: Any element might be too long to be considered atomic. 
Even if you think "Well a value declaration should be small enough" ... Hard coded dictionary, list, or strings longer than function will appear, and there's nothing you can do about it

unstructured/partition/auto.py
MthwRobinson: I'm good with either of these. I'd say avoid `partition_code = partition_text`, just so it's clear later in the code that `partition_text` is being applied and not `partition_code`.
- Add a global switch (or recheck for dependancy) and change the filetype from any code filetype to TXT
- Add a global switch (or recheck for dependancy) and add a nested if in the if filetype in CODE_FILETYPES:

Are the MIME type discrepancies when they're checked as files rather than by filename? If think that's okay and you can mock the correct MIME type in the test. Hopefully most of those files come in with the file extension.

The enforcing minimum logic makes sense to me.

unstructured/partition/code.py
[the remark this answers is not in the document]
MthwRobinson: I'm okay with keeping the indent in the elements for now, even if it creates some bloat. We do have #711 out there to add a metadata field for indentation. At that point, we could add the indentation level of the code as metadata and skip the extra test.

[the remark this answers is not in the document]
scanny: @Plemeur Yeah, on reflection I think you're right about the lack of a naturally-occurring "atomic" element in code. After a day's reflection I can't think of any uniform structure that would always be guaranteed to be shorter than any given threshold.

Also I'm inclined to agree that code is distinctive enough from prose/narrative text that it's not going to be compatible in-general with chunkers that are really designed for narrative text. In particular, there are no `Title` elements in the stream so `chunk_by_title()` wouldn't be applicable and there are no `Table` elements, so a good portion of the chunking code would never get triggered. My working conclusion from that is that if we do have code-element chunkers they would need to be dedicated to code element-streams.

Along the lines of brainstorming, it occurs to me we might one day want elements that as much as possible are whole functions or perhaps [Module | Class | Function]. Like other elements, these would not have a size limit. Then separately a chunker would divide oversized functions into sub-chunks, possibly by a second treesitter parse. Just an idea, but does argue to me for keeping chunking distinct from partitioning, as we do for narrative text. 

Right now I'd say we do have partitioning and chunking, even if they're "welded together" in this implementation:
- Partitioning is generating whole TS nodes that fit in the max_char window.
- Chunking is combining those to be between min and max-chars.

Maybe we'll dream up alternate chunking methods going forward, which would be an argument for keeping these two operations distinct such that they can be flexibly composed into other configurations.

Have you given any thought to the metadata that might make sense for code chunks? I'm inclined to think that hierarchy is readily available and some of that might be worth recording at some point.

All this isn't to do with what your code here per se, just noodling on how code partitioning might evolve so we can preserve flexibilities we might want as it evolves.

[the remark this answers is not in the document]
Plemeur: I understand why you want to keep them separate.

I did at first try to partition more sparsely walking the treesitter tree, grouping elements that have similar nature together if they are close enough, but I had some issues making sense of child elements at times.

For instance 
```python
def LongFunction(
    param_1,
    param_2
   ) :
   function body ....
```

When you arrive to the Long function element, the children nodes are `def` `LongFunction` `(` `param_1` `param_2` `):` `body`. 
So it requires some heavy logic to put everything together properly (and is most likely language dependent)

unstructured/partition/auto.py
Plemeur: Yes, MIME types are not always correctly detected.
Just to make sure people have the option, I'm adding the `programming_language` parameters, and adding some test for the auto partition 

We could try all the available parsers for file that are not detected as code at all, but for the ones where the MIME is mixed up, there's nothing we can really do

Plemeur: After looking up a bit, it seems that most code files do not have a MIME type...
Here is a list of available ones https://www.iana.org/assignments/media-types/media-types.xhtml

Since it is required for the auto partition, I made custom ones, it looks like to me that the common way of doing so is `text/x-{language}` so I went for that

unstructured/partition/code.py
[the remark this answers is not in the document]
scanny: Hmm, yes, that's a good point. I believe the node types from TreeSitter are language-specific, right? So there's no such thing as a "function" node that is the same for all languages. So we would have to maintain some mapping between language and node-type names, which would be perpetually incomplete. Hmm, yes, that would definitely support the idea of not getting too specific with trying to distinguish node types. Glad you mentioned that :)

Plemeur:
Tested a bit with the API, It works for python/js/cpp/c, but mimetype handling is getting a bit out of hand

scanny:
@Plemeur I'm starting a separate PR to bring this code into. I'll credit you with commits that come from this code. I'll let you know once I've got it far enough along to push up, maybe tomorrow or so.

unstructured/utils.py
```diff
@@ -712,3 +714,64 @@ def catch_overlapping_and_nested_bboxes(
...
+    if sys.platform == "win32" and "APPDATA" in os.environ:
+        homedir = os.environ["APPDATA"]
+    else:
+        homedir = os.path.expanduser("~/")
+        if homedir == "~/":
+            raise ValueError("Could not find a default directory for TreeSitter")
+    return homedir
+
+
+def update_tree_sitter(languages: List[str], cache_dir: Optional[str] = None):
```
scanny, line 729: @Plemeur on this bit, I'm inclined to just use the [tree-sitter-languages Python package](https://github.com/grantjenks/py-tree-sitter-languages) as a dependency, rather than enter the business of compiling TreeSitter parsers. Did you consider that option and reject it for some reason?

Plemeur: I did not even check if something like that existed since there was the compiling option, it does seems like an easier solution for distribution.

Plemeur:
> @Plemeur I'm starting a separate PR to bring this code into. I'll credit you with commits that come from this code. I'll let you know once I've got it far enough along to push up, maybe tomorrow or so.

Should I put this back to a draft and stop doing anything to it then ?

scanny:
@Plemeur best case scenario for me would be if you can use this branch to develop it far enough to actually use it and tweak it such that it best suits your use-case.

It might take a little while for me to finish up the branch that would get merged since I have to fit it in-between other work.

There's probably no need for you to attend to release details like getting the requirements.in/txt any better than it already is, that sort of thing since I'll have to go over all those details anyway. Also I'm going to be adding quite a few more tests so probably no need for you to add more than you want for present purposes.

But what I'm not going to be able to do and what would really benefit the library is to actually use this for some real-life use-cases and see how it performs, and perhaps tweak things to optimize that performance.

Is that something this is close enough to for your use-case do you think?

Plemeur:
@scanny 
I might be able to go back to that in a week or two, will keep you updated if I see any performance issue

MthwRobinson:
@Plemeur - Are you still working on this?

Plemeur:
> @Plemeur - Are you still working on this?

No, I thought scanny was taking over
Did not get any time to work on LLMs for code (or any LLMs tasks)

PR can be closed if stale

MthwRobinson:
Got it, thanks @Plemeur !

PR #2430, opened 2024-01-19
fix: fix table index error by bumping `unstructured-inference`

### LINKED ISSUE #2417: bug/IndexError: list index out of range in partition()
**Describe the bug**
When the `partition` func uses the  hi-res strategy with default model (a little hard to parse from the docs, but i think yolox?) and tries to guess table html, it raises `IndexError: list index out of range` when it gets to the line `prediction = recognize(outputs_structure, x, tokens=ocr_tokens)[0]` of run_prediction, which appears to occur when trying to read `tables_agent.predict(cropped_image, ocr_tokens=table_tokens)`. this does **not** occur if i instead specify `detectron2_onnx` as the model_name param to partition(). This bug was also noted over [here](https://github.com/Unstructured-IO/unstructured-inference/issues/306).

**To Reproduce**
```
elements = partition(
    url=pdf_url,
    include_page_breaks=False,
    content_type="application/pdf",
    skip_infer_table_types=[],
    pdf_infer_table_structure=pdf_infer_table_structure,
)
```

**Environment Info**
This is using unstructured 0.12.0, the latest release. 

**Additional context**

MthwRobinson:
### Summary

Closes #2417. Bumps `unstructured-inference` to pull in the fix implemented in https://github.com/Unstructured-IO/unstructured-inference/pull/317


qued (approved):
LGTM!

PR #2435, opened 2024-01-20
fix: check languages args

Coniferish:
This PR is the last in a series of PRs for refactoring and fixing the language parameters (`languages` and `ocr_languages` so we can address incorrect input by users. See #2293

It is recommended to go though this PR commit-by-commit and note the commit message. The most significant commit is "update check_languages..."


Coniferish:
@cragwolfe, I updated the .rst file to use `languages` instead of `ocr_languages`, but noticed that we only mention it for `partition_pdf` and `partition_image`. Should I add something to all the partitioners or create a section about `languages` since it's used by all the partition methods? One point that I think might be important to note somewhere is that `"auto"` is a valid argument for most of the partition methods, but isn't valid for `partition_pdf` or `partition_image` since the `languages` parameter is used for ocr, unlike in the other partition methods where the `languages` parameter is used for language detection (slight difference in how the `languages` param is _used_, but it ends up meaning the same thing in the element metadata).

docs/source/core/partition.rst
```diff
@@ -859,7 +859,7 @@ type for the file. If you do not explicitly pass it, the MIME type will be infer
     elements = partition_via_api(file=f, metadata_filename=filename, api_key="MY_API_KEY")
 
 
-You can pass additional settings such as ``strategy``, ``ocr_languages`` and ``encoding`` to the
+You can pass additional settings such as ``strategy``, ``languages`` and ``encoding`` to the
 API through optional kwargs. These options get added to the request body when the
 API is called.
 See `the API documentation <https://api.unstructured.io/general/docs>`_ for a full list of
```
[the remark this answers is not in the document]
Coniferish: Good catch. Yah, that will need to be a follow-up issue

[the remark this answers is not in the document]
Coniferish: Actually, @shreyanid , I am seeing `languages` as a param for both `/general/v0/general` and `/general/v0.0.62/general` endpoints. Are you referring to something else?

unstructured/partition/lang.py
```diff
@@ -167,34 +167,68 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
...
 
 
-def check_languages(languages: Optional[List[str]], ocr_languages: Optional[str]):
-    """Handle `ocr_languages` and `languages`, defining `languages` to ['eng'] as default and
-    converting `ocr_languages` if needed"""
-    if languages is None:
-        languages = ["eng"]
+def check_language_args(
+    languages: Optional[List[str]], ocr_languages: Optional[str]
+) -> Union[list[str], None]:
```
scanny, line 172: The return type should be `Optional[list[str]]`. If you were going to use `Union` notation for this, which is perfectly valid, just not conventional for this code base, then it would be `list[str] | None`. Both the lowercase version of `List` and the `|` version of `Union` require a `from __future__ import annotations` import (which all our files should have until Python 3.13 or later).

scanny: I would be strongly inclined to cut off optionality of `languages` at the interface function and use `[]` as the zero value for `languages` everywhere else. Rationale:

1. The option to provide None is a convenience to an end-user who is specifying `languages` as a kwarg (dict item) where it it more work to omit the argument. If a default is applied to replace an omitted value, it's the interface function's job to do that substitution.
2. `None` and `[]` have the same semantics (as far as I know). Using both complicates any other non-interface function that uses that value. In general, `Optional` should be used sparingly outside of interface functions and methods.

The general principle here is to identify flexibilities we afford an end-user in calling an interface function and isolate those flexibilities to that interface function. Otherwise we have internal and helper functions attending to peculiarities of calling flexibility which should not be their responsibility (and makes interface changes like this tend to involve much deeper parts of the code than we'd like because they're unnecessarily coupled to the interface particulars.

unstructured/partition/lang.py
```diff
@@ -167,34 +167,68 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
...
+    `ocr_languages` is only a parameter for `auto.partition`, `partition_image`, & `partition_pdf`.
+    `ocr_languages` should not be defined as 'auto' since 'auto' is intended for language detection
+    which is not supported by `partition_image` or `partition_pdf`."""
+    # --- Clean and update defaults
+    if ocr_languages == "auto":
+        raise ValueError(
+            "`ocr_languages` is deprecated but was used to extract text from pdfs and images. "
+            "The 'auto' argument is only for language *detection* when it is assigned "
+            "to `languages` and partitioning documents other than pdfs or images. "
+            "Language detection is not currently supported in pdfs or images."
```
scanny, line 185: When specifying multiline strings like this, I make it my habit to place the space separating each line at the _beginning_ of the second and later lines, like:
```python
"ocr_languages is deprecated but was used to extract text from pdfs and images."
" The 'auto' argument is only for language *detection* when it is assigned"
" to `languages` and partitioning documents other than pdfs or images."
" Language detection is not currently supported in pdfs or images."
```

That way its presence can be confirmed by quickly scanning down the aligned left side rather than the ragged right side, which is much more work for the eye. This leads to fewer merged words in exception output, which is a perennial problem.

unstructured/partition/lang.py
```diff
@@ -167,34 +167,68 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
...
 
-    if not isinstance(languages, list):
+    if ocr_languages:
+        ocr_languages = _clean_ocr_languages_arg(ocr_languages)
+        logger.warning(
+            "The ocr_languages kwarg will be deprecated in a future version of unstructured. "
+            "Please use languages instead.",
+        )
+
+    # raise error if languages is not None of a list
```
scanny, line 195: I think you mean "or" a list rather than "of".

This is an example of the proliferation of multiple zero values.

scanny (commented):
A few remarks. Happy to discuss if you have any questions.

The logic looks okay to me, probably one of the other reviewers is better positioned to say, but I'm happy to approve in a pinch, just let me know.

test_unstructured/partition/test_lang.py
```diff
@@ -146,3 +154,39 @@ def test_clean_ocr_languages_arg(input_ocr_langs, expected):
...
+        (["spa"], "deu", ["spa"]),
+        (["spanish"], "english", ["spa"]),
+        (["spa"], "[deu]", ["spa"]),
+        (["spa"], '"deu"', ["spa"]),
+        (["spa"], ["deu"], ["spa"]),
+        (["spa"], ["[deu]"], ["spa"]),
+        (["spa+deu"], "eng+deu", ["spa", "deu"]),
+    ],
+)
+def test_check_language_args(
```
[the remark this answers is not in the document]
Coniferish: Sounds good. I think I can actually remove the conditional now that I created the separate tests for `None`, but will check.

unstructured/partition/lang.py
```diff
@@ -167,34 +167,68 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
...
 
-    if ocr_languages is not None:
-        if languages != ["eng"]:
-            raise ValueError(
-                "Only one of languages and ocr_languages should be specified. "
-                "languages is preferred. ocr_languages is marked for deprecation.",
-            )
+    # --- If `languages` is a null/default value and `ocr_languages` is defined, use `ocr_languages`
+    if ocr_languages and (
+        languages == ["auto"] or languages == [""] or languages is None or not languages[0]
```
christinestraub: I think `languages == [""]` and  `not languages[0]` have the same semantics.

unstructured/partition/lang.py
```diff
@@ -167,34 +167,68 @@ def prepare_languages_for_tesseract(languages: Optional[List[str]] = ["eng"]):
...
-            logger.warning(
-                "The ocr_languages kwarg will be deprecated in a future version of unstructured. "
-                "Please use languages instead.",
-            )
-    return languages
+            # define as 'auto' for language detection when partitioning non-pdfs or -images
+            languages = ["auto"]
+        return languages
+
+    return None
```
christinestraub, line 225: What do you think about returning a default value (`["eng"]`) instead of `None`, which would avoid setting the default value in `pdf.py` and `image.py`?

christinestraub (changes requested):
Overall, introducing the `check_language_args()` function seems like a good idea to eliminate duplicate code and avoid inconsistencies when checking language parameters. 

In my opinion, the deprecated parameter `ocr_languages` is only needed in interface functions like `partition_pdf()`, `partition_image()`, and `partition_auto()`, and we can remove the parameter from all non-interface functions like `partition_pdf_or_image()`) cc: @cragwolfe

unstructured/partition/lang.py
Coniferish: That might be right--I could have not checked that. I already changed this per Steve's comments, though

unstructured/partition/lang.py
Coniferish: I don't think we can do that because of the other partitioners that default to "auto". `None` is used as the default because we have to be able to resolve the languages/ocr_languages (or just languages) at the auto `partition` level. Downstream, once the document type is determined and the document is passed to the appropriate partitioner, the default for `languages` can be either "auto" or "eng". "eng" for the other partitioners would skip language detection since it'd be interpreted as a user telling us they know the document is in English.

Coniferish:
> Overall, introducing the `check_language_args()` function seems like a good idea to eliminate duplicate code and avoid inconsistencies when checking language parameters.
> 
> In my opinion, the deprecated parameter `ocr_languages` is only needed in interface functions like `partition_pdf()`, `partition_image()`, and `partition_auto()`, and we can remove the parameter from all non-interface functions like `partition_pdf_or_image()`) cc: @cragwolfe

This makes sense. Happy to add those changes to this PR if that sounds good to @cragwolfe

Coniferish:
note: I need to add some additional tests after making some of the requested changes

test_unstructured/partition/test_lang.py
```diff
@@ -146,3 +160,49 @@ def test_clean_ocr_languages_arg(input_ocr_langs, expected):
...
+    returned_langs = check_language_args(languages=languages, ocr_languages=ocr_languages)
+    for lang in returned_langs:  # type: ignore
+        assert lang in expected_langs
+        assert "ocr_languages" in caplog.text
+
+
+@pytest.mark.parametrize(
+    ("languages", "ocr_languages", "expected_langs"),
+    [
+        # raise warning and use `ocr_languages` when `languages` is empty or None
```
christinestraub, line 192: Can we add cases to test `check_language_args()` that returns `None`?
```
(None, None, None),
([], None, None),
(None, "", None),
([], "", None),
...
```

christinestraub:
Can we add more cases to `test_check_language_args_returns_None_with_None_defaults()`?

```
(None, None, None),
([], None, None),
(None, "", None),
([], "", None),
...
```

Coniferish:
> Can we add a test function `test_check_language_args_returns_None()` with more test cases?
> 
> ```
> @pytest.mark.parametrize(
>     ("languages", "ocr_languages", "expected_langs"),
>     [
>       (None, None, None),
>       ([], None, None),
>       (None, "", None),
>       ([], "", None),
>       ...
>    ]
> ```

Per comments from steve, I changed the function so it should not be called with `None` for languages and it is exclusively called with `[]` if `languages` is `None`. But I can add tests to show it will return `None`.

christinestraub:
> Per comments from steve, I changed the function so it should not be called with `None` for languages and it is exclusively called with `[]` if `languages` is `None`. But I can add tests to show it will return `None`.

I'm not sure why the `check_language_args()` should not be called with `None`. The default value for `languages` parameter in `partition()` is `None` so, `check_language_args()` definitely would be called with `None` in `auto.py`

Coniferish:
> I'm not sure why the `check_language_args()` should not be called with `None` for languages. The default value for `languages` parameter in `partition()` is `None` so, `check_language_args()` definitely would be called with `None` for languages in `auto.py`

@christinestraub, here's @scanny's reasoning: https://github.com/Unstructured-IO/unstructured/pull/2435#discussion_r1466706889 lmk if you still think it should be changed, though!

unstructured/partition/lang.py
christinestraub: > I don't think we can do that because of the other partitioners that default to "auto". `None` is used as the default because we have to be able to resolve the languages/ocr_languages (or just languages) at the auto `partition` level. Downstream, once the document type is determined and the document is passed to the appropriate partitioner, the default for `languages` can be either "auto" or "eng". "eng" for the other partitioners would skip language detection since it'd be interpreted as a user telling us they know the document is in English.

@Coniferish @cragwolfe @qued
- By this PR, `None` is used as the default at the auto partition level. Downstream, once the document type is determined and the document is passed to the appropriate partitioner, the `None` for `languages` is passed to all partitioners.
- We already have logic to convert `None` for languages to `auto` in `lang.apply_lang_metadata()` - https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/lang.py#L351-L352

Considering the above two, I think it would be better to set the default value for `languages` to `None` rather than `['auto']` for all partitioners in a future refactoring PR. What do you think?

christinestraub:
> I don't think we can do that because of the other partitioners that default to "auto". `None` is used as the default because we have to be able to resolve the languages/ocr_languages (or just languages) at the auto `partition` level. Downstream, once the document type is determined and the document is passed to the appropriate partitioner, the default for `languages` can be either "auto" or "eng". "eng" for the other partitioners would skip language detection since it'd be interpreted as a user telling us they know the document is in English.

@Coniferish @cragwolfe @qued
- By this PR, `None` is used as the default at the auto partition level. Downstream, once the document type is determined and the document is passed to the appropriate partitioner, the `None` for `languages` is passed to all partitioners.
- We already have logic to convert `None` for languages to `auto` in `lang.apply_lang_metadata()` - https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/lang.py#L351-L352

Considering the above two, I think it would be better to set the default value for `languages` to `None` rather than `['auto']` for all partitioners in a future refactoring PR. What do you think?

Coniferish:
> Considering the above two, I think it would be better to set the default value for `languages` to `None` rather than `['auto']` for all partitioners in a future refactoring PR. What do you think?

@christinestraub, I can't remember offhand why the default was `['auto']`, but changing it to `None` in a follow-up PR sounds good to me. @shreyanid might have something to add on that, though.

PR #2441, opened 2024-01-22
fix: default hi_res model rely on inference setting

badGarnet:
- there are multiple places setting the default `hi_res_model_name` in both `unstructured` and `unstructured-inference`
- they lead to inconsistency and unexpected behaviors
- this fix removes a helper in `unstructured` that tries to set the default hi_res layout detection model; instead we rely on the `unstructured-inference` to provide that default when no explicit model name is passed in

## test

```bash
UNSTRUCTURED_INCLUDE_DEBUG_METADATA=true ipython
```

```python

# find a pdf file
elements = partition("foo.pdf", strategy="hi_res")
assert elements[0].metadata.detection_origin == "yolox"
```


scanny (approved):
LGTM

christinestraub:
@badGarnet The default model on the `unstructured-inference` is ["yolo"`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/base.py#L30C1-L30C24). I think we need to set a different model based on specifying `infer_table_structure` parameter so "yolox" or "yolox_quantized".
Do we just need to set the hi_res default model to "yolox" in the future without relying on the `infer_table_structure` ?parameter? cc: @cragwolfe @qued

badGarnet:
> @badGarnet The default model on the `unstructured-inference` is ["yolox"](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/base.py#L30C1-L30C24). I think we need to set a different model based on specifying `infer_table_structure` parameter so "yolox" or "yolox_quantized". Do we just need to set the hi_res default model to "yolox" in the future without relying on the `infer_table_structure` parameter? cc: @cragwolfe @qued

yes, we want to just use default `yolox` as specified by inference package. Eval so far have not shown the quantized version runs statistically faster so use the same model for all cases is a simpler way to manage the model settings. Additionally we got too many places setting the default models. If down the road we believe there is value for quantized we should add the logic in inference since this is part of inference package's job.

christinestraub (approved):
LGTM!

cragwolfe:
FWIW, the ingest evals in this PR are pretty close but slightly worse:
image

image

Not enough test data to make a definitive call, though.

cragwolfe (approved):

PR #2442, opened 2024-01-22
fix: remove none value keys from flattened dictionary

potter-potter:
When a partitioned or embedded document json has null values, those get converted to a dictionary with None values.

This happens in the metadata. I have not see it in other keys.

Chroma and Pinecone do not like those None values. 

`flatten_dict` has been modified with a `remove_none` arg to remove keys with None values.

Also, Pinecone has been pinned at 2.2.4 because at 3.0 and above it breaks our code.


ahmetmeleq (approved):
Looks good for now, 

let's also discuss (for the future): 
1. if writing metadata values as null to the .json files is what we want (rather than omitting that key altogether). If not, this problem might have an upstream solution where we omit keys with null values altogether rather than writing them with null values.
2. if we actually would like to be able to upload None metadata values (for any reason). If we'd like to upload them, there can be a special string denoting null values to avoid this problem.

PR #2446, opened 2024-01-23
fix: treat yaml files as plain text

### LINKED ISSUE #2412: Add support for `text/yaml` filetype
Currently a `text/yaml` MIME type results in `File type text/yaml is not supported.`. The goal of this issue is to allow for processing YAML files as text files, like we do with other code files.

MthwRobinson:
### Summary

Closes #2412. Adds support for YAML MIME types and treats them as plain text. In response to `500` errors that the API currently returns if the MIME type is `text/yaml`.


qued (approved):
LGTM!

MthwRobinson:
Gotcha yeah may take a pass at updating that to allow any MIME type allowed by `unstructured`.

PR #2455, opened 2024-01-24
WIP: enhancement: detect headers in `partition_pdf` with fast strategy

MthwRobinson:
### Summary

Detects headers and footers when using `partition_pdf` with the fast strategy. Identifies elements that are positioned in the top or bottom 5% of the page as headers or footers. If no coordinate information is available, an element won't be detected as a header or footer.


MthwRobinson:
@qued - Draft in response to a user who had headers and footers identified as titles when using the fast strategy. Wanted to check first on whether I was missing an subtleties about the coordinate system.

unstructured/partition/text.py
```diff
@@ -206,12 +208,56 @@ def is_empty_bullet(text: str) -> bool:
...
+    coordinates: Optional[Tuple[Tuple[float, float], ...]] = None,
+    coordinate_system: Optional[CoordinateSystem] = None,
+    threshold: float = 0.05,
+) -> bool:
+    """Checks to see if the position of the text indicates that the text belongs
+    to a header."""
+    if coordinates is None or coordinate_system is None:
+        return False
+
+    min_y = min([coordinate[1] for coordinate in coordinates])
```
qued, line 222: Am I reading this right that it only checks the y-coordinate of the upper left corner of the bbox? Seems like if the bbox is really tall and only the top edge is in the zone, we could have a false positive.

unstructured/partition/text.py
```diff
@@ -206,12 +208,56 @@ def is_empty_bullet(text: str) -> bool:
...
+    coordinates: Optional[Tuple[Tuple[float, float], ...]] = None,
+    coordinate_system: Optional[CoordinateSystem] = None,
+    threshold: float = 0.95,
+) -> bool:
+    """Checks to see if the position of the text indicates that the text belongs
+    to a footer."""
+    if coordinates is None or coordinate_system is None:
+        return False
+
+    max_y = max([coordinate[1] for coordinate in coordinates])
```
qued, line 238: Same issue as above (but let me know if I'm interpreting it wrong).

unstructured/partition/text.py
MthwRobinson: Hah! Totally forgot about this PR. Yeah it's just checking the top/bottom corner right now. What about checking the mid-point between the y coordinates instead?

MthwRobinson:
@qued - Made those updates. Do you know what commit message I need to trigger the test fixture update flow? Forgot how to do that.

qued (approved):
LGTM!

PR #2460, opened 2024-01-26
fix: link_texts was breaking postgres destination connector

potter-potter:
Formatting of link_texts was breaking metadata storage. Turns out it didn't need any conforming and came in correctly from json.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.3-dev6
+## 0.12.3-dev7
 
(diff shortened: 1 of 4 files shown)


ryannikolaidis (approved):
nice, thanks for the catch and cleanup here!

ryannikolaidis:
actually: should we add a test that would have failed?

potter-potter:
> actually: should we add a test that would have failed?

yeah. was definitely on my mind :)

PR #2462, opened 2024-01-26
feat: add the ability to specify a custom OCR besides the ones natively supported

cjolif:
This is nice to natively support both Tesseract and Paddle. However, one might already use another OCR and might want to keep using it (for quality reasons, for cost reasons etc...). 
This PR adds the ability for the user to specify its own OCR agent implementation that is then called by unstructured.

I am new to unstructured so don't hesitate to let me know if you would prefer this being done differently and I will rework the PR.


badGarnet:
@cjolif to get the unit test running the CI needs to pass changelog test -> which basically requires changelog entry when there are logic changes. Once we got that it can unblock CI runs. And I cloned your branch in case we need to update ingest fixtures (we likely won't need it but just in case)

unstructured/partition/pdf_image/ocr.py
```diff
@@ -692,81 +409,18 @@ def supplement_layout_with_ocr_elements(
...
-
 def get_ocr_agent() -> str:
-    ocr_agent = env_config.OCR_AGENT.lower()
-    if ocr_agent not in [OCR_AGENT_PADDLE, OCR_AGENT_TESSERACT]:
+    ocr_agent_module = env_config.OCR_AGENT
+    # deal with compatibility with origin way to set OCR
+    if ocr_agent_module.lower() == OCR_AGENT_TESSERACT_OLD:
+        ocr_agent_module = OCR_AGENT_TESSERACT
+    elif ocr_agent_module.lower() == OCR_AGENT_PADDLE_OLD:
+        ocr_agent_module = OCR_AGENT_PADDLE
```
badGarnet, line 418: I'd suggest we also add a warning message here to let users know they are using a pattern that is outdated and would be deprecated later like 
```suggestion
        logger.warning("OCR agent name %s is outdated and will be deprecated in a future release; please use %s instead", ocr_agent_module, OCR_AGENT_TESSERACT)
        ocr_agent_module = OCR_AGENT_TESSERACT
    elif ocr_agent_module.lower() == OCR_AGENT_PADDLE_OLD:
        logger.warning("OCR agent name %s is outdated and will be deprecated in a future release; please use %s instead", ocr_agent_module, OCR_AGENT_PADDLE)
        ocr_agent_module = OCR_AGENT_PADDLE
```

unstructured/partition/utils/ocr_models/ocr_interface.py
```diff
@@ -0,0 +1,120 @@
...
+        pass
+
+    @abstractmethod
+    def get_layout_elements_from_image(
+        self, image: PILImage, ocr_languages: str = "eng"
+    ) -> List[LayoutElement]:
+        pass
+
+    @staticmethod
+    @functools.lru_cache(maxsize=None)
```
badGarnet, line 45: My guess is lru cache here is to avoid loading the module multiple times? It does raise a memory lifespan concern: lru cache is not easy to clean; if the python runtime's lifespan and ocr module is the same it is not a big issue; but if the python runtime is supposed to handle followup tasks that might be memory hungry (therefore need cache like this to be released) it can get messy.
One way we can deal with this is to make the `OCRAgent` a singleton pattern https://python-patterns.guide/gang-of-four/singleton/, which doesn't leave cache hidden. Often the memory issue is taken care of by `gc` in this case.

unstructured/partition/utils/ocr_models/ocr_interface.py
```diff
@@ -0,0 +1,120 @@
...
+    def get_instance(ocr_agent_module: str) -> "OCRAgent":
+        module_name, class_name = ocr_agent_module.rsplit(".", 1)
+        module = importlib.import_module(module_name)
```
badGarnet, line 48: since this line is pretty generous in terms of code execution we should add a logic before it to validate the module name is in a whitelist. 
This whitelist could be something can be overwritten by an env so user can expand it.

unstructured/partition/utils/ocr_models/ocr_interface.py
cjolif: Yes, the reason I went with this pattern and the one you mentioned is that I got inspiration from code already here: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/utils/ocr_models/paddle_ocr.py#L8. But if you confirm you would like me to change it I'm happy to do it, just let me know.

unstructured/partition/utils/ocr_models/ocr_interface.py
cjolif: I've pushed a commit that should solve this.

badGarnet:
@cjolif I pushed a change to lint the file `ocr.py`. I would recommend you to install pre-commit via `pre-commit install` for the repo. That will help format and check liniting of changed files. This makes sure the code is formatted in a consistent way. It is also a nice tool to spot misspelled variable names, etc.

badGarnet:
The tests that require unstructured secrets are passed here on the cloned branch: https://github.com/Unstructured-IO/unstructured/pull/2465

unstructured/partition/utils/ocr_models/ocr_interface.py
badGarnet: actually since this is an existing pattern lets get this in first; and refactor that pattern in a future PR

cjolif:
thanks a lot @badGarnet I added a bit more wording to the CHANGELOG in my last commit. Let me know if that is fine with you.

badGarnet (approved):
note: some of the CI requires unstructured secrets to run and they all passed in this cloned branch https://github.com/Unstructured-IO/unstructured/pull/2465

PR #2473, opened 2024-01-29
chore: new black changes

qued:
Update `black` and apply changes to affected files. I separated this PR so we can have a look at the changes and decide whether we want to:
1. Go forward with the new formatting
2. Change the black config to make the old formatting valid
3. Get rid of black entirely and just use `ruff`
4. Do something I haven't thought of


ryannikolaidis (approved):

typings/pptx/oxml/__init__.py
```diff
@@ -6,4 +6,4 @@
 
 
 def parse_xml(xml: Union[str, bytes]) -> etree._Element:
-    ...
+    ...  # fmt: skip
```
qued, line 9: I added a skip here, as the new `black` formatting conflicts with `flake8` E701/E704 (see e.g. [here](https://github.com/alpha-unito/streamflow/pull/362/files/936c8bf78aa5e71051bacbe7ebb7bfaf6bd192d8#diff-b60cf2e5a462ab7c449ed58b3f230db99bdcfa0a22b87022cfb656c20e120a3b)). Resolving this doesn't seem as simple as ignoring these codes in `flake8`; my quick attempt at that caused several other `flake8` errors to surface. I decided the easiest way forward would be to have black ignore affected lines.

Issue #2362, opened 2024-01-07
bug/partition_html only partially extracts text of element with nested inline tag


dgunning:
**Describe the bug**
When a html contains a div with a nested  tag e.g. a **b**  or **span**, the element created from the div contains only the text up to the inline element. 

This is likely a behaviour of lxml related to these issues

- [lxml doesn't get all text if text has br](https://stackoverflow.com/questions/29555452/lxml-doesnt-get-all-text-in-element-if-text-has-br)
- [get all text inside a tag in lxml](https://stackoverflow.com/questions/4624062/get-all-text-inside-a-tag-in-lxml)

**To Reproduce**

```python
html = """
<html>
<body>
    <div>
    the Company issues shares at $<div style="display:inline;"><span>5.22</span></div> per share. 
    There is more text
    </div>
</body>
</html>
"""

elements = partition_html(text=html)
print(''.join(e.text for e in elements))
```

This results in

"the Company issues shares at $5.22"

**Expected behavior**
The following should be printed

"the Company issues shares at $5.22 per share. There is more text"

**Screenshots**

**Environment Info**
Please run `python scripts/collect_env.py` and paste the output here. 
This will help us understand more about the environment in which the bug occurred.

**Additional context**

kingle-zhuang:
![1704791657913](https://github.com/Unstructured-IO/unstructured/assets/142277746/0c382804-79b7-446d-bcf3-bcc9ede8d79c)


same issue with it

kingle-zhuang:
missing in the left navigation bar

LaverdeS:
Apparently the problem extends to other examples not using `<div>`:

```
>>> str(partition_html(text='<p>abc <a href="somewhere">def</a><br><strong></strong><strong>ghi</strong></p>')[0])
'abc'
>>> str(partition_html(text='<p>abc <a href="somewhere">def</a><br>ghi</p>')[0])
'abc'
>>> str(partition_html(text='<p>abc <br>ghi</p>')[0])
'abc'
>>> str(partition_html(text='<p>abc ghi</p>')[0])
'abc ghi'
>>> str(partition_html(text='<p>abc <a href="somewhere">def</a> ghi</p>')[0])
'abc def ghi'
>>> str(partition_html(text='<p>abc <br/>ghi</p>')[0])
'abc'
```

[...]

PR #2483, opened 2024-01-31
convert tail to text/emphasised tag and process all emphasised descendants

stdweird:
cfr discussion at #2362

### DIFF
--- a/unstructured/documents/html.py
+++ b/unstructured/documents/html.py
@@ -37,7 +37,13 @@ from unstructured.partition.text_type import (
 )
 from unstructured.utils import htmlify_matrix_of_cell_texts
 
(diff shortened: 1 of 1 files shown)


cragwolfe:
@stdweird , thanks for the contribution! Do you have an html doc handy that this PR fixes, which could get added to unittests?

stdweird:
@cragwolfe not sure if you want real data or not, but eg
```
<html>
  <body>
    <div>a
      <ul>b
        <li>c1</li>d1
        <li>c2</li>d2
      </ul>e
     </div>f<br>g
  </body>
</html>
```

the main intend should be to keep the result as close as possible to the orignial text (eg `efg` after the list items) , but right now retrieving all text is higher priority (at least for me). i don't think even this PR does that (`g` is still lost i think), but it is already an improvement.

PR #2497, opened 2024-02-02
bug: fix check_connection for 

potter-potter:
fixes check_connection for:
azure
opensearch
postgres

For Azure, the check_connection in fsspec.py actually worked better. 

Adding check_connection for Databricks Volumes


unstructured/ingest/connector/fsspec/azure.py
```diff
@@ -54,16 +54,6 @@ def get_file(self):
...
-    @requires_dependencies(["adlfs"], extras="azure")
-    def check_connection(self):
-        from adlfs import AzureBlobFileSystem
-
-        try:
-            AzureBlobFileSystem(**self.connector_config.get_access_config())
-        except ValueError as connection_error:
-            logger.error(f"failed to validate connection: {connection_error}", exc_info=True)
-            raise SourceConnectionError(f"failed to validate connection: {connection_error}")
-
```
ahmetmeleq, line 66: When it works, works well, however in terms of failing it took 90 seconds for check_connection to fail for Azure in my case when I put an incorrect account name

@ryannikolaidis do 90secs work for this, or do we want ~ < 10 seconds of success/fail time?

potter-potter: I assume you're talking about the check_connection in fsspec since we removed this check_connection from azure.py. (this one just gets the config and doesn't actually try to connect...)

Also, is there an alternative method? Or should I hunt for one? (the fsspec one does an `ls` to determine success/failure.)

unstructured/ingest/connector/opensearch.py
```diff
@@ -140,7 +140,7 @@ def ops(self):
 
     def check_connection(self):
         try:
-            self.ops.ping()
+            assert self.ops.ping()
```
ahmetmeleq, line 143: Works well

unstructured/ingest/connector/sql.py
```diff
@@ -93,9 +93,13 @@ def initialize(self):
...
-        cursor = self.client.cursor()
-        cursor.execute("SELECT 1;")
-        cursor.close()
+        try:
+            cursor = self.client.cursor()
+            cursor.execute("SELECT 1;")
+            cursor.close()
+        except Exception as e:
+            logger.error(f"failed to validate connection: {e}", exc_info=True)
+            raise DestinationConnectionError(f"failed to validate connection: {e}")
```
ahmetmeleq, line 102: Works well

unstructured/ingest/connector/fsspec/azure.py
ahmetmeleq: Yes

I've tried various values from adlfs and it doesn't work. Its upstream library, azure sdk has timeout functionality for these operations, but adlfs doesn't expose them. Something like this might work:

1. Creating an azure blob service client like in here: https://github.com/fsspec/adlfs/blob/576fb7a6a53a55375b4458c09e5bb571d945d410/adlfs/spec.py#L470-L477
2. Maybe ls containers like in [here](https://github.com/Azure/azure-sdk-for-python/blob/9d782b55092bd05782d31c2e4267e34ae26b0f8b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py#L416) but with the timeout [arg](https://github.com/Azure/azure-sdk-for-python/blob/9d782b55092bd05782d31c2e4267e34ae26b0f8b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py#L443)

However this solution would add a dependency and might take additional time to implement

ahmetmeleq: Basically if 90 secs of timeout is reasonable we can go with the current state. 

If not, we might need an additional implementation and dependency like mentioned above. I'd like to hear @ryannikolaidis's opinion on if 90 secs is acceptable

ahmetmeleq: If `check_connection` functions are aimed to be run at the background, 90 secs of timeout could be fine. If not, we might need to fix it

ryannikolaidis: a consumer might literally be waiting for this to complete to validate moving forward. so, ideally no, would prefer the fastest possible method to validate a configuration. is there a quick path to addressing that? otherwise I would suggest checking this in (so we have _a_ way to test this) and opening new ticket to optimize.

potter-potter: @ryannikolaidis @ahmetmeleq I ran the _check_connection_ for azure this morning. Both the success and failure checks were instantaneous. (boot up of unstructured itself always takes about 4 seconds)

potter-potter: All I'm doing to check is to hijack the _instantiate_ to run self.check_connection. With a breakpoint at the beginning of check_connection. and then either letting it go (success) or changing something on the object (failure).

potter-potter: Also added databricks-volumes to this PR since it was easy.

potter-potter: also set fsspec .ls detail to False. Should help with .ls speed

ryannikolaidis (approved):

PR #2507, opened 2024-02-05
added parent_element as output for overlapping cases

erjieyong:
To provide more utility to the `catch_overlapping_and_nested_bboxes` and `identify_overlapping_or_nesting_case` functions, included parent_element as part of the output. 

This would allow user to 
- identify the parent element in the overlapping case: `nested {type*} in {type*}`. Currently, if the element types is similar, an example case output would be `nested Image in Image` which is confusing. 
- easily identify elements to keep or delete


erjieyong:
have added test cases to `test_utils.py`

erjieyong:
@christinestraub , please let me know if there's any changes needed on my end to get this merged. I am quite new to this, so appreciate your guidance.

christinestraub:
@erjieyong Can you please fix the changelog conflicts?

christinestraub (approved):
Note: some of the CI requires unstructured secrets to run and they all passed in this cloned branch - https://github.com/Unstructured-IO/unstructured/pull/2548

PR #2509, opened 2024-02-05
refactor: isolate ingest dependencies into local scopes

ahmetmeleq:
This PR: 
- Moves ingest dependencies into local scopes to be able to import ingest connector classes without the need of installing imported external dependencies. This allows lightweight use of the classes (not the instances. to use the instances as intended you'll still need the dependencies).
- Upgrades the embed module dependencies from `langchain` to `langchain-community` module (to pass CI [rather than introducing a pin])
- Does pip-compile
- Does minor refactors in other files to pass `ruff 2.0` checks which were introduced by pip-compile


potter-potter:
This is neat. But is there more to this? Seems like the bulk of it is just tests.
Also, why did we want to do this? Will this help with check_connection?

ahmetmeleq:
> But is there more to this?

In the future notion connector imports can be moved as well, skipped that for now since the notion connector has up to 63 files to refactor.

These changes basically aim for us to be able to import ingest connector classes without installing external dependencies of them (even though, to use the instances we'll still need the dependencies). 

Being able to import the classes without a requirement on external dependencies provides lightweight schema support - meaning whenever we'd like to obtain information about the class structure, we can get it without installing those dependencies

unstructured/embed/bedrock.py
```diff
@@ -12,7 +12,7 @@
 from unstructured.utils import requires_dependencies
 
 if TYPE_CHECKING:
-    from langchain.embeddings import BedrockEmbeddings
```
potter-potter, line 15: Thanks for doing this. Was tired of seeing the deprecation message.

potter-potter (approved):
Very nice. Thanks for cleaning up all those little enumerations, etc...

PR #2510, opened 2024-02-05
feat: include text from shapes in docx

ds-filipknefel:
Reported bug: Text from docx shapes is not included in the `partition` output.
Fix: Extend docx partition to search for text tags nested inside structures responsible for creating the shape.


ds-filipknefel:
@scanny 

I've settled on this xpath for now: `w:r/descendant::wp:inline[ancestor::w:drawing][1]//w:r`

It'll find first occurence of `w:drawing/wp:inline` for each `w:r` directly under current paragraph and return `w:r` from inside it.
- Each single image should be contained within `w:r` directly under paragraph so I think `w:r/` a reasonable limitation
- In case both choice and fallback use `w:drawing` (I think it's unlikely but maybe not impossible) we need to select the first one `descendant::wp:inline[ancestor::w:drawing][1]`
- We take `wp:inline` to only include inline shapes
- At the end I take `w:r` at the end because we want to use `.text` method of `C_R` the same way we do for other regions.

I've added two example documents and one test for each.

Also this was reported as a bug but I wonder if this should be considered a fix or a feature, what's your take? I'll update changelog accordingly.

scanny (commented):
A few remarks :)

scanny:
Hi @ds-filipknefel that XPath expression looks good to me. I think the trailing `//w:r` is fine because the scope is so narrow by then that the search will be quick.

Btw, to understand the possible parent-child relationships in the XML you can consult the XML Schema for Open XML here: https://github.com/python-openxml/python-docx/blob/master/ref/xsd/wml.xsd. That said I feel pretty confident this XPath will do just what we want.

I'd call this a feature, albeit a small one. This is extending the range of what `partition_docx()` will detect. There are quite a few other corner-caes items it still won't detect and I don't believe we have either promised that it does or committed to make it do so (given the large engineering investment required). So I'd say we celebrate each new extension of its detection capability as a new goodness :)

unstructured/partition/docx.py
```diff
@@ -330,7 +330,12 @@ def _classify_paragraph_to_element(self, paragraph: Paragraph) -> Iterator[Eleme
         does not contribute to the document-element stream and will not cause an element to be
         emitted.
         """
-        text = paragraph.text
+        text = "".join(
+            e.text
+            for e in paragraph._p.xpath(
+                "w:r" "| w:hyperlink" "| w:r/descendant::wp:inline[ancestor::w:drawing][1]//w:r"
```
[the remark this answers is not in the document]
ds-filipknefel: Fixed, didn't realise when formatter snapped these back into single line.

test_unstructured/partition/docx/test_docx.py
```diff
@@ -764,6 +764,31 @@ def test_partition_docx_includes_hyperlink_metadata():
     assert metadata.link_urls is None
 
 
+# -- shape behaviors -----------------------------------------------------------------------------
+
+
+def test_include_inline_shapes_text():
+    elements = partition_docx(example_doc_path("docx-inline-shapes.docx"))
```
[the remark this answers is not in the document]
ds-filipknefel: Merged into a single .docx and included in PR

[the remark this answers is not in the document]
scanny: @ds-filipknefel Okay, now that I see the document I can better see what we're trying to do. So this is like a PowerPoint shape inserted in the text, not an image (PNG, JPG) per se. So that's not going to be complicated. Each shape will have (effectively) a text frame that can contain text, so there's no "text before the image", "text after the image" etc. (because there is no image).

So I think the test cases we have are fine. There just isn't more possible variation that we need to consider.

test_unstructured/partition/docx/test_docx.py
```diff
@@ -764,6 +764,32 @@ def test_partition_docx_includes_hyperlink_metadata():
...
+
+
+def test_exclude_floating_shapes_text():
+    elements = partition_docx(example_doc_path("docx-shapes.docx"))
+
+    assert len(elements) == 3
+
+    paragraph = elements[2]
+
+    assert paragraph.text == "Paragraph with floating shape attached."
```
scanny, line 790: Okay, now I think this simplifies to:
```python
def test_it_considers_text_inside_shapes():
    # -- <bracketed> text is written inside inline shapes --
    assert partition_docx(example_doc_path("docx-shapes.docx")) == [
        NarrativeText("Paragraph with single <inline-image> within.")
        NarrativeText("Paragraph with <inline-image1> and <inline-image2> within.")
        # -- text "<floating-shape>" in floating shape is ignored --
        NarrativeText("Paragraph with floating shape attached.")
    ]
```

You might need to adjust the Element sub-types to match, they might not be classified as NarrativeText.

scanny (commented):
Okay, I think this is good to go. One remark on test that can now be simplified.

If we're ready, you can change it from Draft to "ready" status and let me know and I'll approve it :)

test_unstructured/partition/docx/test_docx.py
ds-filipknefel: I don't now if this is necessarily better.

1. Regardless of how these paragraphs get classified (e.g. here first two are `Title`, third one is `NarrativeText`) their text should remain the same, so looking at text directly avoids test breaking on unrelated change.
2. If this detection breaks and text is different `pytest -vv` will tell you only that the objects differ like this:
  ```
  E         -     <unstructured.documents.elements.Title object at 0x7f2b326af520>,
  E         ?                                                              ^^ ^^
  E         +     <unstructured.documents.elements.Title object at 0x7f2b324f3040>,
  ```
which isn't very helpful until you unpack it.

scanny: You can avoid coupling to the exact element type by changing it to:
```python
assert [element.text for element in partition_docx(...)] == [
    "string 1",
    "string 2",
    "string 3",
]
```
That's probably not a bad idea since this particular behavior should have no direct influence on which element the paragraph is classified as.

ds-filipknefel: Alright, I merged these tests and we're checking only text content.

CHANGELOG.md
```diff
@@ -37,6 +38,7 @@
 * **Add title to Vectara upload - was not separated out from initial connector **
 * **Fix change OpenSearch port to fix potential conflict with Elasticsearch in ingest test **
 
+
```
scanny, line 41: I don't believe we use double-blank-lines between releases. Check the others in this file and make this one consistent.

scanny (approved):
Nice. Amazingly compact :)

One small nit but approving in advance.

PR #2514, opened 2024-02-06
add clean_newline_words

manuelrech:
Since often it happens that words end on a new line, it would be good to have a function for that


MthwRobinson:
@manuelrech - If you're still interested in working on this, could you add a test and a changelog update?

PR #2515, opened 2024-02-06
docs: add information about MIME type of extracted images

ds-filipknefel:
Include information about what mime type is expected when extracting images.


christinestraub (approved):
LGTM

PR #2517, opened 2024-02-07
fix: change opensearch port

potter-potter:
change opensearch port to see if fixes CI. We think there may be a conflict with the elasticsearch docker port.

Also adding simple retry to vector query.


ahmetmeleq (approved):
LGTM

ryannikolaidis:
is there a theory why this broke?

potter-potter:
> is there a theory why this broke?

We think there is some sort of port conflict. Maybe the docker image for elasticsearch isn't fully getting taken down?

PR #2521, opened 2024-02-07
enhancement: process `.p7s` files with `partition_email`

### LINKED ISSUE #2489: email partitioner / partition_eml .ps7 mime type support
When using an eml file with attachment, partition_eml, I get an .UNK filetype error. 
After discussion with unstructured, it seems to require *.ps7 mime attachment* support.

I use Langchain UnstructuredEmailLoader, 
I tried using directly unstructured.io partition_eml, 
got the same result. 

Need this to work , not producing the .UNK filetype error and reading the attachment. 

```python
elements = partition_email(
        filename=eml_file, 
        process_attachments=True, 
        attachment_partitioner=partition)
    
    print(elements)
```

I used python *email* to test if I could read my eml ok and yes. 

**Additional context**

```code

Error msg : 
Une exception s'est produite : ValueError
Invalid file /var/folders/2v/f76959fd4h320hm6wlk20p3w0000gn/T/tmp5ivv6vqw/smime.p7s. The FileType.UNK file type is not supported in partition.
  File "/GitHub/simple-doc/document-classifier/main.py", line 52, in main
    doc = loader.load()
  File "/GitHub/simple-doc/document-classifier/main.py", line 139, in <module>
    main()
ValueError: Invalid file /var/folders/2v/f76959fd4h320hm6wlk20p3w0000gn/T/tmp5ivv6vqw/smime.p7s. The FileType.UNK file type is not supported in partition.

```

MthwRobinson:
### Summary

Closes #2489, which reported an inability to process `.p7s` files. PR implements two changes:

- If the user selected content type for the email is not available and there is another valid content type available, fall back to the other valid content type.
- For signed message, extract the signature and add it to the metadata

### Testing

```python

filename = "example-docs/eml/signed-doc.p7s"
elements = partition(filename=filename) # should get a message about fall back logic
print(elements[0]) # "This is a test"
elements[0].metadata.to_dict() # Will see the signature
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
-## 0.12.4-dev8
+## 0.12.4-dev9
 
 ### Enhancements
 
 * **Apply New Version of `black` formatting** The `black` library recently introduced a new major version that introduces new formatting conventions. This change brings code in the `unstructured` repo into compliance with the new conventions.
 * **Move ingest imports to local scopes** Moved ingest dependencies into local scopes to be able to import ingest connector classes without the need of installing imported external dependencies. This allows lightweight use of the classes (not the instances. to use the instances as intended you'll still need the dependencies).
+* **Add support for `.p7s` files** `partition_email` can now process `.p7s` files. The signature for
+  the signed message is extracted and added to metadata.
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,8 @@
 
 * **Apply New Version of `black` formatting** The `black` library recently introduced a new major version that introduces new formatting conventions. This change brings code in the `unstructured` repo into compliance with the new conventions.
 * **Move ingest imports to local scopes** Moved ingest dependencies into local scopes to be able to import ingest connector classes without the need of installing imported external dependencies. This allows lightweight use of the classes (not the instances. to use the instances as intended you'll still need the dependencies).
-* **Add support for `.p7s` files** `partition_email` can now process `.p7s` files. The signature for
-  the signed message is extracted and added to metadata.
-* **Fallback to valid content types for emails** If the user selected content type does not exist on
-  the email message, `partition_email` now falls back to anoter valid content type if it's
-  available.
+* **Add support for `.p7s` files** `partition_email` can now process `.p7s` files. The signature for the signed message is extracted and added to metadata.
+* **Fallback to valid content types for emails** If the user selected content type does not exist on the email message, `partition_email` now falls back to anoter valid content type if it's available.
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 * **Move ingest imports to local scopes** Moved ingest dependencies into local scopes to be able to import ingest connector classes without the need of installing imported external dependencies. This allows lightweight use of the classes (not the instances. to use the instances as intended you'll still need the dependencies).
 * **Add support for `.p7s` files** `partition_email` can now process `.p7s` files. The signature for
   the signed message is extracted and added to metadata.
+* **Fallback to valid content types for emails** If the user selected content type does not exist on
+  the email message, `partition_email` now falls back to anoter valid content type if it's
+  available.
 
 ### Features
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
...
 
 ### Enhancements
 
 * **Apply New Version of `black` formatting** The `black` library recently introduced a new major version that introduces new formatting conventions. This change brings code in the `unstructured` repo into compliance with the new conventions.
 * **Move ingest imports to local scopes** Moved ingest dependencies into local scopes to be able to import ingest connector classes without the need of installing imported external dependencies. This allows lightweight use of the classes (not the instances. to use the instances as intended you'll still need the dependencies).
+* **Add support for `.p7s` files** `partition_email` can now process `.p7s` files. The signature for
+  the signed message is extracted and added to metadata.
+* **Fallback to valid content types for emails** If the user selected content type does not exist on
+  the email message, `partition_email` now falls back to anoter valid content type if it's
+  available.
```
scanny, line 11: I think these line-breaks will be preserved during rendering the markdown, so probably best to keep each on a single line so the renderer takes care of word-wrap on the display-time device.

test_unstructured/partition/test_auto.py
```diff
@@ -1282,3 +1282,10 @@ def test_partition_image_with_bmp_with_auto(
     table = [el.metadata.text_as_html for el in elements if el.metadata.text_as_html]
     assert len(table) == 1
     assert "<table><thead><th>" in table[0]
+
+
+def test_auto_partition_eml_add_signature_to_metadata(filename="example-docs/eml/signed-doc.p7s"):
+    elements = partition(filename=filename)
```
scanny, line 1288: I know a lot of historical tests do this, but there's no reason to add filename as an argument to the test, at least that I've ever seen. I would just inline it like this:
```python
def test_auto_partition_eml_add_signature_to_metadata():
    elements = partition("example-docs/eml/signed-doc.p7s")
```

It's one less article of state to be kept in mind by the reader and reads very directly, like "partition this" :)

unstructured/partition/email.py
```diff
@@ -118,6 +118,14 @@ def build_email_metadata(
     metadata_last_modified: Optional[str] = None,
 ) -> ElementMetadata:
     """Creates an ElementMetadata object from the header information in the email."""
+    signature = None
+    payload = msg.get_payload()
+    if isinstance(payload, list):
+        for item in payload:
+            if item.get_content_type().endswith("signature"):
+                signature = item.get_payload()
+                break
```
scanny, line 127: This is just complicated enough that it could be extracted to a function of its own, even right here internal to this function:
```python
    def find_signature(msg: Message) -> Optional[str]:
        payload = msg.get_payload()
        if not isinstance(payload, list):
            return None
        for item in payload:
            if item.get_content_type().endswith("signature"):
                return item.get_payload()
        return None
```

The benefit being:
- it encapsulates this particular step
- it's easy to extract to module-level later if we end up needing it elsewhere.
- Gives a place for a docstring to explain the contract, like when it works, when it doesn't, etc.
- a few other more subtle benefits

Then the reference below becomes `signature=find_signature(msg),`

Not a request for a change, just something I would be likely to do in a case like this.

unstructured/partition/email.py
```diff
@@ -367,7 +376,18 @@ def partition_email(
         else:
             content_map[content_type] = part.get_payload()
 
-    content = content_map.get(content_source, "")
+    if content_source in content_map:
+        content = content_map.get(content_source)
+    # NOTE(robinson) - If a the chosen content source is not available and there is
```
scanny, line 381: an extra article in there "a the"

scanny (approved):
A few remarks. Otherwise LGTM, approving in advance.

test_unstructured/partition/test_auto.py
MthwRobinson: Good call on that! Not sure when or why we started doing that.

unstructured/partition/email.py
MthwRobinson: Good suggestion! Pulled that out into its own function.

PR #2524, opened 2024-02-07
rfctr(xlsx): cleaning in prep for XLSX algorithm replacement

scanny:
**Reviewers:** It may be faster to review each of the three commits separately since they are groomed to only make one type of change each (typing, docstrings, test-cleanup).

**Summary**

There are a cluster of bugs in `partition_xlsx()` that all derive from flaws in the algorithm we use to detect "subtables". These are encountered when the user wants to get multiple document-elements from each worksheet, which is the default (argument `find_subtable = True`).

These commits clean up typing, lint, and other non-behavior-changing aspects of the code in preparation for installing a new algorithm that correctly identifies and partitions contiguous sub-regions of an Excel worksheet into distinct elements.

**Additional Context**

This is a summary of the failure cases. There are a few other cases but they're closely related and this was enough evidence and scope for my purposes:
```python
    #
    # -- ✅ CASE 1: There are no leading or trailing single-cell rows.
    #       -> this subtable functions never get called, subtable is emitted as the only element
    #
    #    a b  -> Table(a, b, c, d)
    #    c d

    # -- ✅ CASE 2: There is exactly one leading single-cell row.
    #       -> Leading single-cell row emitted as `Title` element, core-table properly identified.
    #
    #    a    -> [ Title(a),
    #    b c       Table(b, c, d, e) ]
    #    d e

    # -- ❌ CASE 3: There are two-or-more leading single-cell rows.
    #       -> leading single-cell rows are included in subtable
    #
    #    a    -> [ Table(a, b, c, d, e, f) ]
    #    b
    #    c d
    #    e f

    # -- ❌ CASE 4: There is exactly one trailing single-cell row.
    #      -> core table is dropped. trailing single-cell row is emitted as Title
    #         (this is the behavior in the reported bug)
    #
    #    a b  -> [ Title(e) ]
    #    c d
    #      e

    # -- ❌ CASE 5: There are two-or-more trailing single-cell rows.
    #      -> core table is dropped. trailing single-cell rows are each emitted as a Title
    #
    #    a b  -> [ Title(e),
    #    c d       Title(f) ]
    #      e
    #      f

    # -- ✅ CASE 6: There are exactly one each leading and trailing single-cell rows.
    #      -> core table is correctly identified, leading and trailing single-cell rows are each
    #         emitted as a Title.
    #
    #      a  -> [ Title(a),
    #    b c       Table(b, c, d, e),
    #    d e       Title(f) ]
    #    f

    # -- ✅ CASE 7: There are two leading and one trailing single-cell rows.
    #      -> core table is correctly identified, leading and trailing single-cell rows are each
    #         emitted as a Title.
    #
    #    a    -> [ Title(a),
    #    b         Title(b),
    #    c d       Table(c, d, e, f),
    #    e f       Title(g) ]
    #      g

    # -- ✅ CASE 8: There are two-or-more leading and trailing single-cell rows.
    #      -> core table is correctly identified, leading and trailing single-cell rows are each
    #         emitted as a Title.
    #
    #      a  -> [ Title(a),
    #      b       Title(b),
    #    c d       Table(c, d, e, f),
    #    e f       Title(g),
    #    g         Title(h) ]
    #    h

    # -- ❌ CASE 9: Single-row subtable, no single-cell rows above or below.
    #      -> First cell is mistakenly emitted as title, remaining cells are dropped.
    #
    #    a b c  -> [ Title(a) ]

    # -- ❌ CASE 10: Single-row subtable with one leading single-cell row.
    #      -> Leading single-row cell is correctly identified as title, core-table is mis-identified
    #         as a `Title` and truncated.
    #
    #    a      -> [ Title(a),
    #    b c d       Title(b) ]
```

### DIFF
--- a/test_unstructured/partition/xlsx/test_xlsx.py
+++ b/test_unstructured/partition/xlsx/test_xlsx.py
@@ -1,6 +1,7 @@
 import sys
 
 import pytest
+from pytest_mock import MockerFixture
 
 from test_unstructured.partition.test_constants import (
     EXPECTED_TABLE_XLSX,
@@ -17,8 +18,16 @@ EXPECTED_FILETYPE = "application/vnd.openxmlformats-officedocument.spreadsheetml
 EXCEPTED_PAGE_NAME = "Stanley Cups"
 
 
-def test_partition_xlsx_from_filename(filename="example-docs/stanley-cups.xlsx"):
-    elements = partition_xlsx(filename=filename, include_header=False)
+# ------------------------------------------------------------------------------------------------
+# INTEGRATION TESTS
+# ------------------------------------------------------------------------------------------------
+# These test `partition_xlsx()` as a whole by calling `partition_xlsx()` and inspecting the
+# outputs.
+# ------------------------------------------------------------------------------------------------
+
+
+def test_partition_xlsx_from_filename():
+    elements = partition_xlsx("example-docs/stanley-cups.xlsx", include_header=False)
 
     assert sum(isinstance(element, Table) for element in elements) == 2
     assert len(elements) == 4
@@ -32,17 +41,17 @@ def test_partition_xlsx_from_filename(filename="example-docs/stanley-cups.xlsx")
     assert elements[1].metadata.filename == "stanley-cups.xlsx"
 
 
-def test_partition_xlsx_from_filename_with_emoji(filename="example-docs/emoji.xlsx"):
-    elements = partition_xlsx(filename=filename, include_header=False)
+def test_partition_xlsx_from_filename_with_emoji():
+    elements = partition_xlsx("example-docs/emoji.xlsx", include_header=False)
     assert sum(isinstance(element, Text) for element in elements) == 1
     assert len(elements) == 1
     assert clean_extra_whitespace(elements[0].text) == "🤠😅"
 
 
-def test_partition_xlsx_from_filename_with_metadata_filename(
-    filename="example-docs/stanley-cups.xlsx",
-):
-    elements = partition_xlsx(filename=filename, metadata_filename="test", include_header=False)
+def test_partition_xlsx_from_filename_with_metadata_filename():
+    elements = partition_xlsx(
+        "example-docs/stanley-cups.xlsx", metadata_filename="test", include_header=False
+    )
 
     assert sum(isinstance(element, Table) for element in elements) == 2
     assert sum(isinstance(element, Title) for element in elements) == 2
@@ -58,11 +67,10 @@ def test_partition_xlsx_from_filename_with_metadata_filename(
         False,
     ],
 )
-def test_partition_xlsx_infer_table_structure(
-    infer_table_structure,
-    filename="example-docs/stanley-cups.xlsx",
-):
-    elements = partition_xlsx(filename=filename, infer_table_structure=infer_table_structure)
+def test_partition_xlsx_infer_table_structure(infer_table_structure: bool):
+    elements = partition_xlsx(
+        "example-docs/stanley-cups.xlsx", infer_table_structure=infer_table_structure
+    )
     table_elements = [e for e in elements if isinstance(e, Table)]
     for table_element in table_elements:
         table_element_has_text_as_html_field = (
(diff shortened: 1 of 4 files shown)


test_unstructured/partition/xlsx/test_xlsx.py
```diff
@@ -32,17 +41,17 @@ def test_partition_xlsx_from_filename(filename="example-docs/stanley-cups.xlsx")
     assert elements[1].metadata.filename == "stanley-cups.xlsx"
 
 
-def test_partition_xlsx_from_filename_with_emoji(filename="example-docs/emoji.xlsx"):
-    elements = partition_xlsx(filename=filename, include_header=False)
+def test_partition_xlsx_from_filename_with_emoji():
+    elements = partition_xlsx("example-docs/emoji.xlsx", include_header=False)
```
scanny, line 45: Several of the test-cleanup changes are removing the `filename="some-document-path"` parameter from the test function and inlining `"some-document-path"` in the partition call.

The only caller of a test function is `pytest` (the test-runner) and it will never provide a `filename` argument. This simplification makes the test more direct-reading and reduces noise.

unstructured/partition/xlsx.py
```diff
@@ -41,17 +48,17 @@
 @add_chunking_strategy()
 def partition_xlsx(
     filename: Optional[str] = None,
-    file: Optional[Union[IO[bytes], SpooledTemporaryFile]] = None,
+    file: Optional[IO[bytes]] = None,
```
christinestraub, line 51: Does not `partition_xlsx()` expect to pass `SpooledTemporaryFile`?

unstructured/partition/xlsx.py
```diff
@@ -85,26 +92,43 @@ def partition_xlsx(
...
     if filename:
-        sheets = pd.read_excel(filename, sheet_name=None, header=header)
+        sheets = pd.read_excel(  # pyright: ignore[reportUnknownMemberType]
+            filename, sheet_name=None, header=header
+        )
         last_modification_date = get_last_modified_date(filename)
 
     elif file:
-        f = spooled_to_bytes_io_if_needed(
-            cast(Union[BinaryIO, SpooledTemporaryFile], file),
```
christinestraub, line 94: Is there a specific reason to remove the `spooled_to_bytes_io_if_needed()` utility function? IMO, this function is also used for other partitioners.

unstructured/partition/xlsx.py
scanny: @christinestraub `SpooledTemporaryFile` implements the `IO[bytes]` abstract interface, so it is accepted by the type-checker as a legitimate type for the file parameter. So this code type-checks clean:
```python
f = SpooledTemporaryFile(1000)
partition_xlsx(file=f)
```
So there's no need to distinguish `SpooledTemporaryFile` here in the parameter type from any other binary file-like type, like say `BytesIO` or what you get back from `open(file_path, "rb")`.

It's only later that we want to get more specific so we can replace an instance of `SpooledTemporaryFile` with `BytesIO` (because `zipfile.open()` is incompatible with `SpooledTemporaryFile` in Python versions before 3.12 or something). And that we do with `isinstance()`.

unstructured/partition/xlsx.py
scanny: It is, but the typing on it is vexed and fixing that would involve fixing a lot of other files. Basically it's signature should be `IO[bytes] -> IO[bytes]` but it takes `BinaryIO | SpooledTemporaryFile` instead and so it doesn't type-check if you try to pass in a value typed as `IO[bytes]`.

Also its interface is odd because it also accepts `bytes` but then doesn't do anything with it. And what it does do is so simple it's easy to inline.

Anyway, we could undertake that broader fix if you wanted. It looks like it's used in `tsv`, `pdf`, `csv`, and `xml`, so that's not too extensive. Just let me know and I'm happy to do the needful :)

scanny: And btw, I'm not removing the function, just the call to it here. So it's still available for other partitioners to use as they have been :)

christinestraub:
> It is, but the typing on it is vexed and fixing that would involve fixing a lot of other files. Basically it's signature should be `IO[bytes] -> IO[bytes]` but it takes `BinaryIO | SpooledTemporaryFile` instead and so it doesn't type-check if you try to pass in a value typed as `IO[bytes]`.
> 
> Also its interface is odd because it also accepts `bytes` but then doesn't do anything with it. And what it does do is so simple it's easy to inline.
> 
> Anyway, we could undertake that broader fix if you wanted. It looks like it's used in `tsv`, `pdf`, `csv`, and `xml`, so that's not too extensive. Just let me know and I'm happy to do the needful :)


IMO, it would be needful and helpful to do that broader fix in a separate PR later.
CC: @cragwolfe @qued @MthwRobinson

christinestraub (approved):
LGTM!

PR #2529, opened 2024-02-08
Add AstraDB Integration for Unstructured

erichare:
This Pull Request implements an integration with [Astra DB](https://datastax.com) which allows for the Astra DB Vector Database to be compatible with Unstructured's set of integrations.

To create your Astra account and authenticate with your `ASTRA_DB_APPLICATION_TOKEN`, and `ASTRA_DB_API_ENDPOINT`, follow these steps:

1. Create an account at https://astra.datastax.com
2. Login and create a new database
3. From the database page, in the right hand panel, you will find your API Endpoint
4. Beneath that, you can create a Token to be used

Some notes about Astra DB:

- Astra DB is a Vector Database which allows for high-performance database transactions, and enables modern GenAI apps [See here](https://docs.datastax.com/en/astra/astra-db-vector/get-started/concepts.html)
- It supports similarity search via a number of methods [See here](https://docs.datastax.com/en/astra/astra-db-vector/get-started/concepts.html#metrics)
- It also supports non-vector tables / collections


potter-potter:
@erichare This looks super cool! A few things would make it easier for us to get this into Unstructured.
1) Provide some PR description. Including how to get a token if needed. Does it store vectors? Can you do a similarity search on the vectors? Can it store non-vectors (in our case just partitioned, not embedded, documents) If it needs a schema, can you provide an example schema for vector/non-vector.


2) Build a test file. We will be testing this connection with every PR merge to make sure nothing breaks. So something similar to: https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/dest/chroma.sh
This can be a live test (connects to the database). Or if AstraDB has a docker file, it can use that. Also if we are storing vectors, the test should do a similarity search.

Let me know if you have any questions or need help.

erichare:
@potter-potter thank you so much for the feedback. I've added a PR description with some details (sorry about neglecting that in the first place!) and i've also added a test. I tried to model it after the other tests so i'm hopeful I got the formatting right of everything, but will address any feedback as needed.

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+@dataclass
+class AstraDBWriteConfig(WriteConfig):
+    batch_size: int = 20
+
+
+@dataclass
+class AstraDBDestinationConnector(BaseDestinationConnector):
+    write_config: AstraDBWriteConfig
+    connector_config: SimpleAstraDBConfig
+    _astra_db = t.Optional["AstraDB"] = field(init=False, default=None)
```
potter-potter, line 50: should be a : instead of =

erichare: eek, bad typo! Thanks for catching that. Just updated it.

potter-potter:
I'm going over the PR. Will add changes as I go. Astra UI and DB setup and token experience is very nice!

Can you add the appropriate lines to:
`unstructured/ingest/cli/cmds/__init__.py`


Also ideally we delete the the collection when the live test is done. That way we don't end up with thousands of collections :)
The test will run every time we do a PR or a Merge as part of the CI.

erichare:
Thank you once again!! Just added the changes to `unstructured/ingest/cli/cmds/__init__.py` and i'll update the test to delete the collection

unstructured/ingest/cli/cmds/astra.py
```diff
@@ -0,0 +1,61 @@
...
+    def get_cli_options() -> t.List[click.Option]:
+        options = [
+            click.Option(
+                ["--token"],
+                required=True,
+                type=str,
+                help="Astra DB Token with access to the database.",
+            ),
+            click.Option(
+                ["--api_endpoint"],
```
potter-potter, line 22: think this should be --api-endpoint. Click converts to api_endpoint later in the process.

erichare: Yep, definitely. Updated this just now.

unstructured/ingest/cli/cmds/astra.py
```diff
@@ -0,0 +1,61 @@
...
+                required=True,
+                type=Dict(),
```
potter-potter, line 24: Wasn't sure about this but I input the endpoint as a string in my environment. So this should be `str` I believe.

erichare: Yep, this was likely something copied from another parameter and i missed updating it. It's definitely a str - the Astra DB portal gives you the string (on a side note, that string does encode a fair amount of information such that it probably could be a str, but yeah, our client libraries assume its the fully qualified web url so this should absolutely be a str). I've updated this as well.

potter-potter:
Thanks for the fast fixes. I have to stop and come back to this later. Can't wait to see it working!

erichare:
No worries at all. Thanks for the review!! If you have more things you notice i'll address them ASAP. I'm going to resolve the former points just to keep the PR cleaner but definitely re-open if you think any haven't been addressed properly.

test_unstructured_ingest/dest/astra.sh
```diff
@@ -0,0 +1,57 @@
...
+  exit 0
+fi
+
+if [ -z "$ASTRA_DB_API_ENDPOINT" ]; then
+  echo "Skipping Astra DB ingest test because ASTRA_DB_API_ENDPOINT env var is not set."
+  exit 0
+fi
+
+RANDOM_SUFFIX=$((RANDOM % 100000 + 1))
+COLLECTION_NAME="astra-test-output-$RANDOM_SUFFIX"
```
potter-potter, line 26: appears collections will only take: [a-zA-Z][a-zA-Z0-9_]
So let's switch the - to _

unstructured/ingest/cli/cmds/astra.py
```diff
@@ -0,0 +1,61 @@
...
+                ["--api-endpoint"],
+                required=True,
+                type=str,
+                help="The API endpoint for the Astra DB.",
+            ),
+            click.Option(
+                ["--collection-name"],
+                required=False,
+                type=str,
+                help="The name of the Astra DB collection to write into.",
```
potter-potter, line 31: add that only takes letters, numbers and _ symbol to the help

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+class AstraDBAccessConfig(AccessConfig):
+    token: t.Optional[str] = enhanced_field(default=None, sensitive=True)
+    api_endpoint: t.Optional[str] = enhanced_field(default=None, sensitive=True)
+
+
+@dataclass
+class SimpleAstraDBConfig(BaseConnectorConfig):
+    access_config: AstraDBAccessConfig
+    collection_name: str
+    embedding_dimension: int
```
potter-potter, line 38: I believe it needs the embedding dimension to create collection so let's make this required. So lets add it to the destination cmd options. See note above.

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+                collection_name=self.connector_config.collection_name,
+                dimension=self.connector_config.embedding_dimension,
+                options={"indexing": {"deny": NON_INDEXED_FIELDS}},
+            )
+        return self._astra_db_collection
+
+    @requires_dependencies(["astrapy"], extras="astra")
+    @DestinationConnectionError.wrap
+    def initialize(self):
+        _ = self._astra_db_collection
```
potter-potter, line 90: lets switch this to astra_db_collection (without the first underscore) so that it creates the collection right away. simplest way for now.

test_unstructured_ingest/dest/astra.sh
```diff
@@ -0,0 +1,57 @@
...
+  --input-path example-docs/book-war-and-peace-1p.txt \
+  --work-dir "$WORK_DIR" \
+  --chunk-elements \
+  --chunk-max-characters 1500 \
+  --chunk-multipage-sections \
+  --embedding-provider "langchain-huggingface" \
+  astra \
+  --token "$ASTRA_DB_APPLICATION_TOKEN" \
+  --api-endpoint "$ASTRA_DB_API_ENDPOINT" \
+  --collection-name "$COLLECTION_NAME"
```
potter-potter, line 55: add --embedding-dimension 384 (obviously add the \ to the collection name line) and then create the command line flag in cmds

test_unstructured_ingest/python/test-ingest-astra-output.py
```diff
@@ -0,0 +1,56 @@
...
+@click.option("--collection-name", type=str, default="collection_test")
+def run_check(token, api_endpoint, collection_name):
+    print(f"Checking contents of Astra DB collection: {collection_name}")
+
+    # Initialize our vector db
+    astra_db = AstraDB(token=token, api_endpoint=api_endpoint)
+    astra_db.delete_collection(collection_name)
+    astra_db_collection = astra_db.create_collection(collection_name, dimension=5)
+
+    # Insert a document into the test collection
```
potter-potter, line 17: This seems wrong. We should not have to insert a document. or create_collection. We already did that in the shell script that ran before this python file is called.

All we should be doing in this file 
1) do a vector search based on the embedding we uploaded (book-war-and-peace-1p.txt)
2) doing a count
3) deleting the collection

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+
+        astra_batch_size = self.write_config.batch_size
+
+        for chunk in chunk_generator(elements_dict, astra_batch_size):
+            self._astra_db_collection.insert_many(chunk)
+
+    def normalize_dict(self, element_dict: dict) -> dict:
+        return {
+            "id": str(uuid.uuid4()),
+            "vector": element_dict.pop("vector", None),
```
potter-potter, line 111: pop("embeddings")

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+            "content": element_dict.pop("content", None),
```
potter-potter, line 112: If this is supposed to be the document text:
pop("text", None)

test_unstructured_ingest/python/test-ingest-astra-output.py
potter-potter: Here is one way we've been doing the vector search:
https://github.com/Unstructured-IO/unstructured/blob/main/scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
But you can choose your own way.

test_unstructured_ingest/python/test-ingest-astra-output.py
```diff
@@ -0,0 +1,50 @@
...
+    # Initialize our vector db
+    astra_db = AstraDB(token=token, api_endpoint=api_endpoint)
+    astra_db_collection = astra_db.collection(collection_name)
+
+    # Tally up the embeddings
+    docs_count = astra_db_collection.count_documents()
+    number_of_embeddings = docs_count["status"]["count"] 
+
+    # Print the results
+    expected_embeddings = 1
```
potter-potter, line 23: change from 1 to 3 since the chunking produces a document with 3 parts. Which then creates 3 embeddings.

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+                collection_name=self.connector_config.collection_name,
+                dimension=self.connector_config.embedding_dimension,
+                options={"indexing": {"deny": NON_INDEXED_FIELDS}},
+            )
+        return self._astra_db_collection
+
+    @requires_dependencies(["astrapy"], extras="astra")
+    @DestinationConnectionError.wrap
+    def initialize(self):
+        astra_db_collection = self._astra_db_collection
```
potter-potter, line 90: Should be `_ = self.astra_db_collection`
I should have been more clear. I can see the misunderstanding.

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @requires_dependencies(["astrapy"], extras="astra")
+    def check_connection(self):
+        try:
+            _ = self._astra_db_collection
```
potter-potter, line 95: Should be `_ = self.astra_db_collection`

potter-potter:
A few little changes and I think we are good.
I'll end up copying your files into a new branch on Unstructured to get this merged in. (easiest way we have found to do it with our CI/CD)
nice work!

unstructured/ingest/connector/astra.py
erichare: My mistake! so basically all references in that file should be `self.astra_db_collection` - no leading underscore / internal method? I assume it's okay to keep `self._astra_db` right, since this shouldn't need to be interacted with directly?

potter-potter: Just change it in the 2 spots I suggested. Everything works. Don't change anything else :)
I'll give it a tiny little more checkover when I take it over.
In the other spots, the dataclass attribute shoud stay  _astra_db_collection)

potter-potter: And also leave self._astra_db as it is.

erichare: Heh, got it, yep, i see now the property has the right name. I totally missed that. Just pushed the update! @potter-potter

potter-potter: calling self.astra_db_collection will then populate those dataclass attributes. _astra_db and _astra_db_collection

erichare:
> A few little changes and I think we are good. I'll end up copying your files into a new branch on Unstructured to get this merged in. (easiest way we have found to do it with our CI/CD) nice work!

Thanks so much and for the very thorough review. Let me know if I can modify anything else!

potter-potter:
Looks good! (I'll let you know when it is merged. Probably early next week)

PR #2534, opened 2024-02-08
fix(xlsx): xlsx subtable algorithm

scanny:
**Reviewers:** It may be easier to review each of the two commits separately. The first adds the new `_SubtableParser` object with its unit-tests and the second one uses that object to replace the flawed existing subtable-parsing algorithm.

**Summary**

There are a cluster of bugs in `partition_xlsx()` that all derive from flaws in the algorithm we use to detect "subtables". These are encountered when the user wants to get multiple document-elements from each worksheet, which is the default (argument `find_subtable = True`).

This PR replaces the flawed existing algorithm with a `_SubtableParser` object that encapsulates all that logic and has thorough unit-tests.

**Additional Context**

This is a summary of the failure cases. There are a few other cases but they're closely related and this was enough evidence and scope for my purposes. This PR fixes all these bugs:
```python
    #
    # -- ✅ CASE 1: There are no leading or trailing single-cell rows.
    #       -> this subtable functions never get called, subtable is emitted as the only element
    #
    #    a b  -> Table(a, b, c, d)
    #    c d

    # -- ✅ CASE 2: There is exactly one leading single-cell row.
    #       -> Leading single-cell row emitted as `Title` element, core-table properly identified.
    #
    #    a    -> [ Title(a),
    #    b c       Table(b, c, d, e) ]
    #    d e

    # -- ❌ CASE 3: There are two-or-more leading single-cell rows.
    #       -> leading single-cell rows are included in subtable
    #
    #    a    -> [ Table(a, b, c, d, e, f) ]
    #    b
    #    c d
    #    e f

    # -- ❌ CASE 4: There is exactly one trailing single-cell row.
    #      -> core table is dropped. trailing single-cell row is emitted as Title
    #         (this is the behavior in the reported bug)
    #
    #    a b  -> [ Title(e) ]
    #    c d
    #      e

    # -- ❌ CASE 5: There are two-or-more trailing single-cell rows.
    #      -> core table is dropped. trailing single-cell rows are each emitted as a Title
    #
    #    a b  -> [ Title(e),
    #    c d       Title(f) ]
    #      e
    #      f

    # -- ✅ CASE 6: There are exactly one each leading and trailing single-cell rows.
    #      -> core table is correctly identified, leading and trailing single-cell rows are each
    #         emitted as a Title.
    #
    #      a  -> [ Title(a),
    #    b c       Table(b, c, d, e),
    #    d e       Title(f) ]
    #    f

    # -- ✅ CASE 7: There are two leading and one trailing single-cell rows.
    #      -> core table is correctly identified, leading and trailing single-cell rows are each
    #         emitted as a Title.
    #
    #    a    -> [ Title(a),
    #    b         Title(b),
    #    c d       Table(c, d, e, f),
    #    e f       Title(g) ]
    #      g

    # -- ✅ CASE 8: There are two-or-more leading and trailing single-cell rows.
    #      -> core table is correctly identified, leading and trailing single-cell rows are each
    #         emitted as a Title.
    #
    #      a  -> [ Title(a),
    #      b       Title(b),
    #    c d       Table(c, d, e, f),
    #    e f       Title(g),
    #    g         Title(h) ]
    #    h

    # -- ❌ CASE 9: Single-row subtable, no single-cell rows above or below.
    #      -> First cell is mistakenly emitted as title, remaining cells are dropped.
    #
    #    a b c  -> [ Title(a) ]

    # -- ❌ CASE 10: Single-row subtable with one leading single-cell row.
    #      -> Leading single-row cell is correctly identified as title, core-table is mis-identified
    #         as a `Title` and truncated.
    #
    #    a      -> [ Title(a),
    #    b c d       Title(b) ]
```


unstructured/partition/xlsx.py
```diff
@@ -243,6 +244,91 @@ def partition_xlsx(
...
+
+        core_table_stop = len(self._subtable) - len(self._trailing_single_cell_row_indices)
+
+        # -- core-table is what's left in-between --
+        return self._subtable[core_table_start:core_table_stop]
+
+    def iter_leading_single_cell_rows_texts(self) -> Iterator[str]:
+        """Generate the cell-text for each leading single-cell row."""
+        for row_idx in self._leading_single_cell_row_indices:
+            yield self._subtable.iloc[row_idx].dropna().iloc[0]  # pyright: ignore
```
scanny, line 242: `pandas` typing is poor so some of these calls generate multiple typing errors and must be ignored.

test_unstructured/partition/xlsx/test_xlsx.py
```diff
@@ -255,3 +263,153 @@ def test_partition_xlsx_with_more_than_1k_cells():
...
+# ------------------------------------------------------------------------------------------------
+# UNIT TESTS
+# ------------------------------------------------------------------------------------------------
+# These test components used by `partition_xlsx()` in isolation such that all edge cases can be
+# exercised.
+# ------------------------------------------------------------------------------------------------
+
+
+class Describe_SubtableParser:
+    """Unit-test suite for `unstructured.partition.xlsx._SubtableParser` objects."""
```
scanny, line 277: Because subtable-parsing is now extracted from the `partition_xlsx()` monolith we are able to thoroughly test edge cases without needing create a separate XLSX file for each case. Also they run extremely quickly, a few milliseconds for each case.

pyproject.toml
```diff
@@ -1,9 +1,11 @@
 [tool.black]
 line-length = 100
 
-# changes made here should also be made in `.pre-commit-config.yaml` and `Makefile`
 [tool.ruff]
 line-length = 100
+
+# -- changes made here should also be made in `.pre-commit-config.yaml` and `Makefile` --
+[tool.ruff.lint]
```
scanny, line 8: This gets rid of the warning from `ruff`. They recently segregated linting options from those that reformat the code.

test_unstructured/partition/test_auto.py
```diff
@@ -935,7 +935,7 @@ def test_auto_partition_xls_from_filename(filename="example-docs/tests-example.x
     elements = partition(filename=filename, include_header=False, skip_infer_table_types=[])
 
     assert sum(isinstance(element, Table) for element in elements) == 2
-    assert len(elements) == 18
+    assert len(elements) == 14
```
scanny, line 938: The prior code was double-emitting some elements. The correct count is 14.

test_unstructured_ingest/expected-structured-output/onedrive/utic-test-ingest-fixtures/tests-example.xls.json
```diff
@@ -183,27 +183,6 @@
     "text": "http://www.cmu.edu/blackboard/files/evaluate/tests-example.xls",
     "type": "Title"
   },
```
scanny, line 185: These are each a duplicate of the element above, which indicates that particular double-emit behavior is fixed.

christinestraub (approved):
LGTM!

Makefile
```diff
@@ -430,7 +430,7 @@ version-sync:
 
 .PHONY: check-coverage
 check-coverage:
-	coverage report --fail-under=95
+	coverage report --fail-under=94
```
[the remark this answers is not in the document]
scanny: The test coverage for the .xlsx module is 96%. There are other modules that are substantially lower, but increasing those really belongs in a different PR in my view.

I think what happened was it had drifted to be borderline and then fell under on this for some reason.

I can improve coverage here to 100% and maybe that will do it, but I think we have a broader problem where cause and effect are disconnected.

I was going to improve test coverage on chunking in the PR I'm working on now and then bring it back up to 95%, but maybe the metric really needs to be _every_ module has to be 95%+ rather than the average.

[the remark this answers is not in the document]
scanny: It's interesting to note that this didn't show up until I rebased on `main`, which added this commit: https://github.com/Unstructured-IO/unstructured/commit/1a706771facd0adef754e2f87ec58479e42251e6 that adds the `/unstructured/embed/octoai.py` module with 83% coverage (1 test).

The xlsx module only has three missed statements, so I'm not at all sure covering those is going to bring us over the  threshold.

christinestraub (approved):
Added some missing test cases for `pdf_image_utils.py` module to improve average test coverage in PR #2535, and the CI now passes the unit tests in the PR.

Makefile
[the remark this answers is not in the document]
ryannikolaidis: Traditionally everything in [unstructured/ingest](https://github.com/Unstructured-IO/unstructured/tree/main/unstructured/ingest) is covered by end to end shell script tests, but those won't contribute to coverage captured by `check-coverage`. Per discussion outside this thread, we probably need to move to a solution that provides more granularity.

Issue #2482, opened 2024-01-31
PDF links are not recognised and not added to the metadata


MiXiBo:
**Describe the bug**
When we try to load a PDF file with links that was generated by Word then the links are not recognised. In comparison when we load the unstructured sample PDF the links are extracted. When we use a tool such as pdfx we can see that the links are in the document.

**To Reproduce**
Load attached PDF file with unstructured 

**Expected behavior**
Extract the link 'Klick Here' with link to 'https://ebay.com'

**Screenshots**

![Screenshot 2024-01-31 at 11 44 38](https://github.com/Unstructured-IO/unstructured/assets/34360898/43facdb1-5ad4-45f3-997c-1cbeb914a14e)

![Screenshot 2024-01-31 at 11 44 48](https://github.com/Unstructured-IO/unstructured/assets/34360898/e94a29af-757d-4f83-82ec-5547b246a072)

[pdf-link.pdf](https://github.com/Unstructured-IO/unstructured/files/14110595/pdf-link.pdf)

[embedded-link.pdf](https://github.com/Unstructured-IO/unstructured/files/14110573/embedded-link.pdf)

![Screenshot 2024-01-31 at 11 47 23](https://github.com/Unstructured-IO/unstructured/assets/34360898/44a6dac1-148a-44b0-bf0a-b44b6521a205)

We do use unstructured version 0.12.3

MiXiBo:
We can read the links. We had to set the threshold to a lower value:

https://github.com/Unstructured-IO/unstructured/blob/dd6576c603da10f42b39a71d13aef7fcc8b99528/unstructured/partition/pdf.py#L1179

sjoerd222888:
What would be potential side effects of setting a lower threshold value?

sjoerd222888:
Currently it seems this cannot be passed as a parameter, but would be nice to allow to configure this from the outside if needed?

MiXiBo:
I have refactored 'threshold' to 'annotation_threshold' and made it an optional parameter in '_process_pdfminer_pages' by keeping the default at 0.9.

With that we do have the option to make it configurable. It would be very helpful if @christinestraub you could merge that. We do have automated deployments which makes it complicated and cumbersome to workaround.

https://github.com/Unstructured-IO/unstructured/pull/2537


Thanks!

sjoerd222888:
Is there any interest to address this? I know sometimes open-source projects are like this. Dead slow and nobody cares.
[...]

PR #2537, opened 2024-02-09
Refactor threshold to annotation_threshold and make it an optional parameter

MiXiBo:
We are activating to configure the annotation threshold for links as an optional parameter. 

The reason for the change is that we ran into issues extracting simple text links from PDF documents that were created with MS Word. The sample PDF from unstructured worked with a default value of 0.9, and the PDF generated with Word resulted in a threshold of approx 0.67.

We do use unstructured in together with langchain within an automated container deployment and to access by default the setting 'annotation_threshold' (refactored from 'threshold') can be very helpful.


sjoerd222888:
This is to address the issue [#2482](https://github.com/Unstructured-IO/unstructured/issues/2482).

@christinestraub you have assigned yourself to that issue, what do you think about the proposed solution?

sjoerd222888:
What can we do to get the workflows approval? What must we do for the changelog for example?

christinestraub:
> What can we do to get the workflows approval? What must we do for the changelog for example?

Can you add the following lines to `Enhancements` section in the changelog.md?
```
### Enhancements

* **add support for `start_index` in `html` links extraction**
* **improve embedded links extraction in `partition_pdf()` for `fast` strategy**
```

christinestraub (changes requested):
Can you please revert the changes related to support for start_index in html links extraction and use PR #2600 for those changes?

MiXiBo:
> Can you please revert the changes related to support for start_index in html links extraction and use PR #2600 for those changes?

yes, I will prepare it. Thanks

MiXiBo:
> Can you please revert the changes related to support for start_index in html links extraction and use PR #2600 for those changes?

done

christinestraub (approved):
LGTM! We're considering to configure the `annotation_threshold` using an [environment variable](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/utils/config.py) (e.g. `UNSTRUCTURED_PDF_ANNOTATION_THRESHOLD`) after this PR is merged. We'll update you once we make the change.

PR #2544, opened 2024-02-13
fix: don't treat double quote enclosed text as JSON

### LINKED ISSUE #2444: bug/`partition()` treats double quote enclosed text file as JSON and errors
**Describe the bug**
While trying to call `partition()` on a text file which only has text enclosed in double quotes it recognizes it as JSON and fails with the following error
`Detected a JSON file that does not conform to the Unstructured schema. partition_json currently only processes serialized Unstructured output`

**Version used**
`unstructured[all-docs]==0.10.10`

**To Reproduce**
```python
import io
    # `file` passed has no extension though
    def extract_text(file: str) -> str:
        """Extract text from file."""
        try:
            with open(file, mode="rb") as input_file_obj:
                bytes_io = io.BytesIO(input_file_obj.read())
                elements = partition(file=bytes_io)
        except Exception as e:
            print(f"Error partitioning file: {e}")
        text = "\n\n".join([str(el) for el in elements])
        return text

```

**Expected behavior**
The file has to be partitioned and we should be able to extract the text contents from it

**Additional context**
The following contents in a file with no extension can help reproduce the issue.
It fails with the mentioned error in this case
```
"Sample input file that gets mistaken as JSON"
```
However this gets partitioned fine (notice the absence of double-quotes on the text)
```
Sample input file that gets partitioned correctly
```
I also tried the below which works too - if its of any help
```
"Sample input file that also gets partitioned" correctly
```

Thanks in advance :)

MthwRobinson:
### Summary

Closes #2444. Treats JSON serializable content that results in a string as plain text. Even though this is valid JSON per [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt), this is valid JSON, but in almost every cases were really want to treat this as a text file.

### Testing

1. Put `"This is not a JSON"` is a text file `notajson.txt`
2. Run the following

```python
from unstructured.file_utils.filetype import _is_text_file_a_json

_is_text_file_a_json(filename="notajson.txt") # Should be False
```


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@
 
 ### Fixes
 
-* **Fix `check_connection` in opensearch, databricks, postgres, azure connectors ** 
+* **Fix `check_connection` in opensearch, databricks, postgres, azure connectors **
+* **Fix don't treat plain text files with double quotes as JSON ** If a file can be deserialized a JSON but it deserializes as a string, treat it as plain text even though it's valid JSON.
```
qued, line 12:
```suggestion
* **Fix don't treat plain text files with double quotes as JSON ** If a file can be deserialized as a JSON but it deserializes as a string, treat it as plain text even though it's valid JSON.
```

qued (approved):
LGTM with one nit.

PR #2545, opened 2024-02-13
rfctr(chunking): extract chunking-strategy dispatch

scanny:
**Summary**
This is the final step in adding pluggable chunking-strategies. It introduces the `chunk()` function to replace calls to strategy-specific chunkers in the `@add_chunking_strategy` decorator. The `chunk()` function then uses a mapping of chunking-strategy names (e.g. "by_title", "basic") to chunking functions (chunkers) to dispatch the chunking call. This allows other chunkers to be added at runtime rather than requiring a code change, which is what "pluggable" chunkers is.

**Additional Information**
- Move the `@add_chunking_strategy` to the new `chunking.dispatch` module since it coheres strongly with that operation, but publish it from `chunking(.__init__)` (as it was before) so users don't couple to the way we organize the chunking sub-package. Also remove the third level of nesting as it's unrequired in this case.
- Add unit tests for the `@add_chunking_strategy` decorator which was previously uncovered by any direct test.

### DIFF
--- a/unstructured/chunking/__init__.py
+++ b/unstructured/chunking/__init__.py
@@ -6,6 +6,17 @@ Publishes the public aspects of the chunking sub-package interface.
 from __future__ import annotations
 
 from unstructured.chunking.base import CHUNK_MAX_CHARS_DEFAULT, CHUNK_MULTI_PAGE_DEFAULT
-from unstructured.chunking.dispatch import add_chunking_strategy
+from unstructured.chunking.dispatch import (
+    Chunker,
+    add_chunking_strategy,
+    register_chunking_strategy,
+)
 
-__all__ = ["CHUNK_MAX_CHARS_DEFAULT", "CHUNK_MULTI_PAGE_DEFAULT", "add_chunking_strategy"]
+__all__ = [
+    "CHUNK_MAX_CHARS_DEFAULT",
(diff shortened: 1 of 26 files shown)


unstructured/chunking/__init__.py
```diff
@@ -19,82 +19,75 @@
 _P = ParamSpec("_P")
 
 
-def add_chunking_strategy() -> Callable[[Callable[_P, List[Element]]], Callable[_P, List[Element]]]:
+def add_chunking_strategy(func: Callable[_P, List[Element]]) -> Callable[_P, List[Element]]:
```
scanny, line 22: This looks like a lot of changes but it's essentially just sliding the body out by one indent level.

christinestraub:
What do you think about adding a changelog entry for introducing the `chunk()` function?

cragwolfe:
Why not just make it clear it is an internal refactor in the CHANGELOG.
The "why it matters" part of a changelog bullet may also be useful for
developers or anybody in the community that is directly working with the
code base.

scanny:
> Why not just make it clear it is an internal refactor in the CHANGELOG. The "why it matters" part of a changelog bullet may also be useful for developers or anybody in the community that is directly working with the code base.

Sure, sounds good to me, will do :)

christinestraub (approved):
LGTM!

PR #2550, opened 2024-02-14
rfctr(chunking): preparation for plug-in chunkers, Part I

scanny:
**Summary**
In order to accommodate customized chunkers other than those directly provided by `unstructured`, some further modularization is necessary such that a new chunker can be added as a "plug-in" without modifying the `unstructured` library code.

This PR is the straightforward refactoring required for this process like typing changes. There are also some other small changes we've been meaning to make like making all chunking options accept `None` to represent their default value so the broad field of callers (e.g. ingest, unstructured-api, SDK) don't need to determine and set default values for chunking arguments leading to diverging defaults.

Isolating these "noisy" but easy to accept changes in this preparatory PR reduces the noise in the more substantive changes to follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.5-dev7
+## 0.12.5-dev8
 
 ### Enhancements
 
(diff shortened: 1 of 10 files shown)


christinestraub (approved):
LGTM!

PR #2554, opened 2024-02-15
chore: add tests and small fixes in utils.py

Coniferish:
Linting and typing fixes, and add tests to improve test coverage in utils.py

On the main branch, run `coverage run -m pytest test_unstructured/test_utils.py` and then `coverage report -m unstructured/utils.py` to see test coverage for `utils.py`. Check out to this branch and do the same. The percent coverage should increase to 88%


Coniferish:
@LaverdeS, can you take a look at `calculate_shared_ngram_percentage` and fix the type hints or tell me what the return type should be? I was going to add a test for that function, but the type hinting is inconsistent (and in the call to `ngrams()`), so I held off.

LaverdeS:
> @LaverdeS, can you take a look at `calculate_shared_ngram_percentage` and fix the type hints or tell me what the return type should be? I was going to add a test for that function, but the type hinting is inconsistent (and in the call to `ngrams()`), so I held off.

Oh, that is a Set.

Coniferish:
@LaverdeS I think I understand what you were trying to do now and it makes sense to return a `tuple`. Sets should contain only one type, but `calculate_shared_ngram_percentage` was trying to return `float` and a `list`. Please take a look at the commit I just added.

test_unstructured/test_utils.py
```diff
@@ -105,7 +105,7 @@ def test_only_gives_only(iterator):
 @pytest.mark.parametrize("iterator", [[0, 1], (0, 1), range(10)])
```
scanny, line 105: There are a lot of type errors in this file. Want to use this as an opportunity to clean those up?

test_unstructured/test_utils.py
```diff
@@ -223,6 +220,76 @@ def test_only_raises_if_empty(iterator):
                 ],
             ),
         ),
+        (
```
scanny, line 223: The test case inputs are mostly repeated and it makes reading this test a scrolling chore. Can you extract the differentiators to the parameters and place the repeated parts in the test body. I'm thinking this looks something like:
```python
@pytest.paramatrize(
    ("coords", "coords2", "expectation"),
    (
        (
            ((4, 5), (4, 8), ...),
            ((2, 3), (2, 6), ...),
            {
                "overlapping_elements": ...
                ...
            },
            ...
        ),

)
def test...():
    elements = [
         Title(..., coords),
         NarrativeText(... coords2),
    ]

    assert overlapping_flag is True
    assert overlapping_cases == [expectation]
```
This reduces the vertical size by over 150 lines (12 + 1 + 6) * 8 = 152

Then the last case would be a separate test like `test_it_does_not_catch_non_overlapping_cases()`, that way you can parametrize only the dict part of the response and provide the rest in the test body.

While you're a

unstructured/utils.py
```diff
@@ -318,7 +330,7 @@ def scarf_analytics():
         pass
 
 
-def ngrams(s: str, n: int) -> List:
+def ngrams(s: List[str], n: int) -> List[Tuple[str]]:
```
scanny, line 333: Now that we're past 3.7, add a `from __future__ import annotations` import at the top of the module and use lowercase `list`, `dict`, and `tuple` for typing.

scanny (changes requested):
Looks pretty good, a couple of opportunities to clean things up further since we're in here on a refactoring mission :)

test_unstructured/test_utils.py
Coniferish: Sounds good. I was unsure about including that in the scope of this PR, but will do :)

test_unstructured/test_utils.py
```diff
@@ -337,3 +311,66 @@ def test_validate_data_args():
...
+        assert utils.validate_date_args("1990-12-33")
+        assert utils.validate_date_args("2022-02-12T14:30:00-2000")
+
+
+def test_htmlify_matrix_handles_empty_cells():
+    matrix = [["cell1", "", "cell3"], ["", "cell5", ""]]
+    expected_html = (
+        "<table><tr><td>cell1</td><td></td><td>cell3</td></tr>"
+        "<tr><td></td><td>cell5</td><td></td></tr></table>"
+    )
```
scanny, line 336: It's generally best to avoid trivial variables that are used only once. It introduces a level of abstraction that detracts from the direct-reading of the test, like giving this produces that.

In this case:
```python
    assert utils.htmlify_matrix_of_cell_texts(matrix) == (
        "<table><tr><td>cell1</td><td></td><td>cell3</td></tr>"
        "<tr><td></td><td>cell5</td><td></td></tr></table>"
    )
```
removes that level of abstraction. Even this in this case:
```python
    assert utils.htmlify_matrix_of_cell_texts(matrix=[["cell1", "", "cell3"], ["", "cell5", ""]]) == (
        "<table><tr><td>cell1</td><td></td><td>cell3</td></tr>"
        "<tr><td></td><td>cell5</td><td></td></tr></table>"
    )
```
you have the `matrix=` in the call, which is just as informative as an assignment, and the additional step of mentally mapping `expected_html` to that string is removed.

You can reserve using variables like this for when the value is actually variable, such as in a parameterized test.

unstructured/utils.py
```diff
@@ -26,13 +28,18 @@
...
 
 from unstructured.__version__ import __version__
 
 if TYPE_CHECKING:
     from unstructured.documents.elements import Text
 
+Point: TypeAlias = Tuple[float, float]
+Points: TypeAlias = Tuple[Point, ...]
+# Box format: [x_bottom_left, y_bottom_left, x_top_right, y_top_right]
+Box: TypeAlias = Tuple[float, float, float, float]
```
scanny, line 41: Generally a good idea to organize types like this in alphabetical order to ease scanning.

unstructured/utils.py
```diff
@@ -228,8 +235,19 @@ def dependency_exists(dependency: str):
     return True
 
 
-# Copied from unstructured/ingest/connector/biomed.py
-def validate_date_args(date: Optional[str] = None):
+def validate_date_args(date: Optional[str] = None) -> bool:
+    """Validate whether the provided date string satisfies any of the supported date formats.
```
scanny, line 239: PEP 257 specifies a blank line between first line and any continuation text.

unstructured/utils.py
```diff
@@ -547,28 +584,43 @@ def identify_overlapping_case(
     )
 
 
+def _convert_coordinates_to_box(coordinates: Points):
```
scanny, line 587: Definitely worth a docstring.

scanny (changes requested):
Several remarks.

unstructured/utils.py
```diff
@@ -334,28 +352,37 @@ def calculate_shared_ngram_percentage(
...
 
 
-def calculate_largest_ngram_percentage(first_string: str, second_string: str) -> (float, List, str):
+def calculate_largest_ngram_percentage(
+    first_string: str,
+    second_string: str,
+) -> tuple[float, set[tuple[str, ...]], str]:
     """Iteratively calculate_shared_ngram_percentage starting from the biggest
-    ngram possible until is >0.0%"""
+    ngram possible until is >0.0%
```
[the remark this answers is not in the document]
Coniferish: > What is the `>0.0%' bit? Aren't all percentages greater than zero?

It looks like `n` is initialized as the number of words in the longer of the two strings and decremented until it reaches 0 or there's a common word/n-gram (and the percentage is greater than 0).

I updated the first sentence and added the implementation explanation as a comment

unstructured/utils.py
```diff
@@ -334,28 +352,37 @@ def calculate_shared_ngram_percentage(
...
 
     common_ngrams = set(first_string_ngrams) & set(second_string_ngrams)
     percentage = (len(common_ngrams) / len(first_string_ngrams)) * 100
     return percentage, common_ngrams
 
 
-def calculate_largest_ngram_percentage(first_string: str, second_string: str) -> (float, List, str):
+def calculate_largest_ngram_percentage(
+    first_string: str,
+    second_string: str,
```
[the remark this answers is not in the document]
scanny: No. Both parameters will fit on one line, like this:
```python
def calculate_largest_ngram_percentage(
    first_string: str, second_string: str
) -> tuple[float, set[tuple[str, ...]], str]:
```

Note the lack of a trailing comma after `second_string`.

test_unstructured/test_utils.py
```diff
@@ -313,42 +314,40 @@ def test_validate_data_args():
...
-        assert utils.validate_date_args("not a date")
-        assert utils.validate_date_args("1990-12-33")
-        assert utils.validate_date_args("2022-02-12T14:30:00-2000")
+@pytest.mark.parametrize("date", [None, "not a date", "1990-12-33"])
+def test_validate_date_args_raises_for_invalid_formats(date):
+    pattern1 = re.compile(r"The argument.*?(?:is None).*")
+    pattern2 = re.compile(r"The argument.*?(?:does not satisfy the format: YYYY-MM-DD).*")
+    combined_pattern = re.compile(f"({pattern1.pattern}|{pattern2.pattern})")
+    with pytest.raises(ValueError, match=combined_pattern):
+        assert utils.validate_date_args(date)
```
scanny, line 330: Okay, this is fine, but hard to read :)

Leave it this time but in future I'd separate this into two tests, one for each different type, which I expect must be in different places in the code.

Actually I think I'd just change the messages to start with something like "invalid date: ..." and then just inline the match like:
```python
with pytest.raises(ValueError, match="invalid date: "):
    ...
```

But like I say, good enough for present purposes. Just a little bit "cure worse than disease" :)

Also, remember your intent, which is to illuminate for the reader what happens, like "the user gets an error message like this". A complicated regex doesn't serve that intent and we don't really care that the message exactly matches some form, just do your best to give the reader an idea instead of leaving no match.

test_unstructured/test_utils.py
```diff
@@ -313,42 +314,40 @@ def test_validate_data_args():
...
+    )
 
 
 def test_htmlify_matrix_handles_multiple_rows_and_cells():
-    matrix = [["cell1", "cell2"], ["cell3", "cell4"]]
-    expected_html = (
-        "<table><tr><td>cell1</td><td>cell2</td></tr><tr><td>cell3</td><td>cell4</td></tr></table>"
+    assert utils.htmlify_matrix_of_cell_texts([["cell1", "cell2"], ["cell3", "cell4"]]) == (
+        "<table><tr><td>cell1</td><td>cell2</td></tr>"
+        "<tr><td>cell3</td><td>cell4</td></tr></table>"
```
scanny, line 349: Nice. With this simplification you could see this being one parametrized test with a comment like `# -- handles empty cells --` etc. for each case.

That will give us something to look forward to next time we're in here and in the mood.

Nice improvement.

unstructured/utils.py
```diff
@@ -283,8 +282,9 @@ def first(it: Iterable[Any]) -> Any:
 
 
 def only(it: Iterable[Any]) -> Any:
-    """Returns the only element from a singleton iterable. Raises an error if the iterable is not a
-    singleton."""
+    """Returns the only element from a singleton iterable.
+
+    Raises an error if the iterable is not a singleton."""
```
scanny, line 287: PEP 257 specifies last """ goes on new line in the multi-line form. There are a few more of these below.

unstructured/utils.py
```diff
@@ -354,6 +354,7 @@ def calculate_shared_ngram_percentage(
     n: int,
 ) -> tuple[float, set[tuple[str, ...]]]:
     """Calculate the percentage of common_ngrams between string_A and string_B
+
```
scanny, line 357: Okay, blank line, good, but the first line needs to stand on its own, you can't just put the blank line in the middle of a sentence :)

Something like:
```python
"""Percentance of common-ngrams between string_A and string_B.

Common n-grams are computed with reference to ...
"""

File: unstructured/utils.py [hunk 14]
```diff
@@ -440,7 +437,8 @@ def calculate_overlap_percentage(
     box2: Points,
     intersection_ratio_method: str = "total",
 ) -> tuple[float, float, float, float]:
-    """Calculates the percentage of overlapped region with reference to
+    """Calculate the percentage of overlapped region with reference to
+
```
scanny, line 441: Same.

unstructured/utils.py
```diff
@@ -552,11 +551,9 @@ def identify_overlapping_case(
...
 
         else:
-            (
-                largest_ngram_percentage,
-                _,
-                largest_n,
-            ) = calculate_largest_ngram_percentage(text1, text2)
+            (largest_ngram_percentage, _, largest_n) = calculate_largest_ngram_percentage(
+                text1, text2
+            )
```
scanny, line 556: Okay, yeah, no, my mistake. This doesn't fit on one line even without the left-hand-side (LHS) parens. The LHS parens should go though. Black will still wrap but won't add them back in.

scanny (approved):
Some more remarks but approving in advance.

PR #2556, opened 2024-02-16
rfctr(chunking): extract strategy-specific chunking options

scanny:
**Summary**
A pluggable chunking strategy needs its own local set of chunking options that subclasses a base-class in `unstructured`.

Extract distinct `_ByTitleChunkingOptions` and `_BasicChunkingOptions` for the existing two chunking strategies and move their strategy-specific option setting and validation to the respective subclass.

This was also a good opportunity for us to clean up a few odds and ends we'd been meaning to.

Might be worth looking at the commits individually as they are cohesive incremental steps toward the goal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.5-dev8
+## 0.12.5-dev9
 
 ### Enhancements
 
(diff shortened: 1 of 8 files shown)


christinestraub (approved):
LGTM!

scanny:
Thanks Christine! :)

PR #2560, opened 2024-02-17
Fix a parameter name in the js-client example usage

awalker4:
`files` should be `fileName` as noted in https://github.com/Unstructured-IO/unstructured-js-client/issues/24


ron-unstructured (approved):
LGTM

PR #2563, opened 2024-02-20
feat: introduce `date_from_file_object` parameter to partitions

ds-filipknefel:
Introduce `date_from_file_object` to `partition*` functions, by default set to `False`.
If set to `True` and file is provided via `file` parameter, partition will attempt to infer last modified date from `file`'s contents otherwise last modified metadata will be set to `None`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
         language_version: python3
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.0.230"
+    rev: "v0.2.1"
     hooks:
       - id: ruff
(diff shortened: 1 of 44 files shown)


MthwRobinson (approved):
@ds-filipknefel - Is this parameter going to be available in the API as well or just in the library? Intent on the functionality looks good from my end, but I'd get someone from the core lib team to review as well.

ds-filipknefel:
> @ds-filipknefel - Is this parameter going to be available in the API as well or just in the library? Intent on the functionality looks good from my end, but I'd get someone from the core lib team to review as well.

In the discussion this task was based there was no mention of the API. So AFAIK it's not going to be available.

PR #2564, opened 2024-02-20
fix(json): partition_json() does not chunk

scanny:
**Summary**
For whatever reason, the `@add_chunking_strategy` decorator was not present on `partition_json()`. This broke the only way to accomplish a "chunking-only" workflow using the REST API. This PR remedies that problem.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,15 @@
 [tool.black]
 line-length = 100
 
+[tool.pyright]
+pythonPlatform = "Linux"
+pythonVersion = "3.8"
+reportUnnecessaryCast = true
+reportUnnecessaryTypeIgnoreComment = true
+stubPath = "./typings"
+typeCheckingMode = "strict"
+verboseOutput = true
+
 [tool.ruff]
 line-length = 100
 
--- a/unstructured/partition/email.py
+++ b/unstructured/partition/email.py
@@ -7,7 +7,7 @@ import sys
 from email.message import Message
 from functools import partial
 from tempfile import NamedTemporaryFile, SpooledTemporaryFile, TemporaryDirectory
-from typing import IO, Callable, Dict, List, Optional, Tuple, Union
+from typing import IO, Any, Callable, Dict, List, Optional, Tuple, Union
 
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -1,6 +1,15 @@
...
 
+[tool.pyright]
+pythonPlatform = "Linux"
+pythonVersion = "3.8"
+reportUnnecessaryCast = true
+reportUnnecessaryTypeIgnoreComment = true
+stubPath = "./typings"
+typeCheckingMode = "strict"
+verboseOutput = true
+
```
cragwolfe, line 12: is this required for this PR?

scanny: Not strictly speaking @cragwolfe. Is there a problem with it? I didn't figure it worth a whole separate PR and long story but it was convenient for me to add it in here.

unstructured/partition/email.py
```diff
@@ -275,12 +275,12 @@ def partition_email(
     metadata_filename: Optional[str] = None,
     metadata_last_modified: Optional[str] = None,
     process_attachments: bool = False,
-    attachment_partitioner: Optional[Callable] = None,
+    attachment_partitioner: Optional[Callable[..., List[Element]]] = None,
```
Coniferish, line 278: Not that this needs to change, but is there a reason why the ellipse comes before the type?

Coniferish (approved):
LGTM

unstructured/partition/email.py
scanny: The ellipse indicates the parameters of the callable are not specified (can be any set of parameters). The second type is the return value. The syntax is `Callable[[param-types], return-type]`.

In this case, a partitioner can be just about anything parameter-wise. We could define a callback-protocol that stated things like:
- it has to have a `filename: str` parameter as its first parameter
- it needs to have a `languages: list[str]` parameter somewhere
- etc.

I didn't want to get into that here because I just wanted `json.py` to type-check clean and I don't think a more precise type for a partitioner really buys us much just now. But it's something that might make more sense in the future sometime. A good thing to go in a `model` module.

PR #2571, opened 2024-02-21
feat: astra dest connector

potter-potter:
Thanks to Eric Hare @erichare at DataStax we have a new destination connector.

This Pull Request implements an integration with [Astra DB](https://datastax.com) which allows for the Astra DB Vector Database to be compatible with Unstructured's set of integrations.

To create your Astra account and authenticate with your `ASTRA_DB_APPLICATION_TOKEN`, and `ASTRA_DB_API_ENDPOINT`, follow these steps:

1. Create an account at https://astra.datastax.com
2. Login and create a new database
3. From the database page, in the right hand panel, you will find your API Endpoint
4. Beneath that, you can create a Token to be used

Some notes about Astra DB:

- Astra DB is a Vector Database which allows for high-performance database transactions, and enables modern GenAI apps [See here](https://docs.datastax.com/en/astra/astra-db-vector/get-started/concepts.html)
- It supports similarity search via a number of methods [See here](https://docs.datastax.com/en/astra/astra-db-vector/get-started/concepts.html#metrics)
- It also supports non-vector tables / collections


test_unstructured_ingest/python/test-ingest-astra-output.py
```diff
@@ -0,0 +1,57 @@
...
+    random_vector = [round(random.uniform(0, 1), 1) for _ in range(embedding_dimension)]
+
+    find_one = astra_db_collection.find_one()
+    random_vector = find_one["data"]["document"]["vector"]
+    random_text = find_one["data"]["document"]["content"]
+    print(random_text)
+
+    # Perform a similarity search
+    find_result = astra_db_collection.vector_find(random_vector, limit=1)
+    print(find_result)
```
potter-potter, line 45: @erichare Do you know why this is not finding any results? I'm essentially searching with the same vector I just pulled from the database. It should just pull up the same record. But the result is always blank.

erichare: @potter-potter So, i'm looking at the test.... internally, the vector / embeddings column in astra is always named `$vector`, i.e., i would expect line 39 to look like.

```python
random_vector = find_one["data"]["document"]["$vector"]
```

The `find` and `vector_find` methods at the API level are going to look for that field. I'm wondering if we need to make a small change to the integration as a result?

erichare: My guess would have been the `normalize_dict()` method but you'd probably know better? At the time i thought that was normalization for purposes of unstructured, but if thats what is being sent to Astra, with `write_dict`, i think that needs to be changed... that's my guess anyway. `vector` -> `$vector`

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,114 @@
...
+        logger.info(f"Inserting / updating {len(elements_dict)} documents to AstraDB.")
+
+        astra_batch_size = self.write_config.batch_size
+
+        for chunk in chunk_generator(elements_dict, astra_batch_size):
+            self._astra_db_collection.insert_many(chunk)
+
+    def normalize_dict(self, element_dict: dict) -> dict:
+        return {
+            "id": str(uuid.uuid4()),
```
erichare, line 108: Assuming i'm correct that this is the literal document that gets sent to Astra DB, i think let's change `id` to `_id`

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,114 @@
...
+            "vector": element_dict.pop("embeddings", None),
```
erichare, line 109: And here, let's change `vector` to `$vector`

test_unstructured_ingest/python/test-ingest-astra-output.py
```diff
@@ -0,0 +1,57 @@
...
+    assert number_of_embeddings == expected_embeddings, (
+        f"Number of rows in generated table ({number_of_embeddings})"
+        f"doesn't match expected value: {expected_embeddings}"
+    )
+
+    # Generate a random embedding of the appropriate length
+    random_vector = [round(random.uniform(0, 1), 1) for _ in range(embedding_dimension)]
+
+    find_one = astra_db_collection.find_one()
+    random_vector = find_one["data"]["document"]["vector"]
```
erichare, line 39: Here, `vector` to `$vector`

test_unstructured_ingest/python/test-ingest-astra-output.py
erichare: Left a couple comments that I *think* will fix it... this was my oversight, if i'm correct about it.... thinking that the normalization of the dictionary was for the return to Unstructured rather than for insertion into Astra.

I think once these changes are made, `vector_find` should work properly.... as of now i think its failing because its not actually seeing the "proper" `$vector` field in the documents.

potter-potter: Does the free account not allow vector search via api?

erichare: It should allow you to! If you login to `astra.datastax.com` and click your Database, and the Data Explorer tab shows a collection with N dimensions (plus >= 1 records) then `vector_find` should work just fine!

unstructured/ingest/connector/astra.py
potter-potter: Oh. Is it not saving it in the native `id` of AstraDB? What is the actual schema of AstraDB?

test_unstructured_ingest/python/test-ingest-astra-output.py
potter-potter: These changes seemed to work.

erichare: Fantastic. Let me know if there's any further snags. Internally most users never need to know about this, but in this case yeah i think vector_find() was just looking across a field that didn't exist, which was my fault since it should have been part of the document insertion in the first place.

unstructured/ingest/connector/astra.py
```diff
@@ -0,0 +1,114 @@
...
+        astra_batch_size = self.write_config.batch_size
+
+        for chunk in chunk_generator(elements_dict, astra_batch_size):
+            self._astra_db_collection.insert_many(chunk)
+
+    def normalize_dict(self, element_dict: dict) -> dict:
+        return {
+            "_id": str(uuid.uuid4()),
+            "$vector": element_dict.pop("embeddings", None),
+            "content": element_dict.pop("text", None),
```
potter-potter, line 110: @erichare can you confirm that `content` and `metadata` are the correct column names to be stored in the database. (like id and vector had to be adjusted)

unstructured/ingest/connector/astra.py
potter-potter: Got ya. Ideally we are putting the vector in the right place. We could completely remove `_id` if Astra DB makes it's own id.

potter-potter: so where is it storing the text and metadata? We could name those whatever we want?

unstructured/ingest/connector/astra.py
erichare: Yep I can confirm. these are totally fine. And actually if you have a preference for a different name, that's fine too, these fields have no special meaning in Astra DB... just that in some other integrations they are what is used

unstructured/ingest/connector/astra.py
erichare: Correct yep! The only one with the special name is `$vector` if the embeddings have been pre-computed (we're exploring computing them at insert time as an option, but right now that's not in place and we tend to rely on integrations like this to handle that.)

I'm going to be honest i'm a bit less sure about `_id` - but just testing it, it does seem like Astra generates that field automatically with a random uuid if not provided. so i think its safe to remove it if desired.

docs/source/ingest/destination_connectors/astra.rst
```diff
@@ -0,0 +1,32 @@
+Astra
+===========
+
+Batch process all your records using ``unstructured-ingest`` to store structured outputs and embeddings locally on your filesystem and upload those to a Astra DB index.
+
+First you'll need to install the Pinecone dependencies as shown here.
```
ryannikolaidis, line 6: Pinecone? Missed on copy paste?

docs/source/ingest/destination_connectors/code/bash/astra.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+EMBEDDING_PROVIDER=${EMBEDDING_PROVIDER:-"langchain-huggingface"}
+
+unstructured-ingest \
+  local \
+  --input-path example-docs/book-war-and-peace-1p.txt \
+  --output-dir local-output-to-pinecone \
```
ryannikolaidis, line 8: pinecone? nit: this is confusing, rename?

test_unstructured_ingest/python/test-ingest-astra-output.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    # Grab an embedding from the collection and search against itself
+    # Should get the same document back as the most similar
+    find_one = astra_db_collection.find_one()
+    random_vector = find_one["data"]["document"]["$vector"]
+    random_text = find_one["data"]["document"]["content"]
+
+    # Perform a similarity search
+    find_result = astra_db_collection.vector_find(random_vector, limit=1)
```
ryannikolaidis, line 40: 🎉

ryannikolaidis (commented):
a couple very small nits, otherwise this looks GREAT!

docs/source/ingest/destination_connectors/astra.rst
potter-potter: @ryannikolaidis yep. Rushed out the documentation.

ryannikolaidis (approved):
lgtm!

PR #2572, opened 2024-02-21
feat: separate evaluate grouping function

Klaijan:
Separate the aggregating functionality of `text_extraction_accuracy` to a stand-alone function to avoid duplicated eval effort if the granular level eval is already available.

To test:
Run `PYTHONPATH=. pytest test_unstructured/metrics/test_evaluate.py` locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features
 
+* **Add parent_element to overlapping case output** Adds parent_element to the output for `identify_overlapping_or_nesting_case` and `catch_overlapping_and_nested_bboxes` functions.
+
 ### Fixes
 
 * **Add OctoAI embedder** Adds support for embeddings via OctoAI.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,11 @@
 ### Features
 
 * **Add parent_element to overlapping case output** Adds parent_element to the output for `identify_overlapping_or_nesting_case` and `catch_overlapping_and_nested_bboxes` functions.
-* **Separate grouping function for evaluation** Separate the grouping functionality to avoid duplicated evaluation when the granular level eval is already available.
 
 ### Fixes
 
 * **Add OctoAI embedder** Adds support for embeddings via OctoAI.
-* **Fix `check_connection` in opensearch, databricks, postgres, azure connectors **
+* **Fix `check_connection` in opensearch, databricks, postgres, azure connectors**
 * **Fix don't treat plain text files with double quotes as JSON ** If a file can be deserialized as JSON but it deserializes as a string, treat it as plain text even though it's valid JSON.
 * **Fix `check_connection` in opensearch, databricks, postgres, azure connectors **
 * **Fix cluster of bugs in `partition_xlsx()` that dropped content.** Algorithm for detecting "subtables" within a worksheet dropped table elements for certain patterns of populated cells such as when a trailing single-cell row appeared in a contiguous block of populated cells.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 * **Add parent_element to overlapping case output** Adds parent_element to the output for `identify_overlapping_or_nesting_case` and `catch_overlapping_and_nested_bboxes` functions.
+* **Separate grouping function for evaluation** Separate the grouping functionality to avoid duplicated evaluation when the granular level eval is already available.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.5-dev7
+## 0.12.5-dev8
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
-## 0.12.5-dev8
+## 0.12.5-dev9
 
 ### Enhancements
 
 ### Features
 
 * **Add parent_element to overlapping case output** Adds parent_element to the output for `identify_overlapping_or_nesting_case` and `catch_overlapping_and_nested_bboxes` functions.
+* **Add table structure evaluation** Adds a new function to evaluate the structure of a table and return a metric that represents the quality of the table structure. This function is used to evaluate the quality of the table structure and the table contents.
 
 ### Fixes
 
--- a/test_unstructured/metrics/test_evaluate.py
+++ b/test_unstructured/metrics/test_evaluate.py
@@ -6,6 +6,7 @@ import pandas as pd
 import pytest
 
 from unstructured.metrics.evaluate import (
+    group_text_extraction_accuracy,
     measure_element_type_accuracy,
     measure_text_extraction_accuracy,
 )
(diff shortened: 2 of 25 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 * **Add parent_element to overlapping case output** Adds parent_element to the output for `identify_overlapping_or_nesting_case` and `catch_overlapping_and_nested_bboxes` functions.
+* **Separate grouping function for evaluation** Separate the grouping functionality to avoid duplicated evaluation when the granular level eval is already available.
```
yuming-long, line 8: looks like this PR is just moving a block of code to a separate function and i am just curious how this avoid duplicated evaluation?

Klaijan: Sorry I wasn't clear on the description. We are adding a metric evaluation for individual docs and aggregated docs after this. The current structure of the code makes it impossible to get the grouping from existing granular output, but rather need to reevaluate. However, that reminds me I'd want to make edits to the command part of the eval as well. Will convert this back to draft!

unstructured/metrics/evaluate.py
```diff
@@ -187,3 +173,25 @@ def measure_element_type_accuracy(
     _write_to_file(export_dir, "all-docs-element-type-frequency.tsv", df)
     _write_to_file(export_dir, "aggregate-scores-element-type.tsv", agg_df)
     _display(agg_df)
+
+
+def group_text_extraction_acc(
```
scanny, line 178: What does `acc` stand for? Could be accumulation, accumulator, accessory, access, etc.

Can we just spell it out? I'm thinking it's `_accuracy`, but especially since it's a call now it improves readability if the user doesn't need to visit the implementation to make sense of the call.

CHANGELOG.md
scanny: Is this a user-facing function? If not it probably shouldn't have a CHANGELOG text entry. It seems like this is a refactoring to extract a function so we can access it from multiple places. In that case just and updated version string would be sufficient, like `dev8` instead of `dev7` or whatever.

unstructured/metrics/evaluate.py
```diff
@@ -187,3 +173,25 @@ def measure_element_type_accuracy(
...
+        grouped_miss = (
+            df.groupby(grouping)
+            .agg({"cct-%missing": [_mean, _stdev, "count"]})
+            .rename(columns={"_mean": "mean", "_stdev": "stdev"})
+        )
+        df = _format_grouping_output(grouped_acc, grouped_miss)
+        export_filename = f"all-{grouping}-agg-cct"
+    else:
+        print("No field to group by. Returning a non-group evaluation.")
+    return df, export_filename
```
scanny, line 197: In general, avoiding nesting improves readability. In this case it could be accomplished with a "guard", like:
```python
    # -- {{explanation of why other grouping types don't count}} --
    if grouping not in ("doctype", "connector"):
        print("No field to group by. Returning a non-group evaluation.")
        return df, export_filename

    grouped_acc = (
        df.groupby(grouping)
        .agg({"cct-accuracy": [_mean, _stdev, "count"]})
        .rename(columns={"_mean": "mean", "_stdev": "stdev"})
    )
    grouped_miss = (
        df.groupby(grouping)
        .agg({"cct-%missing": [_mean, _stdev, "count"]})
        .rename(columns={"_mean": "mean", "_stdev": "stdev"})
    )
    df = _format_grouping_output(grouped_acc, grouped_miss)
    export_filename = f"all-{grouping}-agg-cct"
    return df, export_filename
```

scanny: Also, why is there a `print()` call? I'm not familiar with this part of the code, so maybe it's part of a CLI or something, but in general a library should not have print statements.

scanny (commented):
A few remarks :)

unstructured/metrics/evaluate.py
Klaijan: The print call is to be shown on CLI terminal if no grouping is done on the df. I'll think of ways to move this elsewhere if possible.

unstructured/metrics/evaluate.py
Klaijan: We'll want to call it through the metrics evaluation script

unstructured/metrics/evaluate.py
scanny: It's fine if this is a CLI, I just didn't realize we had that in the library :)

unstructured/metrics/evaluate.py
scanny: @Klaijan not getting your comment. I'm suggesting you change the name to `group_text_extraction_accuracy()` (if "accuracy" is what is meant by "acc". What does that have to do with being called via the metrics evaluation script?

unstructured/metrics/evaluate.py
```diff
@@ -187,3 +177,46 @@ def measure_element_type_accuracy(
...
+    grouping: str, data_input: Union[pd.DataFrame, str], export_dir: str
+) -> None:
+    """
+    Aggregates accuracy and missing metrics by 'doctype' or 'connector', exporting to TSV.
+
+    Args:
+        grouping (str): Grouping category ('doctype' or 'connector').
+        data_input (Union[pd.DataFrame, str]): DataFrame or path to a CSV/TSV file.
+        export_dir (str): Directory for the exported TSV file.
+    """
```
scanny, line 192: Docstrings should be [PEP 257](https://peps.python.org/pep-0257/) compliant, like this in this case:
```python
    """Aggregates accuracy and missing metrics by 'doctype' or 'connector', exporting to TSV.

    Args:
        grouping (str): Grouping category ('doctype' or 'connector').
        data_input (Union[pd.DataFrame, str]): DataFrame or path to a CSV/TSV file.
        export_dir (str): Directory for the exported TSV file.
    """
```

I encourage you to read the PEP, but the basic idea is:
- first line begins immediately after the """ and extends for no more than one line.
- second line is blank
- remaining lines as needed
- closing """ is on separate line, aligned at same indent level as first """

If only a single line docstring is needed it must fit on one line, like:

```python
    """Aggregate accuracy and missing metrics by 'doctype' or 'connector'."""
```

unstructured/metrics/evaluate.py
```diff
@@ -187,3 +177,46 @@ def measure_element_type_accuracy(
...
+        elif data_input.endswith((".tsv", ".txt")):
+            df = pd.read_csv(data_input, sep="\t")
+        else:
+            raise ValueError("Please provide a .csv or .tsv file.")
+    else:
+        df = data_input
+    if df.empty or grouping not in df.columns or df[grouping].isnull().all():
+        raise SystemExit(
+            f"Data cannot be aggregated by `{grouping}`. \
+                         Check if it's empty or the column is missing/empty."
```
scanny, line 209: In Python, two string literals that immediately follow each other are automatically concatenated. So this error message would be formatted:
```python
        raise SystemExit(
            f"Data cannot be aggregated by `{grouping}`."
            f" Check if it's empty or the column is missing/empty."
        )
```
Note that the blank space needed to separate the first fragment from the second is placed at the _beginning_ of the second (and any later) lines. This makes it easy to confirm by vertically scanning the aligned left of the block rather than the ragged right side. It's a common oversight to forget the separating space such that you get ".. grouping.Check if ..." in the output, so a convention that makes it easy to visually confirm the spaces is welcome :)

Btw, the output from the original version would look like this when it printed out:
"... by grouping.                                                  Check if it's empty ..."

This inclusion of leading space in continuation lines is avoided by using adjacent string concatenation.

test_unstructured/metrics/test_evaluate.py
```diff
@@ -125,3 +140,63 @@ def test_text_extraction_wrong_type():
         measure_text_extraction_accuracy(
             output_dir=output_dir, source_dir=source_dir, export_dir=export_dir, output_type="wrong"
         )
+
+
+@pytest.mark.skipif(is_in_docker, reason="Skipping this test in Docker container")
```
scanny, line 145: This reason just restates the presence of the decorator. I can tell it does this from the `@..skipif(is_in_docker,..)` part. What might be more useful is a statement of _why_ this test is skipped in docker.

scanny (approved):
Okay, a few remarks for you to consider, but approving in advance :)

unstructured/metrics/evaluate.py
Klaijan: Thank you!

test_unstructured/metrics/test_evaluate.py
Klaijan: I will create another ticket to deal with this specifically :) Thank you so much

Klaijan: https://unstructured-ai.atlassian.net/browse/CORE-3980

PR #2574, opened 2024-02-21
fix: Add `pip` as explicit dep in `environment.yml` to prevent warning

tomheaton:
Removes this warning:

> Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.


MthwRobinson (approved):
LGTM

PR #2581, opened 2024-02-24
feat: add support for including/excluding headers and footers when partitioning pdfs

christinestraub:


christinestraub:
we should just bubble up `exclude_element_types` as an option in partition* for now

PR #2586, opened 2024-02-27
bug CORE-3971: fix deserialization in google-drive source connector key path

potter-potter:
Google Drive Service account key can be a dict or a file path(str)

We have successfully been using the path. But the dict can also end up being stored as a string that needs to be deserialized. The deserialization can have issues with single and double quotes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,18 @@
+## 0.12.6-dev0
+
+### Enhancements
+
+### Features
+
 ## 0.12.5
 
 ### Enhancements
 
 ### Features
+
+### Fixes
+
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -13,13 +23,12 @@
 ### Fixes
 
 * **Fix passing list type parameters when calling unstructured API via `partition_via_api()`** Update `partition_via_api()` to convert all list type parameters to JSON formatted strings before calling the unstructured client SDK. This will support image block extraction via `partition_via_api()`.
```
potter-potter, line 25: The octoai below is already in Features.

ryannikolaidis (approved):
Looks good! Nice add with test coverage.

PR #2588, opened 2024-02-27
chore: change table extraction defaults

ds-filipknefel:
Change default values for table extraction - works in pair with [this](https://github.com/Unstructured-IO/unstructured-api/pull/370) `unstructured-api` PR

We want to move away from `pdf_infer_table_structure` parameter, in this PR:
- We change how it's treated wrt `skip_infer_table_types` parameter. Whether to extract tables from pdf now follows from the rule: `pdf_infer_table_structure && "pdf" not in skip_infer_table_types`
- We set it to `pdf_infer_table_structure=True` and `skip_infer_table_types=[]` by default
- We remove it from the examples in documentation
- We describe it as deprecated in favor of `skip_infer_table_types` in documentation

More detailed description of how we want parameters to interact
- if `pdf_infer_table_structure` is False tables will never extracted from pdf
- if `pdf_infer_table_structure` is True tables will be extracted from pdf unless it's skipped via `skip_infer_table_types`
- on default `pdf_infer_table_structure=True` and `skip_infer_table_types=[]`


unstructured/partition/auto.py
```diff
@@ -199,6 +199,8 @@ def partition(
             detect_language_per_element
                 Detect language per element instead of at the document level.
     pdf_infer_table_structure
+        Deprecated! Use `skip_infer_table_types` to opt out of table extraction for any document
+        type.
```
badGarnet, line 203: lets add a warning log about this deprecated setting

ds-filipknefel: I've made it so that the warning is raised when we know someone explicitly used the parameter (so when it's `False`). I am not sure quite if that's effective, e.g. if users mostly do `pdf_infer_table_structure=True` they won't realize it's better to let go of that practice. Also wouldn't want to spam warning every time.

What do you think? Technically we could set it `None` on default, so that would detect every time someone provides value explicitly but maybe made things more confusing, loose thought.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fixes
 
 * **Clarify IAM Role Requirement for GCS Platform Connectors**. The GCS Source Connector requires Storage Object Viewer and GCS Destination Connector requires Storage Object Creator IAM roles.
+* **Change table extraction defaults** Change table extraction defaults in favor of using `skip_infer_table_types` parameter and reflect these changes in documentation.
```
MthwRobinson, line 12: Could we update to `0.13.0` with this changes since we're changing default behavior on a feature that's heavily used?

MthwRobinson (approved):
This looks good from a intent standpoint. Main suggestion is to make sure we do a minor version bump since we're changing the default on table extraction and that's widely used.

PR #2589, opened 2024-02-27
bug: update sharepoint-with-permissions test to fix CI

potter-potter:
Adding `metadata.data_source.permissions_data` to sharepoint-with-permissions.sh --metadata-exclude to prevent sharepoint deprecation warning from ruining test.

Updating expected-structured-output

As per Ahmet's comment. We do want to check sharepoint permissions metadata at some point. But that will take a separate type of test. A file diff test is too unstable. Permissions checking will be later down the road.


test_unstructured_ingest/src/sharepoint-with-permissions.sh
ahmetmeleq, line 1: I'd rename as sharepoint-with-ignored-permissions.sh to keep track of the fact that we ingest, however ignore, the permissions. With the current file name it's misleading. Otherwise the PR LGTM

test_unstructured_ingest/src/sharepoint-with-permissions.sh

potter-potter: I believe the fact that we use permissions in the CLI is why it was called “with permissions “. As opposed to the other one that doesn’t use permissions. I don’t think it refers to the metadata. But I didn’t write the connector.

rbiseck3 (approved):

test_unstructured_ingest/src/sharepoint-with-permissions.sh
potter-potter: This was resolved offline. We will want some sort of testing with permissions metadata in the future. I made note of this twice in the .sh test.

PR #2591, opened 2024-02-27
fix: handling of varied SharePoint date formats

ryannikolaidis:
We are seeing occurrences of inconsistency in the timestamps returned by office365.sharepoint when fetching created and modified dates. Furthermore, in future versions of this library, a datetime object will be returned rather than a string. 

## Changes

- This adds logic to guarantee SharePoint dates will be properly formatted as ISO, regardless of the format provided by the sharepoint library.
- Bumps timestamp format output to include timezone offset (as we do with others)

## Testing

Unit test added to validate this datetime handling across various formats.

### DIFF
--- a/unstructured/ingest/connector/sharepoint.py
+++ b/unstructured/ingest/connector/sharepoint.py
@@ -3,6 +3,8 @@ import os
 import typing as t
 from dataclasses import dataclass
 from datetime import datetime
+from dateutil import parser
+
 from html import unescape
 from pathlib import Path
 from urllib.parse import urlparse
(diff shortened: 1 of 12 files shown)


unstructured/ingest/connector/sharepoint.py
```diff
@@ -202,6 +204,23 @@ def parent_name_matches(parent_type, permissions_filename, ingest_doc_filepath):
 
         return permissions_data
 
+    def _ensure_isoformat_datetime(self, timestamp: t.Union[datetime, str]) -> str:
```
potter-potter, line 207: This is cool. If we end up needing it elsewhere we should put it in a utils file.

potter-potter (approved):
this is great

unstructured/ingest/connector/sharepoint.py
ryannikolaidis: Yep, totally agree!

PR #2593, opened 2024-02-27
fix: Install pandoc consistently, via Makefile recipe (version that supports .rtf files as input format)

micmarty-deepsense:
## Problem Description
In some cases you might find yourselves in a situation when pandoc won't be able to process an `rtf` as input file format, because older versions simply do not support that.

```
RuntimeError: Invalid input format! Got "rtf" but expected one of these: commonmark, creole, csv, docbook, docx, dokuwiki, epub, fb2, gfm, haddock, html, ipynb, jats, jira, json, latex, man, markdown, markdown_github, markdown_mmd, markdown_phpextra, markdown_strict, mediawiki, muse, native, odt, opml, org, rst, t2t, textile, tikiwiki, twiki, vimwiki
```

Basically, some user may install the wrong version. The `README.md` is not be precise enough when mentioning RTF files support: 
https://github.com/Unstructured-IO/unstructured/blob/47b35ccdd61ffbc376c86e9bb08a2039b042cc2b/README.md?plain=1#L120-L122

## Example
Installing `pandoc` from a [stable repository, like Debian](https://packages.debian.org/source/bullseye/pandoc) will give you `2.9` and the official documentation shows clearly that support for rtf was introduced in `2.14` https://pandoc.org/releases.html#pandoc-2.14.2-2021-08-21
image

### Note that `rtf` is not there
image

### More detail
image

## Proposed Solution 
- [x] I've simply added/copied `make install-pandoc` calls, mimicking other recipes in order to ensure that `3.1.2` will be installed in all cases. **Side note**: `make install-pandoc` calls `./scripts/install-pandoc.sh` under the hood.
- [x] Update README file - mention that `make install-pandoc` is recommended (`>=2.14.2`)
- [x] Verify tests that cover `rtf` cases: https://github.com/Unstructured-IO/unstructured/blob/47b35ccdd61ffbc376c86e9bb08a2039b042cc2b/test_unstructured/file_utils/test_file_conversion.py#L14
- [x]  Update `setup_ubuntu.sh` if needed?: https://github.com/Unstructured-IO/unstructured/blob/47b35ccdd61ffbc376c86e9bb08a2039b042cc2b/scripts/setup_ubuntu.sh#L87
-


micmarty-deepsense:
## Suggested error message

```md
RuntimeError: Invalid input format! Got "rtf" but expected one of these: commonmark, creole, csv, docbook, docx, dokuwiki, epub, fb2, gfm, haddock, html, ipynb, jats, jira, json, latex, man, markdown, markdown_github, markdown_mmd, markdown_phpextra, markdown_strict, mediawiki, muse, native, odt, opml, org, rst, t2t, textile, tikiwiki, twiki, vimwiki

Support for RTF files is not available in the current pandoc installation. It was introduced in pandoc 2.14.2.
Reference: https://pandoc.org/releases.html#pandoc-2.14.2-2021-08-21

Current version of pandoc: 2.9.2.1
Please, follow the pandoc installation instructions in README.md to install the right version.
```

Running `make test-extra-pypandoc` will generate the following output
image

## Reinstalling the right version makes all tests passing
```bash
sudo apt remove pandoc
make install-pandoc
make test-extra-pypandoc
```

image

micmarty-deepsense:
> A little related to this PR - Here at this line https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ci.yml#L438 just for cleanliness, I'm thinking is it ok if we add the `pandoc` installation to `unstructured-inference`'s `install-ci` as well since CI will need to run against various file types including `rtf`. This way we can omit the `make install-pandoc` on later line.
> 
> But since it's not directly related to this PR so it's your call :)
> 
> Anyways, just need to update CHANGELOG and update the branch then all LGTM! Switch this to ready for review then I can approve it then!

Good catch @Klaijan ! I removed those `install-pandoc` calls from GH workflows and now it's part of the `install-ci` recipe

PR is no longer in draft mode, so if it's still legitimate, please approve ;)

.github/workflows/ci.yml
```diff
@@ -106,7 +106,6 @@ jobs:
         source .venv/bin/activate
         sudo apt-get update
         sudo apt-get install -y libmagic-dev poppler-utils libreoffice
-        make install-pandoc
```
micmarty-deepsense, line 109: removed that redundant call - see few lines below: there's `install-ci` call -> it ensures that `install-pandoc` will run

.github/workflows/ci.yml
```diff
@@ -437,7 +438,6 @@ jobs:
         # FIXME (yao): sometimes there is cache but we still miss argilla in the env; so we add make install-ci again
         make install-ci
         sudo apt-get update && sudo apt-get install --yes poppler-utils libreoffice
-        make install-pandoc
```
micmarty-deepsense, line 440: same story, but this time `make install-ci` is being called above
https://github.com/Unstructured-IO/unstructured/pull/2593/files#r1507798453

Klaijan (approved):
LGTM!

PR #2596, opened 2024-02-28
Doc: Change Python comment string to JavaScript comment string.

masonbrothers:
JavaScript uses `//` for comments instead of `#`


yuming-long (approved):
Thanks! LGTM

PR #2605, opened 2024-03-01
BUGFIX: fix dependencies in setup.py

rbiseck3:
### Description
Currently the requirements associated with an extra in the `setup.py` is being dynamically generated using the `load_requirements()` method in the same file. This is being passed in all the `.in` files which then get read line by line to generate the requirements associated with an extra. Unless the `.in` file itself has a version pin, this will never respect the `.txt` files being generated by `pip-compile`. This fix updates all the inputs to `load_requirements()` to use the `.txt` files themselves.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,20 @@ jobs:
     - name: Check for dependency conflicts
       run: make check-deps
 
+  check-extras:
+    strategy:
+      matrix:
+        python-version: [ "3.9","3.10","3.11" ]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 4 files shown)


MthwRobinson:
@rbiseck3 - Would this require users who `pip install unstructured` to use the exact version we have in the `.txt` files?

MthwRobinson (approved):
> @rbiseck3 - Would this require users who pip install unstructured to use the exact version we have in the .txt files?

Per offline convo, @rbiseck3 confirmed you can do this in your `requirements.txt` if you want to install a different version of the dependencies than the ones we'll have pinned. Good to go from my end.

```
unstructured
pdfminer2==<not-same-as-unstructured>
```

cooknl:
Is this the same as

> Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.

PR #2611, opened 2024-03-04
Add support for table caption

asadhasan832:
Add support for detecting table caption tags within tables and by themselves.

More on caption tags:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption


MthwRobinson:
@asadhasan832 - Thanks for this PR. If you resolve the conflicts and add tests we can review.

PR #2613, opened 2024-03-05
BUG: name override not passed through in recursive _asdict call

rbiseck3:
### Description
The only real change here is adding this line:
```python
                    apply_name_overload=apply_name_overload,
```

Everything else is from running `make tidy`


potter-potter (approved):
looks good

potter-potter:
Sharepoint-with-permissions passed CI this time... It's so unpredictable.

Issue #2308, opened 2023-12-21
bug/Reading File When IO[bytes] file param is passed , not file name


narenderAwen:
**Describe the bug**
I am using the partition_pdf function, i give the file (IO[bytes]) object as argument in file param, not filename or url. I am recieving file from some remote source, encoded to base 64, i am converting it to BytesIO() , then give as file, i also update its filename arg (because i need to keep it for record downstream), now according to code and error stack i have mentioned below, there is a function which fetches modify date by directly reading file from fs, which i obviously not present
```python
datetime.fromtimestamp(os.path.getmtime(filename))
``` 
So i guess i cant pass file as arg if its not in file system, is this by design, and if yes then how am i supposed to process files fetched from other sources, not necessarily present in my file system

**To Reproduce**
Just pass a BytesIO() object with file name set, and observe

**Expected behavior**
Should ideally not throw an error

Below is the stack trace, i also raised a PR if you like to review and merge.
[https://github.com/Unstructured-IO/unstructured/pull/2307](https://github.com/Unstructured-IO/unstructured/pull/2307)
**Screenshots**
```
  File "/Users/narenderawen/AwenDev/eLLM/eLLM/parsers/pdfparser.py", line 39, in parse
    elements_raw=partition_pdf(
                 ^^^^^^^^^^^^^^
  File "/Users/narenderawen/AwenDev/eLLM/venv/lib/python3.11/site-packages/unstructured/documents/elements.py", line 503, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/narenderawen/AwenDev/eLLM/venv/lib/python3.11/site-packages/unstructured/file_utils/filetype.py", line 591, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/narenderawen/AwenDev/eLLM/venv/lib/python3.11/site-packages/unstructured/file_utils/filetype.py", line 546, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/narenderawen/AwenDev/eLLM/venv/lib/python3.11/site-packages/unstructured/chunking/title.py", line 241, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2614, opened 2024-03-06
fix: raises on file-like object with .name not a valid path

scanny:
**Summary**
Fixes: #2308

**Additional context**
Through a somewhat deep call-chain, partitioning a file-like object (e.g. io.BytesIO) having its `.name` attribute set to a path not pointing to an actual file on the local filesystem would raise `FileNotFoundError` when the last-modified date was being computed for the document.

This scenario is a legitimate partitioning call, where `file.name` is used downstream to describe the source of, for example, a bytes payload downloaded from the network.

**Fix**
- explicitly check for the existence of a file at the given path before accessing it to get its modified date. Return `None` (already a legitimate return value) when no such file exists.
- Generally clean up the implementations.
- Add unit tests that exercise all cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+* **Partitioning raises on file-like object with `.name` not a local file path.** Partitioning a file using the `file=` argument with a file-like object (e.g. io.BytesIO) having a `.name` attribute with a value not a path to a file present on the local filesystem raises `FileNotFoundError`. This prevents setting the `.name` attribute to give a `.metadata.filename` value to elements formed from for example a file retrieved from a network location via HTTP.
 * **Fix SharePoint dates with inconsistent formatting** Adds logic to conditionally support dates returned by office365 that may vary in date formatting or may be a datetime rather than a string.
 * **Include warnings** about the potential risk of installing a version of `pandoc` which does not support RTF files + instructions that will help resolve that issue.
 * **Incorporate the `install-pandoc` Makefile recipe** into relevant stages of CI workflow, ensuring it is a version that supports RTF input files.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.6-dev7
+## 0.12.6-dev8
 
 ### Enhancements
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+* **Partitioning raises on file-like object with `.name` not a local file path.** Partitioning a file using the `file=` argument with a file-like object (e.g. io.BytesIO) having a `.name` attribute with a value not a path to a file present on the local filesystem raises `FileNotFoundError`. This prevents setting the `.name` attribute to give a `.metadata.filename` value to elements formed from for example a file retrieved from a network location via HTTP.
```
Coniferish, line 13: This is a little hard to follow. If I'm following correctly, previously `FileNotFoundError` was raised when a file-like object with a `.name` attribute (that was not a path) was partitioned using `file=`, right?

Coniferish: What you said in the PR description is clearer:
This fix allows a bytes payload to be given a filename in its `metadata` even though such a file never existed on the local filesystem. This fix explicitly checks for the existence of a file at the given path before accessing it to get its modified date. Return None (already a legitimate return value) when no such file exists.

test_unstructured/partition/test_common.py
```diff
@@ -1,3 +1,7 @@
+import datetime as dt
+import io
+import os
+import pathlib
 from dataclasses import dataclass
 from unittest import mock
```
Coniferish, line 7: Only potential change I can think of is adding an integration test to show we can partition a `io.BytesIO` "file" via `file=`. Approving in advance, though.

Coniferish (approved):
Approving in advance (comments are nitpicky).

CHANGELOG.md
scanny: Yeah, the triggering conditions were these:
1. You call partition_{x}() (pretty much any of them).
2. You use the `file` arg to provide the input document.
3. That file value is a file-like object like `io.BytesIO`, not the `io.BufferedReader` you get back from `with open(path) as f`.
4. You assign a name to the file-like object using `file.name = meaningful_name`. You do that so it shows up in the `metadata.filename` field for each element. Maybe you downloaded the bytes of the document over the network so it doesn't have a path on the filesystem, but you still want to indicate what document the elements came from.

Then when the partitioner went to compute the last-modified date for metadata it would try to access the file at `file.name` to get its modified date without checking whether it actually existed on the filesystem or not.

test_unstructured/partition/test_common.py
scanny: Yeah, good call. I added one for XLSX as a representative example.

PR #2617, opened 2024-03-06
rfctr: prepare for fix to raises on file-like-object with name not a path to a file

scanny:
**Summary**
Improve typing and other mechanical refactoring in preparation for fix to issue 2308.


unstructured/partition/common.py
```diff
@@ -148,14 +137,15 @@ def normalize_layout_element(
...
-                text=text if text else "",
+                text=text,
                 coordinates=coordinates,
                 coordinate_system=coordinate_system,
                 metadata=class_prob_metadata,
                 detection_origin=origin,
             )
 
     elif element_type in TYPE_TO_TEXT_ELEMENT_MAP:
+        assert isinstance(element_type, str)
```
Coniferish, line 148: what is this for?

Coniferish (approved):
Have a question about one change, but lgtm. Approving.

unstructured/partition/common.py
scanny: The `element_type` variable can be `None` (set on L119 with `.get()` on dict with no default). `None` is not a valid key for `TYPE_TO_TEXT_ELEMENT_MAP` which is `dict[str, type[Text]]`. This causes the type checker to rightfully complain. However, all the keys of that dict are str and so if it passes the `elif element_type in TYPE_TO_TEXT_ELEMENT_MAP` test then we know it has to be a `str` value.

Adding this assert just communicates to the type-checker this knowledge that it's not sophisticated enough to work out for itself (which would be way more sophisticated than you want a type-checker to be).

I expect there are ways to modify that code so the type of `element_type` is `str` and not `Optional[str]` (Optional is often an unnecessary complication/expense), but I didn't want to get into that in this PR.

unstructured/utils.py
```diff
@@ -263,7 +263,7 @@ def validate_date_args(date: Optional[str] = None) -> bool:
     )
 
 
-def _first_and_remaining_iterator(it: Iterable[Any]) -> tuple[Any, Iterator[Any]]:
+def _first_and_remaining_iterator(it: Iterable[_T]) -> Tuple[_T, Iterator[_T]]:
     iterator = iter(it)
```
scanny, line 267: @Coniferish did you notice these changes? This is a nice example of using a `TypeVar` to narrow `Any`.

The idea is that while it's true that `it` can be an iterable of any type, whatever type that is determines the output type(s). Like you can't send an iterable of `int` and expect `(str, Iterator[bool])` as a return value type, even though using Any for those would allow that. You should get `(int, Iterator[int])` in that case and this way the type-checker can verify that for you.

Kind of a cool example of generics in typing that I think is intuitive to understand.

PR #2620, opened 2024-03-07
feat: redefine table level acc

badGarnet:
This PR redefines the `table_level_acc` metric as follow:
- for each predicted table use sequence matching ratio as its accuracy
- as a prerequisite for the sequence matching we sort the table cells by row then column for both predicted and ground truth to ensure they are ordered the same
- average all predicted table accuracy
- any prediction without a matching ground truth (false positive) would decrease the score
- prediction that splits ground truth into smaller tables would also have low score with perfectly equal splits having lowest score

This new definition makes the new metric a value between 0 and 1 per file. This replaces the existing definition where the metric is defined as (the number of predicted table that has a match to ground truth) to (the number of ground truth table). This existing metric actually gives higher values for predictions that splits tables and can be higher than 1. The new definition prefers predictions that do not split ground truth tables.


unstructured/metrics/table/table_eval.py
```diff
@@ -166,10 +194,8 @@ def process_file(self) -> TableEvaluation:
         )
 
         return TableEvaluation(
-            total_tables=total_tables,
-            table_level_acc=(
-                round(total_predicted_tables / total_tables, 2) if total_tables else np.nan
-            ),
+            total_tables=len(ground_truth_table_data),
+            table_level_acc=round(predicted_table_acc, 2),
```
pawel-kmiecik, line 198: I was thinking about rounding at this level - as we're getting some aggregations on tens of tables in next step - is it ok to round it to 2 decimal places? We're losing some data then and AFAIK numbers shouldn't be rounded before aggregations (rounds can be added just before visualization).

unstructured/metrics/table/table_eval.py
```diff
@@ -166,10 +194,8 @@ def process_file(self) -> TableEvaluation:
...
             element_col_level_index_acc=metrics.get("col_index_acc", np.nan),
             element_row_level_index_acc=metrics.get("row_index_acc", np.nan),
             element_col_level_content_acc=metrics.get("col_content_acc", np.nan),
```
pawel-kmiecik, line 201: BTW I was thinking also about content accuracy and whether we could use average levenshtein distance (this is how we actually measure CCT) instead of the current solution. Right now we're using `token_ratio`.

pawel-kmiecik (approved):
Having 2 comments (questions?) but generally LGTM!

unstructured/metrics/table/table_eval.py
badGarnet: definitely we should explore that as well; for this case I wanted to use similar algorithms we used to do matching a prediction to the actual table to compute the score so choose `difflib`. 
I don't know which will be better yet as the data has a lot of structural text in it (it is a dictionary) so token matching can be tricky (with structural infor tokens always matches).

PR #2623, opened 2024-03-08
fix: docker-publish build test missing key error

ryannikolaidis:
The docker-publish github actions workflow builds amd and arm images of the repository and tests them before publishing. These tests have been failing since [this commit](https://github.com/Unstructured-IO/unstructured/commit/ee8b0f93dcba5adb2e4e032346fbaafbe0f45e54) with an error `UNS_API_KEY environment variable not set`.

The issue is that [this line](https://github.com/Unstructured-IO/unstructured/blob/b27ad9b6aa4cb29d5d3194cb1f1a0bf1fbaab30b/.github/workflows/docker-publish.yml#L62) in the workflow is actually blowing away the value assigned to the file in the previous line

## Changes

* Update line that was overwriting the assignment of UNS_API_KEY to the uns_test_env_file in the docker-publish workflow to leverage the `>>` operator so that UNSTRUCTURED_HF_TOKEN assignment is only appended.
* [bonus]: arithmetic expansion in version-sync.sh to keep shell-check happy

## Testing

To validate, I edited the docker-publish workflow to trigger on push (and to run the test but not publish the workflow) in [this commit](https://github.com/Unstructured-IO/unstructured/pull/2623/commits/0f04f5f0f7912853822239d2c0ba43608e71509a). The successful test results can be reviewed [here](https://github.com/Unstructured-IO/unstructured/actions/runs/8199826803).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NLTK_DATA: ${{ github.workspace }}/nltk_data
-      UNSTRUCTURED_HF_TOKEN: ${{ secrets.HF_TOKEN }}
     needs: [setup, lint]
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


potter-potter:
Good catch. Yep `uns_test_env_file` would be written over. Surprised it didn't show up earlier (line was added 5 months ago).

potter-potter (approved):
looks good.

ryannikolaidis:
> Good catch. Yep `uns_test_env_file` would be written over. Surprised it didn't show up earlier (line was added 5 months ago).

Thanks. Yea, it looks like we were skipping tests that actually needed that key. So it wasn't until a new test leveraged it that this started breaking.

potter-potter:
I'm putting it back in merge queue

PR #2635, opened 2024-03-11
feat: grouping eval takes list of filenames

Klaijan:
Add features to `get_mean_grouping` to allow input as a list of filenames in the format of List of strings or txt file.


unstructured/metrics/evaluate.py
```diff
@@ -173,66 +173,101 @@ def measure_element_type_accuracy(
...
         raise ValueError("Invalid grouping category. Returning a non-group evaluation.")
+
+    if eval_name == "text_extraction":
+        agg_fields = ["cct-accuracy", "cct-%missing"]
+        agg_name = "cct"
+    elif eval_name == "element_type":
+        agg_fields = ["element-type-accuracy"]
+        agg_name = "element-type"
+    else:
+        raise ValueError("Unknown metric. Expected `text_extraction` or `element_type`.")
```
pawel-kmiecik, line 221: shouldn't we add `table_structure` as well?

unstructured/metrics/evaluate.py
```diff
@@ -327,32 +360,79 @@ def measure_table_structure_accuracy(
...
-            "total_tables",
-            "table_level_acc",
-            "element_col_level_index_acc",
-            "element_row_level_index_acc",
-            "element_col_level_content_acc",
-            "element_row_level_content_acc",
-        ]:
-            metric_df = has_tables_df[has_tables_df[metric].notnull()]
-            agg_metric = metric_df[metric].agg([_mean, _stdev, _pstdev, _count]).transpose()
-            if agg_metric.empty:
```
pawel-kmiecik, line 343: That was introduced to cover cases where all values in columns are nans. I don't think we can get back to the previous solution.

unstructured/metrics/evaluate.py
```diff
@@ -327,32 +360,79 @@ def measure_table_structure_accuracy(
...
-                element_metrics_results[metric] = pd.Series(
-                    data=[None, None, None, 0], index=["_mean", "_stdev", "_pstdev", "_count"]
-                )
-            else:
-                element_metrics_results[metric] = agg_metric
-        agg_df = pd.DataFrame(element_metrics_results).transpose().reset_index()
+        # filter out documents with no tables
+        having_table_df = df[df["total_tables"] > 0]
```
pawel-kmiecik, line 366: If we're touching this - it would be good to safely remove this condition. It checks if ground truth file has some tables. But it's possible that it does not contain tables, but we'll find some (false positive). At the same time we should cover various scenarios when data contains NaNs or is empty at all.

unstructured/metrics/evaluate.py
```diff
@@ -327,32 +360,79 @@ def measure_table_structure_accuracy(
...
-            "table_level_acc",
-            "element_col_level_index_acc",
-            "element_row_level_index_acc",
-            "element_col_level_content_acc",
-            "element_row_level_content_acc",
-        ]:
-            metric_df = has_tables_df[has_tables_df[metric].notnull()]
-            agg_metric = metric_df[metric].agg([_mean, _stdev, _pstdev, _count]).transpose()
-            if agg_metric.empty:
-                element_metrics_results[metric] = pd.Series(
```
pawel-kmiecik, line 344: we're doing this separately per metric because some columns might be full of NaNs while others might contains some proper values.

unstructured/metrics/evaluate.py
```diff
@@ -327,32 +360,79 @@ def measure_table_structure_accuracy(
...
+        agg_df.columns = agg_headers
 
-    agg_df.columns = agg_headers
-    _write_to_file(
-        export_dir, "all-docs-table-structure-accuracy.tsv", _rename_aggregated_columns(df)
-    )
-    _write_to_file(
-        export_dir, "aggregate-table-structure-accuracy.tsv", _rename_aggregated_columns(agg_df)
-    )
+    _write_to_file(export_dir, "all-docs-table-structure-accuracy.tsv", df)
```
pawel-kmiecik, line 381: why do we remove renaming?

unstructured/metrics/evaluate.py
```diff
@@ -327,32 +360,79 @@ def measure_table_structure_accuracy(
...
+
+
+def filter_metrics(
+    data_input: Union[str, pd.DataFrame],
+    filter_list: Union[str, List[str]],
+    filter_by: str = "filename",
+    export_filename: Optional[str] = None,
+    export_dir: str = "metrics",
+    return_type: str = "file",
+) -> Optional[pd.DataFrame]:
```
pawel-kmiecik, line 393: I think we need a nice docstring here describing what this function is exactly doing and what each of the parameters means as well as the returned value.

pawel-kmiecik (changes requested):

pawel-kmiecik (approved):
LGTM!

PR #2639, opened 2024-03-11
feat(chunking): composite text gets is_continuation

scanny:
**Summary**
Add `metadata.is_continuation = True` to metadata of second-and-later text-split chunks formed from an oversized non-table element. Previously this metadata was only present on text-split `TableChunk` elements.

This enables downstream filtering of intentionally redundant metadata on chunk elements that may not be desired for all purposes.

### DIFF
--- a/unstructured/chunking/base.py
+++ b/unstructured/chunking/base.py
@@ -536,14 +536,22 @@ class TextPreChunk:
 
     def iter_chunks(self) -> Iterator[CompositeElement]:
         """Split this pre-chunk into one or more `CompositeElement` objects maxlen or smaller."""
+        # -- a pre-chunk containing no text (maybe only a PageBreak element for example) does not
+        # -- generate any chunks.
+        if not self._text:
+            return
+
         split = self._opts.split
-        metadata = self._consolidated_metadata
 
-        remainder = self._text
+        # -- emit first chunk --
+        s, remainder = split(self._text)
+        yield CompositeElement(text=s, metadata=self._consolidated_metadata)
 
+        # -- an oversized pre-chunk will have a remainder, split that up into additional chunks.
+        # -- Note these get continuation_metadata which includes is_continuation=True.
         while remainder:
             s, remainder = split(remainder)
-            yield CompositeElement(text=s, metadata=metadata)
+            yield CompositeElement(text=s, metadata=self._continuation_metadata)
 
     @lazyproperty
     def overlap_tail(self) -> str:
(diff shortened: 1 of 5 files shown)


unstructured/chunking/base.py
```diff
@@ -604,6 +612,20 @@ def _consolidated_metadata(self) -> ElementMetadata:
...
+    def _continuation_metadata(self) -> ElementMetadata:
+        """Metadata applicable to the second and later text-split chunks of the pre-chunk.
+
+        The same metadata as the first text-split chunk but includes `.is_continuation = True`.
+        Unused for non-oversized pre-chunks since those are not subject to text-splitting.
+        """
+        # -- we need to make a copy, otherwise adding a field would also change metadata value
+        # -- already assigned to another chunk (e.g. the first text-split chunk). Deep-copy is not
+        # -- required though since we're not changing any collection fields.
+        continuation_metadata = copy.copy(self._consolidated_metadata)
```
Coniferish, line 625: Only question is if this should be `copy.deepcopy` instead of just `copy.copy`. Otherwise, LGTM!

scanny: `copy.copy` makes a shallow copy. This means that the object is a new instance, but all its fields have the same references (point to the same values/memory-locations). None of those fields are "duplicated". This is enough for us because the only change we are making is changing the value of a scalar field (non-collection field).

Note that scalar fields are immutable. So when you assign a different value you get a new object in a new memory location:
```python
>>> x = 3
>>> y = x
>>> x is y
True
>>> y = 4
>>> x is y
False
>>> x
3
>>> y
4
```

Since `continuation_metadata` is a new `ElementMetadata` instance, when we change the value of its `.is_continuation` field, the value of that field in the original is unchanged, which is what we want and why we needed to make the copy.

Note that if we were modifying a _collection_ field like `.link_urls' by modifying the collection itself (say `continuation_metadata.link_urls.append("https://foo.bar")` then `copy.copy` wouldn't be enough because we would be mutating the same list pointed to by `consolidated_metadata`. We could however replace the whole list and that would be fine, like `continuation_metadata.link_urls = ["https://foo.bar"]`.

This is one of the places where it pays to understand how Python binds names to values. It's internal details and most folks probably don't bother, but it lets you confidently make the right choice in places like this rather than blindly deepcopy everything :)

```python
>>> import copy
>>> from unstructured.documents.elements import ElementMetadata

>>> m = ElementMetadata(filename="foobar.txt", link_texts=["a", "b"])
>>> m2 = copy.copy(m)

>>> m is m2
False
>>> m.filename == m2.filename
True
>>> m.filename is m2.filename
True
>>> m.link_texts is m2.link_texts
True

>>> m2.filename = "barfoo.txt"
>>> m.filename == m2.filename
>>> m.filename
'foobar.txt'
>>> m2.filename
'barfoo.txt'

>>> m2.link_texts.append("c")
>>> m.link_texts
['a', 'b', 'c']

>>> m2.link_texts = ['x', 'y', 'z']
>>> m.link_texts
['a', 'b', 'c']
>>> m2.link_texts
['x', 'y', 'z']
```

Coniferish (approved):
LGTM :)

PR #2645, opened 2024-03-13
rfctr(chunking): simplify chunking opts construction

scanny:
**Summary**
Use omnibus `kwargs` dict for `ChunkingOptions` state rather than explicit option parameters.

**Additional Context**
While articulating explicit options for `ChunkingOptions` and its (now several) sub-classes provides some type-safety, it induces a large amount of redundancy which complicates updates to the base class and especially patches to the base class from client code that adds custom chunkers.

In particular, it makes custom chunkers brittle to any new attributes added to `ChunkingOptions` (the base class).

Use a single omnibus `kwargs` argument to `ChunkingOptions` and its subclasses allowing each to pull out the options it is interested in and happily ignore the rest.

The type safety provided by explicit parameters and types is only afforded to the single place the options item is called from which is the custom chunker itself. Because this is "internal" code and not part of the public interface, this is a manageably small loss in type-safety.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.12.7-dev1
+## 0.12.7-dev2
 
 ### Enhancements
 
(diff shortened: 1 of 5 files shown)


Coniferish (approved):
LGTM

PR #2647, opened 2024-03-14
rfctr: prepare for adding metadata.orig_elements field

scanny:
**Summary**
Some typing modernization in `elements.py` which will get changes to add the `orig_elements` metadata field.

Also some additions to `unit_util.py` to enable simplified mocking that will be required in the next PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ select = [
 ]
 ignore = [
     "COM812",   # -- over aggressively insists on trailing commas where not desireable --
+    "PT005",    # -- flags mock fixtures with names intentionally matching private method name --
     "PT011",    # -- pytest.raises({exc}) too broad, use match param or more specific exception --
     "PT012",    # -- pytest.raises() block should contain a single simple statement --
     "SIM117",   # -- merge `with` statements for context managers that have same scope --
--- a/test_unstructured/unit_utils.py
+++ b/test_unstructured/unit_utils.py
@@ -1,12 +1,41 @@
 """Utilities that ease unit-testing."""
 
+from __future__ import annotations
+
 import datetime as dt
 import difflib
 import pathlib
-from typing import List, Optional
+from typing import Any, List, Optional
+from unittest.mock import (
+    ANY,
+    MagicMock,
+    Mock,
+    PropertyMock,
+    call,
+    create_autospec,
+    mock_open,
+    patch,
+)
+
+from pytest import FixtureRequest, LogCaptureFixture  # noqa: PT013
 
 from unstructured.documents.elements import Element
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -31,6 +31,7 @@ select = [
 ]
 ignore = [
     "COM812",   # -- over aggressively insists on trailing commas where not desireable --
+    "PT005",    # -- flags mock fixtures with names intentionally matching private method name --
     "PT011",    # -- pytest.raises({exc}) too broad, use match param or more specific exception --
```
Coniferish, line 35: Does this change need to be replicated in in `.pre-commit-config.yaml` and `Makefile`? (note at the top)

test_unstructured/unit_utils.py
```diff
@@ -1,13 +1,42 @@
...
+    "Mock",
+    "call",
+    "class_mock",
+    "function_mock",
+    "initializer_mock",
+    "instance_mock",
+    "method_mock",
+    "property_mock",
+)
+
```
Coniferish, line 39: Not that this needs to be changed, but isn't `__all__` used to define what's imported when you use * to import, or is there another use?

test_unstructured/unit_utils.py
```diff
@@ -54,3 +83,136 @@ def example_doc_path(file_name: str) -> str:
 def parse_optional_datetime(datetime_str: Optional[str]) -> Optional[dt.datetime]:
     """Parse `datetime_str` to a datetime.datetime instance or None if `datetime_str` is None."""
     return dt.datetime.fromisoformat(datetime_str) if datetime_str else None
+
+
+# ------------------------------------------------------------------------------------------------
+# MOCKING FIXTURES
+# ------------------------------------------------------------------------------------------------
+# These allow full-featured and type-save mocks to be created simply by adding a unit-test
```
Coniferish, line 91: type-save or type-safe?

pyproject.toml
scanny: Hmm, yeah, good call. I've removed those redundant explicit options definitions. Now they should be picked up from the `pyproject.toml` so it's the single source of truth on options.

test_unstructured/unit_utils.py
scanny: Yes, it is, but it also indicates what you are intentionally "exporting" and prevents ruff errors related to "xyz is imported but not used". That's why I have it here.

test_unstructured/unit_utils.py
scanny: Yep, good catch. Fixed.

Coniferish (approved):
LGTM

PR #2649, opened 2024-03-15
feat(chunking): add include_orig_elements chunking option

scanny:
**Summary**
Add `include_orig_elements: bool = True` as a new chunking option. This PR does not implement _adding_ original elements to chunks, only accepting this parameter as a chunking option and assigning `True` to it as a default value when it is omitted as a keyword argument.

Note this will need to be added in other repositories as well in order to fully support this new option by all access methods. In particular it will need to be added in `unstructured-api` in order to become available via the SDKs.

### DIFF
--- a/unstructured/chunking/base.py
+++ b/unstructured/chunking/base.py
@@ -136,6 +136,15 @@ class ChunkingOptions:
         return arg_value if arg_value is not None else CHUNK_MAX_CHARS_DEFAULT
 
     @lazyproperty
+    def include_orig_elements(self) -> bool:
+        """When True, add original elements from pre-chunk to `.metadata.orig_elements` of chunk.
+
+        Default value is `True`.
+        """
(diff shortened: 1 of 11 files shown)


unstructured/chunking/base.py
```diff
@@ -135,6 +135,15 @@ def hard_max(self) -> int:
...
         return arg_value if arg_value is not None else CHUNK_MAX_CHARS_DEFAULT
 
+    @lazyproperty
+    def include_orig_elements(self) -> bool:
+        """When True, add original elements from pre-chunk to `.metadata.orig_elements` of chunk.
+
+        Default value is `True`.
+        """
+        arg_value = self._kwargs.get("include_orig_elements")
+        return True if arg_value is None else bool(arg_value)
```
Coniferish, line 145: Why use `bool(arg_value)` here?

Coniferish (approved):
LGTM

PR #2655, opened 2024-03-15
Feat: single table structure eval metric

badGarnet:
Creates a compounding metric to represent table structure score. It is an average of existing row and col index and content score.

This PR adds a new property to `unstructured.metrics.table_eval.TableEvaluation`: `composite_structure_acc`, which is computed from the element level row and column index and content accuracy scores. This new metric is meant to offer a single number to represent the performance of table structure extraction model/algorithms. 

This PR also refactors the eval computation logic so it uses a constant `table_eval_metrics` instead of hard coding the name of the metrics in multiple places in the code.


pawel-kmiecik (commented):
If we assume `index_acc` metrics are more important than `content_acc` then LGTM!

pawel-kmiecik (approved):

PR #2657, opened 2024-03-17
refactor title.py

Coniferish:
Minor refactor after conversation with @scanny

Updates docstring and how chunking options are accessed. `self._kwargs.get()` should only be used in the `lazyproperty` definition of an instance's attribute. Other calls should use `self.<attribute>`

### DIFF
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -644,13 +644,11 @@ class Describe_ByTitleChunkingOptions:
     def it_does_not_complain_when_specifying_new_after_n_chars_by_itself(self):
         """Caller can specify `new_after_n_chars` arg without specifying any other options."""
         try:
-            opts = _ByTitleChunkingOptions(new_after_n_chars=200)
-            opts._validate()
+            opts = _ByTitleChunkingOptions.new(new_after_n_chars=200)
         except ValueError:
             pytest.fail("did not accept `new_after_n_chars` as option by itself")
 
         assert opts.soft_max == 200
-        assert opts.combine_text_under_n_chars == 500
 
     @pytest.mark.parametrize(
         ("multipage_sections", "expected_value"),
--- a/test_unstructured/chunking/test_title.py
+++ b/test_unstructured/chunking/test_title.py
@@ -648,19 +648,15 @@ class Describe_ByTitleChunkingOptions:
             )
 
     def it_does_not_complain_when_specifying_new_after_n_chars_by_itself(self):
-        """Caller can specify `new_after_n_chars` arg without specifying any other options.
-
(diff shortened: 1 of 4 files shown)


scanny (approved):
Yep, LGTM :)

test_unstructured/chunking/test_title.py
```diff
@@ -642,19 +642,15 @@ def it_rejects_combine_text_under_n_chars_greater_than_maxchars(
             )
 
     def it_does_not_complain_when_specifying_new_after_n_chars_by_itself(self):
-        """Caller can specify `new_after_n_chars` arg without specifying any other options.
-
-        In particular, `combine_text_under_n_chars` value is adjusted down to the
-        `new_after_n_chars` value when the default for `combine_text_under_n_chars` exceeds the
-        value of `new_after_n_chars`.
-        """
+        """Caller can specify `new_after_n_chars` arg without specifying any other options."""
```
scanny, line 645: Good catch on this docstring :)

test_unstructured/chunking/test_title.py
```diff
@@ -642,19 +642,15 @@ def it_rejects_combine_text_under_n_chars_greater_than_maxchars(
...
         try:
             opts = _ByTitleChunkingOptions(new_after_n_chars=200)
+            opts._validate()
```
scanny, line 648: better to use `_ByTitleChunkingOptions.new(...)` here instead of explicitly calling `opts._validate()` I think. A user should never have occasion to call the constructor directly.

test_unstructured/chunking/test_title.py
```diff
@@ -642,19 +642,15 @@ def it_rejects_combine_text_under_n_chars_greater_than_maxchars(
...
         except ValueError:
             pytest.fail("did not accept `new_after_n_chars` as option by itself")
 
         assert opts.soft_max == 200
-        assert opts.combine_text_under_n_chars == 200
+        assert opts.combine_text_under_n_chars == 500
```
scanny, line 653: No need for an assert here on `.combine_text_under_n_chars`. The default behavior of that property is not under test here. That should be covered by a test on the `.combine_text_under_n_chars` property.

unstructured/chunking/title.py
```diff
@@ -104,6 +104,10 @@ class _ByTitleChunkingOptions(ChunkingOptions):
     multipage_sections
         Indicates that page-boundaries should not be respected while chunking, i.e. elements
         appearing on two different pages can appear in the same chunk.
+    combine_text_under_n_chars
```
scanny, line 107: Keep the options in alphabetical order. I think they show up that way in all the other lists and it becomes more important as more are added.

scanny (approved):
Yep, LGTM. Some small polishing items to fix up before merge but approving in advance.

Issue #2658, opened 2024-03-17
bug/unstructured-ingest produces ModuleNotFoundError: No module named 'unstructured.txtgest


ajjimeno:
**Describe the bug**
Using the latest version of unstructured produces the error "ModuleNotFoundError: No module named 'unstructured.txtgest" when using unstructured-ingest from command line. 

**To Reproduce**
Once unstructured is installed, when running the following command from the command line prompt the error above happens.

```
unstructured-ingest \                          
 s3 \
 --remote-url s3://utic-dev-tech-fixtures/small-pdf-set/ \
 --anonymous \
 --output-dir s3-small-batch-output \
 --num-processes 2
```

**Expected behavior**
The files from s3 are copied to the local folder.

**Screenshots**
N/A

**Environment Info**
N/A

**Additional context**
N/A

MthwRobinson:
@ajjimeno @rbiseck3 - Do you know if this is still an issue?

PR #2661, opened 2024-03-18
bug/unstructured-ingest produces ModuleNotFoundError: No module named 'unstructured.txtgest

rbiseck3:
Quick fix for issue https://github.com/Unstructured-IO/unstructured/issues/2658


MthwRobinson (approved):
LGTM

PR #2662, opened 2024-03-18
initial commit

potter-potter:


unstructured/ingest/cli/common.py
```diff
@@ -6,10 +6,10 @@
 def options_redactions(options: dict) -> dict:
```
potter-potter, line 6: @rbiseck3 Should this whole file cli/common.py be removed since we don't need options_redactions

Or should we just remove options_redactions?

Or should we leave it around just in case? 

it does have lot_options.

rbiseck3: If nothing in this file is being used anywhere, feel free to drop it.

potter-potter: Well. Both source.py and dest.py use log_options. So I will leave it. And delete the options_redactions

rbiseck3 (approved):

PR #2668, opened 2024-03-19
rfctr: prepare to add orig_elements serde

scanny:
**Summary**
The serialization and deserialization (serde) of `metadata.orig_elements` will be located in `unstructured.staging.base` alongside `elements_to_json()` and other existing serde functions. Improve the typing, readability, and structure of that module before adding the new serde functions for `metadata.orig_elements`.

**Reviewers:** The commits are well-groomed and are probably quicker to review commit-by-commit than as all files-changed at once.


test_unstructured/documents/test_elements.py
```diff
@@ -66,13 +65,10 @@ def test_text_element_apply_cleaners():
 
 
 def test_text_element_apply_multiple_cleaners():
-    cleaners = [
-        partial(clean_prefix, pattern=r"\[\d{1,2}\]"),
-        partial(translate_text, target_lang="ru"),
-    ]
-    text_element = Text(text="[1] A Textbook on Crocodile Habitats")
+    cleaners = [partial(clean_prefix, pattern=r"\[\d{1,2}\]"), partial(clean_bullets)]
+    text_element = Text(text="[1] \u2022 A Textbook on Crocodile Habitats")
```
scanny, line 69: This test was running slow, hanging for several seconds on every test run, because it invokes a translation model. Since translation is not the behavior under test here (working with multiple cleaners is), choose a different cleaner that does not incur that excessive overhead.

test_unstructured/documents/test_elements.py
```diff
@@ -82,7 +78,7 @@ def bad_cleaner(s: str):
     text_element = Text(text="[1] A Textbook on Crocodile Habitats")
 
     with pytest.raises(ValueError, match="Cleaner produced a non-string output."):
-        text_element.apply(bad_cleaner)  # pyright: ignore[reportGeneralTypeIssues]
+        text_element.apply(bad_cleaner)  # pyright: ignore[reportArgumentType]
```
scanny, line 81: Looks like the latest version of `pyright` defined some more refined type-error types to use instead of `GeneralTypeIssue`. These narrow the ignore to the more descriptive error actually encountered. In this case we're purposely using an argument of the wrong type to trigger the exception behavior under test.

test_unstructured/staging/test_base.py
```diff
@@ -65,13 +60,14 @@ def test_isd_to_elements():
     ]
 
 
-def test_convert_to_csv(output_csv_file):
+def test_convert_to_csv(tmp_path: str):
+    output_csv_path = os.path.join(tmp_path, "isd_data.csv")
```
scanny, line 64: Inline the fixture this was using since this is the only user and the fixture is trivial.

test_unstructured/staging/test_base.py
```diff
@@ -209,7 +202,7 @@ def test_all_elements_preserved_when_serialized():
     assert base.convert_to_isd(base.isd_to_elements(isd)) == isd
 
 
-def test_serialized_deserialize_elements_to_json(tmpdir):
+def test_serialized_deserialize_elements_to_json(tmpdir: str):
```
Coniferish, line 202: I've noticed we don't type hint as much in tests, but why remove this?

unstructured/staging/base.py
```diff
@@ -216,10 +222,11 @@ def flatten_dict(
                 )
         else:
             flattened_dict[new_key] = value
+
     return flattened_dict
 
 
-def _get_table_fieldnames(rows):
+def _get_table_fieldnames(rows: list[dict[str, Any]]):
```
Coniferish, line 235: Are you using `list` instead of `Sequence` because this is a private method?

Coniferish (approved):
LGTM! Thanks for the organized commits :)

test_unstructured/staging/test_base.py
scanny: Not sure what you mean? We added the type annotation `: str` to `tmpdir` here, so we increased the level of type annotation.

I look to have tests type-check clean on strict, same as the code. Maybe tests tend to start out with less type annotation but they should end up with all they need when I'm done :)

unstructured/staging/base.py
scanny: Hmm, no, this could definitely be `Sequence`. The main reason is the type checker wasn't complaining and we know it will always be this exact type and never `list[some-subtype-of-dict[str, Any]]`. But it wouldn't hurt to change it to `Sequence`.

The reason i use `Sequence` for elements is because `list` is invariant. Invariant means that this function call will type-check:
```python
def f(elements: list[Element]):
    ...

f([Element("abc"), Element("def")])
```
but not this one:
```python
f([Title("abc"), Text("def")])
```

The idea being that a list of `Element` _subtypes_ are not acceptable as a list of `Element`.

`Sequence` on the other hand is _covariant_, which means a sequence of `Element` subtypes is acceptable as `Sequence[Element]`.

The other kind of variance is contra-variance. More on that business in this article: https://www.playfulpython.com/type-hinting-covariance-contra-variance/

And I like this guy's videos on the subject: https://www.youtube.com/watch?v=7Chd5gPHlDg

A good principle is: "be as permissive as possible for input types and as specific as possible for return types". Following that principle, all we need of `rows` is `Iterable[dict[str, Any]]`. So if we were going to change it then probably that would be the best choice.

My intention in these initial mechanical refactorings has just been to get to clean-strict as quickly as possible, so I haven't thought a lot about that, but going forward, as long as we're putting in the effort, it probably makes sense to do as good a job as we can, I'm sure it doesn't take any longer really.

unstructured/staging/base.py
```diff
@@ -78,7 +78,7 @@ def convert_to_isd(elements: Sequence[Element]) -> list[dict[str, Any]]:
 
 
 def convert_to_dict(elements: Sequence[Element]) -> list[dict[str, Any]]:
-    """Converts a list of elements into a dictionary."""
+    """Convert a list of elements into a list of element-dicts."""
```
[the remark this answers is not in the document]
scanny: I think imperative tense is appropriate for docstrings, like "do this" as opposed to "does this". So I'd be more inclined to change any others that say `Converts` to `Convert`.

But my standard with these refactorings is to _improve_ the code rather than to _perfect_ it, mainly because the latter is unachievable and we need to get on with other things. So I'm happy with the Boy Scout rule, which is "leave it cleaner than you found it" :) So I didn't go looking for other cases, although now that you've mentioned it I probably won't be able to stop myself ... :)

test_unstructured/staging/test_base.py
Coniferish: I must have read this backwards when reviewing. Looks good :)

Issue #2461, opened 2024-01-26
Enhancement: better element ID's


cragwolfe:
**Is your feature request related to a problem? Please describe.**

Currently, element_id's are simply a hash of the element's text. This is not great, since id's may then be duplicated within a page or document.

**Proposal**

Deterministic element ID's should be hash of (text, page_num, seq_no_in_page). Then, element_id's would be unique (with extremely high probability) within a document. If processing pages in parallel, element_id's should be consistent as if they were processed in serial (how they are currently processed) instead.

This implies that `metadata_page_number_begin` must also be an optional parameter for `partition()`, and, the API.

**Other considerations**

Hashing with other metadata is potentially fair game, to attempt to keep ID's distinct between documents. Determinism is a must, however.

Initially, this implementation would not effect the partition parameter: [unique_element_ids=True](https://github.com/Unstructured-IO/unstructured/blob/d5a6f4b82c7ac31db10a509ae3e517ab623df777/docs/source/introduction/overview.rst#unique-element-ids).

beyer-martin:
I have an issue with this as well, as parent ids are wrongly set because of this

Starfox64:
The current implementation seems to work pretty well but the deterministic IDs are lost once chunking is applied. Are there plans to extend the feature to chunking ?

PR #2673, opened 2024-03-20
Better element IDs - deterministic and document-unique hashes

micmarty-deepsense:
Part two of: https://github.com/Unstructured-IO/unstructured/pull/2842

Main changes compared to part one:
* hash computation includes element's sequence number on page, page number, document filename and its text
* there are more test for deterministic behavior of IDs returned by partitioning functions + their uniqueness (guaranteed at the document level, and high probability across multiple documents)

This PR addresses the following issue: https://github.com/Unstructured-IO/unstructured/issues/2461


unstructured/partition/common.py
```diff
@@ -560,7 +562,7 @@ def document_to_element_list(
         image_height = page_image_metadata.get("height")
 
         translation_mapping: list[tuple["LayoutElement", Element]] = []
-        for layout_element in page.elements:
+        for index_on_page, layout_element in enumerate(page.elements):
```
micmarty-deepsense, line 565: might want to verify whether this is legit

unstructured/partition/pdf.py
```diff
@@ -519,7 +519,7 @@ def _partition_pdf_or_image_local(
     final_document_layout = clean_pdfminer_inner_elements(final_document_layout)
 
     for page in final_document_layout.pages:
-        for el in page.elements:
+        for page_el_index, el in enumerate(page.elements):
```
micmarty-deepsense, line 522: TODO: unused, gonna revert that change

unstructured/documents/elements.py
```diff
@@ -667,8 +677,12 @@ def __init__(
         # -- defined in a subclass.
         self.text = self.text if hasattr(self, "text") else ""
 
+    @property
+    def id(self) -> str | uuid.UUID | NoID | UUID:
+        return self._id
```
micmarty-deepsense, line 682: `Element` has a very basic behavior, concrete implementation that overrides it is in the `Text` class

unstructured/documents/elements.py
```diff
@@ -772,20 +786,61 @@ def __init__(
...
+        if metadata is not None:
+            # Metadata supported by PDF and HTML files
+            data = f"{self.text}{metadata.page_number}{metadata.index_on_page}{other}"
+        else:
+            data = f"{self.text}{other}"
+
+        return HashValue(hashlib.sha256(data.encode()).hexdigest()[:32])
+
+    @property
+    def id(self) -> str:
```
scanny, line 827: All methods require a docstring.

test_unstructured/partition/test_html_partition.py
```diff
@@ -13,6 +13,9 @@
 from unstructured.documents.elements import EmailAddress, ListItem, NarrativeText, Table, Title
 from unstructured.documents.html import HTMLTitle
 from unstructured.partition.html import partition_html
+from unstructured.staging.base import (
+    convert_to_dataframe,
+)
```
scanny, line 18: This will all fit on one line, won't it?

test_unstructured/partition/test_html_partition.py
```diff
@@ -723,3 +726,43 @@ def test_partition_html_with_table_without_tbody(tag: str, expected: str):
...
     partitions = partition_html(text=table_html)
     assert partitions[0].metadata.text_as_html == expected
+
+
+@pytest.fixture
+def partitioned_html_with_duplicate_elements():
+    filename = "example-docs/fake-html-duplicates.html"
+    with open(filename) as f:
+        elements = partition_html(file=f)
+    return elements
```
scanny, line 736: This resolves to:
```python
return partition_html("example-docs/fake-html-duplicates.html")
```
which can be inlined in the tests that use it so folks don't have to go looking for the fixture..

test_unstructured/partition/test_html_partition.py
```diff
@@ -723,3 +726,43 @@ def test_partition_html_with_table_without_tbody(tag: str, expected: str):
...
+
+
+def test_each_element_has_page_number_and_index_metadata(partitioned_html_with_duplicate_elements):
+    for element in partitioned_html_with_duplicate_elements:
+        assert element.metadata.page_number is not None
+        assert element.metadata.index_on_page is not None
```
scanny, line 742: This test seems a little weak. There's no indication that page_number is an int or index_on_page is not str or something. I think it would be better as something like:
```python
page_index_pairs = [
    (e.metadata.page_number, e.metadata.index_on_page)
    for e in partition_html("example-docs/fake-html-duplicates.html")
]
assert page_index_pairs == [(1, 0), (1, 1), ...]
```

Also this give the user some concrete intuitive sense of how these values proceed along an element-stream.

test_unstructured/partition/test_html_partition.py
```diff
@@ -723,3 +726,43 @@ def test_partition_html_with_table_without_tbody(tag: str, expected: str):
...
+
+
+def test_all_element_ids_are_unique(partitioned_html_with_duplicate_elements):
+    assert convert_to_dataframe(partitioned_html_with_duplicate_elements).element_id.is_unique
```
scanny, line 746: seems the long way around the barn to use `convert_to_dataframe()`, also that couples the test to a dependency outside the unit under test. Could be:
```python
ids = [e.id for e in partition_html("example-docs/fake-html-duplicates.html")]
assert len(ids) == len(set(ids))
```

test_unstructured/partition/test_html_partition.py
```diff
@@ -723,3 +726,43 @@ def test_partition_html_with_table_without_tbody(tag: str, expected: str):
...
+            "a677501bee476d34f96af566d648531e",
+            "7583e5d08909a23e8214f3d1ee87e50b",
+            "490190ffe8519e8cd8b6d750d334397e",
+            "dbf7bcdc973922a41c8c51505b873340",
+            "346cc2595ed5c0b34c7d818b9cc0c891",
+        ],
+    ],
+)
+def test_element_ids_are_deterministic(partitioned_html_with_duplicate_elements, expected_ids):
+    """Test that element IDs are deterministic and match the expected IDs."""
```
scanny, line 767: I think this docstring can go. It really just restates the test name, unless there's something going on I'm not seeing. Matching the expected IDs (on every test run) is how we verify they are deterministic, right?

test_unstructured/partition/test_html_partition.py
```diff
@@ -723,3 +726,43 @@ def test_partition_html_with_table_without_tbody(tag: str, expected: str):
...
+    assert [element.id for element in partitioned_html_with_duplicate_elements] == expected_ids
```
scanny, line 768: Why do you parameterize this test rather than just put the data inline? There's only the one parameter instance here so will only ever run a single test:
```python
assert [e.id for e in partition_html("example-docs/fake-html-duplicates.html")] == [
    "e5fd0a829b734744f52ae195859f7741",
    "a68c75c49dde8b76d6d7208ad6ec9b6a",
    "98a69145a0223bc7f82a75a11293ba3d",
    "f36ded43d030fb5c6146d5961f767a85",
    "d6ca7824f33202a7a86c02c979b42c91",
    "a677501bee476d34f96af566d648531e",
    "7583e5d08909a23e8214f3d1ee87e50b",
    "490190ffe8519e8cd8b6d750d334397e",
    "dbf7bcdc973922a41c8c51505b873340",
    "346cc2595ed5c0b34c7d818b9cc0c891",
]
```

scanny (changes requested):
Quite a few proposed mechanical fixes, but approach-wise I think this is probably the best we can do without cleaning up much more broadly into very old parts of the code (from back when `Document` and `Pages` were a thing) that I think no one wants to get involved with just now.

I think this direction will solve the immediate problem and not increase the internal complexity too much.

I really like that you've managed to narrow `Element.id` to `str` type, which neatly hides all the internal complexity from the user. That's a big improvement I'd say :)

unstructured/documents/elements.py
scanny: `Element` is a base-class (only) as far as I know and is never instantiated directly, only one of its subclasses (at least in a partitioning element-stream). So I'd be inclined to either remove this property or make its return type `str` (consistent with subtypes) and unconditionally raise `AttributeError` with something like `"an instance of the Element base-class cannot have an id"`

unstructured/documents/elements.py
```diff
@@ -772,20 +786,61 @@ def __init__(
...
+        if isinstance(self._id, (NoID, HashValue)):
+            # Ensure that hash is always up-to-date
+            self._id = self.calculate_hash(self.metadata)
+        return str(self._id)
```
scanny, line 831: Here are some underlying domain-model concepts for `Element.id` which I've developed as I've studied these behaviors a little more closely:
### Basic design principles
- **Each `Element` assigns its own ID.** It is not intended for users to choose and assign ids to elements. The code for assigning IDs is internal to the element instance and is exercised on construction. An element deserialized from JSON receives the ID it was assigned by its original `Element` instance before serialization (as a `str` value).
- **Element IDs can be document-unique or globally unique.** An element can have an id based on a hash of its contents that is document-unique and deterministic, or it can have a globally-unique UUID which is not deterministic. Deterministic in this context means repartitioning the document will produce the same ids for all its elements. These are the only two options.
- **Element IDs are type `str`.** Regardless of the type of the `element_id` argument used to construct an element, `element.id` is type `str`. It can never be `None` and a UUID ID is resolved to its standard `str` representation, never the raw `uuid.UUID` object it may have come from. Internal storage is not specified by this and may be non-`str` types at the developer's discretion, but to a user, `element.id` is always a string.
### Construction-time ID behaviors
The `Element` constructor includes an `element_id` argument. This can accept three different values each producing a distinct ID behavior:
- `NoID` (default) - when the `element_id` argument is omitted or otherwise is an instance of `NoID`, the element is assigned a hash-based ID.
- `UUID` - an `element_id` argument that is an instance of `UUID()` (private "dummy" class, not `uuid.UUID`) causes the element to be assigned a UUID ID.
- `str` - a `str` element_id is assumed to be passed only by `unstructured` deserialization code when an element is deserialized from JSON. The element gets this same id initially assigned by the `Element` instance before it was serialized.

Not strictly a "construction-time" behavior, but an element can be assigned a `UUID` ID by calling `element.id_to_uuid()`. This is what the `@process_metadata` decorator does in response to the partitioning argument `unique_element_ids=True`.

Note that passing a `uuid.UUID` object as the `Element` constructor's `element_id` argument is not supported and will raise `ValueError`.

Also note that `Element` itself is a base class and should never be instantiated (in a partition element-stream at least). So such an instance does not (officially) have an `.id`.

scanny: ### `._id_type`

Reasoning with the premises stated above:
- An element implicitly has an `._id_type` attribute, conceptually taking one of the two values `HashValue` or `UUID` (or whatever actual values, this is just the gist).
- The value of this implicit `._id_type` is being inferred in the new `Text.id` property using `isinstance(self._id, (NoID, HashValue))`. Essentially the computation is:
	```python
	id_type = (
	    HashValue
	    if isinstance(self._id, (HashValue, NoID)
	    else UUID
	)
	```
- When `._id_type` is `HashValue`, the id is recomputed on each access, allowing content mutation to be reflected in the id. This is important because in some parts of the code the element is constructed in one step and the metadata added in a separate step.
- When `._id_type` is `UUID` the id is not recomputed.

Here, finally, is the rub. This implicit `._id_type` does not (currently) survive serialization. This means that a deserialized `Element` behaves differently than a "fresh" one that has never been serialized. I consider that a problem. It's way more than I think developers should have to discover for themselves and then be mindful of when debugging some weird bug.

I see three alternatives:
1. Make `Element._id_type` explicit, include it in the serialized JSON, and restore its value on deserialization.
2. Work out the id-type by the form of the string, like length, character set, dashes in certain spots maybe, etc. and restore it on deserialization (`isinstance(element_type, str)` on construction) to `HashValue` at that time rather than `str` such that it will trigger recomputation of the id on each access and behavior is not changed by round-tripping through JSON.
3. Accept that the behavior is way more complicated than we'll be willing or able to explain to users, leave it the way it is, document it carefully here in the code and hope for the best.

In principle I like option 1 but it seems over-engineered. Option 2 might be worthwhile if it's easy and reliable to distinguish a UUID from our SHA-hash. In any case I think we need to document all these quirks carefully (in docstrings and comments) while we understand it so the next developer doesn't have to spend the time we have working out how all this fits together :)

unstructured/documents/elements.py
```diff
@@ -772,20 +786,61 @@ def __init__(
...
-            element_id = hashlib.sha256(text.encode()).hexdigest()[:32]
-
+            self._id = self.calculate_hash(metadata)
         elif isinstance(element_id, UUID):
-            element_id = str(uuid.uuid4())
+            self._id = uuid.uuid4()
+        elif isinstance(element_id, str):
+            self._id = element_id
+        else:
+            raise ValueError("ID must be a string, UUID, or NoID")
```
scanny, line 795: This is real-work in the constructor that can be relocated to the `.id` property like so:
```python
def __init__(...):
    self._id = element_id
    super().__init__(element_id=self.id, ...)  # if we still need the base class to keep track of that for us

...

@property
def id(self) -> str:
    # -- id is hash when element_id=NoID() (default) on construction. This is recomputed on each reference
    # -- because metadata is sometimes added after element construction.
    if isinstance(self._id, NoID):
       # -- not a lot of sense in storing the hash if we're just going to recompute it next time we need it --
        return str(self._calculate_hash())
    # -- id is globally-unique when element_id=UUID() on construction. This is never recomputed --
    if isinstance(self._id, UUID):
        self._id = uuid.uuid4()
    if isinstance(self._id, uuid.UUID):
        return str(self._id)
    # -- element_id is a `str` on construction when deserializing an element from JSON --
    if isinstance(self._id, str)
        return self._id
    raise ValueError ...
```

This prevents "real-work-in-the-constructor" (worth Googling if you're not familiar with the term).

Also, if we do actually need an `.id` setter, it prevents that code from being duplicated.

scanny (commented):
A few more remarks and insights after reflection.

test_unstructured/partition/test_html_partition.py
micmarty-deepsense: 👍 you're right, my implementation was quite naive and needs to be refactored the way you proposed

unstructured/documents/elements.py
```diff
@@ -505,6 +505,43 @@ def field_consolidation_strategies(cls) -> dict[str, ConsolidationStrategy]:
...
+        index_in_sequence: The index of the element in the sequence of elements.
+
+    Returns:
+        The first 32 characters of the SHA256 hash of the concatenated input parameters.
+    """
+    data = f"{text}{page_number}{index_in_sequence}"
+    return hashlib.sha256(data.encode()).hexdigest()[:32]
+
+
+def recalculate_ids(elements: List[Element]) -> List[Element]:
```
micmarty-deepsense, line 524: rename: assign hash ids

unstructured/documents/elements.py
```diff
@@ -505,6 +505,43 @@ def field_consolidation_strategies(cls) -> dict[str, ConsolidationStrategy]:
...
+    """Updates the `id` (and `parent_id`) attributes of each element
+    in the list of elements based on the element's attributes and its index in sequence
+
+    Args:
+        elements: The list of elements whose IDs are to be recalculated.
+
+    Returns:
+        The list of elements with updated IDs.
+    """
+    elements = copy.deepcopy(elements)
```
micmarty-deepsense, line 534: remove

unstructured/documents/elements.py
micmarty-deepsense: done

unstructured/documents/email_elements.py
```diff
@@ -42,6 +42,21 @@ def __init__(
         if isinstance(datestamp, datetime):
             self.datestamp: datetime = datestamp
 
+    def id_to_hash(self, index_in_sequence: int) -> str:
```
micmarty-deepsense, line 45: I discovered that the `Name` and `EmailElement` are almost identical copies of `Text` and `Element`. Honestly, I don't like that design, it requires a refactor - should create a separate issue/PR for thath

unstructured/documents/elements.py
```diff
@@ -505,6 +506,35 @@ def field_consolidation_strategies(cls) -> dict[str, ConsolidationStrategy]:
...
+        The list of elements with updated IDs.
+    """
+
+    old_to_new_id_mapping = {}
+    for idx_in_seq, element in enumerate(elements):
+        old_id = element.id
+        new_id = element.id_to_hash(idx_in_seq)
+        old_to_new_id_mapping[old_id] = new_id
+
+    for element in elements:
```
micmarty-deepsense, line 526: There are cases when elements are in the right order, that's why we need 2 passes.
image

unstructured/documents/elements.py
```diff
@@ -772,20 +786,61 @@ def __init__(
...
+    @id.setter
+    def id(self, value: str | uuid.UUID | NoID | UUID):
+        if isinstance(value, (str, uuid.UUID)):
+            self._id = value
+        elif isinstance(value, UUID):
+            self._id = uuid.uuid4()
+        elif isinstance(value, NoID):
+            self._id = self.calculate_hash(self.metadata)
+        else:
+            raise ValueError("ID must be a string, UUID, or NoID")
```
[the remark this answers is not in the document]
micmarty-deepsense: resolving since this has been already discussed and the code has changed significantly

test_unstructured/partition/test_html_partition.py
micmarty-deepsense: done, can be resolved

test_unstructured/partition/test_html_partition.py
micmarty-deepsense: done, no longer using `conver_to_dataframe`

unstructured/documents/elements.py
```diff
@@ -772,20 +786,61 @@ def __init__(
...
         super().__init__(
-            element_id=element_id,
+            element_id=self._id,
             metadata=metadata,
             coordinates=coordinates,
             coordinate_system=coordinate_system,
             detection_origin=detection_origin,
         )
 
+    def calculate_hash(
```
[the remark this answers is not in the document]
micmarty-deepsense: outdated, please resolve

unstructured/documents/elements.py
micmarty-deepsense: all relevant points are summarized in overview.rst, if I overlooked something crucial, just let me know ;)

test_unstructured/test_utils.py
```diff
@@ -380,3 +381,9 @@ def test_calculate_shared_ngram_percentage_returns_null_vals_for_empty_str():
     percent, common_ngrams = utils.calculate_shared_ngram_percentage(str1, str2, n)
     assert percent == 0
     assert not bool(common_ngrams)
+
+
+def assign_hash_ids(elements: List[Element]) -> List[Element]:
```
scanny, line 386: This needs to go in `unit_utils.py`, not `test_utils.py`. `test_utils` is the test module for the `unstructured.utils` module. Unit-test utilities go in `unit_utils`.

unstructured/documents/elements.py
```diff
@@ -12,7 +12,7 @@
 import re
 import uuid
 from types import MappingProxyType
-from typing import Any, Callable, FrozenSet, Optional, Sequence, cast
+from typing import Any, Callable, FrozenSet, List, Optional, Sequence, cast
```
scanny, line 15: Use lowercase `list` instead of importing and using `Typing.List`. As long as `from __future__ import annotations` appears at the top this will work fine and is Python-version forward-compatible. Occasionally you will need to place the declaration in quotes, like maybe for a `cast()` call or `TypeAlias` instance. The type-checker will tell you where that is necessary but it is infrequent.

unstructured/documents/elements.py
scanny: This doesn't sound quite right to me. Let's discuss.

Two passes sounds right:
1. UUID -> Hash for all elements, saving mapping.
2. Update all parent-ids based on mapping.

I don't see why "already_updated_parent_ids()" is necessary. Wouldn't this get it done?

```python
old_to_new_mapping = {
    e.id: e.id_to_hash(idx)
    for idx, e in enumerate(elements)
}
for e in elements:
    parent_id = e.metadata.parent_id
    if not parent_id:
        continue
    e.metadata.parent_id = old_to_new_mapping(parent_id)
```

unstructured/documents/elements.py
```diff
@@ -651,19 +684,35 @@ def to_dict(cls):
 
 
 class Element(abc.ABC):
-    """An element is a section of a page in the document."""
+    """An element is a section of a page in the document.
```
scanny, line 687: I would change this to: "An element is a portion of a document, generally a paragraph."

The term "section" has a specific meaning in partitioning, a "document-section" which is a "section-heading" and whatever text follows that heading until the next section heading.

Also, in DOCX it has a completely different meaning, basically a collection of pages with the same size and margins.

So better if we avoid that term when we're not talking about one of those two things.

unstructured/documents/email_elements.py
```diff
@@ -1,17 +1,17 @@
-import hashlib
-import uuid
+from __future__ import annotations
+
 from abc import ABC
 from datetime import datetime
-from typing import Callable, List, Union
+from typing import Callable, List
```
scanny, line 5: `List` should not be necessary.

scanny (changes requested):
Looks pretty good. A couple of changes, in particular the post-processing algorithm doesn't look quite right to me.

test_unstructured/documents/test_elements.py
```diff
@@ -29,34 +30,48 @@
...
-    # -- Test that the element is JSON serializable. This shold run without an error --
-    json.dumps(text_element.to_dict())
+@pytest.mark.parametrize(
+    "element",
+    [
+        Element(),
+        Text(text=""),
+        CheckBox(),
+    ],
+)
```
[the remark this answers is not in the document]
micmarty-deepsense: image
image
done

test_unstructured/documents/test_email_elements.py
```diff
@@ -1,26 +1,50 @@
...
-    name_element = Name(name="Example", text="hello there!")
-    assert name_element.id == "c69509590d81db2f37f9d75480c8efed"
+@pytest.mark.parametrize(
+    "element",
+    [
+        EmailElement(text=""),
+        Name(text="", name=""),
+        Subject(text=""),
+    ],
+)
```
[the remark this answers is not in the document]
micmarty-deepsense: image
done

test_unstructured/partition/docx/test_doc.py
```diff
@@ -19,6 +19,39 @@
 from unstructured.partition.docx import partition_docx
 
 
+@pytest.fixture()
```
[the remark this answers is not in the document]
micmarty-deepsense: You're right!

image
done

test_unstructured/partition/docx/test_docx.py
```diff
@@ -864,3 +865,24 @@ def mock_document_file_path(mock_document: Document, tmp_path: pathlib.Path) ->
     filename = str(tmp_path / "mock_document.docx")
     mock_document.save(filename)
     return filename
+
+
+def test_ids_are_unique_and_deterministic(tmpdir):
```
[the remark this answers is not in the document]
micmarty-deepsense: also fixed
image

test_unstructured/test_utils.py
micmarty-deepsense: good catch, I moved it there

unstructured/documents/elements.py
micmarty-deepsense: image
done

unstructured/documents/elements.py
micmarty-deepsense: image

🤔 yeah, "section" may sound misleading, it's fixed now

unstructured/documents/email_elements.py
micmarty-deepsense: done

micmarty-deepsense:
@scanny I've addressed all of your comments, if it looks good to you now, let's merge it today, right after this https://github.com/Unstructured-IO/unstructured/pull/2842

scanny (changes requested):
Let's think again about this in light of parent as object reference rather than parent-id and not merge yet.

unstructured/partition/msg.py
```diff
@@ -120,13 +120,13 @@ def partition_msg(
         last_modification_date = (
             get_last_modified_date_from_file(file) if date_from_file_object else None
         )
-    metadata = build_msg_metadata(
```
micmarty-deepsense, line 123: This fixes a bug: elements were sharing the same metadata object.
Updating an ID for one element used to update IDs for all elements... now each element gets its own metadata

test_unstructured/partition/docx/test_doc.py
```diff
@@ -19,6 +19,23 @@
 from unstructured.partition.docx import partition_docx
 
 
+def test_partition_doc_for_deterministic_and_unique_ids():
+    ids = [
+        element.id for element in partition_doc(filename="example-docs/duplicate-paragraphs.doc")
+    ]
+    assert len(ids) == len(set(ids)), "IDs are not unique"
```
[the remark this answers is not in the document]
micmarty-deepsense: done

test_unstructured/partition/docx/test_docx.py
```diff
@@ -864,3 +864,20 @@ def mock_document_file_path(mock_document: Document, tmp_path: pathlib.Path) ->
     filename = str(tmp_path / "mock_document.docx")
     mock_document.save(filename)
     return filename
+
+
+def test_ids_are_unique_and_deterministic():
+    elements = partition_docx("example-docs/duplicate-paragraphs.docx")
+
+    ids = [e.id for e in elements]
+    assert len(ids) == len(set(ids)), "IDs are not unique"
```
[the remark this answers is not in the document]
micmarty-deepsense: done

test_unstructured/partition/pptx/test_pptx.py
```diff
@@ -470,41 +470,41 @@ def test_partition_pptx_hierarchy_sample_document():
...
 
     # Zip the test cases with the elements
     for element, test_case in zip(elements, test_cases):
-        assert element.metadata.category_depth == test_case[0]
-        assert element.metadata.parent_id == test_case[1]
-        assert element.id == test_case[2]
+        expected_depth, expected_parent_id, expected_id = test_case
+        assert element.metadata.category_depth == expected_depth
+        assert element.metadata.parent_id == expected_parent_id
+        assert element.id == expected_id
```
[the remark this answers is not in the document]
micmarty-deepsense: thx ;)

micmarty-deepsense:
> Let's think again about this in light of parent as object reference rather than parent-id and not merge yet.
I'd like to tackle this in a separate PR, there're too many changes in this PR already I think 🦖

scanny:
@micmarty-deepsense
> I'd like to tackle this in a separate PR, there're too many changes in this PR already I think 🦖

I agree. This PR did a lot for us, but now all that discussion is resolved I think we're far better off just letting this one go and starting a fresh one :)

micmarty-deepsense:
@scanny 
It looks like I need to update the all `element_id` values for the `spring-weather.html.json` file, am I right?
Here's the job that fails: https://github.com/Unstructured-IO/unstructured/actions/runs/8742887283/job/23996742775

image

cragwolfe:
> @scanny It looks like I need to update all the `element_id` values for the `spring-weather.html.json` file, am I right? Here's the job that fails: https://github.com/Unstructured-IO/unstructured/actions/runs/8742887283/job/23996742775
> 

yep.

scanny:
@micmarty-deepsense

> @scanny It looks like I need to update all the `element_id` values for the `spring-weather.html.json` file, am I right?

I'd be inclined to solve this by:
- getting the test running locally
- comment out most of it just leaving a part that does `elements = elements_from_json(filename)`
- call the equivalent of `e.id_to_hash() for e in elements`
- write that out to a new file with something like `elements_to_json(new_filename)`

and see if that does the bulk of the work for you :)

cragwolfe:
> > @scanny It looks like I need to update all the `element_id` values for the `spring-weather.html.json` file, am I right? Here's the job that fails: https://github.com/Unstructured-IO/unstructured/actions/runs/8742887283/job/23996742775
> 
> yep.

actually, wait. this file is in test_unstructured_ingest/expected-structured-output , where element_id's and all json should be updated by https://github.com/Unstructured-IO/unstructured/actions/workflows/ingest-test-fixtures-update-pr.yml (need to click Run Workflow against your branch).

unstructured/partition/docx.py
```diff
@@ -90,6 +90,7 @@ def convert_and_partition_docx(
     metadata_last_modified: Optional[str] = None,
     languages: Optional[List[str]] = ["auto"],
     detect_language_per_element: bool = False,
+    starting_page_number: int = 1,
```
micmarty-deepsense, line 93: I missed that parameter in the previous PR

test_unstructured/partition/test_auto.py
```diff
@@ -212,26 +212,34 @@ def test_auto_partition_html_from_file_rb():
     assert len(elements) > 0
 
 
-def test_auto_partition_json_from_filename():
+def test_auto_partitioned_json_output_maintains_consistency_with_fixture_elements():
```
micmarty-deepsense, line 215: IMO this test asked for more verbose name, I'm open to suggestions if there's a better one

test_unstructured/partition/docx/test_docx.py
```diff
@@ -866,3 +866,18 @@ def mock_document_file_path(mock_document: Document, tmp_path: pathlib.Path) ->
...
+
+    ids = [e.id for e in elements]
+    assert ids == [
+        "2f22d82eea1faf5f40dac60cef52700e",
+        "ca9e1f448e531a5152d960e14eefc360",
+        "9ddeacb172ac17fb45e6f3f15f3c703d",
+        "a4fd85d3f4141acae38c8f9c936ed2f3",
+        "44ebaaf66640719c918246d4ccba1c45",
+        "f36e8ebcb3b6a051940a168fe73cbc44",
+        "532b395177652c7d61e1e4d855f1dc1d",
```
cragwolfe, line 882: don't we just care about the length of distinct element id's rather than the id's themselves? (not a blocker)

micmarty-deepsense: At some point I've been doing this in combination with:
```
assert len(ids) == len(set(ids))
```
Steve suggested that it's redundant since checking hashes explicitly already ensures their uniqueness.

unstructured/documents/elements.py
```diff
@@ -509,8 +510,19 @@ def assign_and_map_hash_ids(elements: list[Element]) -> list[Element]:
     Returns:
         List of updated Element objects with hashes for `id` and `parent_id`.
     """
-    old_to_new_mapping = {e.id: e.id_to_hash() for e in elements}
-
+    # -- generate sequence number for each element on a page --
+    page_numbers = [e.metadata.page_number for e in elements]
+    page_seq_pairs = [
+        seq_on_page for page, group in groupby(page_numbers) for seq_on_page, _ in enumerate(group)
+    ]
```
scanny, line 517: I think this is going to require more testing. The following cases must be verified by tests:
- All `element.metadata.page_number` values are `None`, such as for file-formats that do not support pagination. That would include sometimes-HTML, sometimes-DOCX, probably CSV, maybe TXT, and others.
- Some `element.metadata.page_number` values are `None`, like `PageBreak("")` and maybe `PageNumber()`.

The latter case could cause false groups/restarts.

I think this is the logic we want:
```python
    def iter_element_sequence_number_pairs():
        current_page_number, sequence_number = 1, 1
        for e in elements:
            page_number = e.metadata.page_number
            # -- An occasional element such as PageBreak() that has no page-number assumes the current
            # -- page number. An element stream with no page-numbers gets page_number=1 for each
            # -- of its elements.
            if page_number and page_number != current_page_number:
                current_page_number = page_number
                sequence_number = 1
            yield e, sequence_number
            sequence_number += 1
```

As you know I usually have a preference for comprehensions, but given the special-case logic required here I'm inclined to implement it as a generator like this (can be a function internal to `assign_and_map_hash_ids()`). That way it can be tweaked and commented as required to deal with exception cases.

unstructured/documents/elements.py
```diff
@@ -737,23 +749,27 @@ def convert_coordinates_to_new_system(
...
 
-        Returns:
-            The first 32 characters of the SHA256 hash of the concatenated input parameters.
+        Args:
+            sequence_number: index on page
+
+        Returns: new ID value
         """
-        self._element_id = hashlib.sha256(self.text.encode()).hexdigest()[:32]
+        data = f"{self.metadata.filename}{self.text}{self.metadata.page_number}{sequence_number}"
```
scanny, line 763: I think we're going to need some similar solution here, aren't we? `self.metadata.page_number` can be `None` still, right? And I think we said it should be 1 in that case, but I think it actually needs to be a little more sophisticated to deal with the "occasional None" case; otherwise page_number could go ... 6, 6, 1, 6, 7, 8, 1, ...

I'm inclined to think this method needs to take `page_number` as well as `sequence_number` (which should probably be named `page_sequence_number` btw).

Unless we've somehow established that all elements always have `.metadata.page_number` (and I don't think we have).

scanny (changes requested):
Hi Mike, let's chat about this if we can. Ping when you get in, in case I'm still up.

A few changes needed I think and a couple cases that need to be covered by new tests.

unstructured/documents/elements.py
cragwolfe: are you sure we need to solve this?
we should not be writing `self.metadata.page_number` for this PR, only using it for hash calculations? and if it is None, just assume 1?

scanny: @Crag: A couple things:
- This will be using "None" for the page-number, not "1".
- Therefore the "document-unique" promise is not guaranteed.

I had understood that the "document-unique" guarantee was a hard commit. But if we're satisfied with "usually mostly document-unique" then I suppose this will get it done :)

cragwolfe: if there are None page_number elements in a PDF, that sounds like a bug.
the biggest concern is PDF's. what case are you worried about?

scanny: This code will be executed for all file-types, not just PDF, TIFF, and PPTX that "guarantee" page-numbers.

scanny: So there are two cases across file-types:
- No page numbers
- Occasional omitted page numbers.

No-elements-have-page-numbers happens in formats like TXT and I'm sure several others. Occasional-omitted-page-numbers are like `PageBreak` and perhaps `PageNumber` elements are not assigned page-number metadata.

These are the cases that the proposed generator above addresses.

cragwolfe: > No-elements-have-page-numbers happens in formats like TXT
but then each element should have a different sequence number, as if it was one giant page

> Occasional-omitted-page-numbers are like PageBreak
if a PageBreak with no page number appears in a PDF or PPTX (or other "real page numbers" doc), imo that is a bug. the page number should be the page it is marking the end of.

scanny: >> No-elements-have-page-numbers happens in formats like TXT

> but then each element should have a different sequence number, as if it was one giant page

Hmm, yes, good point. I suppose "None" is as good as "1" in that case, as long as we're consistent. I agree, that handles that case well enough since we aren't page-splitting those formats.

>> Occasional-omitted-page-numbers are like PageBreak

> if a PageBreak with no page number appears in a PDF or PPTX (or other "real page numbers" doc), imo that is a bug.

Okay, that's fair enough. We can check those formats and make sure they assign at least page-number metadata to those elements.

PR #2680, opened 2024-03-21
feat(chunking): add metadata.orig_elements serde

scanny:
**Summary**
This final PR in the "orig_elements" series adds the needful such that `.metadata.orig_elements`, when present on a chunk (element), is serialized to JSON when the chunk is serialized, for instance, to be used in an HTTP response payload.

It also provides for deserializing such a JSON payload into chunks that contain the `.orig_elements` metadata.

**Additional Context**
Note that `.metadata.orig_elements` is always `Optional[list[Element]]` when in memory. However, those original elements are serialized as Base64-encoded gzipped JSON and are in that form (str) when present as JSON or as "element-dicts" which is an intermediate serialization/deserialization format. That is, serialization is `Element -> dict -> JSON` and deserialization is `JSON -> dict -> Element` and `.orig_elements` are Base64-encoded in both the `dict` and `JSON` forms.

### DIFF
--- a/docs/source/core/chunking.rst
+++ b/docs/source/core/chunking.rst
@@ -65,8 +65,8 @@ be specified when a non-default setting is required. Specific chunking strategie
   need to decide based on your use-case whether this option is right for you.
 
 
-Chunking elements
------------------
+Chunking
+--------
 
 Chunking can be performed as part of partitioning or as a separate step after
 partitioning:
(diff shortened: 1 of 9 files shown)


docs/source/core/chunking.rst
```diff
@@ -170,3 +170,45 @@ following behaviors:
...
+
+    >>> [e.metadata.coordinates for e in chunk.metadata.orig_elements]
+    [<CoordinatesMetadata ...>, <CoordinatesMetadata ...>, ...]
+
+    >>> [
+        e.metadata.image_path
+        for e in chunk.metadata.orig_elements
+        if e.metadata.image_path is not None
+    ]
+    ['/tmp/lorem.jpg', '/tmp/ipsum.png']
```
Coniferish, line 214: It looks like the formatting might be off for this last command, but that might only be in the git diff.

unstructured/staging/base.py
```diff
@@ -78,6 +97,28 @@ def elements_from_json(
...
+    element_dicts = elements_to_dicts(precision_adjusted_elements)
+    # -- serialize the dicts to JSON (bytes) --
+    json_bytes = json.dumps(element_dicts, sort_keys=True).encode("utf-8")
+    # -- compress the JSON bytes with gzip compression --
+    deflated_bytes = zlib.compress(json_bytes)
+    # -- base64-encode those bytes so they can be serialized as a JSON string value --
+    b64_deflated_bytes = base64.b64encode(deflated_bytes)
+    # -- convert to a string suitable for serializing in JSON --
+    return b64_deflated_bytes.decode("utf-8")
+
```
Coniferish, line 120: Just a question, but why is this function in staging.base and not utils or somewhere else? I haven't done much in staging, so I'm just unfamiliar with it.

Coniferish (approved):
LGTM

unstructured/staging/base.py
scanny: This is hopefully temporary. All the serde functions really belong in `unstructured.documents.elements` in my view, beside the definitions for elements. The serde function cohere strongly with the element definitions because that is what is being serialized and deserialized and they all depend ultimately on the `.to_dict()` and `.from_dict()` methods of elements and metadata.

I suppose it made sense at the time to put them in staging because that was where they were used, but typically serde coheres with the thing being serialized, not with what uses that serde (because that can be any number of places).

I initially tried putting these two new functions in `.documents.elements` and leaving the rest in `.staging.base` but that produced a nasty circular dependency which was most easily resolved by putting the new functions with the others. Note the local imports needed to resolve the remaining circular dependency in `ElementMetadata.to_dict()` and `.from_dict()`.

Staging is deprecated, so it probably makes sense for us to clean that up in a follow-on PR pretty soon here. We can republish those functions from `.staging.base` so any calls by external packages continue to work.

docs/source/core/chunking.rst
scanny: Hmm, probably tabs vs. spaces. I'll have a look :)

docs/source/core/chunking.rst
```diff
@@ -170,3 +170,45 @@ following behaviors:
   ``combine_text_under_n_chars`` argument. This defaults to the same value as ``max_characters``
   such that sequential small sections are combined to maximally fill the chunking window. Setting
   this to ``0`` will disable section combining.
+
+
+Recovering Chunk Elements
+-------------------------
+
+In general, a chunk consolidates multiple document elements to maximally fill a chunk of the desired
+size. Information is naturally lost in this consolidation, for example which element a passage of
```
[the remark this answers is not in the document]
scanny: Mmm, I'll have a noodle on that and rework that "passage" :)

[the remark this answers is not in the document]
scanny: k, I went with "portion" :)

docs/source/core/chunking.rst
scanny: Ah, okay, that last line is the _output_ of the command on the previous line(s), not a command itself. This format for the snippet with the ">>>" prefix represents the Python REPL where commands are entered after ">>> " and the output shows up on the next line (without the prefix).

PR #2687, opened 2024-03-22
feat: add --include-orig-elements option to Ingest CLI

scanny:
**Summary**
Add an `--include-orig-elements` option to the Ingest CLI to allow users to specify that corresponding new chunking parameter.

**Reviewer** A lot of this is cleanup, the second commit is where the actual adding of this option are. The first commit fixes a number of inaccuracies in the documentation and does some other clean-up.


cragwolfe:
Please update or create an ingest test to use this parameter (doesn't have to be many docs, of course).

Coniferish (changes requested):
LGTM once @cragwolfe's comment is addressed

scanny:
> Please update or create an ingest test to use this parameter (doesn't have to be many docs, of course).

@Crag yep, good call. Turns out there's only one test that does any chunking, so we're a bit under-covered in that area of functionality. While I was in there I noticed a bug in one of the other chunking parameters (no way to turn off multipage-sections in "by-title" chunking) and fixed it and covered it with the same new test while I was at it.

But chunking tests in ingest probably deserve some more attention going forward. I'll be back in here in a day or two to add chunk-by-similarity support and will give it some more attention then.

Coniferish (approved):
LGTM

PR #2700, opened 2024-03-28
feat: element types extension

pawel-kmiecik:
This PR adds some new element types that can be used especially by pdf/image parition.

### DIFF
--- a/unstructured/documents/elements.py
+++ b/unstructured/documents/elements.py
@@ -629,6 +629,7 @@ class ElementType:
     FOOTER = "Footer"
     FOOTNOTE = "Footnote"
     PAGE_FOOTER = "Page-footer"
+    OTHER = "Other"
 
     @classmethod
     def to_dict(cls):
@@ -770,6 +771,16 @@ class RadioButton(Checkable):
         """Serialize to JSON-compatible (str keys) dict."""
         out = super().to_dict()
         out["type"] = "RadioButton"
+        return out
+
+
+class Other(Element):
+    """An element that does not fit into any other category. Not expected to contain a text"""
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize to JSON-compatible (str keys) dict."""
+        out = super().to_dict()
+        out["type"] = "Other"
         return out
 
 
--- a/unstructured/documents/elements.py
+++ b/unstructured/documents/elements.py
@@ -600,11 +600,13 @@ class ElementType:
     ABSTRACT = "Abstract"
     THREADING = "Threading"
     FORM = "Form"
+    PARAGRAPH_IN_FORM = "ParagraphInForm"
     FIELD_NAME = "Field-Name"
     VALUE = "Value"
     LINK = "Link"
     COMPOSITE_ELEMENT = "CompositeElement"
     IMAGE = "Image"
+    PARAGRAPH_IN_IMAGE = "ParagraphInImage"
     PICTURE = "Picture"
     FIGURE_CAPTION = "FigureCaption"
     FIGURE = "Figure"
@@ -621,6 +623,7 @@ class ElementType:
     PAGE_BREAK = "PageBreak"
     FORMULA = "Formula"
     TABLE = "Table"
+    PARAGRAPH_IN_TABLE = "ParagraphInTable"
     HEADER = "Header"
     HEADLINE = "Headline"
     SUB_HEADLINE = "Subheadline"
@@ -629,6 +632,8 @@ class ElementType:
     FOOTER = "Footer"
     FOOTNOTE = "Footnote"
     PAGE_FOOTER = "Page-footer"
+    PAGE_NUMBER = "PageNumber"
+    CODE_SNIPPET = "CodeSnippet"
     OTHER = "Other"
 
     @classmethod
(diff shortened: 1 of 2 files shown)


badGarnet (approved):

scanny:
@pawel-kmiecik @badGarnet Can we discuss? I think it's going to have further-reaching implications to add elements that do not have a text attribute.

When we were doing some aspect of chunking we gave `CheckBox` a `.text` attribute to make it uniform with other elements. It always gets the text `""` but it does not raise `AttributeError` when `checkbox.text` is referenced.

Personally my initial inclination is to make all elements a subtype of `Text`, maybe moving `.text` up to the `Element` base-class, but that probably bears further thought.

In any case we should probably get together to discuss and also make sure we add tests that attempt to chunk element streams that include these new element types and ensure it produces the desired behavior.

scanny (changes requested):
Let's discuss.

badGarnet:
> Let's discuss.

We got two options:

- like @scanny suggested just add a default attribute `text` to all elements; this is not necessary a clean solution because it changes existing elements' definition and if some user already using the existence of `text` attribute to identify different types of attributes this will break that
- in the code where we interact with elements and tries to get `text` attribute we always use the `getattr` method with fallback empty string -> this doesn't change the existing element's definition but requires scouring through the code
- design wise I want to say less default is better: we clearly have element types that do not have text and making `text` default for elements would deviate from this principal

scanny:
@badGarnet I'm pretty sure all the existing element types have a `.text` attribute (because they subclass `Text`). The only one that didn't was `CheckBox` and we added that attribute during the chunking refactor a few months back.

scanny:
@badGarnet @pawel-kmiecik Hmm, my mistake. Actually we didn't add it to `CheckBox`, we added it to `Element`. That means the work is already done. So I think this means my initial concern was a false alarm. All subclasses of `Element` will have a `.text` attribute that defaults to `""` for instances not a subtype of `Text`.

Still probably worth having a test or two that ensures (and defends going forward) that chunking works as expected when the element-stream contains the new element types, but I expect it will.

The expected behavior has two cases:
1. The prior element is not oversized: an element with zero-length text will be added to the prior chunk (because it cannot increase the chunk's size beyond the size limit.
2. The prior element is oversized (by itself to large to fit in `max_characters`): The next element will start a new chunk.

I think there's an edge case for `new_after_n_chars` that's an exception to case 1 but that's kind of a finer point. Note these behaviors are well-enough tested already, I just mention them here so you know what to expect.

pawel-kmiecik:
@scanny I've added a test for serialization of non-text elements and thrown-in `RadioBox`es, `CheckBox`es and `Other`s  to chunking tests.
It looks they have no influence on the `CompositeElement`s test but they are visible in `orig_elements` metadata.

@badGarnet 
> design wise I want to say less default is better: we clearly have element types that do not have text and making text default for elements would deviate from this principal

Well, it looks the current design is that all elements have `text` attribute so I've moved `__str__` method from `Text` to `Element` - then all elements (including new ones) are serializable to `str`.

unstructured/documents/elements.py
```diff
@@ -614,6 +615,10 @@ class ElementType:
     LIST_ITEM_OTHER = "List-item"
     CHECKED = "Checked"
     UNCHECKED = "Unchecked"
+    CHECK_BOX_CHECKED = "CheckBoxChecked"
+    CHECK_BOX_UNCHECKED = "CheckBoxUnchecked"
+    RADIO_BUTTON_CHECKED = "RadioButtonChecked"
+    RADIO_BUTTON_UNCHECKED = "RadioButtonUnchecked"
```
badGarnet, line 621: we still need to add mapping of those to `CHECKED/UNCHECKED` below, right?

scanny (approved):
Okay, I think I'm grokking this now. I'm going to approve because I don't want to stand in the way of urgent progress, but I do have a couple remarks:

1. It looks to me like `documents.elements.ElementType` is problematic. I don't think we know what it is or its semantics and it is being used in conflicting ways. I think it's worth articulating our best understanding of its semantics in a docstring in this PR so we at least understand the role it's being pressed into for present purposes.
2. The business of "normalizing" a collection of `LayoutElement` into `list[Elements]` does not appear to me to be the business of the `documents.elements` module. It seems to me it coheres much more tightly with inference. In particular, the `Element` object hierarchy is unaffected by normalization. The rules and mappings for it can safely be extracted to a place that is convenient for `normalize_layout_element()` and any other logic that does normalization. Removing any concern for interacting with `LayoutElement` would be an improvement to the `documents.elements` module in my view.

badGarnet (changes requested):
blocking pending resolution of design discussion

badGarnet (approved):

unstructured/documents/elements.py
pawel-kmiecik: CheckBox subclasses from `Element`, not `Text` - so these have a dedicated "if branch" here:
https://github.com/Unstructured-IO/unstructured/pull/2700/files#diff-9eb4b1248793302fa3b6bc9db25756503ea28f53ca556417c27f0be85d7a709bR164

We have a `checked` attribute in `CheckBox` even if its base class is moved to `Text`, we cannot cover it with standard `Text` elements mapping.

PR #2702, opened 2024-03-29
Add HTMLImage class

heya5:


cragwolfe:
@heya5 , can you describe what you had in mind after this element class is added?

heya5:
I would like to identify image tag and keep the image source url in `page.element`. 
And then I can deal this kind of element to a markdown format with image.

PR #2722, opened 2024-03-29
Update Connectors README.md to fix broken hyperlink

CodesLikeIcarus:
Fixed broken hyperlink to destination connector example 'delta_table.py' vs BROKEN: 'delta-table.py'


CodesLikeIcarus:
Canceling this -- the issue is more widespread -- many links are broken presumably from a directory refactor

PR #2842, opened 2024-04-03
Preparing the foundation for better element IDs

micmarty-deepsense:
Part one of the issue described here: https://github.com/Unstructured-IO/unstructured/issues/2461

It does not change how hashing algorithm works, just reworks how ids are assigned:
> Element ID Design Principles
> 
> 1. A partitioning function can assign only one of two available ID types to a returned element: a hash or UUID.
> 2. All elements that are returned come with an ID, which is never None.
> 3. No matter which type of ID is used, it will always be in string format.
> 4. Partitioning a document returns elements with hashes as their default IDs.

Big thanks to @scanny for explaining the current design and suggesting ways to do it right, especially with chunking.

Here's the next PR in line: https://github.com/Unstructured-IO/unstructured/pull/2673


test_unstructured/documents/test_elements.py
```diff
@@ -383,9 +374,14 @@ def and_it_serializes_a_data_source_sub_object_to_a_dict_when_it_is_present(self
         }
 
     def and_it_serializes_an_orig_elements_sub_object_to_base64_when_it_is_present(self):
+        elements = [Title("Lorem"), Text("Lorem Ipsum")]
+
+        for element in elements:
+            element.id_to_hash()
```
micmarty-deepsense, line 380: I think it's better than doing this:
```
elements = assign_hash_ids(elements)
```
which has more logic underneath, and would nee to be imported

unstructured/documents/email_elements.py
```diff
@@ -24,20 +24,13 @@ def __init__(
         self,
         name: str,
         text: str,
-        datestamp: Union[datetime, NoDatestamp] = NoDatestamp(),
-        element_id: Union[str, uuid.UUID, NoID, UUID] = NoID(),
+        datestamp: datetime | NoDatestamp = NoDatestamp(),
+        element_id: str | NoID = NoID(),
     ):
         self.name: str = name
         self.text: str = text
```
scanny, line 31: I think this line goes away because `Text` takes care of adding this attribute (self.text) and setting its value when you call `super()` below.

test_unstructured/documents/test_elements.py
scanny: Yeah, this one is borderline I think. We should definitely not use `documents.elements.assign_hash_ids()` because this is not its intended use and is not under test here.

But it might not hurt to add a test-utility function to `test_unstructured.unit_utils` to do this. Unfortunately the name that really makes sense to me for that is `assign_hash_ids()`, which could be a little confusing, but still, I think that's probably worthwhile because this same code passage appears a dozen times in this PR and maybe more in future. Then instead of three new lines each time it would just be:
```python
elements = assign_hash_ids([Title("Lorem"), Text("Lorem Ipsum")]
```

I think that's worth doing. We can adjust the name if we really don't like it. `using_hash_ids()` could be a good name, or maybe `with_hash_ids()`.

Or we might prefer to adjust the name of the original to include the fact that it's mapping metadata ids. Maybe something like `assign_and_map_hash_ids()`.

scanny (approved):
LGTM. A few comments to attend to before merging but approving in advance.

Not sure if Crag wants to take another look before merging, probably worth asking.

test_unstructured/documents/test_elements.py
```diff
@@ -32,31 +31,23 @@
 )
 
 
-def test_text_id():
+def test_text_hash_id():
     text_element = Text(text="hello there!")
-    assert text_element.id == "c69509590d81db2f37f9d75480c8efed"
+    assert text_element.id_to_hash() == "c69509590d81db2f37f9d75480c8efed"
```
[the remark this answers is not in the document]
micmarty-deepsense: done

test_unstructured/documents/test_elements.py
```diff
@@ -32,31 +31,23 @@
...
-    assert isinstance(element.id, UUID)
+@pytest.mark.parametrize(
+    "element",
+    [
+        Element(),  # should default to UUID
+        Text(text="hello there!"),  # should default to UUID
+        Text(text="hello there!", element_id=NoID()),
+    ],
+)
+def test_text_uuid(element: Element):
```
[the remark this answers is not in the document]
micmarty-deepsense: good point, I'll try to not forget about that in the future. 
test name updated - done

unstructured/documents/elements.py
```diff
@@ -651,19 +675,35 @@ def to_dict(cls):
...
+    5. Assigning a string ID manually is possible, but is meant to be used
+        only for deserialization purposes.
+    """
 
     text: str
 
     def __init__(
         self,
-        element_id: str | uuid.UUID | NoID | UUID = NoID(),
+        element_id: str | NoID = NoID(),
```
qued, line 694: If the purpose of the `NoID` class is to be a placeholder indicating we want the `Element` to self-assign an id, do we need the class at all? Could we just use `None` instead (i.e. `element_id` could have type `Optional[str]`)? Or do we need `NoID` for backward compatibility with code out in the wild (in which case maybe we could deprecate)?

scanny (changes requested):
Don't merge yet, we're rethinking our approach on assigning element-ids and in particular the idea of having `element.metadata.parent` as an object reference when the element is in-memory and `element.metadata.parent_id` only appearing in serialized forms (dict, JSON).

unstructured/documents/elements.py
scanny: I think that's an excellent question @qued. I'm inclined to agree that we should get rid of `NoID` and fix up whatever internal calls we have in a partitioner that might use it.

It doesn't add any new information, at least not anymore. I think at one time you could "pre-select" the "`id_type`" by specifying either `UUID()` (not the stdlib `uuid.UUID`) or `NoID()`, but that's no longer the case and I think this should go away now and just use `None` as the default.

scanny (changes requested):
~Hi Mike, I'm a little confused. I thought this PR was to _prepare_ for better element-ids, but not to actually implement them yet. That the bit that changed the hash algorithm to include page_number and page_sequence_number was going to be separate (because that's what triggers all the ingest-test changes).~

~Did we change our mind on that?~

scanny (approved):
Hmm, okay, nevermind. I mistook something else for a change to the hash algorithm.

LGTM :)

PR #2843, opened 2024-04-03
bug: Add options to SFTP

potter-potter:
Noticed authentication errors when connected to a non localhost SFTP. It errored out when looking for ssh keys.

This gives us the option to not look for those. Which is correct if we are giving it user/password.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Fixes
 
 * **Fix `partition_html()` swallowing some paragraphs**. The `partition_html()` only considers elements with limited depth to avoid becoming the text representation of a giant div. This fix increases the limit value.
+* **Fix SFTP** Adds more options to prevent looking for ssh files if only using username password.
```
ahmetmeleq, line 10:
```suggestion
* **Fix SFTP** Adds flag options to SFTP connector on whether to use ssh keys / agent, with flag values defaulting to False. This is to prevent looking for ssh files when using username and password. Currently, username and password are required, making that always the case.
```

ahmetmeleq:
- Ingest test on sftp source runs on localhost, which should be why we weren't able to catch this
- Spinning / keeping a remote server alive just to catch similar issues in CI tests is possible, but not urgent
- Defaults for `--allow-agent` and `--look-for-keys` being `false` makes sense, given that `--username` and `--password` are required

LGTM

ahmetmeleq (approved):

PR #2851, opened 2024-04-04
fix: Add mongodb env variables to ingest-test-fixtures-update-pr.yaml

potter-potter:
ingest-test-fixtures-update-pr.yaml was missing mongodb vars. And the workflow was failing.


ryannikolaidis:
...and can we make sure in the future it would?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 * **Fix `partition_html()` swallowing some paragraphs**. The `partition_html()` only considers elements with limited depth to avoid becoming the text representation of a giant div. This fix increases the limit value.
 * **Fix SFTP** Adds flag options to SFTP connector on whether to use ssh keys / agent, with flag values defaulting to False. This is to prevent looking for ssh files when using username and password. Currently, username and password are required, making that always the case.
+* **Fix Add Mongodb Env Vars to ingest-test-fixtures-update-pr** Mongodb env vars were never added. This caused errors when trying to update test fixtures.
```
ryannikolaidis, line 15: non-code / CI bits don't need to bump the changelog

scanny (approved):
LGTM :)

CHANGELOG.md
potter-potter: @ryannikolaidis  I removed all changes to changelog and `__version__`. I hope thats what you meant.

ryannikolaidis: yep, thank you!

ryannikolaidis (approved):
nit: can we not bump the changelog since this doesn't impact the end user?

scanny:
> nit: can we not bump the changelog since this doesn't impact the end user?

@ryannikolaidis I believe the Changelog workflow will fail if there are no changes to CHANGELOG and if `__version__` does not correspond. So the minimal change to get a mergeable PR is to bump the changelog and update __version__ to match.

At least that's the way it was when I formed that habit months ago :)

potter-potter:
> > nit: can we not bump the changelog since this doesn't impact the end user?
> 
> @ryannikolaidis I believe the Changelog workflow will fail if there are no changes to CHANGELOG and if `__version__` does not correspond. So the minimal change to get a mergeable PR is to bump the changelog (version number) and update **version** to match.
> 
> At least that's the way it was when I formed that habit months ago :)
Thats what I thought. Looks like CI didn't like it without one.
 there is already a commit associated with version 0.13.1.

ryannikolaidis:
> @potter-potter you need to bump the version number in `CHANGELOG` and change the version in `unstructured/__version__.py` to match it. No need for an actual `CHANGELOG` "line-item" if it doesn't affect the user, but you do need to change the version.
> 
> In this case, since Ryan just cut a new release, you need to add a new version block named `0.13.2-dev0` with empty `Enhancements`, `Fixes`, etc. blocks underneath it.

Yep! That ☝️ exactly.

PR #2853, opened 2024-04-05
rfctr(pptx): extract _PptxPartitionerOptions

scanny:
**Reviewers:** Likely quicker to review commit-by-commit.

**Summary**

In preparation for adding a PPTX `Picture` shape _sub-partitioner_, extract management of PPTX partitioning-run options to a separate `_PptxPartitioningOptions` object similar to those used in chunking and XLSX partitioning. This provides several benefits:
- Extract code dealing with applying defaults and computing derived values from the main partitioning code, leaving it less cluttered and focused on the partitioning algorithm itself.
- Allow the options set to be passed to helper objects, prominently including sub-partitioners, without requiring a long list of parameters or requiring the caller to couple itself to the particular option values the helper object requires.
- Allow options behaviors to be thoroughly and efficiently tested in isolation.


Makefile
```diff
@@ -398,7 +398,7 @@ check-flake8-print:
 .PHONY: check-ruff
 check-ruff:
     # -- ruff options are determined by pyproject.toml --
-	ruff .
+	ruff check .
```
scanny, line 401: This fixes a deprecation warning in CI. `ruff` now also does reformatting, like Black, so there is `ruff check ...` and `ruff format ...`. We only want the lint part (check) for present purposes.

test_unstructured/partition/pptx/test_files/README.md
```diff
@@ -1,3 +0,0 @@
-PPTX files in this directory are made by hand using PowerPoint. This most faithfully represents source files likely to be used by users of this library. Files produced by `python-pptx` are "cleaner" in certain respects and may pass tests that a "real" PowerPoint file would not.
```
scanny, line 1: I added this directory back during my first couple weeks at Unstructured before I came to learn that test-files go in `/example-docs`. The one .pptx test file from here moved to `/example-docs` and this directory goes away (so this README it contained is no longer relevant).

test_unstructured/partition/pptx/test_pptx.py
```diff
@@ -89,12 +89,6 @@ def test_partition_pptx_from_file_with_metadata_filename():
         assert element.metadata.filename == "test"
 
 
-def test_partition_pptx_raises_with_both_specified():
```
scanny, line 96: This behavior is no longer implemented. If both filename and file are provided, filename is used.

unstructured/partition/pptx.py
```diff
@@ -53,11 +53,9 @@ def partition_pptx(
     file: Optional[IO[bytes]] = None,
     include_page_breaks: bool = True,
     metadata_filename: Optional[str] = None,
-    include_metadata: bool = True,
```
Coniferish, line 56: Do we need to create a follow-up issue to add this as a functioning parameter?

Coniferish (approved):
LGTM :)

unstructured/partition/pptx.py
scanny: No. I asked Matt about it and it is not a published (end-user) option. It has something to do with file-conversion he said so if we're not doing anything with it in a particular partitioner we can just let it get collected by `kwargs` like other options the particular partitioner is insensitive to.

Coniferish (approved):

PR #2861, opened 2024-04-05
feat: improve load_requirements in setup.py

rbiseck3:
### Description
This PR adds a number of enhancements around how we dynamically generate the requirements defined in the `setup.py` file. All logic around this was moved out into a `setup_utils.py` file to support unit tests around the logic:
 
* `load_requirements`: Current approach to dynamically generate the requirements was causing them to maintain the newline character at the end. This was removed. Also added support for the `-r` character in `*.in` files which recursively pull in other `.in` files which would currently be ignored. This isn't being used at the moment so it wasn't causing a problem, but this is future looking. It also used to support a list of filenames but that was never used, isolated to take in only a string filename of a Path object.
* New methods added to generate the dictionaries fo requirements, one for connector requirements and another for doc type requirements. 
* Safegaurds added throughout to raise an error if something incorrect was added. These include the file not being found or it not having the expected `.in` suffix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,36 +18,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from pathlib import Path
-from typing import List, Optional, Union
-
 from setuptools import find_packages, setup
 
+from setup_utils import load_requirements
 from unstructured.__version__ import __version__
-
-
-def load_requirements(file_list: Optional[Union[str, List[str]]] = None) -> List[str]:
-    if file_list is None:
-        file_list = ["requirements/base.in"]
-    if isinstance(file_list, str):
-        file_list = [file_list]
-    requirements: List[str] = []
-    for file in file_list:
-        path = Path(file)
-        file_dir = path.parent.resolve()
-        with open(file, encoding="utf-8") as f:
-            raw = f.read().splitlines()
-            requirements.extend([r for r in raw if not r.startswith("#") and not r.startswith("-")])
-            recursive_reqs = [r for r in raw if r.startswith("-r")]
-            if recursive_reqs:
-                filenames = []
-                for recursive_req in recursive_reqs:
-                    file_spec = recursive_req.split()[-1]
-                    file_path = Path(file_dir) / file_spec
-                    filenames.append(str(file_path.resolve()))
-                requirements.extend(load_requirements(file_list=filenames))
-    return requirements
-
 
 csv_reqs = load_requirements("requirements/extra-csv.in")
 doc_reqs = load_requirements("requirements/extra-docx.in")
--- a/setup.py
+++ b/setup.py
@@ -20,45 +20,8 @@ limitations under the License.
 
 from setuptools import find_packages, setup
 
-from setup_utils import load_requirements
+from setup_utils import get_base_reqs, get_extras
 from unstructured.__version__ import __version__
-
-csv_reqs = load_requirements("requirements/extra-csv.in")
-doc_reqs = load_requirements("requirements/extra-docx.in")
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -32,11 +33,19 @@ def load_requirements(file_list: Optional[Union[str, List[str]]] = None) -> List
...
+            raw = f.read().splitlines()
+            requirements.extend([r for r in raw if not r.startswith("#") and not r.startswith("-")])
+            recursive_reqs = [r for r in raw if r.startswith("-r")]
+            if recursive_reqs:
+                filenames = []
+                for recursive_req in recursive_reqs:
+                    file_spec = recursive_req.split()[-1]
+                    file_path = Path(file_dir) / file_spec
+                    filenames.append(str(file_path.resolve()))
+                requirements.extend(load_requirements(file_list=filenames))
```
qued, line 48: We don't need to keep the file open after we've read it, so this block can be moved out of the `with` clause.

setup.py
```diff
@@ -32,11 +33,19 @@ def load_requirements(file_list: Optional[Union[str, List[str]]] = None) -> List
...
+        file_dir = path.parent.resolve()
         with open(file, encoding="utf-8") as f:
-            requirements.extend(f.readlines())
-    requirements = [
-        req for req in requirements if not req.startswith("#") and not req.startswith("-")
-    ]
+            raw = f.read().splitlines()
+            requirements.extend([r for r in raw if not r.startswith("#") and not r.startswith("-")])
+            recursive_reqs = [r for r in raw if r.startswith("-r")]
+            if recursive_reqs:
```
qued: Small nit: the code iterates over `recursive_reqs`, so if `recursive_reqs` is empty all the code within this `if` block is a noop anyway. In such cases I think it's cleaner and more readable to not include the `if` statement.

rbiseck3: I actually did this for readability because I only instantiate the `filenames = []` if there are any recursive ones to look through.

rbiseck3: But I could just have it call it one at a time.

test_unstructured/test_setup_utils.py
```diff
@@ -0,0 +1,9 @@
+from setup_utils import load_requirements
+
+
+def test_load_requirements():
+    file = "./files/example.in"
```
qued, line 5:
```suggestion
    file = f"{os.path.dirname(__file__)}/files/example.in"
```
When dealing with tests, the working directory gets wonky. In many situations, it's the project base directory. So this test fails when, for example, I run `pytest test_unstructured/test_setup.py`. Maybe it passes when I `make test`... but I don't always want to run the whole suite. If you follow this suggestion you'd also need to `import os`.

setup_utils.py
```diff
@@ -0,0 +1,115 @@
...
+            raw = f.read().splitlines()
+            requirements.extend([r for r in raw if not r.startswith("#") and not r.startswith("-")])
+            recursive_reqs = [r for r in raw if r.startswith("-r")]
+            if recursive_reqs:
+                filenames = []
+                for recursive_req in recursive_reqs:
+                    file_spec = recursive_req.split()[-1]
+                    file_path = Path(file_dir) / file_spec
+                    filenames.append(str(file_path.resolve()))
+                requirements.extend(load_requirements(file_list=filenames))
```
qued: Just reiterating we don't want this section in the `with` clause since that keeps the file open unnecessarily.

rbiseck3: Yep, this was moved out

rbiseck3:
@qued, repo uploaded to test pypi here:  [unstructured/0.13.3.dev3/](https://test.pypi.org/project/unstructured/0.13.3.dev3/)

PR #2862, opened 2024-04-05
feat: Remove constraint on unstructured client from .in file

rbiseck3:
### Description
Don't limit the version of the unstructured client for all users of the repo


Coniferish (approved):
LGTM

PR #2879, opened 2024-04-11
feat(pptx): partition_pptx() accepts strategy arg

scanny:
**Summary**
As we move to adding pluggable sub-partitioners, `partition_pptx()` will need to become sensitive to the `strategy` argument, in particular when it is set to "hi_res". Up until now there were no expensive operations (inference, OCR, etc.) incurred while partitioning PPTX so this argument was ignored.

After this PR, `partition_pptx()` still won't do anything with that value, other than pass it along to `_PptxPartitionerOptions` for safe-keeping, but now its ready for use by a `PicturePartitioner` (to come in a subsequent PR).

### DIFF
--- a/unstructured/partition/pptx.py
+++ b/unstructured/partition/pptx.py
@@ -39,6 +39,7 @@ from unstructured.partition.text_type import (
     is_possible_narrative_text,
     is_possible_title,
 )
+from unstructured.partition.utils.constants import PartitionStrategy
 from unstructured.utils import lazyproperty
 
 DETECTION_ORIGIN = "pptx"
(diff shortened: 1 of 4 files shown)


unstructured/partition/pptx.py
```diff
@@ -49,15 +50,17 @@
...
     include_page_breaks: bool = True,
-    metadata_filename: Optional[str] = None,
-    metadata_last_modified: Optional[str] = None,
     include_slide_notes: bool = False,
     infer_table_structure: bool = True,
     languages: Optional[list[str]] = ["auto"],
-    detect_language_per_element: bool = False,
-    date_from_file_object: bool = False,
+    metadata_filename: Optional[str] = None,
+    metadata_last_modified: Optional[str] = None,
```
scanny, line 62: Adding the "*" just makes explicit that all arguments other than `filename` are keyword-only. This has always been the case because `filename` and `file` can't both be used, but explicit is better than implicit. The remaining line changes are just sorting the kwargs alphabetically for the sake of good order which makes it easier to find the one you're looking for and tell whether one is missing.

Coniferish (approved):
lgtm :)

PR #2880, opened 2024-04-11
feat(pptx): add pluggable PPTX Picture sub-partitioner

scanny:
**Summary**
Delegate partitioning of PPTX Picture (image, to a first approximation) shapes to a distinct sub-partitioner and allow the default picture sub-partitioner to be replaced at run-time by one of the user's choosing.


CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
-## 0.13.3-dev3
+## 0.13.3-dev4
 
 ### Enhancements
 
 * **Add `strategy` arg value to `_PptxPartitionerOptions`.** This makes this paritioning option available for sub-partitioners to come that may optionally use inference or other expensive operations to improve the partitioning.
+* **Support pluggable sub-partitioner for PPTX Picture shapes.** Use a distinct sub-partitioner for partitioning PPTX Picture (image) shapes and allow the default picture sub-partitioner to be replaced at run-time by one of the user's choosing.
```
Coniferish, line 6: Is this feature only intended for partitioning images that contain text? Wondering if we will eventually support users plugging in object detection models here so they can have text descriptions of images in a powerpoint.

unstructured/partition/pptx.py
```diff
@@ -307,6 +347,14 @@ def _shape_is_off_slide(self, shape: Shape) -> bool:
 class _PptxPartitionerOptions:
     """Encapsulates partitioning option validation, computation, and application of defaults."""
 
+    _PicturePartitionerCls = None
+    """Sub-partitioner used to partition PPTX Picture (Image) shapes.
+
+    This value has module lifetime and is updated by calling the `register_picture_partitioner()`
+    function defined at the module level to replace its value with a pluggable sub-partitioner
+    implementing the `AbstractPicturePartitioner` interface.
```
Coniferish, line 355: That's a long sentence. Can we break it up into... "This value has module lifetime and is updated by calling the `register_picture_partitioner()` function defined at the module level. `register_picture_partitioner()` replaces `_PicturePartitionerCls` with a pluggable sub-partitioner that implements the `AbstractPicturePartitioner` interface."?

Coniferish (approved):
Optional comment to address, but lgtm

CHANGELOG.md
scanny: This _specific_ feature supports any of those behaviors. Basically `AbstractPicturePartitioner` takes a `pptx.Picture` object and emits zero or more document elements: `iter_elements(picture: Picture) -> Iterator[Element]`. And, any sub-partitioner that implements this interface (signature) can be registered as the picture subpartitioner. So a subpartitioner that does OCR on the image can be created or one that one that does object detection, or one that does both or does something cool we haven't though of yet :)

unstructured/partition/pptx.py
scanny: Yep, good call. Done.

PR #2881, opened 2024-04-11
Fix: ingest test fixtures update pr

christinestraub:
This PR aims to update "Ingest Test Fixtures Update PR" CI to update the ingest test fixtures only if the OVERWRITE_FIXTURES variable is not `false` and the OUTPUT_DIR directory is not empty.


test_unstructured_ingest/check-diff-expected-output.sh
```diff
@@ -37,7 +37,7 @@ function cleanup() {
 trap cleanup EXIT
 
 # to update ingest test fixtures, run scripts/ingest-test-fixtures-update.sh on x86_64
-if [ "$OVERWRITE_FIXTURES" != "false" ]; then
+if [ "$OVERWRITE_FIXTURES" != "false" ] && [ "$(ls -A "$OUTPUT_DIR")" ]; then
```
cragwolfe, line 40: probably worth a comment here like:

```
# only overwrite fixtures if the OUTPUT_DIR directory is not empty (there are potential additions or updates)
```
?

cragwolfe (approved):

PR #2885, opened 2024-04-12
bug: fix pip-compile

rbiseck3:
### Description
Currently wasn't compiling `base.in` first, which is required because others use the generated `.txt` file as a constraint.


MthwRobinson (approved):
LGTM! Confirmed `make pip-compile` runs successfully for me now

PR #2893, opened 2024-04-15
rfctr: improve typing in OCR modules

scanny:
**Summary**
In preparation for using OCR for partitioners other than PDF, clean up typing in the OCR module.

### DIFF
--- a/unstructured/partition/utils/ocr_models/ocr_interface.py
+++ b/unstructured/partition/utils/ocr_models/ocr_interface.py
@@ -1,12 +1,14 @@
+from __future__ import annotations
+
 import functools
 import importlib
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, List
+from typing import TYPE_CHECKING
 
 from unstructured.partition.utils.constants import OCR_AGENT_MODULES_WHITELIST
 
 if TYPE_CHECKING:
-    from PIL import PILImage
+    from PIL import Image as PILImage
     from unstructured_inference.inference.elements import TextRegion
     from unstructured_inference.inference.layoutelement import (
         LayoutElement,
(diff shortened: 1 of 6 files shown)


unstructured/partition/utils/ocr_models/ocr_interface.py
```diff
@@ -1,44 +1,41 @@
...
     )
 
 
 class OCRAgent(ABC):
-    def __init__(self):
-        self.agent = self.load_agent()
-
-    @abstractmethod
-    def load_agent(self, language: str) -> Any:
-        pass
```
Coniferish, line 22: Why was the __init__ method (and the related abstractmethod) removed?

unstructured/partition/utils/ocr_models/tesseract_ocr.py
```diff
@@ -26,21 +28,17 @@
 
 
 class OCRAgentTesseract(OCRAgent):
-    def load_agent(self):
-        pass
```
Coniferish, line 30: marking since it's related to my other comment

Coniferish (approved):
Approving in advance

unstructured/partition/utils/ocr_models/ocr_interface.py
scanny: In a nutshell, because this class specifies the interface for an "OCRAgent" and neither of those methods is actually part of the interface.

Some background facts:
- The only call to `.load_agent()` is from the initializer (`.__init__()`).
- `.load_agent() is implemented in the Paddle OCR-agent.
- `.load_agent()` is _not_ implemented in the Tessaract OCR-agent.

From those three facts we can infer:
- `.load_agent()` is not part of the interface (because you can be an agent without implementing that method).
- `.load_agent()` is part of the implementation (not interface) because it is only called internally.

For this reason, I removed these from the `OCRAgent` ABC (interface definition) and moved them to the `Paddle` implementation.

My engineering motivations were:
- An interface should be as small as possible while remaining coherent and useful.
    - If I'm implementing an interface I want to know the minimum methods I need to implement.
    - If I'm using an interface I want to know what is guaranteed to be available to me.
- An interface specifies:
    - what can be called
    - what it is called with
    - what it returns

    Everything else should be left to the discretion of the implementor.

In fairness, an ABC can also act as a base class, and `.get_instance()` is an example of an ABC providing a common method, so these principles must be applied with some judgement. But in this case, it was clear `.load_agent()` was an implementation detail so I moved it to the implementation that used it.

unstructured/partition/utils/ocr_models/tesseract_ocr.py
scanny: This `.load_agent()` implementation, even though a no-op, needed to be here because `.load_agent()` was marked as an `@abstractmethod` in the ABC. So once that was gone we no longer needed to have an implementation here.

I vaguely suspect that the original intent was to allow OCR agents to be loaded on-demand since they're a little resource-intensive. But there's no documentation to that effect or any calls to this from outside the `Paddle` implementation, so we can easily add it back in later if the design evolves that way with the addition of some other OCR platform or a change in loading strategy. Until then, this is YAGNI and dead code.

PR #2901, opened 2024-04-17
rfctr(pptx): make PptxPartitionerOptions public

scanny:
**Summary**
A few additional small, mechanical odds and ends required for PPTX image extraction.

The big one is removing the leading underscore from `PptxPartitionerOptions` because now client code that implements a custom Picture-shape sub-partitioner will need to reference this class.


Coniferish (approved):
lgtm

Coniferish:
Tagging #2903, which this PR works towards address

PR #2905, opened 2024-04-18
feat: Enable remote chunking via unstructured-ingest

Coniferish:
Update: The cli shell script works when sending documents to the free api, but the paid api is down, so waiting to test against it.

- The first commit adds docstrings and fixes type hints.
- The second commit reorganizes `test_unstructured_ingest` so it matches the structure of `unstructured/ingest`.
- The third commit contains the primary changes for this PR.
  - The `.chunk()` method responsible for sending elements to the correct method is moved from `ChunkingConfig` to `Chunker` so that `ChunkingConfig` acts as a config object instead of containing implementation logic. `Chunker.chunk()` also now takes a json file instead of a list of elements. This is done to avoid redundant serialization if the file is to be sent to the api for chunking.


test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+class DescribeChunker:
+    """Unit tests for ingest.pipeline.reformat.chunking.Chunker"""
+
+    # -- Chunker.run() -----------------------------------------------------------------------------
+
+    # -- integration test --
+    def it_creates_json(self, _ingest_docs_map_: Mock):
+        chunking_config = ChunkingConfig(chunking_strategy="by_title")
+        pipeline_context = PipelineContext()
+        partition_config = PartitionConfig()
```
scanny, line 35: These are all short and can be inlined in the `Chunker` call immediately below. In general, the only reason you want to introduce a local variable for something like this is if you need access to the object later in the test.

Besides compactness, placing them inline avoids implying to the reader that these need to be accessed later. Like the time I spent discovering they aren't used anywhere else would be unnecessary.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+        chunker = Chunker(
+            chunking_config=chunking_config,
+            pipeline_context=pipeline_context,
+            partition_config=partition_config,
+        )
+        with tempfile.TemporaryDirectory() as tmpdir:
```
scanny, line 41: Hmm, no, I don't think this is how you want to do the temporary directory thing.

`tmpdir` is a "built-in" `pytest` fixture. So if you include `tmpdir` in the parameter list for `it_creates_json()` then `pytest` will do all the needful for you. The rest of this can be the same, you just ditch the `with tempfile.TemporaryDirectory()` bit. If you use that you need to do everything in the "with" context and there's no way to inspect the results afterward. Also it's more work.

If you look into it in the `pytest` documentation you'll see `pytest` does something like store the temp-dirs for the last five test runs. So it never gets out of control but you can always inspect the results if you want.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+            # -- `Chunker.chunk()` defaults to writing to "{work_dir}/chunked", which is located in
+            # -- "/.cache" of a user's profile.
+            # -- Define `work_dir` add the "/chunked" subdirectory to it:
+            chunker.pipeline_context.work_dir = tmpdir
+            os.makedirs(os.path.join(tmpdir, "chunked"), exist_ok=True)
```
scanny, line 46: Do you have to create the `chunked` directory? Doesn't ingest do that for itself?

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+        chunker = Chunker(
+            chunking_config=chunking_config,
+            pipeline_context=pipeline_context,
+            partition_config=partition_config,
+        )
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # -- `Chunker.chunk()` defaults to writing to "{work_dir}/chunked", which is located in
+            # -- "/.cache" of a user's profile.
+            # -- Define `work_dir` add the "/chunked" subdirectory to it:
+            chunker.pipeline_context.work_dir = tmpdir
```
scanny, line 45: If you use the `pytest` `tmpdir` fixture you can assign this in the constructor above.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+            os.makedirs(os.path.join(tmpdir, "chunked"), exist_ok=True)
+
+            filename = chunker.run(ELEMENTS_JSON_FILE)
+            head, tail = os.path.split(filename if filename else "")
```
scanny, line 49: Under what circumstances would `filename` be `None`? Seems like worth a separate assert, like:
```python
assert filename is not None, "message explaining what this means and why it's a failure"
```

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+            # -- Check that a json file was created in `/chunked` --
+            assert head.endswith("chunked")
+            assert tail.endswith(".json")
```
scanny, line 52: Seems weak. Why wouldn't you load the JSON file and assert something about its contents? It's a list of element, right? So you could do length == n and `all(e["type"]=="CompositeElement" for e in elements)`, that sort of thing.

Chunked elements are always one of `CompositeElement`, `Table`, or `TableChunk` (which I think still has JSON "type": "Table"), so that's a solid indicator that chunking happened.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def it_logs_error_with_invalid_remote_chunking_args(
+        self, _ingest_docs_map_: Mock, caplog: LogCaptureFixture
+    ):
+        chunking_config = ChunkingConfig(chunking_strategy="by_invalid")
+        pipeline_context = PipelineContext()
+        partition_config = PartitionConfig(partition_by_api=True)
```
scanny, line 59: Same inlining thing.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+            chunker.pipeline_context.work_dir = tmpdir
+            os.makedirs(os.path.join(tmpdir, "chunked"), exist_ok=True)
+
+            filename = chunker.run(ELEMENTS_JSON_FILE)
+            head, tail = os.path.split(filename if filename else "")
+            # -- Check that a json file was created in `/chunked` --
+            assert head.endswith("chunked")
+            assert tail.endswith(".json")
+
+    def it_logs_error_with_invalid_remote_chunking_args(
```
scanny, line 54: The name says invalid remote chunking args but the test only tests invalid chunking strategy. Those two should agree. I'm thinking change the name.

It also makes me wonder how it knows the chunking strategy is invalid but I'll read on.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        chunker.run(ELEMENTS_JSON_FILE)
+        assert "Input should be 'basic', 'by_page', 'by_similarity'" in caplog.text
+
+    def it_warns_with_nonlocal_chunking_strategy_and_partition_by_api_False(
+        self, _ingest_docs_map_: Mock, caplog: LogCaptureFixture
+    ):
+        chunking_config = ChunkingConfig(chunking_strategy="by_similarity")
+        pipeline_context = PipelineContext()
+        partition_config = PartitionConfig(partition_by_api=True)
```
scanny, line 73: inline

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+        chunking_config = ChunkingConfig(chunking_strategy="by_similarity")
+        pipeline_context = PipelineContext()
+        partition_config = PartitionConfig(partition_by_api=True, api_key="aaaaaaaaaaaaaaaaaaaaa")
+        chunker = Chunker(
+            chunking_config=chunking_config,
+            pipeline_context=pipeline_context,
+            partition_config=partition_config,
+        )
+
+        chunker.chunk(ELEMENTS_JSON_FILE)
```
scanny, line 120: All tests have three parts, setting up the test-fixture, running the unit under test, and then verifying the results. There are different names for it around like Arrange-Act-Assert and others you can find around.

I find it very helpful to the reader to separate the three sections with a blank line (and keep the sections together with no blank lines but sometimes comment lines.

The first question when reading a test or trying to find the test you're interested in is always: What is the unit under test? (unit being function or method generally). Making the line that exercises the unit stand out by being surrounded by a blank line above and below eases that search.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,144 @@
...
+            api_key="aaaaaaaaaaaaaaaaaaaaa",
+            api_url="https://api.unstructured.io/general/v0/general",
+            chunking_strategy="by_similarity",
+            combine_under_n_chars=None,
+            max_characters=None,
+            multipage_sections=None,
+            new_after_n_chars=None,
+            overlap=None,
+            overlap_all=None,
+        )
```
scanny, line 132: Nice.

test_unstructured_ingest/unit/test_interfaces.py
```diff
@@ -255,3 +258,24 @@ def test_process_file_flatten_metadata(mocker, partition_test_results):
...
     for elem in isd_elems:
         assert expected_keys == set(elem.keys())
+
+
+class DescribeChunkingConfig:
+    """Unit tests for unstructured.ingest.interfaces.ChunkingConfig"""
+
+    def it_accepts_chunking_strategy_by_itself(self):
+        config = ChunkingConfig(chunking_strategy="basic")
+        assert config.chunking_strategy == "basic"
```
scanny, line 268: Btw, I usually make an exception to the "blank-line-above-and-below" convention for short ones line this where the Act and Assert steps are combined. So I write them just as you have here, although I probably would have tried to fit this one all on one line and avoid the `config` local variable, but that's just me :)

unstructured/ingest/pipeline/reformat/chunking.py
```diff
@@ -62,3 +85,49 @@ def run(self, elements_json: str) -> Optional[str]:
...
+                new_after_n_chars=self.chunking_config.new_after_n_chars,
+                overlap=self.chunking_config.overlap,
+                overlap_all=self.chunking_config.overlap_all,
+            )
+        # -- Warn that the defined chunking_strategy is not locally available --
+        logger.warning(
+            "There is no locally available chunking_strategy:"
+            f" {self.chunking_config.chunking_strategy}."
+            "If trying to partition remotely, check that `partition_by_api`, `api_url`,"
+            "and `api_key` are correctly defined."
```
scanny, line 132: When using a multi-line string like this, make them all `f-` strings if any of them need to be.

Also, remember that continuation lines should start with the space that separates the segments. In this case there are no spaces between line 2 and 3 or 3 and 4.

Placing the spaces all at the front of lines 2+ makes it easy to see they're all there because they're all lined up in a column. Including the `f` prefix helps with that too.

unstructured/ingest/cli/cli.py
```diff
@@ -9,10 +9,15 @@ def ingest():
 
 
 def get_cmd() -> click.Command:
+    """Construct and return a Click command object representing the main command for the CLI.
```
scanny, line 12: This as a separate commit is good but a separate PR would be better. There's a lot of noise in the `files changed` version otherwise.

Also, the commit message on this commit seems off. Something about AstraDB that doesn't seem to have to do with what you've done in this PR.

scanny (commented):
Okay, quite a few remarks.

I like what you did with the typing and docstrings but I wonder if you could pull that into a separate "prep" PR so the meaty changes are more focused.

I want to take another look in the morning but here are these initial comments in the meantime.

Btw, generally tests should appear in the same commit as the changes that make them pass. We should probably talk about this a little bit. It has to do with the test-driven approach, like first you write the test and then write the code to pass the test (and only the code it takes to pass the test). Let's chat about that next time we get together.

unstructured/ingest/cli/cli.py
Coniferish: I didn't notice the AstraDB message, but removed it now. Thanks.
When I rebased or something a while ago, it somehow pulled in info from the last commit on main. That's why the first commit includes Ahmet. I don't know what I did exactly, though.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
Coniferish: No, without creating it `FileNotFoundError: [Errno 2] No such file or directory` is raised

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,147 @@
...
+        # -- "/.cache" of a user's profile.
+        # -- Define `work_dir` add the "/chunked" subdirectory to it:
+        os.makedirs(os.path.join(tmpdir, "chunked"), exist_ok=True)
+
+        filename = chunker.run(ELEMENTS_JSON_FILE)
+
+        head, tail = os.path.split(filename if filename else "")
+        # -- Check that a json file was created in `/chunked` --
+        assert head.endswith("chunked")
+        assert tail.endswith(".json")
```
scanny, line 48: Hmm, this seems like a weak test. I'd be inclined to open that file, parse the contents (`elements = json.load(...)`), and then make some assertions about the contents. A couple go-to facts about chunking that could make more convincing assertions:

- Chunk count should be consistent, like `len(chunks) == 42`.
- A chunk can only ever be a `CompositeElement`, a `Table`, or a `TableChunk`. `CompositeElement` and `TableChunk` never show up in partitioner output, so something like:
    ```python
    assert all(e["type"] in ("CompositeElement, ...) for e in elements)
    ```
- Maybe a spot check on text like `assert elements[4].text.startswith("Demographic: Base file ...")
- The assertions on the filename are fine, those can stay, just not really enough by themselves.

Then I would name this: `it_chunks_JSON_elements()` and you will have demonstrated that convincingly.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,147 @@
...
+    ):
+        chunker = Chunker(
+            chunking_config=ChunkingConfig(),
+            pipeline_context=PipelineContext(),
+            partition_config=PartitionConfig(),
+        )
+        caplog.set_level(logging.INFO)
+
+        assert chunker.run(ELEMENTS_JSON_FILE) is None
+        assert "chunking_strategy is None, skipping chunking for" in caplog.text
```
scanny, line 61: Nice.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
scanny: How about "it_logs_error_on_invalid_remote_chunking_strategy()`.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
```diff
@@ -0,0 +1,147 @@
...
+    ):
+        chunker = Chunker(
+            chunking_config=ChunkingConfig(chunking_strategy="by_invalid"),
+            pipeline_context=PipelineContext(),
+            partition_config=PartitionConfig(partition_by_api=True),
+        )
+
+        chunker.run(ELEMENTS_JSON_FILE)
+
+        assert "Input should be 'basic', 'by_page', 'by_similarity'" in caplog.text
```
scanny, line 74: Why is this the output and where does it come from? Only the remote chunker can tell what chunking strategies it has registered. Seems like `'by_title'` should be in the list too.

And `Input should be ...` isn't very helpful as an error message goes. Is there more that was cut off? I'd be looking to see something saying "chunking_strategy" to give a hint where the problem is. No reason this string can't be as long as it needs to be to clearly represent the log entry.

unstructured/ingest/pipeline/reformat/chunking.py
```diff
@@ -62,3 +85,51 @@ def run(self, elements_json: str) -> Optional[str]:
 
     def get_path(self) -> Path:
         return (Path(self.pipeline_context.work_dir) / "chunked").resolve()
+
+    def chunk(self, elements_json_file: str) -> Optional[list[Element]]:
+        """Called by Chunker.run() to properly execute the defined chunking_strategy."""
+        # -- No chunking_strategy means no chunking --
+        if self.chunking_config.chunking_strategy is None:
+            return
+        # -- Chunk locally --
```
scanny, line 94: Maybe:
```python
# -- Chunk locally for open-source chunking strategies, even when partitioning remotely --
```

unstructured/ingest/pipeline/reformat/chunking.py
```diff
@@ -62,3 +85,51 @@ def run(self, elements_json: str) -> Optional[str]:
...
+                # -- If api_key or api_url are None, partition_via_api will raise an error, which
+                # -- will be caught and logged by Chunker.run()
+                api_key=self.partition_config.api_key,  # type: ignore
+                api_url=self.partition_config.partition_endpoint,  # type: ignore
+                chunking_strategy=self.chunking_config.chunking_strategy,
+                combine_under_n_chars=self.chunking_config.combine_text_under_n_chars,
+                # -- These are not currently supported by the sdk and raise an error.
+                # -- They will need to be updated once we bump the client version.
+                # combine_text_under_n_chars=self.chunking_config.combine_text_under_n_chars,
+                # include_orig_elements=self.chunking_config.include_orig_elements,
```
scanny, line 120: What does being support by `unstructured-client` have to do with calling `partition_via_api()` via the `unstructured` library? `unstructured-client` doesn't even need to be installed, does it?

unstructured/ingest/pipeline/reformat/chunking.py
```diff
@@ -62,3 +85,51 @@ def run(self, elements_json: str) -> Optional[str]:
...
+                max_characters=self.chunking_config.max_characters,
+                multipage_sections=self.chunking_config.multipage_sections,
+                new_after_n_chars=self.chunking_config.new_after_n_chars,
+                # -- These are not supported by older versions of unstructured-client, which we have
+                # -- pinned to <=0.18.0
+                # overlap=self.chunking_config.overlap,
+                # overlap_all=self.chunking_config.overlap_all,
+            )
```
scanny, line 128: What is the behavior when there is no chunking strategy registered with the requested name on the remote host?

Like how does the user get a message like: `No such chunking strategy: 'foobar'`?

scanny (changes requested):
Ok, I think this is almost there. A few questions too.

Let me know when it's ready and I'll take a quick look and approve.

test_unstructured_ingest/unit/pipeline/reformat/test_chunking.py
Coniferish: `unstructured_client.models.errors.httpvalidationerror.HTTPValidationError` raises the error and `by_title` is in the list. I was trying to keep it to one line while still conveying what the logged message entailed.

The full output:
```
'ERROR    unstructured.ingest:chunking.py:83 failed to run chunking on file /Users/johnjennings/src/unstructured/example-docs/test_evaluate_files/unstructured_output/Bank Good Credit Loan.pptx.json, {"detail":[{"loc":["body","chunking_strategy"],"msg":"Input should be \'basic\', \'by_page\', \'by_similarity\' or \'by_title\'","type":"literal_error"}]}
  File "/Users/johnjennings/src/unstructured/unstructured/ingest/pipeline/reformat/chunking.py", line 64, in run
   chunked_elements = self.chunk(elements_json)
  File "/Users/johnjennings/src/unstructured/unstructured/ingest/pipeline/reformat/chunking.py", line 109, in chunk
    return partition_via_api(
  File "/Users/johnjennings/src/unstructured/unstructured/partition/api.py", line 103, in partition_via_api\n    response = sdk.general.partition(req)
  File "/Users/johnjennings/.pyenv/versions/unstructured/lib/python3.10/site-packages/unstructured_client/utils/_human_utils.py", line 86, in wrapper
    return func(*args, **kwargs)
  File "/Users/johnjennings/.pyenv/versions/unstructured/lib/python3.10/site-packages/unstructured_client/utils/_human_split_pdf.py", line 40, in wrapper
    return func(*args, **kwargs)
  File "/Users/johnjennings/.pyenv/versions/unstructured/lib/python3.10/site-packages/unstructured_client/general.py", line 90, in partition
    raise out
unstructured_client.models.errors.httpvalidationerror.HTTPValidationError: {"detail":[{"loc":["body","chunking_strategy"],"msg":"Input should be \'basic\', \'by_page\', \'by_similarity\' or \'by_title\'","type":"literal_error"}]}'
```

unstructured/ingest/pipeline/reformat/chunking.py
Coniferish: `unstructured_via_api()` uses `unstructured-client` to call the api.

Coniferish: I opened this as a follow-up issue:
https://github.com/Unstructured-IO/unstructured/issues/2924

PR #2927, opened 2024-04-24
Extend write_config with additional metadata chroma.py

0xjgv:
Allow users to set additional metadata values to expand on metadata filtering capabilities


MthwRobinson:
Closing since this is related to connectors. Please reopen in https://github.com/Unstructured-IO/unstructured-ingest if needed.

PR #2928, opened 2024-04-24
feat: Astra DB Source Connector Support

erichare:
This pull requests adds support for Astra DB as a source connector, pulling in data from Astra DB for use in Unstructured and applying appropriate metadata.

### DIFF
--- a/test_unstructured_ingest/src/astra.sh
+++ b/test_unstructured_ingest/src/astra.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+set -e
+
+SRC_PATH=$(dirname "$(realpath "$0")")
+SCRIPT_DIR=$(dirname "$SRC_PATH")
+cd "$SCRIPT_DIR"/.. || exit 1
+OUTPUT_FOLDER_NAME=astra-src
+OUTPUT_ROOT=${OUTPUT_ROOT:-$SCRIPT_DIR}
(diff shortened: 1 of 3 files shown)


potter-potter:
@erichare I'm taking a look at this.

Can you give me a description on why someone would use Astra DB as a source connector?  We don't have any other examples of a vector database as a source. Unless this is acting as a non-vector db?

erichare:
> @erichare I'm taking a look at this.
> 
> Can you give me a description on why someone would use Astra DB as a source connector? We don't have any other examples of a vector database as a source. Unless this is acting as a non-vector db? Just want to know the use case.

@potter-potter You are spot on, the use case is essentially that Astra DB can act as a non-vector DB, and in fact up until quite recently was solely a non-vector DB. There are customers that do have standard tabular databases that are trying to migrate to vector, and so the source connector would aid in that migration. I think (correct me if i'm wrong) that it's kinda similar to Mongo being a source connector in that sense, and the description of [Mongo's source connector](https://unstructured-io.github.io/unstructured/ingest/source_connectors/mongodb.html) is pretty much the same as we were thinking for Astra.

I actually had the same question at first and had to clarify as well haha. Let me know if that makes sense or there's any other clarification needed.

potter-potter:
@erichare Cool. Thanks for the explanation. I will combine this in a PR later today with the Destination updates.

potter-potter:
@erichare Let's work on this here:
https://github.com/Unstructured-IO/unstructured/pull/2964

PR #2935, opened 2024-04-26
fix: reqs arm64 friendly again. release 0.13.4

cragwolfe:
Cut a release.

Run pip-compile on mac to avoid `nvidia-*` requirements creeping into `requirements/extra-pdf-image.txt`. This should fix arm64 image builds that have been breaking on main.


scanny:
@cragwolfe what causes the `nvidia` requirements to appear? Is that running on Intel, or Linux maybe?

cragwolfe:
> @cragwolfe what causes the `nvidia` requirements to appear? Is that running on Intel, or Linux maybe?

yes, someone ran `pip-compile` on intel locally earlier

Issue #2931, opened 2024-04-25
bug/KeyError with PDF partition fast strategy element ID — in old_to_new_mapping[parent_id]


adieuadieu:
**Describe the bug**
When partitioning [this](https://github.com/Unstructured-IO/unstructured/files/15109052/a1977-backus.pdf) PDF document with the `fast` strategy, the following `KeyError` occurs:

```
{
	"name": "KeyError",
	"message": "'782eec119b3409ea1a0bc8abf8f059ac'",
	"stack": "---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[32], line 6
      1 from unstructured.partition.auto import partition
      4 file = \"/Users/marco/Downloads/a1977-backus.pdf\"
----> 6 elements = partition(
      7     filename=file,
      8     strategy=\"fast\"
      9 )

File ~/miniconda3/envs/ai-tasks/lib/python3.11/site-packages/unstructured/partition/auto.py:423, in partition(filename, content_type, file, file_filename, url, include_page_breaks, strategy, encoding, paragraph_grouper, headers, skip_infer_table_types, ssl_verify, ocr_languages, languages, detect_language_per_element, pdf_infer_table_structure, extract_images_in_pdf, extract_image_block_types, extract_image_block_output_dir, extract_image_block_to_payload, xml_keep_tags, data_source_metadata, metadata_filename, request_timeout, hi_res_model_name, model_name, date_from_file_object, starting_page_number, **kwargs)
    421 elif filetype == FileType.PDF:
    422     _partition_pdf = _get_partition_with_extras(\"pdf\")
--> 423     elements = _partition_pdf(
    424         filename=filename,  # type: ignore
    425         file=file,  # type: ignore
    426         url=None,
    427         include_page_breaks=include_page_breaks,
    428         infer_table_structure=infer_table_structure,
    429         strategy=strategy,
    430         languages=languages,
    431         hi_res_model_name=hi_res_model_name or model_name,
    432         extract_images_in_pdf=extract_images_in_pdf,
    433         extract_image_block_types=extract_image_block_types,
    434         extract_image_block_output_dir=extract_image_block_output_dir,
    435         extract_image_block_to_payload=extract_image_block_to_payload,
    436         starting_page_number=starting_page_number,
    437         **kwargs,
    438     )
    439 elif filetype in IMAGE_FILETYPES:
    440     elements = partition_image(
    441         filename=filename,  # type: ignore
[...]

PR #2940, opened 2024-04-26
Quickfix for elements sharing the same memory address

micmarty-deepsense:
This PR attempts to fix a memory issue, which resulted in errors like this: https://github.com/Unstructured-IO/unstructured/issues/2931
The root cause seems to be in how ListItems are being combined, not in how hashes or parent IDs are updated.

When `assign_and_map_hash_ids()` is called and elements (or elements' metadata) do not have unique memory addresses, then updating the parent_id of one element will also overwrite the parent_id of some other element.


Coniferish:
I checked out to your branch and verified this does not raise the KeyError, but can you add a test? I suppose it could be a follow-up PR. Approving in advance

Coniferish (approved):

cragwolfe:
https://github.com/Unstructured-IO/unstructured/actions/runs/8850164463/job/24304063524
```
=========================== short test summary info ============================
FAILED test_unstructured/partition/pdf_image/test_pdf.py::test_partition_pdf_word_bbox_not_char - assert 18 == 17
 +  where 18 = len([<unstructured.documents.elements.Header object at 0x7ff8e4d241d0>, <unstructured.documents.elements.Header object at ...NarrativeText object at 0x7ff8e4d0eb10>, <unstructured.documents.elements.NarrativeText object at 0x7ff8f8a89ad0>, ...])
= 1 failed, 2231 passed, 13 skipped, 2 deselected, 2 xfailed, 9 xpassed, 8 warnings in 1170.99s (0:19:30) =
```
. either this test was incorrect before, or the `def _combine_list_elements` change broke something.
let's make this a real release (not `-dev`) when tests pass.

adieuadieu:
Amazing. Thank you for the quick turnaround on this fix @micmarty-deepsense @cragwolfe! 🙇 🙏
I can confirm that unstructured 0.13.5 fixes the issue in #2931.

PR #2942, opened 2024-04-26
chore CORE-4775: remove html page number metadata field

yuming-long:
### Summary

Rip off page_number metadata fields until we have page counting for all kinds of html files (not just limited to news articles with multiple `<article>` tag)

### Test
Unit tests `test_add_chunking_strategy_on_partition_html_respects_multipage` and `test_add_chunking_strategy_title_on_partition_auto_respects_multipage` removed since they relay on the `page_number` fields from the SEC html file - now test moved to mock test for chunk_by_title -> revisit those tests when we find test file for this

Also changed the element ids from partition outputs for html files - element id change due to page number change (in element id hashing) -> todo ticket: update other deterministic element id tests per crag's comment

### DIFF
--- a/test_unstructured/partition/test_html_partition.py
+++ b/test_unstructured/partition/test_html_partition.py
@@ -733,12 +733,12 @@ def test_all_element_ids_are_unique():
 def test_element_ids_are_deterministic():
     ids = [e.id for e in partition_html("example-docs/fake-html-with-duplicate-elements.html")]
     assert ids == [
-        "cba9e551ed975e0f8a1956095894e92a",
-        "f540ea3b6569aafeb433df6616e79971",
-        "f4a34ee0fac26589fffdb53d0dfedbaf",
-        "15168aeddbd19da60791109a5a45af65",
-        "0c027f66120dd96271489dd0bb69bff5",
-        "abe89090c2e46dda8fff81053cc79f17",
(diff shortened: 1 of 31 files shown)


test_unstructured/partition/test_html_partition.py
```diff
@@ -733,12 +733,12 @@ def test_all_element_ids_are_unique():
...
-        "f4a34ee0fac26589fffdb53d0dfedbaf",
-        "15168aeddbd19da60791109a5a45af65",
-        "0c027f66120dd96271489dd0bb69bff5",
-        "abe89090c2e46dda8fff81053cc79f17",
+        "5899179e882d799d869a1d98fe7c7e77",
+        "88e47a42516af650afdcedfe098c3e6c",
+        "884be260a86bbdf265c248d5fff5ea00",
+        "0a23b3ae6bd812b3d90e47fec1df9fe0",
+        "1e9e5be33c99f7bbf2e569b2430e16cf",
+        "333e32df62a0ec81a8df07d52dd73c99",
```
cragwolfe, line 741: it's a bit of a chore to have to update these.
i realized after we merged this pattern we could just call ` partition_html` twice and make sure element_id's are the same. that should be a separate PR wherever that pattern is followed, though.

cragwolfe (approved):

Issue #709, opened 2023-06-09
bug/_convert_table_to_text index out of range


igoforth:
**Describe the bug**
A list index out of range occurs in _convert_table_to_text during docx parsing.

**To Reproduce**
I was operating on 1360 docx files from this source: https://www.3gpp.org/ftp/Specs/latest/Rel-17
In the case of doc's, I first converted to docx using the below command:

`C:\"Program Files"\LibreOffice\program\soffice.exe --headless --convert-to docx --outdir out in\<filename>`

**Expected behavior**
_convert_table_to_text to correctly convert all docx tables

**Screenshots**
![Capture](https://github.com/Unstructured-IO/unstructured/assets/20114381/055d1137-f147-4d05-971f-0427ec3b5daa)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Python version Python 3.10.10

**Additional context**
21101-h10.docx
21202-h00.docx
21205-h10.docx
21905-h10.docx
22003-h00.docx
22004-h00.docx
22011-h60.docx
22022-h00.docx
22030-h00.docx
22031-h00.docx
22032-h00.docx
22041-h00.docx
22042-h00.docx
22057-h00.docx
22071-h00.docx
22072-h00.docx
22081-h00.docx
22084-h00.docx
22087-h00.docx
22090-h00.docx
22094-h00.docx
22096-h00.docx
22097-h00.docx
22101-h50.docx
22104-h70.docx
22115-h10.docx
22119-h00.docx
22125-h60.docx
22135-h00.docx
22142-h00.docx
22146-h00.docx
22182-h00.docx
22185-h00.docx
22186-h00.docx
22220-h00.docx
22226-h00.docx
22242-h00.docx
22246-h00.docx
22259-h00.docx
22261-hb0.docx
22263-h40.docx
22268-h00.docx
22278-h20.docx
22279-h00.docx
22282-h00.docx
22346-h00.docx
22368-h00.docx
22468-h01.docx
22519-h00.docx
22826-h20.docx
22829-h10.docx
22832-h40.docx
22836-h10.docx
22866-h10.docx
22873-020.docx
22881-020_cl.docx
22889-h40.docx
22912-h00.docx
22936-h00.docx
22944-h00.docx
22948-h00.docx
22967-h00.docx
22973-h00.docx
22978-h00.docx
22979-h00.docx
22986-h00.docx
22987-h00.docx
23035-h00.docx
23041-h40.docx
23172-h00.docx
[...]

PR #2943, opened 2024-04-27
fix(docx): fix short-row DOCX table

scanny:
**Summary**
The DOCX format allows a table row to start late and/or end early, meaning cells at the beginning or end of a row can be omitted. While there are legitimate uses for this capability, using it in practice is relatively rare. However, it can happen unintentionally when adjusting cell borders with the mouse. Accommodate this case and generate accurate `.text` and `.metadata.text_as_html` for these tables.


scanny:
Fixes #709.

cragwolfe (approved):

PR #2947, opened 2024-04-29
feat: upload all fields to qdrant

ahmetmeleq:
Since [Qdrant allows](https://qdrant.tech/documentation/concepts/payload/#payload) any content that can be represented as JSON as the "payload", upload all element fields to the destination. Previous approach included redundant parsing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.13.6-dev0
+
+### Enhancements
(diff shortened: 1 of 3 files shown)


MthwRobinson:
Closing since this is related to connectors. Please reopen in https://github.com/Unstructured-IO/unstructured-ingest if needed.

PR #2963, opened 2024-05-02
rfctr improve partitioner typing

scanny:
**Summary**
Remedy the persistent type errors when importing `unstructured`. Give the partitioner type annotations a general scrubbing while we're at it.

### DIFF
--- a/unstructured/partition/csv.py
+++ b/unstructured/partition/csv.py
@@ -1,6 +1,7 @@
+from __future__ import annotations
+
 import csv
-from tempfile import SpooledTemporaryFile
-from typing import IO, BinaryIO, List, Optional, Union, cast
+from typing import IO, Any, Optional, cast
 
 import pandas as pd
 from lxml.html.soupparser import fromstring as soupparser_fromstring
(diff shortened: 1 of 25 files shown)


unstructured/partition/csv.py
```diff
@@ -29,18 +30,18 @@
 @add_chunking_strategy
 def partition_csv(
     filename: Optional[str] = None,
-    file: Optional[Union[IO[bytes], SpooledTemporaryFile]] = None,
```
Coniferish, line 32: just curious, but why were you able to get rid of the SpooledTemporaryFile type?

unstructured/partition/odt.py
```diff
@@ -15,18 +14,18 @@
 @add_chunking_strategy
 def partition_odt(
     filename: Optional[str] = None,
-    file: Optional[BinaryIO] = None,
+    file: Optional[IO[bytes]] = None,
```
Coniferish, line 17: This is changed to also accept in-memory byte streams?

unstructured/partition/pdf.py
```diff
@@ -239,23 +225,23 @@ def partition_pdf(
 
 def partition_pdf_or_image(
     filename: str = "",
-    file: Optional[Union[bytes, BinaryIO, SpooledTemporaryFile]] = None,
+    file: Optional[bytes | IO[bytes]] = None,
```
Coniferish, line 228: Does `IO[bytes]` not fully encompass the `bytes` types?

Coniferish (approved):
minor questions, but lgtm

unstructured/partition/csv.py
scanny: `IO[bytes]` is implemented by`SpooledTemporaryFile`, i.e. `SpooledTemporaryFile` _is a_ `IO[bytes]`. So specifying `SpooledTemporaryFile` along with `IO[bytes]` would be redundant in the same way that `Fruit | Oranges` would be. Oranges are a fruit, so are already covered as a permissible type by `Fruit`.

unstructured/partition/pdf.py
scanny: `bytes` does not have a `seek()` or `read()` or `close()` method. `bytes` is just like `str` but individual bytes instead of "multi-byte" characters. So no, `IO[bytes]` doesn't encompass `bytes`. This is just a kind of weird historical signature because folks can receive a block of bytes from, say, an HTTP response (downloading a file) and hand it directly to this method. In that case `file` isn't really a "file", it is the "bytes-of-a-file". But the author chose to combine both those cases in a single parameter.

PR #2964, opened 2024-05-02
Potter/astradb updates

potter-potter:
Keeping this as DRAFT for now.


unstructured/ingest/connector/astra.py
```diff
@@ -111,7 +133,89 @@ def normalize_dict(self, element_dict: dict) -> dict:
...
+@dataclass
+class AstraIngestDoc(IngestDocCleanupMixin, BaseSingleIngestDoc):
+    connector_config: SimpleAstraConfig
+    meta: t.Dict[str, str] = field(default_factory=dict)
+    registry_name: str = "astra"
+
+    def update_source_metadata(self):
+        self.source_metadata = SourceMetadata(
+            exists=True,
+        )
```
potter-potter, line 148: @erichare Could you look deeper into this? Best example might be delta_table.py since it is also reading from a database.

unstructured/ingest/connector/astra.py
```diff
@@ -111,7 +133,89 @@ def normalize_dict(self, element_dict: dict) -> dict:
...
+
+        if hasattr(self_cp, "_astra_db_collection"):
+            setattr(self_cp, "_astra_db_collection", None)
+
+        return _asdict(self_cp, **kwargs)
+
+    @property
+    @requires_dependencies(["astrapy"], extras="astra")
+    def astra_db_collection(self) -> "AstraDBCollection":
+        if self._astra_db_collection is not None: ### FIX THIS IS NEVER NONE
```
potter-potter, line 175: @erichare could you look deeper into this. For some reason `self._astra_db_collection` is never None. It should be when it instantiates. Destination connector part instantiates this as None.

test_unstructured_ingest/src/astra.sh
```diff
@@ -0,0 +1,55 @@
...
+    cleanup_dir "$DOWNLOAD_DIR"
+  fi
+}
+trap cleanup EXIT
+
+RANDOM_SUFFIX=$((RANDOM % 100000 + 1))
+COLLECTION_NAME="astra_test_output_$RANDOM_SUFFIX"
+EMBEDDING_DIMENSION=384
+
+RUN_SCRIPT=${RUN_SCRIPT:-./unstructured/ingest/main.py}
```
potter-potter, line 41: @erichare I made a non-vector collection in the unstructured database called `ingest_test_collection`. You might want to do the same to test this out.

erichare:
@potter-potter starting work on this now! thank you!

unstructured/ingest/connector/astra.py
erichare: @potter-potter this is a strange one for me too since as far as i can tell it's the same logic being used for the destination connector. I'm looking into it

unstructured/ingest/connector/astra.py
erichare: Got it. Yes, this is a great example - basically pulling the data into Pandas dataframes, writing them as CSVs, which presumably then become the input which gets converted into JSON for Unstructured.

I think we can follow the exact paradigm with minor changes, and i've done that (will push shortly)... I don't know the general protocol here but I would be happy also to credit them in some way with the inspiration for that solution

PR #2965, opened 2024-05-03
rfctr: extract OCRAgent.get_agent() out of PDF subtree

scanny:
**Summary**
File-types other than PDF need to use OCR on extracted images. Extract `OCRAgent.get_agent()` such that any file-type partitioner can use it without risking dependency on PDF-only extras.


Coniferish (approved):
lgtm

christinestraub (approved):
LGTM

PR #2969, opened 2024-05-03
First pass at Astra Doc Ingest logic

erichare:
@potter-potter i don't have write access to your branch, so opened up a PR for my changes here

### DIFF
--- a/unstructured/ingest/connector/astra.py
+++ b/unstructured/ingest/connector/astra.py
@@ -18,7 +18,6 @@ from unstructured.ingest.interfaces import (
     BaseSingleIngestDoc,
     BaseSourceConnector,
     IngestDocCleanupMixin,
-    IngestDocSessionHandleMixin,
     SourceConnectorCleanupMixin,
     SourceMetadata,
     WriteConfig,
@@ -119,7 +118,7 @@ class AstraDestinationConnector(BaseDestinationConnector):
             _ = self.astra_db_collection
         except Exception as e:
             logger.error(f"Failed to validate connection {e}", exc_info=True)
-            raise DestinationConnectionNetworkError(f"failed to validate connection: {e}")
+            raise DestinationConnectionError(f"failed to validate connection: {e}")
 
     def write_dict(self, *args, elements_dict: t.List[t.Dict[str, t.Any]], **kwargs) -> None:
         logger.info(f"Inserting / updating {len(elements_dict)} documents to Astra.")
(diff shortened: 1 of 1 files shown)


potter-potter:
@erichare Let's continue working on this from here.

Is this running successfully for you when you test it?

erichare:
@potter-potter i had something come up earlier and haven't had a chance to fully test, i will get back to you on that soon. Also sorry for the very undescriptive branch name here haha, I forgot to change the default 😄 

I'll follow up soon

potter-potter:
I don't think we can fully sort of copy delta-table. I was just trying to point you to something that read from a "database table"-like source. There isn't really a good example of a source connector that parallels AstraDB as far as I can tell.

potter-potter:
It actually looks like this gets the entire table:
astra_docs = list(self.astra_db_collection.paginated_find())

In most of the connectors, since they are not table based, `get_ingest_docs` just gets a pointer to the docs and then we can potentially fan out downloaders to download the docs with `get_files`

That being said, this is not a "docs" based connector (its table based) so it is a little different. I will look through our source connectors again to see if there is a better connector to model.

erichare:
> It actually looks like this gets the entire table: astra_docs = list(self.astra_db_collection.paginated_find())
> 
> In most of the connectors, since they are not table based, `get_ingest_docs` just gets a pointer to the docs and then we can potentially fan out downloaders to download the docs with `get_files`
> 
> That being said, this is not a "docs" based connector (its table based) so it is a little different. I will look through our source connectors again to see if there is a better connector to model.

That is a really great point. Yep, it will pull the full table. paginated_find itself can take a filter condition, if there's any way to pass kwargs to `get_ingest_docs`?

MthwRobinson:
@potter-potter - Do you know if this PR is still being actively worked?

erichare:
@MthwRobinson I got pulled into some other stuff, but would like to get this merged... with that said, we sort of had two simultaneous PRs going, and i think this particular PR is out of date. If @potter-potter agrees, i'll open a new PR with just the Astra DB Source connector logic and we can close this one. I have some updates to make to it anyway as a result of recent changes on the Astra end.

MthwRobinson:
Thanks @erichare !

potter-potter:
@erichare Thanks. If you could close this and open a new one. We merged in the updates to the Astra destination, but this source one needed more work.

erichare:
> @erichare Thanks. If you could close this and open a new one. We merged in the updates to the Astra destination (via a different PR), but this source connector needed more work.

I'm on it @potter-potter . Thanks!

PR #2973, opened 2024-05-06
Support for concurrent processing of documents during evaluation

micmarty-deepsense:
Currently, CCT eval takes a long time for any of the test_metrics CI runs. Documents in an eval set are evaluated sequentially, and It appears that a max of 1 cpu core is currently utilized. This implies there could be a large speedup by running eval across multiple docs concurrently (probably with multiprocessing).

Things done in this PR:
- [x] concurrent.futures.ProcessPoolExecutor instead of sequential for-loop
- [x] refactor/reorganization of redundant pieces of code without changing the inner logic too much. Without that we'd have 3 places where documents are being processed. Take a look at `BaseMetricsCalculator` class and classes that inherit from it.
- [x] string paths manipulation is now reworked and relies on `pathlib.Path()`


unstructured/metrics/evaluate.py
```diff
@@ -41,16 +46,318 @@
...
+    def calculate(
+        self,
+        executor: Optional[concurrent.futures.Executor] = None,
+        export_dir: Optional[str | Path] = None,
+        visualize_progress: bool = True,
+        display_agg_df: bool = True,
+    ) -> pd.DataFrame:
+        if executor is None:
+            executor = self._default_executor()
+        rows = self._process_all_documents(executor, visualize_progress)
```
micmarty-deepsense, line 78: This is where the documents will be processed in parallel

unstructured/metrics/evaluate.py
```diff
@@ -41,165 +45,391 @@
...
+                    total=len(self._document_paths),
+                    leave=False,
+                    disable=not visualize_progress,
+                )
+                if row is not None
+            ]
+
+    def _try_process_document(self, doc: Path) -> Optional[list]:
+        """Safe wrapper around the document processing method."""
+        logger.info(f"Processing {doc}")
```
micmarty-deepsense, line 153: If we don't want to see this log message in GH, just let me know

badGarnet (approved):
tested and working as advertised

PR #2977, opened 2024-05-07
feat(evaluation): skip accuracy calculation

amadeusz-ds:
Skip accuracy calculation for files for which output and ground truth sizes differ greatly.

~10% speed up on local machine, keeping the same metrics.


unstructured/metrics/evaluate.py
```diff
@@ -116,7 +116,12 @@ def measure_text_extraction_accuracy(
             except Exception:
                 # if any of the output/source file is unable to open, skip the loop
                 continue
-            accuracy = round(calculate_accuracy(output_cct, source_cct, weights), 3)
+            # NOTE(amadeusz): Levenshtein distance calculation takes too long
+            # skip it if file sizes differ wildly
+            if 0.5 < len(output_cct.encode()) / len(source_cct.encode()) < 2.0:
+                accuracy = round(calculate_accuracy(output_cct, source_cct, weights), 3)
+            else:
+                accuracy = 0.0
```
cragwolfe, line 124: maybe something small and recognizable like 0.01?

amadeusz-ds: ok

amadeusz-ds: done

unstructured/metrics/evaluate.py
```diff
@@ -116,7 +116,13 @@ def measure_text_extraction_accuracy(
...
                 continue
-            accuracy = round(calculate_accuracy(output_cct, source_cct, weights), 3)
+            # NOTE(amadeusz): Levenshtein distance calculation takes too long
+            # skip it if file sizes differ wildly
+            if 0.5 < len(output_cct.encode()) / len(source_cct.encode()) < 2.0:
+                accuracy = round(calculate_accuracy(output_cct, source_cct, weights), 3)
+            else:
+                # 0.01 to distinguish it was set manually
+                accuracy = 0.01
             percent_missing = round(calculate_percent_missing_text(output_cct, source_cct), 3)
```
Klaijan, line 126: Do we want to calculate `percent_missing` still, if we are skipping `accuracy`?

amadeusz-ds: We can but for those examples accuracy was always `0.0` but percent_missing wasn't `1.0` so it would change current results.
And it's quicker to calculate than accuracy.

Klaijan: Maybe a question for @cragwolfe ?

cragwolfe: yeah, assuming it is much quicker to calculate than accuracy, i'm good to leave `percent_missing` as is

Klaijan: Thank you. Then this looks good to me!

Klaijan (approved):

PR #2978, opened 2024-05-07
remove unnecessary warning log

badGarnet:
The warning log about default model not set is no longer needed. This PR removes this log to reduce confusion.


scanny (approved):
LGTM :)

PR #2982, opened 2024-05-07
feat: add chunking to partition_tsv

### LINKED ISSUE #2980: bug: chunking doesn't work on TSV
`partition_tsv()` ignores the `chunking_strategy` argument. This is because the `@add_chunking_strategy` decorator is missing from the `partition_tsv()` function definition.

**To Reproduce**
```python
def it_chunks_the_partitioned_elements_when_so_instructed():
    elements = partition_tsv(
        "example-docs/stanley-cups.tsv", chunking_strategy="basic", max_characters=50
    )
    assert len(elements) == 10, f"expected 10, got {len(elements)}"

>>> it_chunks_the_partitioned_elements_when_so_instructed()
AssertionError: expected 10, got 1
```

**Additional context**
Even though `partition_tsv()` produces a single `Table` element, chunking should still be available (as it is on the similar `partition_csv()`) because the `Table` element can and often will be larger than the desired chunk size and must be divided into smaller chunks.

Coniferish:
Closes #2980

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,6 @@
 * **Remove `page_number` metadata fields** for HTML partition until we have a better strategy to decide page counting.
 * **Extract OCRAgent.get_agent().** Generalize access to the configured OCRAgent instance beyond its use for PDFs.
 * **Add calculation of table related metrics which take into account colspans and rowspans**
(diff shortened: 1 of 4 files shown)


scanny (approved):
LGTM

PR #2987, opened 2024-05-08
Add starting_page_num param to partition_image

### LINKED ISSUE #2985: bug: `starting_page_number` parameter is missing from `partition_image()`
`starting_page_number` argument should be present in all partitioning functions that assign `element.metadata.page_number`. This small change is useful for users who partition only fragments of their PDF documents or images. This is also related to the following issue: https://github.com/Unstructured-IO/unstructured/issues/2461

Coniferish:
Add missing starting_page_num param to partition_image

Closes #2985


Coniferish:
@micmarty-deepsense, would you know why this test is generating different hashes for the docker test, but passes in the unit tests?
https://github.com/Unstructured-IO/unstructured/actions/runs/9005311385/job/24740389787?pr=2987

Coniferish:
@christinestraub, tagging you to double-check that there wasn't a reason why `starting_page_num` was left out of `partition_image`

test_unstructured/partition/pdf_image/test_image.py
```diff
@@ -737,3 +741,71 @@ def test_partition_image_works_on_heic_file(
...
+        "9f87d002da6a94daec6a1b957b2fedfe",
+        "8211d66d55646be7f798172ea5596979",
+        "aa7c780f8442901becaf1c4f7067708e",
+        "9516561be9106b4867dd7bfeeafa895d",
+        "77b2a946442eef307412d62f284757ff",
+        "e557fb5a2472f5c9be22e6bd07cc6b7a",
+        "23d5472327ea19d8331cd5b0ed970e3f",
+        "6c67796c499a68160c107006a99a9392",
+        "7468807f52b225b01404f9f3d7c9e6e2",
+    ]
```
cragwolfe, line 791: i know we have this pattern elsewhere in the tests, but it is not great to manually have to update these when the ID hashing algo changes.
instead, let's run partition twice (even though the test will take longer) and make sure we are getting back the same element_id's.

cragwolfe:
> @christinestraub, tagging you to double-check that there wasn't a reason why `starting_page_num` was left out of `partition_image`

i assume there was no reason, as this is relatively new. the big question is if multi-image .tiff's will correctly start counting from `starting_page_num` -- it would be good to prove that is working (though also could be a follow on PR).

cragwolfe (approved):

scanny (approved):
LGTM :)

PR #2988, opened 2024-05-08
fix param types for partition_image and _pdf

Coniferish:
Make the `filename` and `file` params for `partition_image` and `partition_pdf` match the other partitioners


cragwolfe (approved):

scanny (approved):
LGTM :)

PR #2996, opened 2024-05-09
fix: ppt parameters include_page_breaks and include_slide_notes

Coniferish:
Pass the parameters `include_slide_notes` and `include_page_breaks` to `partition_pptx` from `partition_ppt`.

Also update the .ppt example doc we use for testing so it has slide notes and a PageBreak (and second page)


test_unstructured/partition/pptx/test_ppt.py
```diff
@@ -181,9 +182,15 @@ def test_add_chunking_strategy_by_title_on_partition_ppt():
...
-    elements = partition_ppt(example_doc_path("fake-power-point.ppt"))
+def test_partition_ppt_params():
+    """Integration test of params: languages, include_page_break, and include_slide_notes."""
+    elements = partition_ppt(
+        example_doc_path("fake-power-point.ppt"), include_page_breaks=True, include_slide_notes=True
+    )
     assert elements[0].metadata.languages == ["eng"]
+    assert any(isinstance(element, PageBreak) for element in elements)
+    # The example doc contains a slide note with the text "This is a slide note"
+    assert any(element.text == "This is a slide note" for element in elements)
```
scanny, line 193: I don't think this is going to pass, is it? `fake-power-point.ppt` only has a single slide as far as I know so it won't have any page-breaks. Also I doubt it has any slide notes. Did you add one? That could disrupt other tests because they all use this file and compare results against a known set of elements listed at the top of the file.

You could make a copy under a different name and use that. Or maybe use `eng_spa_mult.ppt`.

Coniferish: Some tests did just fail in `auto`, though. I'll see if it won't be too complicated to update them as well and add a new file with two slides if it is

scanny (approved):
LGTM :)

PR #3001, opened 2024-05-10
remove unused links param

Coniferish:
The `links` param in `partition_pdf` was never used by the partitioner, but added when that metadata element was created. This removes the unused parameter since `links` are extracted during partitioning.


scanny (approved):
LGTM :)

PR #3002, opened 2024-05-10
paragraph grouper type hint

Coniferish:
Fix type hint for paragraph_grouper param. `paragraph_grouper` can be set to `False`, but the type hint did not not reflect this previously.


scanny (approved):
LGTM :)

PR #3008, opened 2024-05-13
build(deps): bump versions for security hygiene

MthwRobinson:
### Summary

Version bumps to keep on top of security scans.


Coniferish (approved):
lgtm

PR #3009, opened 2024-05-13
feat: refactor ingest (WIP)

rbiseck3:
### Description
This refactors the current ingest CLI process to support better granularity in how the steps are ran
* Both multiprocessing and async now supported. Given that a lot of the steps are IO-bound, such as downloading and uploading content, we can achieve better parallelization by using async here
* Destination step broken up into a stager step and an upload step. This will allow for steps that require manipulation of the data between formats, such as converting the elements json into a csv format to upload for tabular destinations, to be pulled out of the step that does the actual upload. 
* The process of writing the content to a local destination was now pulled out as it's own dedicated destination connector, meaning you no longer need to persist the content locally once the process is done if the content was uploaded elsewhere.


unstructured/ingest/v2/logger.py
```diff
@@ -0,0 +1,111 @@
...
+import json
+import os
+from logging import Formatter, Logger, StreamHandler, getLevelName, getLogger
+from typing import Any, Callable
+
+log_level = os.getenv("INGEST_LOG_LEVEL", "INFO")
+LOGGER_NAME = "unstructured.ingest.v2"
+
+
+def default_is_data_sensitive(k: str, v: Any) -> bool:
```
potter-potter, line 11: Are we still able to use ` enhanced_field(sensitive=True)` ?  I really liked how that worked.

potter-potter:
This is really cool! And its also a lot to understand. Could you add notes on how to test it out in the description?

unstructured/ingest/v2/logger.py
rbiseck3: This was mostly a copy-paste of what was already being used by the current ingest code.

ryannikolaidis:
are there doc changes that need to happen for users of the connectors via Python?

unstructured/ingest/cli/cli.py
```diff
@@ -15,10 +17,16 @@ def get_cmd() -> click.Command:
     to the main command as nested subcommands.
     """
     cmd = ingest
+    src_dict = {s.name: s for s in src}
+    dest_dict = {d.name: d for d in dest}
+    for s in src_v2:
+        src_dict[s.name] = s
+    for d in dest_v2:
+        dest_dict[d.name] = d
```
ryannikolaidis, line 25: got it, so we just defer to v2 for cli if there exists a v2 command

rbiseck3: Yep, at least right now as soon as the new connector exists, it replaces the old one.

ryannikolaidis: Yea, this is actually pretty nice because it means that nothing breaks as far as users know. Also means we can add deprecations warnings for the Python users for whom it will break once we switch over.

ryannikolaidis: Maybe worth doing that for the ones that you're bumping here? And probably bump the docs for the respective connectors?

unstructured/ingest/v2/cli/base/cmd.py
```diff
@@ -0,0 +1,158 @@
...
+            destination_options = destination_options or {}
+            if "output_dir" not in destination_options:
+                destination_options["output_dir"] = source_options["output_dir"]
+            pipeline_kwargs["uploader"] = self.get_default_uploader(options=destination_options)
+        return Pipeline(**pipeline_kwargs)
+
+    @staticmethod
+    def get_default_uploader(options: dict[str, Any]) -> UploaderT:
+        uploader_config = extract_config(flat_data=options, config=LocalUploaderConfig)
+        return LocalUploader(upload_config=uploader_config)
```
ryannikolaidis, line 78: this is clever

unstructured/ingest/cli/cli.py
rbiseck3: Where do we manage the docs for connectors now? I thought that changed recently and not sure if that's still being managed by this repo.

unstructured/ingest/v2/logger.py
rbiseck3: To be explicit, only the logging code was copied. Everything else in this PR is new code.

unstructured/ingest/cli/cli.py
ryannikolaidis: good question! so, we now host the docs themselves in this repo: https://github.com/Unstructured-IO/docs 

and it pulls source code snippets from here: https://github.com/Unstructured-IO/unstructured/tree/main/docs/source/ingest/source_connectors/code
and
https://github.com/Unstructured-IO/unstructured/tree/main/docs/source/ingest/destination_connectors/code

ryannikolaidis:
Overall I think there are a lot of good changes in here! This does come at the cost of quite a bit of complexity (though to be fair some of this is coming back in a bit cold). My biggest concern is that it's going to be tough for an average person to walk in and debug or be effective here, and currently you have to do a lot of tracing to grok what's going on. My main ask: can we equivalently (in size and scope) push the documentation? Including a lot more comments and docstrings as well as a comprehensive breakdown of everything happening with maybe some diagrams as well?

unstructured/ingest/v2/cli/base/dest.py
```diff
@@ -0,0 +1,54 @@
...
+            )
+            pipeline.run()
+        except Exception as e:
+            logger.error(f"failed to run destination command {self.cmd_name}: {e}", exc_info=True)
+            raise click.ClickException(str(e)) from e
+
+    def get_cmd(self) -> click.Command:
+        # Dynamically create the command without the use of click decorators
+        fn = self.cmd
+        fn = click.pass_context(fn)
```
ryannikolaidis, line 41: maybe I'm missing something, why can't we just `fn = click.pass_context(self.cmd)`

unstructured/ingest/README.md
```diff
@@ -1,4 +1,5 @@
-# Batch Processing Documents
+# Batch Processing Documents [DEPRECATED]
+For the latest approach, go to: [v2](./v2)
```
ryannikolaidis, line 2: I guess not quite deprecated until we add all of v2, right?

ryannikolaidis: warning of "soon to be deprecated"?

unstructured/ingest/v2/README.md
```diff
@@ -0,0 +1,119 @@
...
+`utic-dev-tech-fixtures` bucket with a prefix of `small-pdf-set/`.
+
+    unstructured-ingest \
+       s3 \
+       --remote-url s3://utic-dev-tech-fixtures/small-pdf-set/ \
+       --anonymous \
+       --output-dir s3-small-batch-output \
+       --num-processes 2
+
+Naturally, --num-processes may be adjusted for better instance utilization with multiprocessing.
```
ryannikolaidis, line 17: worth calling out mp + asyncio support here (per changes)?

unstructured/ingest/v2/README.md
```diff
@@ -0,0 +1,119 @@
...
+  - [ ] Add the decorator `unstructured.utils.requires_dependencies` on top of each class instance or function that uses those connector-specific dependencies e.g. for `GitHubConnector` should look like `@requires_dependencies(dependencies=["github"], extras="github")`
+  - [ ] Run `make tidy` and `make check` to ensure linting checks pass.
+- [ ] Update ingest documentation [here](https://github.com/Unstructured-IO/docs)
+- [ ] For team members that are developing in the original repository:
+  - [ ] If there are secret variables created for the connector tests, make sure to:
+    - [ ] add the secrets into Github (contact someone with access)
+    - [ ] include the secret variables in [`ci.yml`](https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ci.yml) and [`ingest-test-fixtures-update-pr.yml`](https://github.com/Unstructured-IO/unstructured/blob/main/.github/workflows/ingest-test-fixtures-update-pr.yml)
+    - [ ] add a make install line in the workflow configurations to be able to provide the workflow machine with the required dependencies on the connector while testing
```
ryannikolaidis, line 95: assuming this is the github workflow configuration for ci right? maybe link here?

unstructured/ingest/v2/README.md
```diff
@@ -0,0 +1,119 @@
...
+* `Chunker`: Optionally chunk the partitioned content. Can also be run locally or via the api, with async/multiprocessing set in the same approach as the partitioner.
+* `Embedder`: Create vector embeddings for each element in the structured output. Most of these are via an API call (i.e. AWS Bedrock) and therefor run async but can also use a local huggingface model which will run via multiprocessing.
+* `Stager`: This is an optional step that won't apply for most pipelines. If the data needs to be modified from the existing structure to better support the upload, such as converting it to a csv for tabular-based destinations (sql dbs).
+* `Uploader`: Write the local content to the destination. If none if provided, the local one will be used which writes the final result to a location on the local filesystem. If batch uploads are needed, this will run in a single process with access to all docs. If batch is not supported, all docs can be uploaded concurrently using the async approach.
+
+
+
+### Parallel Execution/Concurrency
```
ryannikolaidis, line 118: The above is a good birds eye view; I think would be a really helpful addition is to diagram and explain the entire journey in depth. i.e. trace granularly (like function level) incoming request -> constructing command -> building pipeline -> executing pipeline so that a developer can drop into code an understand exactly where it fits in and how to debug.

unstructured/ingest/README.md
rbiseck3: I guess the README should be considered fully deprecated because I don't want users coming here for docs on how to add a connector using this once the PR is merged in.

unstructured/ingest/v2/cli/base/dest.py
rbiseck3: I wanted to make it intuitive that `click.pass_context()` simply returns a modified version of what was passed in. 
```python
fn = click.pass_context(self.cmd)
```
Doesn't make that as apparent.

unstructured/ingest/v2/README.md
ryannikolaidis: hmm, if neither of us are certain what this is referring to ... that's not good. actually digging in, wonder if this pre-dated the iteration in the `install-all-ingest` [target](https://github.com/Unstructured-IO/unstructured/blob/915aa0f386bcd5f04a55e646a51b6a87a61c7058/Makefile#L108-L111) and we required manually adding `make install` commands?

unstructured/ingest/v2/cli/base/dest.py
ryannikolaidis: oh funny, yea, I assumed as much and then was just confused why we were setting it above. I guess dealers choice, don't have strong feelings.

unstructured/ingest/v2/README.md
rbiseck3: This is related to the make targets for each connector dependency install, such as this: [install-ingest-google-drive](https://github.com/Unstructured-IO/unstructured/blob/main/Makefile#L111)

ryannikolaidis (approved):
Probably should mention in the PR description the bug fix (which is actually what caused the PR's surface area to bloat a bunch)

Thanks for the additions so far. Just noting here from conversation, more extensive documentation (mentioned in comments) will be added before we switch over from original code and current flow. In the meantime this should be non-breaking for users.

PR #3011, opened 2024-05-13
rfctr(docx): move docx unit tests to bottom

scanny:
No code changes, strictly this single block move.

Move `Describe_DocxPartitioner` unit-test class to bottom so `DescribeDocxPartitionerOptions` unit-test to follow in subsequent commit will be together with it. Integration tests first, then unit tests, for consistency with other test modules e.g. test_pptx.

I added `Describe_DocxPartitioner` soon after I arrived, before we adopted the convention of placing unit-tests after integration tests. Move this so we can maintain that consistency with the block of tests to follow in a closely subsequent PR.


Coniferish (approved):
LGTM

PR #3013, opened 2024-05-14
Astra DB: explicit projection when reading from Astra DB

hemidactylus:
In view of upcoming changes in the Astra DB Data API, this PR explicitly sets a projection every time a `find` command (or equivalent) is executed, in order to ensure all necessary fields of the document are returned by the API.

This occurs in the two method calls `find_one` and `vector_find` in the test `test-ingest-astra-output.py`.

With this change, future versions of the Data API, which may implement an exclude-vector-by-default policy, will not cause any disruption in these tests.

### DIFF
--- a/test_unstructured_ingest/python/test-ingest-astra-output.py
+++ b/test_unstructured_ingest/python/test-ingest-astra-output.py
@@ -49,12 +49,16 @@ def check(ctx):
 
     # Grab an embedding from the collection and search against itself
     # Should get the same document back as the most similar
-    find_one = astra_db_collection.find_one()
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):

potter-potter:
@hemidactylus Are there any code changes in Unstructured needed for this? Or is it just in that test script that we are using to check our results?

hemidactylus:
Hi, @potter-potter, it is just the test script indeed. Please feel free to close this and apply the patch within another  PR :)
 ... and thank you for taking care of it!

potter-potter:
Closing.
update added here:
https://github.com/Unstructured-IO/unstructured/pull/3003

PR #3018, opened 2024-05-14
rfctr(docx): extract DocxPartitionerOptions

scanny:
**Reviewers:** Probably easier to review first and second commits separately as the first one adds all the new code and tests (without installing it), and the second one installs it into the partitioner along with the required changes to code and tests.

**Summary**
Enable communication of partitioning options to sub-partitioners, in particular to the pluggable `PicturePartitioner` coming in a closely subsequent PR to implement image-extraction and OCR for DOCX, DOC, and ODT formats.

**Additional Context**
In general, validation of partitioning options as well as assigning default values and computing derived partitioning settings can be extracted from partitioners into a neatly encapsulated separate object. This simplifies the core partitioning code by removing the noise associated with computing metadata values and deciding how to access the source document, etc.

However, better factoring aside, having the partition-time "settings" available in a single object allows partitioning of certain document features, for example images, to be readily _delegated_ to a sub-partitioner while still giving it access to all the relevant partitioning settings for the current document. This is particularly important when a sub-partitioner is "pluggable" at runtime and must rely on a clearly-defined (and simple as possible) interface to operate smoothly.

### DIFF
--- a/unstructured/partition/docx.py
+++ b/unstructured/partition/docx.py
@@ -217,19 +217,19 @@ def partition_docx(
         Assign this number to the first page of this document and increment the page number from
         there.
     """
-    # -- verify that only one file-specifier argument was provided --
-    exactly_one(filename=filename, file=file)
-
-    elements = _DocxPartitioner.iter_document_elements(
-        filename,
-        file,
-        metadata_filename,
-        include_page_breaks,
-        infer_table_structure,
-        metadata_last_modified,
-        date_from_file_object,
+    opts = DocxPartitionerOptions(
+        date_from_file_object=date_from_file_object,
+        file=file,
+        file_path=filename,
+        include_page_breaks=include_page_breaks,
+        infer_table_structure=infer_table_structure,
+        metadata_file_path=metadata_filename,
+        metadata_last_modified=metadata_last_modified,
         starting_page_number=starting_page_number,
     )
+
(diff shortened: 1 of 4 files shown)


unstructured/partition/docx.py
```diff
@@ -238,6 +238,159 @@ def partition_docx(
...
+        return bool(self.document.element.xpath(xpath))
+
+    @lazyproperty
+    def _docx_file(self) -> str | IO[bytes]:
+        """The Word 2007+ document file to be partitioned.
+
+        This is either a `str` path or a file-like object. `python-docx` accepts either for opening
+        a document file.
+        """
+        if self._file_path:
```
Coniferish, line 376: file_path is filename, right?

unstructured/partition/docx.py
```diff
@@ -217,19 +217,19 @@ def partition_docx(
...
-        file,
-        metadata_filename,
-        include_page_breaks,
-        infer_table_structure,
-        metadata_last_modified,
-        date_from_file_object,
+    opts = DocxPartitionerOptions(
+        date_from_file_object=date_from_file_object,
+        file=file,
+        file_path=filename,
```
Coniferish, line 223: Seeing this now. My other comment can be ignored

Coniferish (approved):
LGTM

PR #3019, opened 2024-05-15
fix(doc): fix disk-space leak

scanny:
**Summary**
Remedy disk-space leak where `partition_doc()` would leave a copy of each `.doc` file passed as a file-like object on disk.

**Additional Context**
`partition_doc()` creates a temporary file in which it writes each source-document provided as a file-like object. This file is not deleted and disk consumption grows without bound.

The `convert_office_doc()` function used to convert DOC->DOCX uses a command-line program provided with LibreOffice to convert do the conversion. Because this command-line program operates in a different memory space, the source file cannot be passed as an in-memory object and needs to be on the filesystem. When the DOC file is passed as a file-like object, it is written to disk so the conversion program has access to it. It is not deleted afterward.

Fix this by writing the temporary source DOC file in the TemporaryDirectory already being used to write the conversion-target DOCX file. That directory is automatically removed when `partition_doc()` completes.

### DIFF
--- a/unstructured/partition/doc.py
+++ b/unstructured/partition/doc.py
@@ -63,52 +63,74 @@ def partition_doc(
         This information will be reflected in elements' metadata and can be be especially
         useful when partitioning a document that is part of a larger document.
     """
-    # Verify that only one of the arguments was provided
-    if filename is None:
-        filename = ""
     exactly_one(filename=filename, file=file)
 
-    if len(filename) > 0:
-        _, filename_no_path = os.path.split(os.path.abspath(filename))
-        base_filename, _ = os.path.splitext(filename_no_path)
-        if not os.path.exists(filename):
-            raise ValueError(f"The file {filename} does not exist.")
+    last_modified = get_last_modified(filename, file, date_from_file_object)
 
-        last_modification_date = get_last_modified_date(filename)
+    # -- validate file-path when provided so we can provide a more meaningful error --
+    if filename is not None and not os.path.exists(filename):
+        raise ValueError(f"The file {filename} does not exist.")
 
-    elif file is not None:
-        tmp = tempfile.NamedTemporaryFile(delete=False)
-        tmp.write(file.read())
-        tmp.close()
-        filename = tmp.name
-        _, filename_no_path = os.path.split(os.path.abspath(tmp.name))
-        base_filename, _ = os.path.splitext(filename_no_path)
+    # -- `convert_office_doc` uses a command-line program that ships with LibreOffice to convert
+    # -- from DOC -> DOCX. So both the source and the target need to be file-system files. Put
+    # -- transient files in a temporary directory that is automatically removed so they don't
+    # -- pile up.
+    with tempfile.TemporaryDirectory() as target_dir:
 
-        last_modification_date = (
-            get_last_modified_date_from_file(file) if date_from_file_object else None
-        )
+        source_file_path = f"{target_dir}/document.doc" if file is not None else filename
+        assert source_file_path is not None
 
-    with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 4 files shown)


unstructured/partition/doc.py
```diff
@@ -63,52 +63,74 @@ def partition_doc(
         This information will be reflected in elements' metadata and can be be especially
         useful when partitioning a document that is part of a larger document.
     """
-    # Verify that only one of the arguments was provided
-    if filename is None:
-        filename = ""
```
scanny, line 68: There is no need for a `None` filename value to be converted to `str`. Ditch this.

unstructured/partition/doc.py
```diff
@@ -63,52 +63,74 @@ def partition_doc(
...
     exactly_one(filename=filename, file=file)
 
-    if len(filename) > 0:
-        _, filename_no_path = os.path.split(os.path.abspath(filename))
-        base_filename, _ = os.path.splitext(filename_no_path)
```
scanny, line 73: These two lines are part of computing the target "converted" file-path so that path can be sent to `partition_docx()`. They appear both here in the "filename" branch and below in the "file" branch. Move these to just above where that value is used and the source filename is resolved.

unstructured/partition/doc.py
```diff
@@ -63,52 +63,74 @@ def partition_doc(
...
-        if not os.path.exists(filename):
-            raise ValueError(f"The file {filename} does not exist.")
+    last_modified = get_last_modified(filename, file, date_from_file_object)
```
scanny, line 68: Extract the `last_modified` computation to its own function. This same logic appears in every partitioner and implementing it separately in each partitioner invites bugs arising from inconsistency. The extracted function can be moved to `partition.common` and used by other partitioners in a separate PR.

unstructured/partition/doc.py
```diff
@@ -63,52 +63,74 @@ def partition_doc(
...
 
-        last_modification_date = get_last_modified_date(filename)
+    # -- validate file-path when provided so we can provide a more meaningful error --
+    if filename is not None and not os.path.exists(filename):
+        raise ValueError(f"The file {filename} does not exist.")
 
-    elif file is not None:
-        tmp = tempfile.NamedTemporaryFile(delete=False)
```
scanny, line 80: This is the line that gives rise to the disk-space leak. Note the `delete=False` argument which causes the file to remain after use.

unstructured/partition/doc.py
```diff
@@ -63,52 +63,74 @@ def partition_doc(
...
+    # -- from DOC -> DOCX. So both the source and the target need to be file-system files. Put
+    # -- transient files in a temporary directory that is automatically removed so they don't
+    # -- pile up.
+    with tempfile.TemporaryDirectory() as target_dir:
 
-        last_modification_date = (
-            get_last_modified_date_from_file(file) if date_from_file_object else None
-        )
+        source_file_path = f"{target_dir}/document.doc" if file is not None else filename
+        assert source_file_path is not None
```
scanny, line 81: Instead of assigning it a random temporary file name, use `document.doc`. Because it is added in the Temporary directory added by this `with` context, it will be deleted with the directory when the `with` context is exited.

unstructured/partition/doc.py
```diff
@@ -63,52 +63,74 @@ def partition_doc(
...
             languages=languages,
-            detect_language_per_element=detect_language_per_element,
+            metadata_filename=metadata_filename,
+            metadata_last_modified=metadata_last_modified or last_modified,
             starting_page_number=starting_page_number,
         )
-        # remove tmp.name from filename if parsing file
-        if file:
-            for element in elements:
-                element.metadata.filename = metadata_filename
```
scanny, line 112: No need for this to be within the `with` context.

Coniferish (approved):
LGTM!

PR #3024, opened 2024-05-15
rfctr(doc): spruce up test_doc.py

scanny:
**Summary**
In preparation for adding more tests related to image extraction, improve the `partition_doc()` test suite:

- Remove redundant DOCX -> DOC file conversions on most tests.
- Add type annotations to type-check clean on strict mode.
- Improve test names.
- Simplify tests where possible.
- Remove one duplicated test

Speed was roughly doubled: 24 tests in 20s -> 23 tests in 8s.


test_unstructured/partition/docx/test_doc.py
```diff
@@ -1,214 +1,179 @@
...
-from unstructured.partition.common import convert_office_doc
 from unstructured.partition.doc import partition_doc
 from unstructured.partition.docx import partition_docx
 
 
-def test_partition_doc_matches_partition_docx(mock_document, expected_elements, tmpdir):
-    docx_filename = os.path.join(tmpdir.dirname, "mock_document.docx")
-    doc_filename = os.path.join(tmpdir.dirname, "mock_document.doc")
-    mock_document.save(docx_filename)
-    convert_office_doc(docx_filename, tmpdir.dirname, "doc")
```
scanny, line 27: These four lines creating the same DOCX file from scratch and then convert it to `.doc` for the test. This code is repeated on almost every test and is not the behavior under test. Replace these documents, which are the same on every test run, with `simple.doc` and `simple.docx` stored in `example-docs/`, then use them directly rather than remake them each time.

This is where the 50% time savings comes from, plus it's a lot easier to read.

test_unstructured/partition/docx/test_doc.py
```diff
@@ -1,214 +1,179 @@
...
-    elements = partition_doc(filename=doc_filename)
+def test_partition_doc_from_filename(expected_elements: list[Element], capsys: CaptureFixture[str]):
+    elements = partition_doc(example_doc_path("simple.doc"))
+
     assert elements == expected_elements
-    assert elements[0].metadata.filename == "mock_document.doc"
-    assert elements[0].metadata.file_directory == tmpdir.dirname
```
scanny, line 42: Just as easy to test all the elements as to spot-check the first one.

Coniferish (approved):
LGTM

PR #3026, opened 2024-05-15
feat(docx): add strategy parameter to partition_docx()

scanny:
**Summary**
The behavior of an image sub-partitioner can be partially determined by the partitioning strategy, for example whether it is "hi_res" or "fast". Add this parameter to `partition_docx()` so it can pass it along to `DocxPartitionerOptions` which will make it available to any image sub-partitioners.

### DIFF
--- a/unstructured/partition/docx.py
+++ b/unstructured/partition/docx.py
@@ -57,6 +57,7 @@ from unstructured.partition.text_type import (
     is_possible_title,
     is_us_city_state_zip,
 )
+from unstructured.partition.utils.constants import PartitionStrategy
 from unstructured.utils import (
     dependency_exists,
     is_temp_file_path,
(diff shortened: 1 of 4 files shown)


unstructured/partition/docx.py
```diff
@@ -345,6 +351,15 @@ def page_number(self) -> int:
...
         return self._page_counter
 
+    @lazyproperty
+    def strategy(self) -> str:
+        """The partitioning strategy for this document.
+
+        One of "hi_res", "fast", and a few others. These are available as class attributes on
+        `unstructured.partition.utils.constants.PartitionStrategy` but resolve to str values.
+        """
+        return PartitionStrategy.HI_RES if self._strategy is None else self._strategy
```
Coniferish, line 361: Just want to double-check that it should default to hi_res, otherwise looks good

Coniferish (approved):
lgtm

unstructured/partition/docx.py
scanny: k, yep, confirmed :)

PR #3028, opened 2024-05-15
chore: move logger error to debug when pdfminer extract fails

yuming-long:
### Summary

We are seeing logger error `Invalid dictionary construct` for hosted APIs, move this logger error to debug level - we still continue partition when pdfminer text extraction fails as before (just don't throw the log error anymore)

### Test
I was able to reproduce the logger error with an internal only file (please DM me if needed) and the error trace look like
```
 File "/Users/yumingl/develops/unstructured/unstructured/partition/pdf.py", line 709, in _process_pdfminer_pages
    annotation_list = get_uris(page.annots, height, coordinate_system, page_number)
  File "/Users/yumingl/develops/unstructured/unstructured/partition/pdf.py", line 1049, in get_uris
    resolved_annots = annots.resolve()
...
```
we also won't be able to repair pdf structure on `get_uris` (not a page level) so move this exception to debug level.


cragwolfe (approved):

unstructured/partition/pdf.py
```diff
@@ -269,7 +269,7 @@ def partition_pdf_or_image(
                 isinstance(el, Text) and el.text.strip() for el in extracted_elements
             )
         except Exception as e:
-            logger.error(e)
+            logger.debug(e)
```
cragwolfe, line 272: should the below still be a warning, or info?

christinestraub (approved):

PR #3033, opened 2024-05-16
improve: add Python 3.12 support

### LINKED ISSUE #2959: chore/Allow python 3.12
Currently unstructured pins the [python version to >=3.9.0,<3.12](https://github.com/Unstructured-IO/unstructured/blob/main/setup.py#L87C22-L87C35). It would be great if 3.12 would be allowed too. Seems like there were two attempts to enable it before:
* https://github.com/Unstructured-IO/unstructured/pull/1703
* https://github.com/Unstructured-IO/unstructured/pull/1902

But the logs from the CI/CD are gone, so I am not sure why they failed.

nicoloboschi:
Draft: currently blocked by argilla not supporting 3.12
Waiting for this PR to be released: https://github.com/argilla-io/argilla/pull/4837

Fixes #2959


MthwRobinson:
Thanks for this @nicoloboschi ! Just turned on the CI pipeline. If I recall correctly, it was `onnx` that was blocking 3.12 support before, but that should be resolved now.

MthwRobinson:
CI failure was due to this. We may be able to work around that though.

```python
ERROR: Cannot install argilla==1.28.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    argilla 1.28.0 depends on numpy<1.24.0
    The user requested (constraint) numpy>=1.26.0
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

MthwRobinson:
Ahh disregard, I see you have a PR open for 3.12 support in Argilla as well.

gokturkDev:
@MthwRobinson  hi I experience the below error when trying to use python 3.12 with unstructured

gokturkDev:
Meanwhile are there any hotfixes I can apply?

MthwRobinson:
@gokturkDev - In the meantime, I'd try manually running `make install` on the branch for this PR. The version updates @nicoloboschi added to `contraints.txt` are what you need to enable 3.12. As long as you aren't using the Argilla integration I don't think it should be an issue, though I haven't tested it myself yet.

gokturkDev:
Hey thanks for the response. I ended up installing the latest onnx manually instead of using the one installed as an unstructured dependency and its working so far 👍

MthwRobinson:
@nicoloboschi - Going to merge this into a feature branch and then get it into `main` from there. Failing tests are because env vars are missing on the fork. Thank you for this contribution!

MthwRobinson:
New PR is #3047

PR #3035, opened 2024-05-16
BREAKING CHANGE: revert table extraction off by default for PDFs and images

### LINKED ISSUE #3021: Turn table extraction off by default for PDFs and images
The goal of this issue is to update the default value of `skip_infer_table_types` to `skip_infer_table_types=["pdf", "jpg", "png", "xls", "xlsx", "heic"]`. Prior to version `0.13`, table extraction was off by default. Since the default behavior was changed, we've found there are users who don't know that setting exists, and observed slow downs in PDF processing times.

MthwRobinson:
### Summary

Closes #3021 . Turns table extraction for PDFs and images off by default. The default behavior originally changed in #2588 . The reason for reversion is that some users did not realize turning off table extraction was an option and experience long processing times for PDFs and images with the new default behavior.


unstructured/partition/auto.py
```diff
@@ -141,12 +141,12 @@ def partition(
     encoding: Optional[str] = None,
     paragraph_grouper: Optional[Callable[[str], str]] | Literal[False] = None,
     headers: dict[str, str] = {},
-    skip_infer_table_types: list[str] = [],
+    skip_infer_table_types: list[str] = ["pdf", "jpg", "png", "xls", "xlsx", "heic"],
```
scanny, line 144: I recommend we only include PDF and image types in this list, not `xls` or `xlsx`.

### Rationale
The "infer_table" behavior is historically somewhat confused when it comes to non-PDF/Image file-types.

The `infer_table_structure` argument (predecessor to `skip_infer_table_types`) controlled two things and we often want those two controlled separately for non-PDF/Image types.

1. For PDF/Image file-types, `infer_table_types` controlled partitioning "expense", both money and time, by controlling whether the extra compute of using a table-inference model was engaged. This was important (and still is) for PDF/Image because doing this can take a lot more CPU and elapsed-time.
2. It _also_ determined whether `.metadata.text_as_html` was added to `Table` elements. For PDF/Image this was obvious enough because if you didn't infer tables there was no `Table` element to put the `.metadata.text_as_html` on.
3. However, for non-PDF/Image file-types, no inference was ever required or used to detect tables. So `Table` elements always appear for those file-types when they occur in the document.
4. In that non-PDF/Image case, `infer_table_structure` came to mean: "include `.metadata.text_as_html` on `Table` elements".
### My POV
- In my opinion, any "infer_table" option, regardless of how it is specified, should be limited to PDF/Image types.
- In addition, we can have a `suppress_metadata_fields: list[str] = []` parameter that is applicable to all partitioners and handled in post-partitioning processing (implemented in one of the decorators), and that removes any metadata the user prefers not to appear in the payload.
- In the meantime, I believe only PDF/Image partitioners are actually sensitive to the value of `skip_infer_table_types`, so probably not a big deal, but still, better to remove `xls` and `xlsx` from the list I think.

scanny (approved):
Hi Matt, just the one comment but I think it's worth reading and removing Excel types from the list.

Approving in advance, you can make whatever changes you think appropriate based on my remarks before you merge :)

unstructured/partition/auto.py
qued: Agree with all this.

MthwRobinson: Updated!

unstructured/partition/auto.py
```diff
@@ -141,12 +141,12 @@ def partition(
     encoding: Optional[str] = None,
     paragraph_grouper: Optional[Callable[[str], str]] | Literal[False] = None,
     headers: dict[str, str] = {},
-    skip_infer_table_types: list[str] = [],
+    skip_infer_table_types: list[str] = ["pdf", "jpg", "png", "heic"],
```
MalteHB, line 144: You should really never do this. Initializing mutable default parameters, such as `skip_infer_table_types`, with a mutable data type (e.g., list, dict, etc.) can lead to unexpected behavior. The default value is shared across all instances of the function call, meaning that any modification to this parameter within one function call will affect subsequent calls. This can introduce subtle bugs that are difficult to trace and debug. Instead, initialize the parameter with `None` and set the default value inside the function if it is `None`.

MthwRobinson: Thanks for the suggestion @MalteHB ! Opened #3063 to address.

PR #3042, opened 2024-05-16
feat(docx): add strategy parameter to DOC and ODT

scanny:
**Summary**
Because DOCX now supports the `strategy` argument to control aspects of image extraction, `partition_doc()` and `partition_odt()` will need to support it to because they delegate partitioning to `partition_docx()`. This will allow image extraction to work the same way for those two additional document-types.


Coniferish (approved):
lgtm

PR #3061, opened 2024-05-20
feat: `partiton_pdf()` set inferred elements text

christinestraub:
This PR adds the ability to fill inferred elements text from embedded text (`pdfminer`) without depending on `unstructured-inference` library. This PR is the second part of moving embedded text related code from `unstructured-inference` to `unstructured` and works together with https://github.com/Unstructured-IO/unstructured-inference/pull/349.


unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -102,6 +103,10 @@ def merge_inferred_with_extracted_layout(
     )
     from unstructured_inference.models.detectron2onnx import UnstructuredDetectronONNXModel
 
+    # If the model is a chipper model, we don't want to order the
+    # elements, as they are already ordered
+    order_elements = not hi_res_model_name.startswith("chipper")
```
MthwRobinson, line 108: Are there any other future models we'd need to skip ordering for? Key value was the one I had in mind. cc @leah1985 @LaverdeS

MthwRobinson (approved):
Just one question on if this impacts the forms model, otherwise looks good to me. No further review required from my end once we get confirmation on that, so giving it an approve!

christinestraub:
> Just one question on if this impacts the forms model, otherwise looks good to me. No further review required from my end once we get confirmation on that, so giving it an approve!

This PR is just for refactoring PR so it shouldn't impact the forms model.

unstructured/partition/pdf_image/pdfminer_processing.py
christinestraub: I only know `chipper` models as of now, but if there are any other future models, we could skip ordering for those models in a separate PR later.

MthwRobinson: 👍 - works for me, thanks!

PR #3066, opened 2024-05-21
feat: Migrate over fsspec connectors

rbiseck3:
### Description
Move over all fsspec connectors to the new framework

Variety of bug fixes found and fixed in this PR as well:
* custom json mixin being used for the enhanced dataclass would break if typing was quoted. That was fixed. A check was also added to the enhanced dataclass to prevent `InitVar` from being used in the root dataclass since this breaks serialization. 
* hashing for partitioner was using the filename of the raw file being partitioned rather than the file name of the file data generated from indexing. This means that mutliple files could result in the same partition hash when recursive flag is passed in. This was updated to use the file data file name instead.


scripts/opensearch-test-helpers/destination_connector/test-ingest-opensearch-output.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-import sys
```
ryannikolaidis, line 2: what was the deal with changes here? noticed a bug? general cleanup

test_unstructured_ingest/expected-structured-output/gcs/nested-2/stanley-cups.xlsx.json
```diff
@@ -1,96 +1,96 @@
 [
```
ryannikolaidis, line 1: why are all of the expected results getting bumped? it looks like different order of elements and datetime format? is there anything else going on? and reason for those changes?

ryannikolaidis:
Just a couple quick questions above.

scripts/opensearch-test-helpers/destination_connector/test-ingest-opensearch-output.py
rbiseck3: For some reason this was failing for me so I started tweaking it some to get some more information. In the end, seems like it was a transient issue in the CI but left my changes in here anyways.

test_unstructured_ingest/expected-structured-output/gcs/nested-2/stanley-cups.xlsx.json
rbiseck3: datetime is not being converted to an int for all fsspec connectors which lets us check if modification date from the source is newer than files that were potentially downloaded already. Also the `url` is the full remote path where the `remote_file_path` is the one passed in, so if recursive, this is the root of where the file was grabbed from. Open to either approach. Not sure why the order of elements would change, I haven't noticed that.

ryannikolaidis (approved):

PR #3074, opened 2024-05-22
chore: bump unstructured-inference 0.7.33

christinestraub:
Summary:
- bump unstructured-inference to `0.7.33`
- cut a release for `0.14.2`
- add some dependencies that previously came through from the layoutparser extras

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
-## 0.14.2-dev1
+## 0.14.2
 
 ### Enhancements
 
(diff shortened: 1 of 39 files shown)


MthwRobinson (approved):
LGTM - thanks!

MthwRobinson:
@christinestraub - Actually looks like we may need to add some dependencies (like `pytesseract` from the chipper test failure) that previously came through from the `layoutparser` extras.

PR #3081, opened 2024-05-23
feat(docx): add pluggable picture sub-partitioner

scanny:
**Summary**
Allow registration of a custom sub-partitioner that extracts images from a DOCX paragraph.

**Additional Context**
- A custom image sub-partitioner must implement the `PicturePartitionerT` interface defined in this PR. Basically have an `.iter_elements()` classmethod that takes the paragraph and generates zero or more `Image` elements from it.
- The custom image sub-partitioner must be registered by passing the class to `register_picture_partitioner()`.
- The default image sub-partitioner is `_NullPicturePartitioner` that does nothing.
- The registered picture partitioner is called once for each paragraph.


test_unstructured/partition/test_docx.py
```diff
@@ -622,6 +630,45 @@ def test_it_considers_text_inside_shapes():
...
+def test_partition_docx_uses_registered_picture_partitioner():
+    class FakeParagraphPicturePartitioner:
+        @classmethod
+        def iter_elements(
+            cls, paragraph: Paragraph, opts: DocxPartitionerOptions
+        ) -> Iterator[Image]:
+            call_hash = hashlib.sha1(f"{paragraph.text}{opts.strategy}".encode()).hexdigest()
+            yield Image(f"Image with hash {call_hash}, strategy: {opts.strategy}")
+
+    register_picture_partitioner(FakeParagraphPicturePartitioner)
```
Coniferish, line 651: It kind of bothers me about having a standalone function that alters a class method like this and not having the same pattern for de-registering the picture partitioner, but I'll defer to you on this

Coniferish (approved):
One question, but LGTM

test_unstructured/partition/test_docx.py
scanny: From the caller's perspective there is only saying "send paragraphs to this custom partitioner for image extraction". You can either register a sub-partitioner or not, and that persists until the program ends. This is the extent of the behavior we require to support "pluggable" sub-partitioners.

However, it has occurred to me that there's nothing stopping a user from "re-registering" to change the sub-partitioner more than once and that might be an interesting way to provide different behaviors. Like you could have one sub-partitioner that does OCR and another that uses a model to summarizes the image instead.

But if a caller wants to "de-register" a picture partitioner they can call `register_picture_partitioner()` with `_NullPicturePartitioner` or even `None` (although that latter value wouldn't type-check).

The short answer I suppose though is that the only behavior we're interested in at the moment is set-it-and-forget-it.

PR #3083, opened 2024-05-23
build: bump amd64 image to python 3.12

### LINKED ISSUE #3051: Update Docker images to use Python 3.12
Currently the AMD image uses Python 3.11 and the ARM image using Python 3.10. Since we support Python 3.12 as of #3047, we can now update these containers to use Python 3.12 instead. This will keep us on the latest version and reduce the risk that our build will break of Python 3.11 is dropped in `wolfi-base:latest`.

MthwRobinson:
### Summary

Closes #3051. Updates the AMD64 docker image to use Python 3.12 instead of Python 3.11. This is important for making sure the AMD64 images based on Chainguard continue to build in future, because the Chainguard latest images are updated frequently and could drop Python 3.11 support in the future.

This PR swaps from `wolfi-base` to `python:latest-dev` because building on `wolfi-base` could not install `pycocotools` with Python 3.12, though it worked for Python 3.11. We could likely explore slimming down the image by building on `wolfi-base` with a similar set of system dependencies from `python:latest-dev`.

As part of this PR, we mount the test directories instead of copying them into the Dockerfile.

### Testing

- CI, and specifically `test_dockerfile` should pass on the feature branch
- `make docker-start-dev` should work with a new container, and you should be able to process a file


Dockerfile-amd64
```diff
@@ -1,4 +1,4 @@
-FROM cgr.dev/chainguard/wolfi-base:latest
+FROM cgr.dev/chainguard/python:latest-dev
```
MthwRobinson, line 1: Building off of `wolfi-base` gave me issues related to `pycocotools`. We could probably do this off of `wolfi-base` by adding some of the `apk` packages listed [here](https://edu.chainguard.dev/chainguard/chainguard-images/reference/python/), but for this image I think `python:latest` works fine. We could likely also slim down the image size doing something like the following from the docs. Did confirm the `-dev` version didn't have any CVEs.

```
FROM cgr.dev/chainguard/python:latest-dev as builder

WORKDIR /app

COPY requirements.txt .

RUN pip install -r requirements.txt --user

FROM cgr.dev/chainguard/python:latest
```

Makefile
```diff
@@ -471,18 +471,20 @@ docker-start-bash:
 docker-start-dev:
 	docker run --rm \
 	-v ${CURRENT_DIR}:/mnt/local_unstructued \
-	-ti ${DOCKER_IMAGE}
+	-ti --entrypoint /bin/bash ${DOCKER_IMAGE}
```
MthwRobinson, line 474: @tabossert - This worked when I specified the entrypoint in `docker run`, but didn't work with `/bin/bash` as `CMD` in the `Dockerfile`. Any idea what would cause that?

Makefile
```diff
@@ -471,18 +471,20 @@ docker-start-bash:
...
-	-v ${CURRENT_DIR}/test_unstructured:/home/notebook-user/test_unstructured \
-	-v ${CURRENT_DIR}/test_unstructured_ingest:/home/notebook-user/test_unstructured_ingest \
-	$(if $(wildcard uns_test_env_file),--env-file uns_test_env_file,) \
-	$(DOCKER_IMAGE) \
-	bash -c "CI=$(CI) \
-	UNSTRUCTURED_INCLUDE_DEBUG_METADATA=$(UNSTRUCTURED_INCLUDE_DEBUG_METADATA) \
-	pytest -m 'not chipper' $(if $(TEST_FILE),$(TEST_FILE),test_unstructured)"
+		$(if $(wildcard uns_test_env_file),--env-file uns_test_env_file,) \
+		$(DOCKER_IMAGE) \
+		/home/nonroot/.local/bin/pytest -m 'not chipper' test_unstructured
```
MthwRobinson, line 481: @tabossert - And another mystery here: I added `/home/nonroot/.local/bin` to `PATH` in the `Dockerfile` but it didn't find it here without the explicit path. Feels like its related to the other comment above but can't figure out why.

Dockerfile-amd64
```diff
@@ -7,10 +7,9 @@ USER root
 COPY ./docker-packages/*.apk packages/
 COPY ./requirements requirements/
 COPY unstructured unstructured
-COPY test_unstructured test_unstructured
```
MthwRobinson, line 10: Moved to mounting this directory for the tests

MthwRobinson:
Going to close this for, but can pick back up later if `wolfi-base` drops Python 3.11 support

PR #3086, opened 2024-05-23
Fix: Chroma Upsert instead of Add

potter-potter:
Thanks to @0xjgv we have upserting instead of adding in Chroma. This will prevent duplicate embeddings.

Also including a huggingface example. We had examples for all the other embedders.


ryannikolaidis:
can the huggingface example be a separate PR? it's entirely unrelated

ryannikolaidis (approved):
looks good!

PR #3090, opened 2024-05-23
fix: decide table extraction

christinestraub:
This PR aims to add backward compatibility for the deprecated `pdf_infer_table_structure` parameter. A missing part of turning table extraction for PDFs and Images off by default in https://github.com/Unstructured-IO/unstructured/pull/3035, which was turned on in https://github.com/Unstructured-IO/unstructured/pull/2588.


MthwRobinson (approved):
LGTM - thanks

PR #3092, opened 2024-05-23
fix: add missing params to ElementMetadata

scanny:
A couple of parameters needed for DOCX image extraction were not added as parameters to the `ElementMetadata` constructor when they were added as known fields.

Also repair a couple gaps in alphabetical ordering cause by recent additions.


Coniferish (approved):
lgtm

PR #3047, opened 2024-05-17
improve: add Python 3.12 support (#3033)

MthwRobinson:
### Summary

Closes #2959. Updates the dependency and CI to add support for Python 3.12. 

The MongoDB ingest tests were disabled due to jobs like [this one](https://github.com/Unstructured-IO/unstructured/actions/runs/9133383127/job/25116767333) failing due to issues with the `bson` package. `bson` is a dependency for the AstraDB connector, but `pymongo` does not work when `bson` is installed from `pip`. This issue is documented by MongoDB [here](https://pymongo.readthedocs.io/en/stable/installation.html). Spun off #3049 to resolve this. Issue seems unrelated to Python 3.12, though unsure why this didn't surface previously.

Disables the `argilla` tests because `argilla` does not yet support Python 3.12. We can add the `argilla` tests back in once the PR references below is merged. You can still use the `stage_for_argilla` function if you're on `python<3.12` and you install `argilla` yourself.
- https://github.com/argilla-io/argilla/pull/4837

PR #3104, opened 2024-05-28
fix: uninstall bson for mongo connector

### LINKED ISSUE #3049: fix `bson` so MongoDB and AstraDB dependencies are compatible
Mongo ingest jobs like [this one](https://github.com/Unstructured-IO/unstructured/actions/runs/9133383127/job/25116767333) are failing because `pymongo` ships with its own version of `bson`, and install `bson` from `pip` breaks `pymongo`, as documented by MongoDB [here](https://pymongo.readthedocs.io/en/stable/installation.html). `bson` is a dependency of AstraDB, as seen in the Astra [requirements file](https://github.com/Unstructured-IO/unstructured/blob/main/requirements/ingest/astra.txt#L13-L14). The goal of this issue is to figure out a way to make these two sets of dependencies compatible, then reenable the MongoDB tests in [test-ingest-src.sh](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-src.sh) and [test-ingest-dest.sh](https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/test-ingest-dest.sh)

MthwRobinson:
### Summary

Closes #3049. Reenables the MongoDB connector test, which was disabled previously in #3047 due to incompatibility between the `pymongo` and the `bson` package from `pip`, which is a dependency for the Astra connector. Per the `pymongo` docs below, `pymongo` ships with its own version of `bson` and installing `bson` from `pip` breaks `pymongo`.

- https://pymongo.readthedocs.io/en/stable/installation.html

### Testing

Ingest tests ran successfully for the [source connector](https://github.com/Unstructured-IO/unstructured/actions/runs/9273154676/job/25512636315) and the [destination connector](https://github.com/Unstructured-IO/unstructured/actions/runs/9273154676/job/25512635546).


scanny (approved):
LGTM

PR #3107, opened 2024-05-28
rfctr: clean MSG partitioner and tests as prep

scanny:
**Summary**
Fix type errors and generally prepare `partition_msg()` and its tests for refactoring to use `python-oxmsg` library instead of the problematic `msg_parser` library for partitioning Outlook MSG files.


Coniferish (approved):
I like the asserts in unstructured/partition/msg.py
LGTM

PR #3110, opened 2024-05-29
build: pin python-docx

qued:
Since we incorporate a newer feature from `python-docx` [here](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/docx.py#L521), we should make the version of `python-docx` that first supports that method an explicit requirement.

I didn't pip recompile since our generated dependencies already have `python-docx==1.1.2`, but I can do that if someone thinks it's necessary.


scanny (approved):
LGTM

PR #3120, opened 2024-05-30
docs: cleanup readme; add python 3.12

MthwRobinson:
### Summary

Updates documentation references in the README to point to https://docs.unstructured.io and cleans up a few sections of the README. Specifically:

- Removes an old API announcement
- Removes the section mentioning Chipper as a beta feature. Chipper is only available through the SaaS API.

Also adds a Python 3.12 tag to `setup.py` since we now support Python 3.12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,9 +182,7 @@ that are helpful for new users to review:
 
 
 ### PDF Document Parsing Example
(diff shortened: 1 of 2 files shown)


MKhalusova (approved):
LGTM, we should also update the new docs wrt to Chipper. I've seen it mentioned as being in beta, and in the OS docs somewhere. 
Can you do this or would you like me to do it? 
Approving this PR, as the updates to the new docs will have to be in another repo.

MthwRobinson:
@MKhalusova - Yeah let's update that over there as well. I can do those updates.

PR #3142, opened 2024-06-03
fix(msg): use python-oxmsg for MSG email parsing

### LINKED ISSUE #2481: Garbled Text When Using Outlook Files in *.msg Format(non-unicode)
image

When saving a file using Outlook and selecting the (*.msg) format instead of Unicode format, the text loaded by UnstructuredFileLoader will appear as garbled characters.

Like this:
```
[Document(page_content='\u0a0d\u0a0d톰宥\ueab8\ue6ae\u0a0dꆬ쪰\uedb7\ue9a4ㄨ㌱约ꐲ㋫\ue9a4꘩쉢ꖾꙂ쎳꾺뫇Ꟗ꩑ꓷꖧ외끗ꗏ엾ꛩꑐꆯൃഊ\u200a\u0a0d\u0a0d\ue2a9謁\ue8a4ꆦ\u0a0d\u0a0d떤约亱쒱캥疡욼\ueca6人涱皡䆡䎨䢤튬뎦䶱疡\ue2a9謁잧릸皡䆡人涱ꆬ쪰亱\uf3a9箲\uf5b3\ue2a9墥튩뎦꒤謁잧릸䎡\u0a0d\u0a0d\u0a0d\u0a0d\ue2a9謁잧릸撬\u0a0d펭䢤䶱疡\ue2a9謁잧릸皡亱\uf3a9ꐱ㋫ꐹ⣩䂤ꤩ곳낡뫊꿴뚸\ua97d곱롤ꇟൃഊ\u200a\u0a0d\u0a0d꒤謁\ue2bb直\ueab8\ue6ae\u0a0d꒤謁傦꾤랶\uf3a9炤约嶩ꆬ쪰늵\uf4a7斫䆡첾뮥䢤퇃侧틃\uf3a9箲\uf5b3疡涱纫䦧ꮴ䊳皡뇃\ue2bb䆡侹즮\uf8b5傦\uf1a9\uf3b1䎡\u0a0d䂡낡붤墥傦꾤亱侫撯謁떶䆡璥澵熳뺪\ue2bb直䎡\u0a0d\u0a0d䂡낡瞩솴疸떹䢤ﮭꎤ\ue3a8ꆬ쪰\uedb7톤疡厯侧宥墽謁떶皡Ꞥ\ue2a9謁\ueab8\ue6ae䎡\u0a0d\u0a0dഠഊ갊낡곊뢢ෟ먊꧖띥ꩼ끁뇈뵍ㅵ㠸㠸\ue0c2ഴഊ', metadata={'source': 'C:\\Users\\10908306\\Desktop\\email格式\\Yearend_party_notUnicode.msg', 'file_directory': 'C:\\Users\\10908306\\Desktop\\email格式', 'filename': 'Yearend_party_notUnicode.msg', 'languages': ['kor'], 'sent_to': ['䡗⁑敗晬牡⽥䡗⽑楗瑳潲 (None)'], 'subject': '榡溽킳炤约嶩檡〲㌲夠慥\u2d72湥\u2064慐瑲⁹人涱ꆬ쪰\uecbf殪䎡', 'filetype': 'application/vnd.ms-outlook', 'category': 'Title'})]
```
 However, when using the Outlook format - Unicode, the normal content can be read. 😂
The following is my code:
```
data = UnstructuredFileLoader(
            file_path="mydocument_file_path",
            mode="elements",
            content_source="text/html",
            strategy="fast",
        )

document_elements = data.load()
document_elements
```
Is there a way to read msg files in different formats?🤔

---
unstructured: 0.12.2
python: 3.11.4

### LINKED ISSUE #3006: `partition_msg` is unable to process attachments
To reproduce

```python
import traceback

filename = "example-docs/fake-email-multiple-attachments.msg"
try:
  elements = partition_msg(
    filename=filename, process_attachments=True, attachment_partitioner=partition
  )
except RuntimeError as e:
  print(e)
except:
    print("ERROR about attachments")
    traceback.print_exc()

output_filename = "msg_mul_attach.json"
elements_to_json(elements, filename=output_filename)
```

The error tha arise is:
ValueError: Invalid file /tmp/tmpo99fe8l4/Engineering Onboarding.pptx. The FileType.ZIP file type is not supported in partition.

scanny:
**Summary**
`partition_msg()` previously used the `msg_parser` library for parsing Outlook MSG email files (.msg files). The `msg_parser` library is unmaintained and has several major shortcomings such as not being able to parse MSG files with 8-bit encoded strings and not reliably extracting attachments.

Use the new and permissively licenced `python-oxmsg` library instead.

**Additional Context**
For reviewability purposes, this PR temporarily places the new `partition_msg()` implementation in `new_msg.py` and references that implementation from `msg.py`. `new_msg.py` will be renamed to `msg.py` in a closely following PR. This avoids a very messy interleaving of hunks in a diff between the old and re-written `partition_msg()` implementation.

Fixes #2481 
Fixes #3006


test_unstructured/partition/test_msg.py
```diff
@@ -121,17 +105,9 @@ def test_partition_msg_from_file_with_metadata_filename():
         assert element.metadata.filename == "test"
 
 
-def test_extract_attachment_info():
-    filename = example_doc_path("fake-email-attachment.msg")
-    attachment_info = extract_msg_attachment_info(filename)
-    assert len(attachment_info) > 0
-    assert attachment_info == ATTACH_EXPECTED_OUTPUT
```
scanny, line 128: `extract_attachment_info()` is no longer used. Those fields are directly available from `oxmsg.Attachment`.

test_unstructured/partition/test_msg.py
```diff
@@ -156,102 +132,49 @@ def test_partition_msg_from_file_exclude_metadata():
...
-    assert elements[-2].text == attachment_elements[-2].text
-    for element in elements:
-        if element.metadata.attached_to_filename is not None:
-            assert len(element.text) <= 12
-            assert len(element.text) >= 6
-
-
-def test_partition_msg_raises_with_no_partitioner():
-    with pytest.raises(ValueError):
-        partition_msg(example_doc_path("fake-email-attachment.msg"), process_attachments=True)
```
scanny, line 228: `attachment_partitioner` parameter was removed and is now unconditionally `partition()`.

test_unstructured/partition/test_msg.py
```diff
@@ -156,102 +132,49 @@ def test_partition_msg_from_file_exclude_metadata():
...
-    attachment_filename = str(tmp_path / str(ATTACH_EXPECTED_OUTPUT[0]["filename"]))
-
-    attachment_elements = partition_text(
-        filename=attachment_filename,
-        metadata_filename=attachment_filename,
-        min_partition=6,
-        max_partition=12,
-    )
-
+def test_partition_msg_can_process_attachments():
```
MthwRobinson, line 135: Do you know how this test was passing before if attachment processing was broken?

test_unstructured/partition/test_msg.py
```diff
@@ -156,102 +132,49 @@ def test_partition_msg_from_file_exclude_metadata():
...
-    expected_last_modification_date = "2022-12-16T17:04:16-05:00"
-
-    mocker.patch(
-        "unstructured.partition.msg.get_last_modified_date",
-        return_value=None,
-    )
-    mocker.patch(
-        "unstructured.partition.msg.get_last_modified_date_from_file",
-        return_value=None,
+        example_doc_path("fake-email-attachment.msg"), process_attachments=True
```
MthwRobinson, line 137: Not related to this PR per se, but I don't think there's a way to flip on attachment processing in the API right now. We may want to enable that in the API or make `process_attachments=True` the default.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+
+### Enhancements
+
+* **Use `python-oxmsg` for `partition_msg()`.** Outlook MSG emails are now partitioned using the `python-oxmsg` package which resolves some shortcomings of the prior MSG parser.
+
+### Features
+
+### Fixes
+
+* **8-bit string Outlook MSG files are parsed.** `partition_msg()` is now able to parse non-unicode Outlook MSG emails.
```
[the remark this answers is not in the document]
scanny: Yep, good call, added.

unstructured/partition/msg.py
```diff
@@ -1,239 +1,17 @@
-from __future__ import annotations
-
-import os
-import tempfile
-from typing import IO, Any, Callable, Optional
+# pyright: reportPrivateUsage=false
```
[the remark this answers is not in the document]
scanny: This suppresses a type-error on the publication of `_AttachmentPartitioner` and `_MsgPartitioner` that are private classes. The type checker (`pyright`) would otherwise rightly complain "hey, you shouldn't be publishing private classes or methods". This will go away in the next PR when we rename `partition/new_msg.py` to `partition/msg.py` and get rid of this transitional version of `msg.py`.

This pragma sometimes legitimately appears in a test module because we're testing private classes or methods, but generally should not appear in a code module. The only reason it does here is because I did this "redirect old->new" thing to keep the diffs from being scrambled.

unstructured/partition/new_msg.py
```diff
@@ -0,0 +1,309 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
scanny: This is mentioned in the PR summary text above:
> **Additional Context**
> For reviewability purposes, this PR temporarily places the new partition_msg() implementation in new_msg.py and references that implementation from msg.py. new_msg.py will be renamed to msg.py in a closely following PR. This avoids a very messy interleaving of hunks in a diff between the old and re-written partition_msg() implementation.

Since there are only a few lines of the original left, I found it a lot easier to review like : "Here's what it was, that's all gone now", and "Here's the new version". I figured other reviewers would too.

The plan is to follow this PR immediately with one that changes `new_msg.py` into `msg.py`. The reviewability concern was the only driver here.

test_unstructured/partition/test_msg.py
scanny: Hmm, good question.

The test file here: `example-docs/fake-email-attachment.msg`, has one plain-text attachment: `fake-attachment.txt`.

My hypothesis would be it would have to do with `msg_parser` improperly applying some encoding to the attachment bytes. The encoding for the HTML body for this message is UTF-8 and if the attachment was "decoded" with that codec it would definitely corrupt the attachment but leave the "PK" prefix intact, which is what we observed in #3006. In the MSG format, attachment bytes are opaque `bytes` and should be detached intact.

I didn't try to unravel the spaghetti to pinpoint the cause of corruption, figuring it wasn't going to be as simple as just submitting a fix PR to `msg_parser` :)

test_unstructured/partition/test_msg.py
scanny: Hmm, good point. Not much sense in having it if we can't turn it on :)

I'd be inclined to default it to `True` then, in this PR, and then see about following up with a set of PRs to add an API parameter to the SDKs, API, ingest, etc.

Another idea that occurs to me is we could tie this to `strategy`, which is already a supported parameter, and have `fast` skip attachments and anything else (`auto`, `hi_res`, `ocr_only` I suppose too) partition any attachments.

What do you think of that idea?

test_unstructured/partition/test_msg.py
MthwRobinson: Yeah, was mainly confused why it wasn't failing before since that was the test file that was causing the issue. Don't think we need to unravel why that was happening in `msg_parser`, but it would be nice to have a test for that file that fails on `main` but passes on this feature branch to verify we've fixed #3006. Unless you think that's captured in one of the other tests in here.

test_unstructured/partition/test_msg.py
MthwRobinson: Since this is already a big PR, let's keep the default the same for this one and we can work a separate PR to update the default for that parameter. One quirk is that `strategy` could also apply to how the attachment gets processed, and if you said `process_attachments=True` if `strategy="hi_res"`, you can never process PDF attachments with `"fast"`. We can sort that out in a follow on PR though and get input from the API crew (cc @rakepat @amanda103 )

unstructured/partition/msg.py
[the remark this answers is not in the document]
MthwRobinson: Gotcha, thanks!

unstructured/partition/new_msg.py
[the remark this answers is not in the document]
MthwRobinson: Gotcha, thanks! Shoulda read the description closer haha

test_unstructured/partition/test_msg.py
scanny: The test file that was failing was `fake-email-multiple-attachments.msg`. That one has three attachments, one of which is a PPTX and that's the one the failed. This `fake-email-attachment.msg` just has one plain-text attachment.

I can change this test to use the one with the PPTX instead, sounds like a great idea to me.

I'd have to make a different PR to add an XFAIL version if we actually wanted it on main, but if you're satisfied with me trying it on main and seeing it fail there then we can just pop it in here.

scanny: k, I went ahead and did that here: https://github.com/Unstructured-IO/unstructured/pull/3142/files?show-viewed-files=true&file-filters%5B%5D=#diff-28bb41388f1944285315370fad3dad611289870db16e7962e8c94565bdd534a1R135

MthwRobinson: Awesome, thanks!

test_unstructured/partition/test_msg.py
scanny: Hmm, yes, good point, I hadn't thought of that.

On a related note, we currently don't pass along _any_ parameters to the attachment partitioner, so there's no way to specify a non-default option.

That gap presented itself as I was doing the rewrite and I added a `.partitioning_kwargs` property to `MsgPartitionerOptions` that is passed along to `partition()` for each attachment. It's currently empty but we can populate it with whatever we think makes sense. https://github.com/Unstructured-IO/unstructured/pull/3142/files?show-viewed-files=true&file-filters%5B%5D=#diff-ff46992888c6719e6d8f804d3ed6c8c1b736be31528a359c7987191b904bf012R143-R148

We have the second PR in this series (new_msg.py -> msg.py) that I'll put up as soon as this merges, so we can add some pass-through arguments there if we want. Either way it's simple to do here, maybe not so much from the API or whatever.

`partition_msg()` really only takes `process_attachments` which no other file-format uses, and remaining "generic" parameters (e.g. `date_from_file_object`, etc.) wouldn't apply to attachments. So maybe we can just pass along `kwargs`.

But yeah, maybe we need to add a `partition_attachments: bool = False` parameter to the API, SDK, ingest, etc.

MthwRobinson (approved):
LGTM! Follow on PR for `msg.py` -> `new_msg.py` to follow :)

PR #3151, opened 2024-06-04
chore: Weaviate pyv4 example

dudanogueira:
Update Unstructured example for Weaviate, now using latest python v4 client.


MthwRobinson (approved):
LGTM! Thanks for the update.

PR #3154, opened 2024-06-05
fix API-297: List parameters incorrectly passed to API requests

ds-filipknefel:
In two places parameters passed to the python client when using either Ingest workflow and `partition_via_api` function directly we parse the parameters with list values to strings e.g. 
```python
extract_image_block_types=["image"] -> extract_image_block_types='["image"]'
```
as of now these parameters are parsed incorrectly when given as strings and correctly when given as lists.

This PR removes parsing from `PartitionConfig` and `partition_via_api`.


awalker4 (approved):

christinestraub (approved):
LGTM!

PR #3162, opened 2024-06-06
rfctr(html): improve SNR in HTMLDocument

scanny:
**Summary**
Remove dead code and organize helpers of HTMLDocument in preparation for improvements and bug-fixes to follow

### DIFF
--- a/test_unstructured/documents/test_html.py
+++ b/test_unstructured/documents/test_html.py
@@ -1,4 +1,5 @@
 # pyright: reportPrivateUsage=false
+# pyright: reportUnusedFunction=false
 
 """Test suite for `unstructured.documents.html` module."""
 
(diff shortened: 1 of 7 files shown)


test_unstructured/documents/test_html.py
```diff
@@ -287,9 +287,6 @@ def test_read_html_doc(tmp_path: pathlib.Path):
         f.write(
             "<html>\n"
             "  <body>\n"
-            "    <header>\n"
-            "      <p>Here is a header. We want to ignore anything that is in this section.</p>\n"
-            "    </header>\n"
```
Coniferish, line 292: Why remove the extra tags in the html?

scanny: This test used the dead `.doc_after_cleaners()` method. One of the "cleaners" was to remove header and footer elements. That's done a different way now, using `filter_headers_and_footers()`.

So the text "we want to ignore anything..." would be misleading now.

But the test seemed pretty good otherwise and I didn't want to ditch the whole thing, so I just removed the parts that had to do with verifying `.doc_after_cleaners()` behavior.

I could have changed the text and then changed the expected elements in the asserts, but I thought this was the best balance. Adding those wouldn't really make this general "can read an HTML doc" test more valuable as far as I could tell.

Coniferish: Where is `filter_headers_and_footers`? I'm not finding it

Coniferish (approved):

PR #3165, opened 2024-06-07
rfctr(html): drop now dead XMLDocument and Document

scanny:
**Summary**
`HTMLDocument` is the class handling the core of HTML parsing. This is critical code because 8 of the 20 file-type partitioners end up using this code (`partition_html()` + 7 brokering partitioners like EPUB, MD, and RST).

For historical reasons, `HTMLDocument` subclassed `XMLDocument` which in turn subclassed `Document`, both of which are no longer relevant and unnecessarily complicate reasoning about `HTMLDocument` behavior.

Remove that inheritance and dependency and drop both `XMLDocument` and `Document` modules which become dead code after no longer being used by `HTMLDocument`.


test_unstructured/documents/test_html.py
```diff
@@ -329,15 +328,6 @@ def test_read_html_doc(tmp_path: pathlib.Path):
     ]
 
 
-# -- HTMLDocument.from_pages() -------------------------------------------------------------------
-
-
-def test_HTMLDocument_can_construct_from_existing_pages():
-    page = Page(number=0)
-    html_document = HTMLDocument.from_pages([page])
-    assert html_document.pages == [page]
```
Coniferish, line 338: Why can this be removed?

Coniferish (approved):
LGTM

test_unstructured/documents/test_html.py
scanny: Constructing an `HTMLDocument` instance from `Page` objects was a legacy capability, probably doubtful even then because no HTML source code would then be in the object, and now only used by this single test. So this test and the alternate constructor that enabled it are no longer useful.

PR #3174, opened 2024-06-10
Feat/migrate elasticsearch src connector

rbiseck3:
### Description
Migrate elasticsearch connector with support for what used to be batch ingest docs but not it support for the download step to generate additional file data.


unstructured/ingest/v2/cli/cmds/elasticsearch.py
```diff
@@ -0,0 +1,112 @@
...
+                default=[],
+                help="If provided, will limit the fields returned by Elasticsearch "
+                "to this comma-delimited list",
+            ),
+        ]
+        return options
+
+
+@dataclass
+class ElasticsearchCliIndexerConfig(CliConfig):
```
potter-potter, line 87: Interesting that we now have an Indexer config. Cool

potter-potter (approved):
works well.

PR #3176, opened 2024-06-10
feat: Kafka source and destination connector

potter-potter:
Thanks to @tullytim we have a new Kafka source and destination connector. It also works with hosted Kafka via Confluent.

Documentation will be added to the Docs repo.


unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,314 @@
...
+
+        consumer = self.kafka_consumer
+        running = True
+
+        collected = []
+        num_messages_to_consume = self.connector_config.num_messages_to_consume
+        logger.info(f"Config set for blocking on {num_messages_to_consume} messages")
+        # Consume specified number of messages
+        while running:
+            msg = consumer.poll(timeout=1.0)
```
rbiseck3, line 166: Should this timeout be set as a variable to be passed in?

unstructured/ingest/cli/cmds/kafka.py
```diff
@@ -0,0 +1,106 @@
...
+@dataclass
+class KafkaCliConfig(SimpleKafkaConfig, CliConfig):
+    @staticmethod
+    def get_cli_options() -> t.List[click.Option]:
+        options = [
+            click.Option(
+                ["--bootstrap-server"], required=True, type=str, help="Broker server hostname"
+            ),
+            click.Option(
+                ["--port"],
```
potter-potter, line 22: @rbiseck3 How careful do we have to be with options that can be in multiple connectors? I know sometimes we have to prefix them to prevent a source connector from using the option of a destination connector that is named the same.

unstructured/ingest/connector/kafka.py
potter-potter: Added.

unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,315 @@
...
+                    # End of partition event
+                    logger.error(
+                        "%% %s [%d] reached end at offset %d\n"
+                        % (msg.topic(), msg.partition(), msg.offset())
+                    )
+            else:
+                collected.append(json.loads(msg.value().decode("utf8")))
+                if len(collected) >= num_messages_to_consume:
+                    logger.debug(f"Found {len(collected)} messages, stopping")
+                    running = False
```
rbiseck3, line 182: Don't need this line since you call `break` after `consumer.commit()`

unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def check_connection(self):
+        try:
+            self.kafka_producer
+        except Exception as e:
+            logger.error(f"failed to validate connection: {e}", exc_info=True)
+            raise DestinationConnectionError(f"failed to validate connection: {e}")
+
+    @staticmethod
+    def chunks(iterable, batch_size=100):
```
rbiseck3, line 259: You actually already have this written as a generic util [here](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/utils/data_prep.py#L5)
```python
def chunk_generator(iterable, batch_size=100)
```

unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,315 @@
...
+        """A helper function to break an iterable into chunks of size batch_size."""
+        it = iter(iterable)
+        chunk = tuple(itertools.islice(it, batch_size))
+        while chunk:
+            yield chunk
+            chunk = tuple(itertools.islice(it, batch_size))
+
+    @DestinationConnectionError.wrap
+    def upload_msg(self, batch) -> int:
+        logger.debug(f"Uploding batch: {batch}")
```
rbiseck3, line 269:
```suggestion
        logger.debug(f"Uploading batch: {batch}")
```

unstructured/ingest/connector/kafka.py
```diff
@@ -0,0 +1,315 @@
...
+        logger.info(f"Writing {len(dict_list)} documents to Kafka")
+        BATCH_SIZE = 4
+        num_uploaded = 0
+
+        """
+        if self.write_config.num_processes == 1:
+           #self.upload_msg("asdf")
+           for chunk in self.chunks(dict_list, BATCH_SIZE):
+                num_uploaded += self.upload_msg(chunk)  # noqa: E203
+        else:
```
rbiseck3, line 290: Is this meant to still be commented out? Looks like it's never using multiprocessing here.

unstructured/ingest/connector/kafka.py
potter-potter: Oh yeah. Now I remember that.

unstructured/ingest/cli/cmds/kafka.py
rbiseck3: It’ll raise an error when creating the CLI if there are any conflicting names. For now, this should be fine.

PR #3177, opened 2024-06-10
rfctr(html): promote HTMLDoc candidate methods

scanny:
**Summary**
Make `._find_articles()` and `._find_main` into `._articles` and `._main` properties on HTMLDocument, respectively.

**Additional Context**
After prior refactorings, these two functions now each require only `self` and can become `@lazyproperty`s on `HTMLDocument`. This ensures they are computed at most once. In addition, their close relationship to `HTMLDocument` is indicated by their membership as methods rather than "loose" functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.14.6-dev0
+## 0.14.6-dev1
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


Coniferish (approved):
LGTM

PR #3179, opened 2024-06-10
FIX: Use proper Astra env vars name to match internal docs

erichare:
Updates env var names to match Astra internal docs. Keeping as draft temporarily due to a 500 error which may or may not be related.


erichare:
@potter-potter I opened this pr first, before the refactored source connector PR (which i have ready) because this standardizes the environment variables to match what our other integrations do, as well as what our documentation lists as the astra variables, for the endpoint and token.

I hope i caught all the instances where this needs to change - let me know where i can open a corresponding PR for the docs  if you'd like me too.

Sorry about the confusion on this, just thought it might be easier to address it now rather than later - i left a TODO item as a comment which is that, if desired, we could update the github secret to match the env var name - but (hopefully) this PR should work without that change as is.

potter-potter:
@erichare This seems fine. And thanks for adjusting the docs in our new Docs repo. 

I'll add this to the Astra Source connector PR when it is created. Just to keep things simple.

erichare:
> @erichare This seems fine. And thanks for adjusting the docs in our new Docs repo.
> 
> I'll add this to the Astra Source connector PR when it is created. Just to keep things simple.

Thanks so much.  I’ll try to have that opened up today. Appreciate it.

PR #3180, opened 2024-06-11
rfctr(html): break coupling to DocumentLayout

scanny:
**Summary**
Remove use of `partition.common.document_to_element_list()` by `HTMLDocument`. The transitive coupling with layout-inference through this shared function have been the source of frustration and a drain on engineering time and there's no compelling reason for the two to share this code.

**Additional Context**
`partition_html()` uses `partition.common.document_to_element_list()` to get finalized elements from `HTMLDocument` (pages). This gives rise to a very nasty coupling between `DocumentLayout`, used by `unstructured_inference`, and `HTMLDocument`. `document_to_element_list()` has evolved to work for both callers, but they share very few common characteristics with each other.

This coupling is bad news for us and also, importantly, for the inference and page layout folks working on PDF and images.

Break that coupling so those inference-related functions can evolve whatever way they need to without being dragged down by legacy `HTMLDocument` connections.

The initial step is to extract a `document_to_element_list()` function of our own, getting rid of the coordinates and other
`DocumentLayout`-related bits we don't need. As you'll see in the next few PRs, all of this `document_to_element_list()` code will end up either going away or being relocated closer to where it's used in `HTMLDocument`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.14.6-dev1
+## 0.14.6-dev2
 
 ### Enhancements
 
--- a/test_unstructured/partition/test_html.py
+++ b/test_unstructured/partition/test_html.py
@@ -28,7 +28,13 @@ from unstructured.documents.elements import (
     TableChunk,
     Title,
 )
-from unstructured.documents.html_elements import HTMLTable, TagsMixin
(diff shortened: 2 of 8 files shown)


Coniferish (approved):
lgtm

PR #3183, opened 2024-06-11
3105/office image fix

### LINKED ISSUE #3105: `partition_doc` fails the first time it is run in the AMD64 container
**Describe the bug**

**To Reproduce**
- Run `make docker-build`
- Run `make docker-start-dev`
- Run the following code
```python
elements = partition_doc("example-docs/fake.doc")
```
- You'll see the error that looks like:
```python
docx.opc.exceptions.PackageNotFoundError: Package not found at '/tmp/tmpmdlglc5n/fake.docx'
```
- Run `elements = partition_doc("example-docs/fake.doc")` again. It will work and you'll see output from `elements[0].text`. This can be in the same or a new python session.

**Expected behavior**
I'd expect the command to run the first time.

**Environment Info**
This is in the AMD64 Python docker image for `unstructured==0.14.2`

MthwRobinson:
### Summary

Closes #3105. Runs `libreoffice` once in the `wolfi` image to avoid the "cold start" problem where `libreoffice` fails on the first run. Reenables a test that was turned off due to this problem. Thanks @micmarty-deepsense for the solution!

### Testing

Reenabled test should run in `test_dockerfile`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.14.6-dev1
+## 0.14.6-dev2
 
 ### Enhancements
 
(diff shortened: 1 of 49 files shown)


MthwRobinson:
Closing in favor of fixing upstream in https://github.com/Unstructured-IO/base-images/pull/22

PR #3184, opened 2024-06-11
rfctr(html): drop page concept

scanny:
**Summary**
Pagination of HTML documents is currently unused. The `Page` class and concept were deeply embedding in the legacy organization of HTML partitioning code due to the legacy `Document` (= pages of elements) domain model. Remove this concept from the code such that elements are available directly from the partitioner.

**Additional Context**
- Pagination can be re-added later if we decide we want it again. A re-implementation would be much simpler and much lower impact to the structure of the code and introduce much less additional complexity, similar to the approach we take in `partition_docx()`.

### DIFF
--- a/test_unstructured/documents/test_html.py
+++ b/test_unstructured/documents/test_html.py
@@ -6,7 +6,7 @@
 from __future__ import annotations
 
 import pathlib
-from typing import Any, cast
+from typing import Any
 
 import pytest
 from lxml import etree
@@ -356,7 +356,7 @@ def test_parses_tags_correctly(opts_args: dict[str, Any]):
     opts = HtmlPartitionerOptions(**opts_args)
     doc = HTMLDocument.load(opts)
 
-    element = cast(TagsMixin, doc.elements[0])
+    element = doc._html_elements[0]
     assert element.tag == "table"
 
 
--- a/test_unstructured/documents/test_html.py
+++ b/test_unstructured/documents/test_html.py
@@ -298,7 +298,6 @@ def test_read_html_doc(tmp_path: pathlib.Path):
             "        </tr>\n"
             "      </tbody>\n"
             "    </table>\n"
-            "    <hr>\n"
             "    <h2>A New Beginning</h2>\n"
             "    <div>Here is the start of a new page.</div>\n"
             "  </body>\n"
@@ -307,22 +306,16 @@ def test_read_html_doc(tmp_path: pathlib.Path):
 
     html_document = HTMLDocument.from_file(filename)
 
-    assert len(html_document.pages) == 2
+    assert len(html_document.pages) == 1
     assert all(isinstance(p, Page) for p in html_document.pages)
-    # --
     p = html_document.pages[0]
-    assert len(p.elements) == 5
+    assert len(p.elements) == 7
     assert p.elements == [
         Title("A Great and Glorious Section"),
         NarrativeText("Dear Leader is the best. He is such a wonderful engineer!"),
         Title("Another Magnificent Title"),
         NarrativeText("The prior element is a title based on its capitalization patterns!"),
         Table("I'm in a table"),
-    ]
-    # --
-    p = html_document.pages[1]
-    assert len(p.elements) == 2
-    assert p.elements == [
         Title("A New Beginning"),
         NarrativeText("Here is the start of a new page."),
     ]
--- a/test_unstructured/documents/test_html.py
+++ b/test_unstructured/documents/test_html.py
@@ -6,7 +6,7 @@
 from __future__ import annotations
 
 import pathlib
-from typing import cast
+from typing import Any, cast
 
 import pytest
 from lxml import etree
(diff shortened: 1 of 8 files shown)


unstructured/documents/html.py
```diff
@@ -205,17 +201,7 @@ def _parse_pages_from_element_tree(self) -> list[Page]:
...
-                if element == PageBreak(""):
-                    # -- ignore leading page-break and second and later consecutive page-breaks --
-                    if not page.elements:
-                        continue
-                    pages.append(page)
-                    page_number += 1
-                    page = Page(number=page_number)
-                else:
-                    page.elements.append(element)
+            page.elements = list(self._parse_article_to_elements(article))
```
Coniferish: Looking at the second commit.
In the first commit, you added to the `_parse_pages_from_element_tree` docstring that "A horizontal-rule (`<hr/>`) element also triggers a page break", but you're removing that functionality here (removing lines 186-188)?

test_unstructured/documents/test_html.py
```diff
@@ -785,44 +785,6 @@ def test_process_list_item_ignores_deep_divs():
...
-            "  <article>\n"
-            "    <h2>Another Wonderful Section!</h2>\n"
-            "    <p>Look at this other amazing section!</p>\n"
-            "  </article>\n"
-            "</body>\n"
-        )
-        assert len(html_document._articles) == 2
-        assert all(a.tag == "article" for a in html_document._articles)
-
-    def but_it_returns_the_root_element_when_no_articles_are_present(self):
```
Coniferish, line 815: Is the behavior tested in this test not going to stay the same?

test_unstructured/partition/test_html.py
```diff
@@ -906,4 +901,187 @@ def __init__(self, text: str, status_code: int, headers: dict[str, str] = {}):
...
+        assert opts.html_str == "<html><body><p>I just flew over the internet!</p></body></html>"
+
+    def but_it_raises_when_no_path_or_file_or_text_or_url_was_provided(
+        self, opts_args: dict[str, Any]
+    ):
+        opts = HtmlPartitionerOptions(**opts_args)
+
+        with pytest.raises(ValueError, match="Exactly one of filename, file, text, or url must be"):
+            opts.html_str
+
```
Coniferish, line 901: Should the `test_partition_html_raises_with_too_many_specified` that was removed be moved here?

unstructured/documents/html.py
Coniferish: *noting for myself to check that the docstring is updated again in a later commit.

unstructured/documents/html.py
```diff
@@ -60,8 +60,12 @@ def load(cls, opts: HtmlPartitionerOptions) -> HTMLDocument:
 
     @lazyproperty
     def elements(self) -> list[Element]:
-        """Gets all elements from pages in sequential order."""
-        return [el for page in self.pages for el in page.elements]
+        """All "regular" elements (e.g. Title, NarrativeText, etc.) parsed from document.
+
+        Elements appear in document order. Elements that appeared in a header or footer are filtered
+        out.
```
Coniferish, line 65: Are we adding back header and footer elements later?

unstructured/documents/html_elements.py
```diff
@@ -16,6 +17,23 @@
 )
 
 
+class HtmlElement(Protocol):
+    """Interface provided by HTML-specific elements like HTMLNarrativeText."""
```
Coniferish, line 21: What is this class used for?

test_unstructured/documents/test_html.py
scanny: This "returns_the_root_element" behavior is tested by `DescribeHTMLDocument.it_can_find_the_main_element_in_the_document()` and `.but_it_returns_the_root_when_no_main_element_is_present()`.

Those tests were added in the prior PR, when `main()` became `HTMLDocument._main`, which is why they don't appear as new adds in this commit.

unstructured/documents/html.py
```diff
@@ -196,20 +176,12 @@ def _parse_pages_from_element_tree(self) -> list[Page]:
...
-        for article in self._articles:
-            page.elements = list(self._parse_article_to_elements(article))
+        page = Page(number=0)
 
-            # -- create new page for each article, but not when article was empty --
-            if len(page.elements) > 0:
-                pages.append(page)
-                page_number += 1
-                page = Page(number=page_number)
+        page.elements = list(self._parse_article_to_elements(self._main))
```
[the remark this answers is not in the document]
scanny: Yeah, I did exactly that :) That method becomes `._iter_elements()`.

test_unstructured/partition/test_html.py
scanny: The "raises_with_too_many_specified" behavior is no longer present. It just uses the first one it finds in `filename`, `file`, `text`, `url` order.

If someone can show me support cases where folks are providing more than one and then being confused about the behavior we can add that back in, but my assumption is it was just speculation that such a behavior would be worthwhile and I don't believe it is.

unstructured/documents/html.py
scanny: Yep, good catch. Not sure where that came from but it's gone now :)

Coniferish (approved):
few comments, but lgtm

PR #3186, opened 2024-06-11
fix: relative path / permissions issues with v2 fsspec connectors

ryannikolaidis:
When the v2 fsspec connectors currently generate the relative path, they may introduce a path with a leading slash (this happens in the case of the Box connector, which is a subclass of fsspec). When this happens this results in the paths unintentionally being treated as absolute paths. As a result, the ingest pipeline attempts to write files to directories at root level, which in turn raises permission issues.

Note: Box expected results needed to update now that it's no longer failing.

Aside: found that our tests were unintentionally skipping `box.sh` tests because we were intending to skip `dropbox.sh` and we use regex to match if a given test is in skip tests. This adds changes to force an exact match.

## Changes

* Strip leading slashes during the creating of relative paths in fsspec connectors
* Add expected results for Box connector
* (bonus): `make tidy` altered an unrelated file by removing an unnecessary call of `pass`
* (bonus): check exact match for skipped ingest tests which fixes Box tests getting skipped

## Testing

[Tests](https://github.com/Unstructured-IO/unstructured/actions/runs/9461928289/job/26093475612#step:7:2085) for the Box connector was failing. It was accidentally getting skipped (see changes above). It is now no longer skipped and passing.


scanny (approved):
Looks fine to me :)

PR #3189, opened 2024-06-12
fix: dropbox source connector file path bugs

ryannikolaidis:
The Dropbox source connector currently raises exceptions when indexing files due to two issues: a path formatting idiosyncrasy of the Dropbox library and a divergence in the definition of the Dropbox libraries fs.info method, expecting a 'url' parameter rather than 'path'. 

## Changes

* add a `/` prefix to file path used by DropboxIndexer
* override the fsspec sterilize_info method in DropboxIndexer to call `self.fs.info` with `url` rather than `path`; to accommodate for the fact that `dropboxdrivefs` diverges with this signature
* remove `dropbox.sh` from ignored source tests
* update test fixtures (now that the dropbox connector has been fixed and not skipped)

## Testing
`dropbox.sh` source ingest test now succeeds (and is no longer ignored)


MthwRobinson (approved):
Wahoo! LGTM - test is passing now.

PR #3199, opened 2024-06-12
feat/tqdm ingest support

rbiseck3:
### Description
Add in tqdm support to show progress bar of status of each job when being run. Supported for each mode (serial, async, multiprocess). Also small timing wrapper around jobs to print out how long it took in total.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features
 
+* **tqdm ingest support** add optional flag to ingest flow to print out progress bar of each step in the process.
+
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


ryannikolaidis:
neat! should we add the option to a test for general coverage and not regressing if this is a feature we support?

rbiseck3:
@ryannikolaidis given that this change is only visual, not sure how to add this to the ingest tests and validate the log output? But I would love to have a test for it somehow.

ryannikolaidis:
> @ryannikolaidis given that this change is only visual, not sure how to add this to the ingest tests and validate the log output? But I would love to have a test for it somehow.

don't want to hold this up, though if you up for validating, something like this: 
add ` &> app.log` to the end of a ingest call. 

then just validate we see the completion for steps:
```
partition_completion_present=$(grep -o 'partition:  *100%|' app.log)
```

ryannikolaidis (approved):
dropped a thought on validation, but don't want you blocked on this.

PR #3200, opened 2024-06-12
ci: update to Node 20 actions

scanny:
**Summary**
Silence the long list of warnings we get in CI from using Node 16 actions by updating to Node 20 versions.


ryannikolaidis (approved):
wonderful, thanks for the tidying here!

PR #3203, opened 2024-06-13
Roman/dry ingest pipeline step

rbiseck3:
### Description
The main goal of this was to reduce the duplicate code that was being written for each ingest pipeline step to support async and not async functionality. 

Additional bug fixes found and fixed:
* each logger for ingest wasn't being instantiated correctly. This was fixed to instantiate in the beginning of a pipeline run as soon as the verbosity level can be determined.
* The `requires_dependencies` wrapper wasn't wrapping async functions correctly. This was fixed so that `asyncio.iscoroutinefunction()` gets trigger correctly.

### DIFF
--- a/unstructured/ingest/v2/pipeline/interfaces.py
+++ b/unstructured/ingest/v2/pipeline/interfaces.py
@@ -6,13 +6,13 @@ from dataclasses import dataclass
 from functools import wraps
 from pathlib import Path
 from time import time
-from typing import Any, Optional, TypeVar
+from typing import Any, Callable, Optional, TypeVar
 
 from tqdm import tqdm
 from tqdm.asyncio import tqdm as tqdm_asyncio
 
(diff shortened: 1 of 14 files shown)


unstructured/ingest/v2/pipeline/interfaces.py
```diff
@@ -61,7 +61,7 @@ def _wrap_mp(self, input_kwargs: dict) -> Any:
 
     def _set_log_level(self, log_level: int):
         # Set the log level for each spawned process when using multiprocessing pool
-        logger.setLevel(log_level)
+        make_default_logger(log_level)
```
potter-potter, line 64: This causes logs to be created for every process I think. Though I think it is worth it if it may surface errors better.

rbiseck3: Fixed this in the latest commit

potter-potter (approved):
looks good. tqdm is a bonus

PR #3207, opened 2024-06-13
rfctr(html): drop HTML-specific elements

scanny:
**Summary**
Remove HTML-specific element types and return "regular" elements like `Title` and `NarrativeText` from `partition_html()`.

**Additional Context**
- An aspect of the legacy HTML partitioner was the use of HTML-specific element types used to track metadata during partitioning.
- That role is no longer necessary or desireable.
- HTML-specific elements like `HTMLTitle` and `HTMLNarrativeText` were returned from partitioning HTML but also the seven other file-formats that broker partitioning to HTML (convert-to-HTML and partition_html()). This does not cause immediate breakage because these are still `Text` element subtypes, but it produces a confusing developer experience.
- Remove the prior metadata roles from HTML-specific elements and remove those element types entirely.


test_unstructured/documents/test_html.py
```diff
@@ -83,9 +77,9 @@ def test_it_can_parse_a_bare_bones_table_to_an_HTMLTable_element(opts_args: dict
     opts = HtmlPartitionerOptions(**opts_args)
     html_document = HTMLDocument.load(opts)
 
-    # -- there is exactly one element and it's an HTMLTable instance --
+    # -- there is exactly one element and it's a TMLTable instance --
```
Coniferish, line 80: Typo: "Table"

scanny: Ah, thanks! Fixed.

Coniferish (approved):
LGTM

PR #3210, opened 2024-06-14
Roman/fix ingest async connectors

rbiseck3:
### Description
Choosing to use async needs to be very careful because if a connector is set to use async, the pipeline will not fan out the inputs via multiprocessing but instead it will be limited to run in a single process under the assumption it has more benefit from async due to heavy network traffic. This means the exact same code that is not optimized for async and is blocking will force the pipeline to perform worse than simply never marking the connector to use async since the pipeline will fan that out using multiprocessing. 

All connectors and processes in the pipeline we revisited to make sure this criteria was met and updated accordingly:
* Currently the unstructured client does not support making requests async, so this was moved over to use multiprocessing
* fsspec connector was updated to use the async client from the fsspec library. This also required that the client be a `@property` fetched on demand, otherwise the client would break the multiprocessing pool since it maintains a thread lock and that can't be pickled when the fsspec connector doesn't support async.
* elasticsearch was also updated to use the async client
* weaviate only recently came out with async support in their SDK at a version that is higher than we can use in the open source repo, so a TODO was left but otherwise moved to use multiprocessing
* all underlying embedders don't use async to embedder step must be multiprocessing for now. TODO left to update underlying embedder classes to optionally support async. 
* Chunking parameters were not accurately being passed through from cli to chunker params, this was fixed


unstructured/ingest/v2/processes/connectors/elasticsearch.py
```diff
@@ -158,73 +171,76 @@ def get_identifier(self, index_name: str, record_id: str) -> str:
...
-            except Exception as e:
-                logger.error(
-                    f"failed to download from index {index_name} "
-                    f"and id {record_id} to {download_path}: {e}",
-                    exc_info=True,
-                )
-                raise SourceConnectionNetworkError(
-                    f"failed to download file {file_data.identifier}"
-                )
+        async for result in async_scan(
```
potter-potter, line 235: We now get 
```
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x13658ee60>, 271841.176414041)
```
in the logging. Not sure if we can address this. You would think it would close out the connector when it is done automatically?

rbiseck3: I'll look into this

rbiseck3: resolved in latest commits

test_unstructured_ingest/expected-structured-output/local-single-file-basic-chunking/handbook-1p.docx.json
potter-potter, line 1: Is there a good reason that these expected-structured-output files for chunking have changed so much?

potter-potter: Nevermind.!

potter-potter (approved):
looks good.

Will be interesting to see if it passes this new (2 months old) expected-text-output test.
https://github.com/Unstructured-IO/unstructured/blob/2815226b544d0accf577cdb863ad9dfd1e1d9c1e/test_unstructured_ingest/.gitignore#L3

PR #3213, opened 2024-06-14
build: pull from wolfi base image

### LINKED ISSUE #3105: `partition_doc` fails the first time it is run in the AMD64 container

MthwRobinson:
### Summary

Updates the `wolfi` image to pull from the upstream `wolfi-base` base image to avoid maintaining the base layers in both locations. Closes #3105 by pulling in the fix from upstream.

### Testing

`test_dockerfile` should continue to pass with the changes.


christinestraub (approved):
LGTM!

PR #3214, opened 2024-06-14
rfctr: chroma destination migrated to V2

potter-potter:
Moving Chroma destination to the V2 version of connectors.


unstructured/ingest/v2/processes/connectors/chroma.py
```diff
@@ -0,0 +1,284 @@
...
+            data["metadata"]["data_source"]["version"] = str(version)
+
+        if page_number := data.get("metadata", {}).get("page_number"):
+            data["metadata"]["page_number"] = str(page_number)
+
+        if regex_metadata := data.get("metadata", {}).get("regex_metadata"):
+            data["metadata"]["regex_metadata"] = str(json.dumps(regex_metadata))
+
+    @classmethod
+    def normalize_dict(cls, data: dict) -> dict:
```
rbiseck3, line 140: Makes sense why you have `conform_dict` and `normalize_dict` but maybe there's a better naming schema for these? On name alone, I can't tell how the intent is different without really looking through the code.

unstructured/ingest/v2/processes/connectors/chroma.py
```diff
@@ -0,0 +1,284 @@
...
+
+@dataclass
+class ChromaAccessConfig(AccessConfig):
+    settings: Optional[Dict[str, str]] = None
+    headers: Optional[Dict[str, str]] = None
+
+
+@dataclass
+class ChromaConnectionConfig(ConnectionConfig):
+    access_config: ChromaAccessConfig = enhanced_field(sensitive=True)
```
rbiseck3, line 48: Seeing this error when running using `python3.9`:
```
Fields with a default value must come after any fields without a default. 
```

unstructured/ingest/v2/processes/connectors/chroma.py
```diff
@@ -0,0 +1,284 @@
...
+        return False
+
+    @requires_dependencies(["chromadb"], extras="chroma")
+    def create_collection(self) -> "ChromaCollection":
+        import chromadb
+
+        if self.connection_config.path:
+            chroma_client = chromadb.PersistentClient(
+                path=self.connection_config.path,
+                settings=self.connection_config.settings,
```
rbiseck3, line 198: I don't think there's a `settings` field defined on the connection config

unstructured/ingest/v2/processes/connectors/chroma.py
```diff
@@ -0,0 +1,284 @@
...
+        else:
+            raise ValueError("Chroma connector requires either path or host and port to be set.")
+
+        collection = chroma_client.get_or_create_collection(
+            name=self.connection_config.collection_name
+        )
+        return collection
+
+    @DestinationConnectionError.wrap
+    def upsert_batch(self, batch):
```
rbiseck3, line 222: Might want to have this configurable if a user wants for an error to be thrown if it's a duplicate. Use upsert by default?

unstructured/ingest/v2/processes/connectors/chroma.py
```diff
@@ -0,0 +1,284 @@
...
+@dataclass
+class ChromaUploaderConfig(UploaderConfig):
+    batch_size: int = 100
+
+
+@dataclass
+class ChromaUploader(Uploader):
+    upload_config: ChromaUploaderConfig
+    connection_config: ChromaConnectionConfig
+    _collection: Optional["ChromaCollection"] = None
```
rbiseck3, line 183: Why are we creating a collection as part of this process? Is a collection like a table in the DB sense? If so, we should assume a user already has this before running the upload process.

unstructured/ingest/v2/processes/connectors/chroma.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def __post_init__(self):
+        self._collection = self.create_collection()
+
+    def is_async(self) -> bool:
```
rbiseck3, line 188: This is redundant since the default is already to return `False`

unstructured/ingest/v2/processes/connectors/chroma.py
rbiseck3, line 183: Should mark `field(init=False)` so that it doesn't expect this field to exist when instantiating the uploader

unstructured/ingest/v2/processes/connectors/chroma.py
potter-potter: Removed the conform dict one and replaced it with the normalize_dict one. All the metadata was a "nice to have" but just added confusion.

unstructured/ingest/v2/processes/connectors/chroma.py
potter-potter: A little over a month ago we decided that upsert was the preferred default behaviour. Passed this through Ryan and Matt. https://github.com/Unstructured-IO/unstructured/pull/2926

unstructured/ingest/v2/processes/connectors/chroma.py
potter-potter: I moved things around. Its just Client here now. 

Still allowing people to create an adhoc collection though upon upload. 

I think it would be breaking behaviour to require people to have a collection already made. Chroma allows both options.

rbiseck3 (approved):

PR #3215, opened 2024-06-15
rfctr(html): drop convert_and_partition_html()

scanny:
**Summary**
Remove `unstructured.partition.html.convert_and_partition_html()`. Move file-type conversion (to HTML) responsibility to each brokering partitioner that uses that strategy and let them call `partition_html()` for themselves with the result.

**Additional Context**

Rationale:
- `partition_html()` does not want or need to know which partitioners might broker partitioning to it.
- Different brokering partitioners have their own methods to convert their format to HTML and quirks that may be involved for their format. Avoid coupling them so they can evolve independently.
- The core of the conversion work is already encapsulated in `unstructured.partition.common.convert_file_to_html_text_using_pandoc()`.
- `convert_and_partition_html()` represents an additional brokering layer with the entailed complexities of an additional site for default parameter values to be (mis-)applied and/or dropped and is an additional location for new parameters to be added.

### DIFF
--- a/test_unstructured/partition/test_common.py
+++ b/test_unstructured/partition/test_common.py
@@ -568,6 +568,43 @@ def test_ocr_data_to_elements(
         )
 
 
+class Describe_get_last_modified:
+    """Isolated unit-tests for `unstructured.partition.common.get_last_modified()."""
+
+    def it_pulls_last_modified_from_the_filesystem_when_a_path_is_provided(
+        self, file_and_last_modified: tuple[str, str]
+    ):
+        file_path, last_modified = file_and_last_modified
+        last_modified_date = common.get_last_modified(str(file_path), None, False)
+        assert last_modified_date == last_modified
+
+    def and_it_pulls_last_modified_from_the_file_like_object_when_one_is_provided(
+        self, file_and_last_modified: tuple[str, str]
+    ):
+        file_path, last_modified = file_and_last_modified
+        with open(file_path, "rb") as f:
+            last_modified_date = common.get_last_modified(None, f, True)
+        assert last_modified_date == last_modified
+
+    def but_not_when_date_from_file_object_is_False(self, file_and_last_modified: tuple[str, str]):
+        file_path, _ = file_and_last_modified
+        with open(file_path, "rb") as f:
+            last_modified_date = common.get_last_modified(None, f, False)
+        assert last_modified_date is None
+
+    # -- fixtures --------------------------------------------------------------------------------
+
+    @pytest.fixture()
+    def file_and_last_modified(self, tmp_path: pathlib.Path) -> tuple[str, str]:
+        modified_timestamp = dt.datetime(
+            year=2024, month=6, day=14, hour=15, minute=39, second=25
+        ).timestamp()
+        file_path = tmp_path / "some_file.txt"
+        file_path.write_text("abcdefg")
+        os.utime(file_path, (modified_timestamp, modified_timestamp))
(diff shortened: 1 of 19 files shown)


test_unstructured/partition/test_common.py
```diff
@@ -568,6 +568,43 @@ def test_ocr_data_to_elements(
         )
 
 
+class Describe_get_last_modified:
+    """Isolated unit-tests for `unstructured.partition.common.get_last_modified()."""
+
+    def it_pull_last_modified_from_the_filesystem_when_a_path_is_provided(
```
Coniferish, line 574:
```suggestion
    def it_pulls_last_modified_from_the_filesystem_when_a_path_is_provided(
```

test_unstructured/partition/test_doc.py
```diff
@@ -161,61 +159,6 @@ def test_partition_doc_from_filename_prefers_metadata_last_modified_when_provide
...
-    with open(example_doc_path("simple.doc"), "rb") as f:
-        sf = tempfile.SpooledTemporaryFile()
-        sf.write(f.read())
-        sf.seek(0)
-        elements = partition_doc(file=sf, date_from_file_object=True)
-
-    assert all(e.metadata.last_modified is None for e in elements)
-
-
-def test_partition_doc_from_file_prefers_metadata_last_modified_when_provided(mocker: MockFixture):
```
Coniferish, line 202: So, these are integration tests that aren't needed any more since we have the unit tests for `get_last_modified`, right? I just want to double-check that we aren't losing coverage of any kind of conditions, though since this one and `test_partition_doc_from_file_suppresses_last_modified_from_file_by_default` sound like they're testing some extra conditions.

unstructured/partition/org.py
```diff
@@ -15,14 +17,12 @@
 def partition_org(
     filename: Optional[str] = None,
     file: Optional[IO[bytes]] = None,
```
Coniferish, line 19:
```suggestion
    *,
    file: Optional[IO[bytes]] = None,
```

unstructured/partition/rst.py
```diff
@@ -16,11 +18,8 @@
 def partition_rst(
     filename: Optional[str] = None,
     file: Optional[IO[bytes]] = None,
```
Coniferish, line 20:
```suggestion
    *,
    file: Optional[IO[bytes]] = None,
```

unstructured/partition/rtf.py
```diff
@@ -16,11 +18,8 @@
 def partition_rtf(
     filename: Optional[str] = None,
     file: Optional[IO[bytes]] = None,
```
Coniferish, line 20:
```suggestion
    *,
    file: Optional[IO[bytes]] = None,
```

Coniferish (approved):
LGTM

test_unstructured/partition/test_common.py
scanny: Done.

test_unstructured/partition/test_doc.py
scanny: That was my rationale, yes.

1. These fine-grained behaviors are not implemented here so we can't easily test them.
2. These behaviors are all now tested in `test_common.py`.

Less decisive but also a consideration are:
- These tests are repeated copypasta for almost all partitioners. We should be reducing that if we can.

unstructured/partition/rst.py
scanny: Yep, good call, changed these :)

PR #3220, opened 2024-06-17
build(deps): version bumps for 2024-06-17

MthwRobinson:
### Summary

Version bumps for the week of 2024-06-17. There is a now a pin on `numpy` due to a breaking change in the latest version that we'll need to investigate and remove in a subsequent PR.


Coniferish (approved):
LGTM

PR #3224, opened 2024-06-17
Roman/migrate es dest

rbiseck3:
### Description
Migrate elasticsearch destination connector to new v2 ingest framework

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,6 +427,29 @@ jobs:
         tesseract --version
         ./test_unstructured_ingest/test-ingest-dest.sh
 
+  test_ingest_help:
+    environment: ci
+    strategy:
+      matrix:
+        python-version: ["3.9","3.10","3.11", "3.12"]
+    runs-on: ubuntu-latest
+    needs: [setup_ingest, lint]
+    steps:
+      - uses: 'actions/checkout@v4'
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Setup virtual environment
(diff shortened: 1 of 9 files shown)


unstructured/ingest/v2/cli/cmds/elasticsearch.py
```diff
@@ -104,9 +104,56 @@ def get_cli_options() -> list[click.Option]:
         return options
 
 
+@dataclass
+class ElasticsearchCliUploadStagerConfig(CliConfig):
+    @staticmethod
+    def get_cli_options() -> list[click.Option]:
+        options = [
+            click.Option(
+                ["--index-name"],
```
potter-potter, line 113: How does this work since we already have --index-name in line 92? I guess it will work because this is part of the destination connector. But what if we are going from ES to ES?

rbiseck3: It would still exist in two different parts of the CLI call, one would be before the destination connector, one would be after. So the dictionary coming back from click wouldn't conflict.

.github/workflows/ci.yml
```diff
@@ -427,6 +427,28 @@ jobs:
         tesseract --version
         ./test_unstructured_ingest/test-ingest-dest.sh
 
+  test_ingest_help:
```
potter-potter, line 430: can you add to description what this and the test-help are and why they were added?

potter-potter (approved):
looks good. Just explain the new test-helper file in the description

.github/workflows/ci.yml
rbiseck3: Wasn't sure if I was even going to add it yet. If it actually provides some benefit, ya I'll add some docs around it.

PR #3225, opened 2024-06-17
refactor: remove download packages step

christinestraub:
This PR aims to remove the download packages step since all of that gets installed in the base images. This PR also updates the base `wolfi` image because the original base image can not be found anymore: https://github.com/Unstructured-IO/unstructured/actions/runs/9555654898/job/26339587945


MthwRobinson (approved):
LGTM! Thanks for the cleanup

PR #3238, opened 2024-06-18
fix: docker image publishing error

christinestraub:
This PR aims to fix a docker image publishing error caused by user changes when pulling the `amd64` image from the `unstructured` `wolfi-base` image. (https://github.com/Unstructured-IO/unstructured/pull/3213).


MthwRobinson (approved):
LGTM - good catch!

PR #3239, opened 2024-06-18
feat/migrate gdrive source connector

rbiseck3:
### Description
Migrate the google drive source connector over to the new v2 ingest framework and include a variety of improvements as part of the refactor:
* The ID is no longer limited to a drive id but can also be the id of a subfolder within a drive or a file directly and each case is handled appropriately
* More metadata is pulled in from google drive to enrich the partitioned elements downstream and now the modified date is being set to not reprocess if the ingest pipeline already has the file cached
* timing information is set on the file created when downloaded based on the last modified data retrieved from google drive


unstructured/ingest/v2/cli/base/cmd.py
```diff
@@ -34,6 +34,10 @@ class BaseCmd(ABC):
     def cmd_name_key(self):
         return self.cmd_name.replace("-", "_")
 
+    @property
+    def cli_cmd_name(self):
+        return self.cmd_name.replace("_", "-")
```
potter-potter, line 39: Is this in case people use a CLI arg like `--collection_name` accidentally? Or is this something that happens with every CLI arg?

rbiseck3: Nah, this is when the connector id is something like `google_drive` to make the command name `google-drive`

potter-potter (approved):
Looks good. Works well. We are gonna want some unit tests soon because there are a lot of very detailed steps.

PR #3246, opened 2024-06-19
fix: fix `IndexError` when partioning a pdf with `starting_page_number`

awalker4:
The Issue:

When extracting images from pdfs, we use the metadata page number to index into a list of the images. However, the metadata page number can now be changed via `starting_page_number`. To get the true page index, we need to subtract this value.

Testing:

Run this snippet in a python shell. Before the fix, this throws an IndexError. On this branch, it will return the elements.
```
filename = "example-docs/layout-parser-paper-with-table.pdf"
partition(filename, strategy="hi_res", extract_image_block_types=["Image", "Table"], starting_page_number=20)
```

### DIFF
--- a/test_unstructured/partition/pdf_image/test_pdf.py
+++ b/test_unstructured/partition/pdf_image/test_pdf.py
@@ -1229,16 +1229,17 @@ def test_partition_pdf_element_extraction(
                 extract_image_block_to_payload=extract_image_block_to_payload,
                 extract_image_block_output_dir=tmpdir,
             )
-            assert elements[0].metadata.page_number == 20
         else:
             with open(filename, "rb") as f:
                 elements = pdf.partition_pdf(
                     file=f,
+                    # Image extraction shouldn't break by setting this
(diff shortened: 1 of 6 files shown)


test_unstructured/partition/pdf_image/test_pdf.py
```diff
@@ -1223,6 +1223,8 @@ def test_partition_pdf_element_extraction(
         if file_mode == "filename":
             elements = pdf.partition_pdf(
                 filename=filename,
+                # Image extraction shouldn't break by setting this
+                starting_page_number=20,
```
MthwRobinson, line 1227: Could we also add a test that asserts that the resulting page number in `metadata` is correct?

MthwRobinson (changes requested):
Looks good! Just requesting one additional test assertion.

MthwRobinson (approved):
LGTM

christinestraub (approved):
LGTM!

PR #3257, opened 2024-06-20
rfctr(html): prepare for new html parser

scanny:
**Summary**
Extract as much mechanical refactoring from the HTML parser change-over into the PR as possible. This leaves the next PR focused on installing the new parser and the ingest-test impact.

**Reviewers:** Commits are well groomed and reviewing commit-by-commit is probably easier.

**Additional Context**
This PR introduces the rewritten HTML parser. Its general design is recursive, consistent with the recursive structure of HTML (tree of elements). It also adds the unit tests for that parser but it does not _install_ the parser. So the behavior of `partition_html()` is unchanged by this PR. The next PR in this series will do that and handle the ingest and other unit test changes required to reflect the dozen or so bug-fixes the new parser provides.


test_unstructured/partition/html/test_partition.py
```diff
@@ -30,7 +30,7 @@
     Title,
 )
 from unstructured.file_utils.encoding import read_txt_file
-from unstructured.partition.html import HtmlPartitionerOptions, partition_html
+from unstructured.partition.html.partition import HtmlPartitionerOptions, partition_html
```
Coniferish: Can this import statement be `from unstructured.partition.html import HtmlPartitionerOptions, partition_html` since we have the line `__all__ = ["HtmlPartitionerOptions", "partition_html"]` in `unstructured/partition/html/__init__.py`?

unstructured/partition/html/partition.py
```diff
@@ -70,9 +70,6 @@ def partition_html(
         The last modified date for the document.
     skip_headers_and_footers
         If True, ignores any content that is within <header> or <footer> tags
-    source_format
```
Coniferish: Is this param no longer applicable or is it being pass through in **kwargs?

test_unstructured/partition/html/test_partition.py
scanny: It could be. Actually, thinking about it for a minute, I think it would be better that way actually.

The import `from unstructured.partition.html import partition_html` is the one we expect end users to use. Adding the actual location is an implementation detail that is subject to change. I'll change it for `partition_html`, that way if importing from `...partition.html` stopped working for some reason (which would break user code) then the tests would let us know.

On `HtmlPartitionerOptions` I think I'll leave that in the explicit location import and remove it from `__init__.py` since we don't have any users for that yet and better not to over-publish.

unstructured/partition/html/partition.py
scanny: Once I got the code simplified, it became apparent that no code was using that argument. It's a relevant argument elsewhere, like for specifying the format you want to convert into HTML, and it may have had a metadata role at one time, but no longer. In any case, it for sure doesn't do what is says it does so I dropped it from the docstring.

test_unstructured/partition/html/test_parser.py
```diff
@@ -0,0 +1,600 @@
...
+# pyright: reportUnknownArgumentType=false
+
+"""Test suite for `unstructured.partition.html.parser` module."""
+
+from __future__ import annotations
+
+from collections import deque
+
+import pytest
+from lxml import etree
```
Coniferish, line 11: I'm getting warnings from Pylance about methods of etree being partially unknown. Do we need to add a stub file for lxml?

scanny: Hmm, that's a good point. I have `types_lxml` installed locally so I'm not getting those, but it would be better for me to just add the few stubs we need into `/typings`. I'll do that.

test_unstructured/partition/html/test_parser.py
```diff
@@ -0,0 +1,600 @@
...
+
+        e = next(pre.iter_elements())
+
+        assert e.text == "You're turning into a penguin."
+        assert e.metadata.emphasized_text_contents == ["turning"]
+        assert e.metadata.emphasized_text_tags == ["b"]
+        assert e.metadata.link_texts == ["penguin"]
+        assert e.metadata.link_urls == ["http://eie.io"]
+
+
```
Coniferish, line 331: Should there be tests for TableBlock?

test_unstructured/partition/html/test_parser.py
```diff
@@ -0,0 +1,600 @@
...
+        text, annotation = next(text_segments)
+        assert text == "rhombus"
+        assert annotation == {
+            "emphasized_text_contents": "rhombus",
+            "emphasized_text_tags": "i",
+        }
+        text, annotation = next(text_segments)
+        assert text == " pentagon"
+        assert annotation == {}
```
Coniferish, line 600: Do we need tests for the LineBreak, RemovedPhrasing, and DefaultElement classes?

unstructured/partition/html/partition.py
```diff
@@ -151,7 +157,7 @@ def encoding(self) -> str | None:
         return self._encoding
 
     @lazyproperty
-    def html_str(self) -> str:
+    def html_text(self) -> str:
```
Coniferish, line 160: Is this a public attribute that we need to tell users is changing?

Coniferish (approved):
A couple of questions still outstanding, but generally LGTM

scanny:
Thanks @Coniferish, I'll get those changes in before I merge :)

test_unstructured/partition/html/test_parser.py
scanny: k, done.

test_unstructured/partition/html/test_parser.py
scanny: `TableBlock` is essentially a transplant from `HTMLDocument`, so it still has all the tests that it did there. I purposely didn't make any changes to that implementation so I didn't see the need to add more tests.

We'll probably be back in there at some time and if so that would be a good time to translate some of those integration tests into isolated unit tests and maybe add some more.

test_unstructured/partition/html/test_parser.py
scanny: Yep, good call, added.

unstructured/partition/html/partition.py
scanny: No. `HtmlPartitionerOptions` hasn't been published or documented, so there's no contract as to its behavior. There still isn't actually. It's theoretically public but I wouldn't say it's part of the interface. As long as `partition_html()` fulfills its contract, `HtmlPartitionerOptions` is an implementation detail. Also, I can't think of what use it would be to anyone just now :)

PR #3264, opened 2024-06-20
Roman/bugfix conflicting event loop ingest

rbiseck3:
### Description
In use cases where an external system (such as code being run in a jupyter notebook) already has a running event loop, run the async code in a dedicated thread pool to not conflict with the existing event loop.

This also has a variety of fixes that were found when putting together a demo leveraging the elasticsearch destination connector

### DIFF
--- a/scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
+++ b/scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 
-import sys
+from time import sleep, time
 from typing import List
 
+import click
 from elasticsearch import Elasticsearch
 from es_cluster_config import (
     CLUSTER_URL,
@@ -12,8 +13,6 @@ from es_cluster_config import (
 )
 
 from unstructured.embed.huggingface import HuggingFaceEmbeddingConfig, HuggingFaceEmbeddingEncoder
-
-N_ELEMENTS = 1404
 
 
 def embeddings_for_text(text: str) -> List[float]:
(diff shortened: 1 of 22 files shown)


scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
```diff
@@ -39,32 +38,96 @@ def query(client: Elasticsearch, search_text: str):
...
+    print(f"Validating that the count of documents in index {INDEX_NAME} is {num_elements}")
     count = int(client.cat.count(index=INDEX_NAME, format="json")[0]["count"])
-    try:
-        assert count == N_ELEMENTS
-    except AssertionError:
-        sys.exit(
-            "Elasticsearch dest check failed:"
-            f"got {count} items in index, expected {N_ELEMENTS} items in index."
-        )
+    consistent = False
```
potter-potter, line 44: What was the motivation for this? We were getting different counts?

unstructured/ingest/v2/pipeline/steps/upload.py
```diff
@@ -44,18 +44,12 @@ def __post_init__(self):
...
-        logger.info("processing content async")
-        return asyncio.run(self._process_async(iterable=iterable))
-
+    @timed
     def __call__(self, iterable: iterable_input):
         logger.info(
             f"Calling {self.__class__.__name__} " f"with {len(iterable)} docs",  # type: ignore
         )
-        if self.process.is_async():
+        if self.process.is_async() and self.process.is_async():
```
potter-potter, line 52: This seems strange. why test for is_async twice?

scripts/elasticsearch-test-helpers/destination_connector/test-ingest-elasticsearch-output.py
rbiseck3: If there's anything surrounding chunking that affects this, it is no longer a static count. I did find that this count changed based on fixes to things like the chunker but this test is meant to isolate only the destination connector. These changes made it a bit more dynamic and based on the results that came out of the stager step in the pipeline.

unstructured/ingest/v2/pipeline/steps/upload.py
rbiseck3: Good catch, must have been a copy-paste error.

unstructured/ingest/v2/processes/partitioner.py
```diff
@@ -116,14 +116,25 @@ def create_partition_parameters(self, filename: Path) -> "PartitionParameters":
         from unstructured_client.models.shared import Files, PartitionParameters
 
         partition_request = self.config.to_partition_kwargs()
+        possible_fields = [f.name for f in fields(PartitionParameters)]
+        filtered_partition_request = {
+            k: v for k, v in partition_request.items() if k in possible_fields
+        }
+        if len(filtered_partition_request) == len(partition_request):
```
potter-potter, line 123: I don't understand this. I ran a test with 
--strategy fast \
--partition-by-api \

And got this:
`Following fields were omitted due to not being supported by the currently used unstructured client: `
with nothing past the `:`

rbiseck3: Had the boolean logic wrong, updated it from `==` to `!=` so this shouldn't show up anymore if no fields were omitted.

potter-potter (approved):
works well

PR #3267, opened 2024-06-21
Compatibility Issue with Chinese Text in Document Parsing

JIAQIA:
FYI: https://github.com/Unstructured-IO/unstructured/pull/3096

Thanks for your review.


JIAQIA:
Sorry for the delayed response. I've recently had some changes at work and have just formed a team dedicated to LLM-related development. This has caused some delays, and I couldn't address the previous issues in time.

I just finished testing on my Mac, and the test results are consistent with the main branch (possibly due to some OCR packages or other issues, even the main branch can't fully pass the tests on my Mac).

So, please review it again.

JIAQIA:
@MthwRobinson  FYI

MthwRobinson:
@JIAQIA - could you resolve the merge conflicts on the HTML docs? Once those are resolved we can get the tests running in CI.

JIAQIA:
@MthwRobinson Mth OK, merge main into this branch now.

MthwRobinson:
CI is running on #3339

examples/pgvector/pgvector.ipynb
```diff
@@ -34,7 +34,6 @@
    "execution_count": 1,
    "id": "8a538b14",
    "metadata": {},
-   "outputs": [],
```
MthwRobinson, line 37: Could you revert these example files to main if there aren't any changes? Looks like the notebooks were just restarted with no output.

unstructured/partition/text.py
```diff
@@ -257,6 +257,7 @@ def element_from_text(
     text: str,
     coordinates: Optional[tuple[tuple[float, float], ...]] = None,
     coordinate_system: Optional[CoordinateSystem] = None,
+    languages: Optional[list[str]] = None,
```
MthwRobinson, line 260: This kwarg was in some partitioning functions but not all of them before, correct? @scanny would you rather go with `**kwargs` here?

MthwRobinson:
Looks good overall, left a couple of inline comments and otherwise just waiting on CI.

unstructured/partition/text.py
scanny: I think this is fine for the time being. This argument is omissible, like `coordinates` and `coordinate_system` are, so if the caller doesn't have it they can just leave it out of the call.

But I think all the partitioners have this parameter explicitly defined because it's used in the `apply_lang_metadata()` call which I believe is global to all partitioners.

JIAQIA:
OK，I run make check and make tidy.

```bash
(unstructured) ➜  unstructured git:(feature/zh_adaptation) make tidy
ruff . --fix-only || true
warning: `ruff <path>` is deprecated. Use `ruff check <path>` instead.
autoflake --in-place .
black --line-length=100 .
All done! ✨ 🍰 ✨
640 files left unchanged.
(unstructured) ➜  unstructured git:(feature/zh_adaptation) make check
ruff check .
All checks passed!
black . --check --line-length=100
All done! ✨ 🍰 ✨
640 files would be left unchanged.
flake8 .
scripts/version-sync.sh -c \
                -f "unstructured/__version__.py" semver
From github.com:JIAQIA/unstructured
 * branch              main       -> FETCH_HEAD
Checking sed version...
Detected sed version: 4.9
version sync would make no changes to unstructured/__version__.py.
flake8 --per-file-ignores "" ./unstructured/ingest
```

Note that I improved the make check script to support lint checks on MacOS. On MacOS, gsed (installed via brew) replaces the default sed. The script includes platform checks to use gsed on MacOS and sed on Linux. Additionally, awk is used for version extraction. Preliminary tests indicate the script works correctly on both Linux and MacOS.

JIAQIA:
@MthwRobinson  May I ask if there are any other issues that I need to address? I noticed some errors in the CI/CD process, such as the chipper model not being found. However, I'm not familiar with that part of the code. If there are any errors in the CI/CD process that I need to resolve, we can discuss them here. Additionally, if there are other requirements that are not yet met for merging, please bring them up. This feature is very important to our team. Thank you.

examples/pgvector/pgvector.ipynb
Coniferish: @JIAQIA, it looks like there are still some merge conflicts that need to be resolved and the changes to the notebooks need to be reverted if there aren't any changes to them.

JIAQIA: @Coniferish Con, Could you please check again? I made some merges, but it seems that due to some features of PyCharm, the .ipynb files have minor changes even if I didn't intentionally modify them. I’ve tried to handle these changes as best as I could. If there are still any merge conflicts, I'll try to resolve them.

Thank you!

Coniferish:
@JIAQIA, You'll need to update the CHANGELOG and `__version__` in `unsructured/` for the tests. You will bump the `-devX` version up one in both files (ex. `-dev1` >> `-dev2`) and add a description of the changes in the CHANGELOG.

JIAQIA:
@Coniferish  OK, I have bump to dev10, and add a ChangeLog description

Coniferish:
> @Coniferish OK, I have bump to dev10, and add a ChangeLog description

@JIAQIA, it looks like you need to merge with main and bump this again since `main` is on [0.15.3](https://github.com/Unstructured-IO/unstructured/blob/main/CHANGELOG.md). In this circumstance, you would bump it to `0.15.4-dev0`. Note that you have to make this change in both the `CHANGELOG.md` file and `__version__.py` [file](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/__version__.py)

JIAQIA:
@Coniferish   I've updated the main branch to version 0.15.4 and bumped it to 0.15.5-dev0. However, I encountered an error in the CI/CD pipeline related to test_chipper. It seems like the test is trying to find a model, but I'm not sure how to fix it. Any ideas?

JIAQIA:
> > @Coniferish I've updated the main branch to version 0.15.4 and bumped it to 0.15.5-dev0. However, I encountered an error in the CI/CD pipeline related to test_chipper. It seems like the test is trying to find a model, but I'm not sure how to fix it. Any ideas?
> 
> Hey @JIAQIA, it seems we don't have a good process in place to resolve this for contributors, so I will create a duplicate PR with your changes as a branch from our repo so it has the appropriate keys.

@Coniferish Con .Thank you very much. Also, it seems that I couldn't find a guide for contributing code. During my recent code contribution, I encountered many issues that could have been clarified by documentation, such as upgrading the version number, or the previous problem about the use of `make tidy` to format the code, and so on.

I'm not entirely sure if such a guide exists and I just missed it, but if possible, we could maintain a "Contribution Guide" document to help everyone improve their code contribution efficiency.

unstructured/partition/text_type.py
```diff
@@ -57,6 +59,10 @@ def is_possible_narrative_text(
         If True, conducts checks that are specific to the chosen language. Turn on for more
         accurate partitioning and off for faster processing.
     """
+    if languages is None or languages == [""]:
+        languages = ["auto"]
+    if isinstance(languages, list) and "auto" in languages and text:
+        languages = detect_languages(text) or []
```
Coniferish, line 65: @JIAQIA, this section needs to change. `detect_languages` should only be called when `detect_language_per_element=True` and with this change it is being called for almost every element.

`apply_lang_metadata` is the entry point all partitioners use for language detection and it determines if language detection is done at all. How we handle language detection admittedly needs to be refactored, but detecting the language of every element would slow down the processing of all documents, which we can't do.

If we just need to modify the condition for calling `exceeds_cap_ratio` to take into account languages that don't contain capitalization, we can modify this to just check the `language` parameter and you can set that when partitioning documents in other languages. Would work for you?

JIAQIA: @Coniferish Con The issue of detect_languages being called repeatedly is indeed a problem, and I’ve also noticed that it’s called again during apply_lang_metadata. However, as you mentioned, language detection is an area that needs deep refactoring. Given the current situation, undertaking such a deep refactor might be beyond what I can handle within a reasonable timeframe. Therefore, I chose the following solution, which is based on the following considerations:

1. If the user explicitly specifies the language as "eng", the repeated call to detect_language will be skipped. detect_language is only triggered when the user sets the language to auto or "" (an empty string). It’s worth noting that using "" to represent a specific meaning is hard to understand and process, and this should also be refactored.

2. To ensure that existing users are not affected, detect_language will not be called repeatedly if set language to "eng". If we want to enable this capability, I will specify language="auto" when using Unstructured. It’s important to note that if language="auto" is specified, then repeated calls to detect_language will indeed occur.

3. As I mentioned when I first proposed this solution, it will result in repeated calls to detect_languages. I believe the way to address this issue is to pass through detect_language_per_element during the partition process. Additionally, we need to consider the different behaviors when language="auto" and detect_language_per_element=False are configured, which should be illustrated using a Cartesian product to explain the program’s behavior under different combinations. If language detection has already been completed for an element in the first half of the partition process, this information should be stored in the element’s metadata. Furthermore, we need to optimize the behavior of apply_lang_metadata so that if the element’s metadata contains language information, the detection is skipped.

However, I am currently very limited on time, so I am unable to complete step 3. Step 3 would require extensive testing and coverage, and I believe that the extent of these changes would significantly increase the difficulty of merging.

Therefore, with my changes, the capabilities will only be activated when the user specifies that the document language is a non-English language like Chinese or "auto". If the user specifies "eng" or does not specify a language, my changes will not take effect.

Additionally, I hope the Unstructured team understands that metadata at the Element level is extremely important to us. If this metadata is incorrect, it is almost unusable in most scenarios. Therefore, we hope that this initial small step can be merged first, allowing non-English-speaking countries to use it before we discuss deep optimization.

To reiterate, this change ensures that when the user sets the language to "eng", the number of detect_language calls remains the same. But the number of detect_language calls will only change when the language belongs to the "languages without capitalization" list or **is set to "auto" (but sadlly, default value is auto)**, where there may be repeated calls (depending on the detect_language_per_element setting).

P.S. Regarding your suggestion to modify the conditions in the exceeds_cap_ratio function to account for languages that do not use capitalization, wouldn’t this also require calling detect_language? My opinion is that the root of this issue lies in the fact that the language parameter is not correctly passed during the partition process, and that the func call timing and logic of storing the results of detect_language are not ideal. So, based on my current understanding, modifying exceeds_cap_ratio might also introduce potential issues down the line.

Coniferish: Hi @JIAQIA, I'm not proposing you to refactor the language behavior. That is something we would take care of ourselves, so please don't worry about that.

To your other points:
> 1. If the user explicitly specifies the language as "eng", the repeated call to detect_language will be skipped. detect_language is only triggered when the user sets the language to auto or "" (an empty string). It’s worth noting that using "" to represent a specific meaning is hard to understand and process, and this should also be refactored.

I agree that using an empty string is bad practice and something that needs to be changed. To clarify, though, it is being used to skip language detection by partitioners that depend on other partitioners (see [this comment](https://github.com/Unstructured-IO/unstructured/blob/1f8030dd0ee76389c619603197fc5f7f4e546d09/unstructured/partition/lang.py#L383))

> 2. To ensure that existing users are not affected, detect_language will not be called repeatedly if set language to "eng". If we want to enable this capability, I will specify language="auto" when using Unstructured. It’s important to note that if language="auto" is specified, then repeated calls to detect_language will indeed occur.

The default parameter for `languages` for all the document specific partitioners is `["auto"]`. Your changes would require users' behavior to change by having them passing `languages=["eng"]` if they do not want to detect the language per element.

When `language="auto"`, partitioning should not automatically detect language per element. That is the main point I am raising. There is a separate parameter, `detect_language_per_element` for controlling if language is detected per element and it should be the only parameter that has that responsibility. The `languages` parameter is provided for users to declare what language the document is in, mostly for the purposes of OCR and metadata, but it could also be used to detect the language before the element type is determined. I thought that the main issue you were trying to fix with this PR was the element type that was used for languages that do not use capitalization (reducing the number of `Title` and `UncategorizedText` elements that were detected and correctly categorizing them as `NarrativeText`). Please let me know if I'm mistaken about that.

I'm proposing that we keep your changes that make `languages` and `detect_language_per_element` accessible through the partition process. We could then use those parameters to more accurately determine if `exceeds_cap_ratio` needs to be called (which seemed to be the main change in determining the element type in this PR). Instead of relying on the default `languages=['auto']`, I'm saying that users partitioning non-English language documents could define the languages param (ex. `languages=['zho']`) and that could be used in `is_possible_narrative_text`. Or, users could leave the `languages` parameter alone and set `detect_language_per_element=True` and that could trigger `detect_languages` from within `is_possible_narrative_text`. 

> 3. As I mentioned when I first proposed this solution, it will result in repeated calls to detect_languages. I believe the way to address this issue is to pass through detect_language_per_element during the partition process. Additionally, we need to consider the different behaviors when language="auto" and detect_language_per_element=False are configured, which should be illustrated using a Cartesian product to explain the program’s behavior under different combinations. If language detection has already been completed for an element in the first half of the partition process, this information should be stored in the element’s metadata. Furthermore, we need to optimize the behavior of apply_lang_metadata so that if the element’s metadata contains language information, the detection is skipped.

Here is a table describing the current behavior:
| `detect_language_per_element` | `languages` | behavior |
|---|---|---|
| `True` | ["auto"] (default) |  Language is detected per element |
| `False` (default) | ["auto"] (default) |  Language is detected based on the text of the whole document |
| `True` | ["zho"] |  Language is detected per element (ignore `languages`) |
| `False` (default) | ["zho"] |  Language is not detected and "zho" is used |

PR #3268, opened 2024-06-21
build: switch arm64 image to wolfi-base

MthwRobinson:
### Summary

Updates the `arm64` build to use the same `Dockerfile` as `amd64`, since there are now upstream base images for `wolfi-base` for both architectures. The legacy `rockylinux-9.4` is now stashed in a subdirectory the `docker` subdirectory and is no longer built in CI, but is available is users would like to build it themselves.

Additionally, this PR includes a fix to symlink `python3` to `python3.11`, which had caused a CI failure [here](https://github.com/Unstructured-IO/unstructured/actions/runs/9619486931/job/26535697755).

BREAKING CHANGE: the `arm64` image no longer supports `.doc`, `.pptx`, or `.xls` because we do not yet have a `libreoffice` `apk` built for `wolfi-base`. We intend to address that as a follow on. All other filetypes work.

### Testing

Successfully docker builds, tests, and smoke tests for [amd64](https://github.com/Unstructured-IO/unstructured/actions/runs/9619458140/job/26535610735?pr=3268) and [arm64](https://github.com/Unstructured-IO/unstructured/actions/runs/9619458140/job/26535610341?pr=3268) on the feature branch (with publish disabled).


.github/workflows/docker-publish.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       run: |
         ARCH=$(cut -d "/" -f2 <<< ${{ matrix.docker-platform }})
         DOCKER_BUILDKIT=1 docker buildx build --platform=$ARCH --load \
```
christinestraub, line 51: Doesn't it matter --platform=$ARCH or --platform=${{ matrix.docker-platform }}?

christinestraub (approved):
LGTM! left a small comment.

.github/workflows/docker-publish.yml
MthwRobinson: Thanks - updated!

hughesadam87:
Appreciate this work - these changes to dockefile seem to improve security scans tremendously.  Is the goal here to get the dockerfile secure for vulnerability scans?

MthwRobinson:
Thanks @hughesadam87 ! Yeah the change to `wolfi` is intended to improve the security posture for our images.

hughesadam87:
> Thanks @hughesadam87 ! Yeah the change to `wolfi` is intended to improve the security posture for our images.

Ah great.  Let me ask this - your full installation depends on libreoffice. Wolfi recently started supported libreoffice.  Is the libreoffice bundled with the full install of unstructured currently (or going to) use the wofli libreoffice build?

MthwRobinson:
Coming soon! We have a PR in our base images repo that will switch to using the `libreoffice` package from the `wolfi` package manager. We'll have `libreoffice` available for the `arm64` build once we switch to the new upstream base image. Planning to have that in before the end of the week.

- https://github.com/Unstructured-IO/base-images/pull/32

PR #3269, opened 2024-06-21
OD metrics for CI

mariannaparzych:
OD metrics for CI


unstructured/metrics/object_detection.py
```diff
@@ -0,0 +1,665 @@
...
+                # Extract coordinates, confidence, and class label from each prediction
+                class_label = element["type"]
+                class_idx = class_labels.index(class_label)
+                if prediction:
+                    confidence = element["prob"]
+                    x1, y1, x2, y2 = element["bbox"]
+                    page_elements.append([x1, y1, x2, y2, confidence, class_idx])
+                else:
+                    x, y, w, h = element["bbox"]
+                    page_elements.append([class_idx, x, y, w, h])
```
pawel-kmiecik, line 128: is the order here correct? Above the coordinates are first?

unstructured/metrics/object_detection.py
```diff
@@ -0,0 +1,665 @@
...
+            predictions_data["object_detection_classes"]
+            == ground_truth_data["object_detection_classes"]
+        ), "Classes in predictions and ground truth do not match."
+        assert len(predictions_data["pages"]) == len(
+            ground_truth_data["pages"]
+        ), "Pages number in predictions and ground truth do not match."
+        for pred_page, gt_page in zip(predictions_data["pages"], ground_truth_data["pages"]):
+            assert (
+                pred_page["size"] == gt_page["size"]
+            ), "Page sizes in predictions and ground truth do not match."
```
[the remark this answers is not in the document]
mariannaparzych: Maybe... I was thinking that id pages sizes are different it indicates those files are not compatible and shouldn't be compared

unstructured/metrics/object_detection.py
```diff
@@ -0,0 +1,665 @@
...
+            for element in page["elements"]:
+                # Extract coordinates, confidence, and class label from each prediction
+                class_label = element["type"]
+                class_idx = class_labels.index(class_label)
+                if prediction:
+                    confidence = element["prob"]
+                    x1, y1, x2, y2 = element["bbox"]
+                    page_elements.append([x1, y1, x2, y2, confidence, class_idx])
+                else:
+                    x, y, w, h = element["bbox"]
```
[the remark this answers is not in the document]
mariannaparzych: gts were in x,y,w,h format. You think it should be saved into xyxy in json already

unstructured/metrics/object_detection.py
mariannaparzych: yes, in super-gradients their functions use two different formats for gt and pred

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +586,58 @@ def filter_metrics(
...
+
+    def default_agg_tsv_name(self):
+        return "aggregate-object-detection-metrics.tsv"
+
+    def _process_document(self, doc: Path) -> list:
+        doc_path = Path(doc)
+        out_filename = doc_path.stem
+        doctype = Path(out_filename).suffix[1:]
+
+        src_gt_filename = out_filename + ".json"  # TODO gt file is not ready yet
```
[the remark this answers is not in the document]
mariannaparzych: Don't understand, please give more info? I believe it is the same as in TextExtractionMetricsCalculator?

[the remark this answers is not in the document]
pawel-kmiecik: Nope - for the OD dump the structure differs (as it takes files from the analysis feature).
But should be simple to form these paths.

Screenshot 2024-06-27 at 14 49 54
Screenshot 2024-06-27 at 14 50 26

unstructured/metrics/object_detection.py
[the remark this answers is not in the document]
pawel-kmiecik: for sure, the OD result dump does it in form `x1, y1, x2, y2`

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +586,93 @@ def filter_metrics(
...
+        ground_truth_file = self.ground_truths_dir / src_gt_filename
+        if not ground_truth_file.exists():
+            logger.warning(f"Ground truth file {ground_truth_file} does not exist, skipping")
+            return None
+
+        processor = ObjectDetectionEvalProcessor.from_json_files(
+            prediction_file_path=prediction_file,
+            ground_truth_file_path=ground_truth_file,
+        )
+        metrics = processor.get_metrics()
```
pawel-kmiecik, line 650: The `get_metrics` function returns a dict with types like `tensor` and `np.array` while it's expected to be a single float value per metric (dict key).

unstructured/metrics/evaluate.py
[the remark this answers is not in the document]
mariannaparzych: I believe you set proper paths in your commit and this comment can be resolved?

unstructured/metrics/object_detection.py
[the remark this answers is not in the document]
mariannaparzych: this is changed to be consistent

pawel-kmiecik (approved):
LGTM but as I've pushed few commits - let's have a second pair of eyes to look at this.

unstructured/ingest/evaluate.py
```diff
@@ -249,6 +250,55 @@ def measure_table_structure_accuracy_command(
     )
 
 
+@main.command()
+@click.option("--output_dir", type=str, help="Directory to structured output.")
+@click.option("--source_dir", type=str, help="Directory to structured source.")
```
ds-filipknefel, line 255: Could use clicks [Path type](https://click.palletsprojects.com/en/8.1.x/api/#click.Path) here for more control over the parameter's value e.g. only directory, existance

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +604,95 @@ def filter_metrics(
...
+    def default_tsv_name(self):
+        return "all-docs-object-detection-metrics.tsv"
+
+    @property
+    def default_agg_tsv_name(self):
+        return "aggregate-object-detection-metrics.tsv"
+
+    def _find_file_in_ground_truth(self, file_stem: str):
+        for path in self._ground_truth_paths:
+            if Path(path.stem).stem == file_stem:
```
ds-filipknefel, line 641: Looks like the intention was simply `path.stem`. If that's not the case I think it may require a comment.

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +604,95 @@ def filter_metrics(
...
+    def _process_document(self, doc: Path) -> list:
+        od_dump_path = Path(doc)
+        file_stem = od_dump_path.parts[-3]
+
+        src_gt_filename = self._find_file_in_ground_truth(file_stem)
+
+        if src_gt_filename not in self._ground_truth_paths:
+            return None
+
+        doctype = Path(src_gt_filename.stem).suffix[1:]
```
ds-filipknefel, line 654: What's the reason to use the `.stem` and `Path` over `src_gt_filename.suffix`?
If there's multiple suffixes perhaps `.suffixes` would be more transparent.

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +604,95 @@ def filter_metrics(
...
+    def default_agg_tsv_name(self):
+        return "aggregate-object-detection-metrics.tsv"
+
+    def _find_file_in_ground_truth(self, file_stem: str):
+        for path in self._ground_truth_paths:
+            if Path(path.stem).stem == file_stem:
+                return path
+        return None
+
+    def _process_document(self, doc: Path) -> list:
```
ds-filipknefel, line 645: Return type hint doesn't represent often returned `None`

unstructured/metrics/evaluate.py
```diff
@@ -73,6 +74,20 @@ def __post_init__(self):
...
+        """Default name for the per-document metrics TSV file."""
+
+    @property
+    @abstractmethod
+    def default_agg_tsv_name(self):
+        """Default name for the aggregated metrics TSV file."""
+
+    @abstractmethod
+    def _generate_dataframes(self, rows: list) -> tuple[pd.DataFrame, pd.DataFrame]:
+        """Generates pandas DataFrames from the list of rows."""
```
ds-filipknefel, line 89: Can we add the info on what are these data frames? Nice to see which one is aggregate without looking into the implementation.

unstructured/ingest/evaluate.py
```diff
@@ -249,6 +250,55 @@ def measure_table_structure_accuracy_command(
...
+@click.option(
+    "--output_list",
+    type=str,
+    multiple=True,
+    help="Optional: list of selected structured output file names under the \
+        directory to be evaluate. If none, all files under directory will be used.",
```
micmarty-deepsense, line 261:
```suggestion
    help=(
        "Optional: list of selected structured output file names under the"
        "directory to be evaluated. If none, all files under directory will be used."
    ),
```

unstructured/metrics/utils.py
```diff
@@ -98,23 +98,25 @@ def _display(df):
...
     """
     Save the metrics report to tsv file. The function allows an option 1) to choose `mode`
     as `w` (write) or `a` (append) and 2) to `overwrite` the file if filename existed or not.
     """
     if mode not in ["w", "a"]:
         raise ValueError("Mode not supported. Mode must be one of [w, a].")
-    if dir and not os.path.exists(dir):
-        os.makedirs(dir)
+    if directory and not os.path.exists(directory):
+        os.makedirs(directory)
```
ds-filipknefel, line 110: As we're using `pathlib` elsewhere maybe it's a good opportunity to use it here as well over `os`.

unstructured/ingest/evaluate.py
micmarty-deepsense: typo in "evaluate" -> evaluated

ds-filipknefel (changes requested):
Mostly looking for clarification on slightly confusing code with paths, once these are addressed I'll re-review. Other suggestions optional.

unstructured/metrics/evaluate.py
ds-filipknefel: This one I care for wrt to approval

unstructured/metrics/utils.py
micmarty-deepsense, line 110: May want to use `os.makedirs` with `exist_ok=True`, or even refactor to `pathlib.Path`
```suggestion
    if directory:
        Path(directory).mkdir(exist_ok=True)
```

unstructured/metrics/object_detection.py
```diff
@@ -0,0 +1,677 @@
...
+        eval_processor = ObjectDetectionEvalProcessor.from_json_files(
+            prediction_file_path, ground_truth_file_path
+        )
+
+        metrics: ObjectDetectionEvaluation = eval_processor.get_metrics()
+        print(f"Metrics for {ground_truth_file_path.name}:")
+        from dataclasses import asdict
+
+        print(asdict(metrics))
+        print("\n")
```
mpolomdeepsense, line 677: Why not `logging`?

unstructured/metrics/object_detection.py
```diff
@@ -0,0 +1,677 @@
...
+    for prediction_file_path, ground_truth_file_path in zip(
+        prediction_file_paths, ground_truth_file_paths
+    ):
+        eval_processor = ObjectDetectionEvalProcessor.from_json_files(
+            prediction_file_path, ground_truth_file_path
+        )
+
+        metrics: ObjectDetectionEvaluation = eval_processor.get_metrics()
+        print(f"Metrics for {ground_truth_file_path.name}:")
+        from dataclasses import asdict
```
mpolomdeepsense, line 674: Not actually sure how this works in runtime but I think it would be best to either move this import outside of this loop or even to the top of the file

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +604,95 @@ def filter_metrics(
...
+        )
+        metrics = processor.get_metrics()
+
+        return [
+            src_gt_filename.stem,
+            doctype,
+            None,  # connector
+        ] + [getattr(metrics, metric) for metric in self.supported_metric_names]
+
+    def _generate_dataframes(self, rows):
```
mpolomdeepsense, line 678: No return type

unstructured/ingest/evaluate.py
pawel-kmiecik: True but this should be applied to all ingest commands then - worth a separate PR.

unstructured/metrics/evaluate.py
pawel-kmiecik: That's tricky and I think requires a comment because looks misleading.
Generally the ground truth files are json but the original extension is kept.
So for source files:
```
some_document.pdf
poster.jpg
```
ground truth files will look like:
```
some_document.pdf.json
poster.jpg.json
```

So to take real `stem` we need to apply it twice. I'll add a comment for that.

unstructured/metrics/evaluate.py
pawel-kmiecik: That needs a comment too :) 
So for ground truth file like:
```
some_document.pdf.json
```
we need to take the middle - `pdf` in this case - as a source file type. This operation supports more complicated cases like:
```
some_document.tar.gz.pdf.json
```
-> `tar.gz.pdf`
(they happen sometimes in datasets)

unstructured/metrics/object_detection.py
pawel-kmiecik: That's probably left as debug (?) - I'll compare with other calculators.

unstructured/metrics/object_detection.py
pawel-kmiecik: I didn't want to import it globally as it's only used by this example in `__main__` but true - I'll move it up outside the loop.

unstructured/metrics/object_detection.py
mariannaparzych: Yes, I used it to debug, I left it as example usage, can be removed

unstructured/metrics/evaluate.py
pawel-kmiecik: Added a descriptive docstring

unstructured/metrics/object_detection.py
pawel-kmiecik: I left the print as it's only for showup purposes.

unstructured/metrics/evaluate.py
pawel-kmiecik: I've added descriptive docstrings and comments here and there - I think it's clear now :)

ds-filipknefel (approved):
I'd put the note about stems in comment (as in line 687) rather then doc-string but LGTM

unstructured/metrics/evaluate.py
```diff
@@ -585,3 +609,130 @@ def filter_metrics(
...
+    - recall
+    - average precision (mAP)
+    It also calculates aggregated metrics.
+    """
+
+    def __post_init__(self):
+        super().__post_init__()
+        self._document_paths = [
+            path.relative_to(self.documents_dir)
+            for path in self.documents_dir.rglob("analysis/*/layout_dump/object_detection.json")
```
mpolomdeepsense, line 629: Maybe this path should be declared in some common file.

mpolomdeepsense (approved):

unstructured/metrics/evaluate.py
micmarty-deepsense: this function should return `list`, not `Optional[list]`. If something's wrong, it should raise an exception.
`_try_process_document` is the function that executes `_process_document` safely, catching exceptions and logging them

PR #3278, opened 2024-06-22
fix(auto): partition() passes strategy to DOC,ODT

scanny:
**Summary**
Remedy gap where `strategy` argument passed to `partition()` was not forwarded to `partition_doc()` or `partition_odt()` and so was not making its way to `partition_docx()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.14.9-dev4
+## 0.14.9-dev5
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


test_unstructured/partition/test_auto.py
```diff
@@ -187,7 +188,17 @@ def test_auto_partition_doc_with_file(mock_docx_document, expected_docx_elements
         PartitionStrategy.OCR_ONLY,
     ],
 )
-def test_partition_forwards_strategy_arg_to_partition_docx(request: FixtureRequest, strategy: str):
+def test_partition_forwards_strategy_arg_to_partition_docx_and_its_brokers(
```
MthwRobinson, line 191: Is there a test we can add that tests this for all brokering partitioners? So if something isn't passed we're able to catch it.

scanny: Maybe, but I think the solution is not being selective about what parameters we pass along from `partition()` to delegate partitioners. I go into more details in the write-up I sent, but there shouldn't be a good reason for us to be explicitly calling out which parameters get forwarded. We can do something like:
```python
        elements = _partition_doc(filename, file=file, **kwargs)
```
instead of what we're doing now which looks like this:
```python
        elements = _partition_doc(
            filename=filename,
            file=file,
            languages=languages,
            starting_page_number=starting_page_number,
            strategy=strategy,
            **kwargs,
        )
```

There's some work to get there, which I went into in my write-up, but the result is basically maintenance-free.

All partitioners get all arguments from the call, they pick out which ones they want and ignore the rest.

Any new parameter that a partitioner starts using will already be in its `kwargs`. All it has to do is create a parameter definition for that one and it will get populated on each call.

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.14.8-dev4
+## 0.14.8-dev5
```
MthwRobinson, line 1: We released `0.14.8` so I think you'll need to update this

MthwRobinson (approved):
LGTM

PR #3293, opened 2024-06-25
rfct [P6M]-392: OpenSearch V2 Destination Connector

potter-potter:
Migrates OpenSearch destination connector to V2. Relies a lot on the Elasticsearch connector where possible. (this is expected)


unstructured/ingest/v2/processes/connectors/opensearch.py
```diff
@@ -0,0 +1,124 @@
...
+
+@dataclass
+class OpenSearchConnectionConfig(ConnectionConfig):
+    hosts: Optional[list[str]] = None
+    username: Optional[str] = None
+    access_config: OpenSearchAccessConfig = enhanced_field(sensitive=True)
+
+    def to_dict(self, **kwargs) -> dict[str, json]:
+        d = super().to_dict(**kwargs)
+        d["use_ssl"] = self.access_config.use_ssl
```
potter-potter, line 54: @rbiseck3 is there a more elegant way to "unpack" the `access_config` since its a nested dictionary at this level?

potter-potter: I guess I can just update the dictionary with a dict comprehension.

unstructured/ingest/v2/processes/connectors/opensearch.py
```diff
@@ -0,0 +1,118 @@
...
+                elements = json.load(elements_file)
+                elements_dict.extend(elements)
+        logger.info(
+            f"writing document batches to destination"
+            f" index named {self.upload_config.index_name}"
+            f" at {self.connection_config.hosts}"
+            f" with batch size (in bytes) {self.upload_config.batch_size_bytes}"
+            f" with {self.upload_config.thread_count} (number of) threads"
+        )
+        from opensearchpy.helpers import parallel_bulk
```
rbiseck3, line 84: Given how much overlap this has with the elasticsearch connector, should we find a way to extract how the method is fetched (from the lasticsaerch sdk or the opensearch one) and have this one inherit from the elasticsearch connector?

potter-potter: We definitely could. The only holdback is due to the logging being misleading if it says Elasticsearch. 

Maybe we remove the `elasticsearch` word from the logging in the ES connector? That work? Then I can pretty much just inherit from ES.

rbiseck3: Or dynamically use the class name to populate that part of the log?

rbiseck3: Also want to call out given the logic itself is the same, I made some changes to the elasticsaerch connector logic on writing the data and by inheriting from that, we don't need to worry about copy-paste drift between the two.

potter-potter: @rbiseck3 I shifted some things. It did affect the upload stager test.

unstructured/ingest/v2/processes/connectors/opensearch.py
potter-potter: Got ya. I'll stick with what I have. That one looks less fun.

unstructured/ingest/v2/processes/connectors/elasticsearch.py
```diff
@@ -331,7 +331,13 @@ class ElasticsearchUploader(Uploader):
     upload_config: ElasticsearchUploaderConfig
     connection_config: ElasticsearchConnectionConfig
 
+    def load_parallel_bulk(self):
```
rbiseck3, line 334: This needs the `requires_dependencies` wrapper.

unstructured/ingest/v2/processes/connectors/opensearch.py
```diff
@@ -0,0 +1,82 @@
...
+    client_key: Optional[str] = None
+
+
+@dataclass
+class OpenSearchConnectionConfig(ConnectionConfig):
+    hosts: Optional[list[str]] = None
+    username: Optional[str] = None
+    access_config: OpenSearchAccessConfig = enhanced_field(sensitive=True)
+
+    def to_dict(self, **kwargs) -> dict[str, json]:
```
rbiseck3, line 49: To keep things consistent, can we use similar logic to what's done in the elastisearch connector configs for this: [elasticsearch.py#L72](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/v2/processes/connectors/elasticsearch.py#L72)

potter-potter: Definitely. But its so messy. Whats the advantage? Is Elasticsearch more picky on what is in the dictionary?

rbiseck3: Allows us to strictly type to expected value to match what's expected by the client and also log exactly what parameters are being passed in with support for sensitive fields.

unstructured/ingest/v2/processes/connectors/opensearch.py
```diff
@@ -0,0 +1,117 @@
...
+    @requires_dependencies(["opensearchpy"], extras="opensearch")
+    def get_client(self) -> "OpenSearch":
+        from opensearchpy import OpenSearch
+
+        return OpenSearch(**self.get_client_kwargs())
+
+
+@dataclass
+class OpenSearchUploader(ElasticsearchUploader):
+    connection_config: OpenSearchConnectionConfig
```
rbiseck3, line 99: Seeing this in my IDE: `Non-default argument(s) follows default argument(s) defined in 'ElasticsearchUploader' `

rbiseck3: Also missing the connector type being set which I think might be causing the above problem.

potter-potter: I couldn't seem to reproduce the error above. Even with python 9 . 
I assume its some sort of warning in your ide?
Do you see the same for the elasticsearch connector which has a default arg above non-default?

Also I did add connector_type now. Thought it might pick it up from the inheritance. But it picks up the wrong one. So thanks.

rbiseck3 (approved):

PR #3298, opened 2024-06-25
revert unstructured-client pin and make pip-compile

Coniferish:
Change unstructured-client pin to setting minimum version instead of max version and `make pip-compile`. 

Integration tests that were dependent on the old version of the client are removed. These tests should be replicated in/moved to the SDK repo(s).


test_unstructured/partition/test_api.py
```diff
@@ -73,34 +73,6 @@ def test_partition_via_api_from_filename(monkeypatch):
     assert elements[0].metadata.filetype == "message/rfc822"
 
 
-def test_partition_via_api_custom_url(monkeypatch):
-    """
-    Assert that we can specify api_url and requests are sent to the right place
-    """
```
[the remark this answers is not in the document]
Coniferish: *this test is now failing with the updated SDK client since it is raising an error about not having an api_key. When you define an api_key, another error is still raised

Coniferish, line 79: It seems to me this test should be in the SDK repos instead since it just passes `base_url` through to the SDK. This test is failing because it gets stuck in retries before failing.

test_unstructured/partition/test_api.py
```diff
@@ -202,38 +201,6 @@ def test_partition_via_api_image_block_extraction():
         assert isinstance(image_data, bytes)
 
 
-def test_partition_via_api_pass_list_type_parameters(monkeypatch):
-    mock_request = Mock(return_value=MockResponse(status_code=200))
-    monkeypatch.setattr(requests.Session, "request", mock_request)
```
Coniferish, line 235: This test was still an integration test despite the patching because it only patched the eventual request made with `requests`. The request is made by the sdk client, though, which has made a number of preprocessing changes before making the request.

in client v0.18.0:
```
def do_request():
    return client.request('POST', url, data=data, files=form, headers=headers)
```

In the current version of the client:
```
def do_request():
    nonlocal req
    try:
        req = client.prepare_request(requests_http.Request('POST', url, params=query_params, data=data, files=form, headers=headers))
        req = self.sdk_configuration.get_hooks().before_request(BeforeRequestContext(hook_ctx), req)
        http_res = client.send(req)
    except Exception as e:
        _, err = self.sdk_configuration.get_hooks().after_error(AfterErrorContext(hook_ctx), None, e)
        if err is not None:
            raise err from e
        raise e
```

test_unstructured/partition/test_api.py
MthwRobinson: Yeah that makes sense to me. You could probably update the mock response to prevent the retries, but I don't think the effort's worth and, like you said, test makes more sense on the SDK side.

test_unstructured/partition/test_api.py
```diff
@@ -214,13 +185,17 @@ def test_partition_via_api_valid_request_data_kwargs():
     assert isinstance(elements, list)
 
 
+@pytest.mark.skipif(skip_outside_ci, reason="Skipping test run outside of CI")
+@pytest.mark.skipif(skip_not_on_main, reason="Skipping test run outside of main branch")
```
MthwRobinson, line 189: Is there a reason you're skipping this test outside of `main`?

MthwRobinson (commented):
Just one question on the test where you added "skip on `main`". Otherwise looks good!

test_unstructured/partition/test_api.py
Coniferish: I saw that this decorator was used on all the tests that required an api key and added it to this one as well. It did strike me as a potential problem as well, but this is an integration test that makes a real request to the API, so I assumed that was why it was skipped on main.

Also, I met with @scanny yesterday to look over this a little and he pointed out that the mock response at the top of the file is more a fake response and we should have asserts against an actual mock in the tests, so I'm going to work on updating those before merging this, so I'm happy to make changes to these tests as well while I'm at it if we want to do something other than skipping on main.

MthwRobinson: Gotcha, if this is a test that requires an API that checks out. Just wanted to make sure we weren't disabling a test we wanted to run on branches. Thanks!

MthwRobinson (approved):
LGTM!

unstructured/ingest/pipeline/reformat/chunking.py
```diff
@@ -107,23 +107,19 @@ def chunk(self, elements_json_file: str) -> Optional[list[Element]]:
...
+                # -- error, which will be caught and logged by Chunker.run()
                 api_key=self.partition_config.api_key,  # type: ignore
                 api_url=self.partition_config.partition_endpoint,  # type: ignore
                 chunking_strategy=self.chunking_config.chunking_strategy,
                 combine_under_n_chars=self.chunking_config.combine_text_under_n_chars,
-                # -- These are not currently supported by the sdk and raise an error.
-                # -- They will need to be updated once we bump the client version.
-                # combine_text_under_n_chars=self.chunking_config.combine_text_under_n_chars,
-                # include_orig_elements=self.chunking_config.include_orig_elements,
+                combine_text_under_n_chars=self.chunking_config.combine_text_under_n_chars,
```
Coniferish, line 116: @scanny, tagging you for your awareness. Resolves this conversation a while back: https://github.com/Unstructured-IO/unstructured/pull/2905/files#r1577236582

MthwRobinson:
@Coniferish - Make sure to re-`make pip-compile` on this one since #3310 had some dependency updates

PR #3300, opened 2024-06-26
Fix not counting false negatives and false positives in table metrics

plutasnyy:
This pull request fixes counting tables metric for three cases:
- False Negatives: when table exist in ground truth but any of the predicted tables doesn't match the table, the table should count as 0 and the file should not be completely skipped (before it was np.NaN). 
- False Positives: When there is a predicted table that didn't match any ground truth table it should be counted as 0, right now it is skipped in processing (matched_indices==-1)
- The file should be completely skipped only if there is no tables in ground truth and in prediction

In short we can say that previous metric calculation didn't consider OD mistakes


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 ### Fixes
+- Fix counting false negatives and false positives in table structure evaluation
```
pawel-kmiecik, line 8: You need to bump the version to `0.14.10-dev0` and run `make version-sync`. This way  you should have 2 files to commit - `CHANGELOG.md` and `unstructured/__version__.py`

unstructured/metrics/table/table_eval.py
```diff
@@ -200,37 +200,55 @@ def process_file(self) -> TableEvaluation:
...
-        matched_indices = TableAlignment.get_table_level_alignment(
-            predicted_table_data,
-            ground_truth_table_data,
-        )
-        if matched_indices:
+        is_table_in_gt = bool(ground_truth_table_data)
+        is_table_predicted = bool(predicted_table_data)
+        if not is_table_in_gt:
+            # There is no table data in ground truth, you either got perfect score or 0
+            score = 0 if is_table_predicted else np.nan
```
pawel-kmiecik, line 207: I'm thinking if we cover that in aggregation function there:
https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/metrics/evaluate.py#L270

Probably worth checking before merging.

pawel-kmiecik (approved):
LGTM

PR #3303, opened 2024-06-26
Bugfix/ingest pipeline check

rbiseck3:
### Description
Using a `isinstance` on the destination registry mapping breaks when inheritance is used for the associated uploader types. This adds a connector type field to all uploaders so that the entry can be deterministically fetched when running check for associated stager in pipeline.


potter-potter (approved):
looks good

PR #3306, opened 2024-06-26
feat(docx): differentiate no-file from not-ZIP

scanny:
**Summary**
The `python-docx` error `docx.opc.exceptions.PackageNotFoundError` arises both when no file exists at the given path and when the file exists but is not a ZIP archive (and so is not a DOCX file).

This ambiguity is unwelcome when diagnosing the error as the two possible conditions generally indicate a different course of action to resolve the error.

Add detailed validation to `DocxPartitionerOptions` to distinguish these two and provide more precise exception messages.

**Additional Context**
- `python-pptx` shares the same OPC-Package (file) loading code used by `python-docx`, so the same ambiguity will be present in `python-pptx`.
- It would be preferable for this distinguished exception behavior to be upstream in `python-docx` and `python-pptx`. If we're willing to take the version bump it might be worth considering doing that instead.

### DIFF
--- a/unstructured/partition/docx.py
+++ b/unstructured/partition/docx.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import html
 import io
 import itertools
+import os
 import tempfile
+import zipfile
 from typing import IO, Any, Iterator, Optional, Protocol, Type
 
 # -- CT_* stands for "complex-type", an XML element type in docx parlance --
@@ -155,7 +157,7 @@ def partition_docx(
         Assign this number to the first page of this document and increment the page number from
         there.
     """
-    opts = DocxPartitionerOptions(
+    opts = DocxPartitionerOptions.load(
         date_from_file_object=date_from_file_object,
         file=file,
         file_path=filename,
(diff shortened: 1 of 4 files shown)


unstructured/partition/docx.py
```diff
@@ -358,12 +365,26 @@ def _docx_file(self) -> str | IO[bytes]:
             self._file.seek(0)
             return io.BytesIO(self._file.read())
 
-        if self._file:
-            return self._file
+        assert self._file is not None  # -- assured by `._validate()` --
```
MthwRobinson, line 368: Would we want a more informative error that the assertion error here?

MthwRobinson: I think we're okay here based on `_validate`

MthwRobinson (approved):
LGTM!

unstructured/partition/docx.py
scanny: We should never get an error here (outside of tests maybe). This assert is to indicate to the type-checker we have other ways of knowing this value is not None here, even though type-wise it is `IO[bytes] | None`.

Used as expected, a new `opts` object is constructed using `opts = DocxPartitionerOptions.load(...)`, which runs `._validate()` which would raise the appropriate detailed exception before we ever got here.

This is what the `# -- assured by ._validate() --` comment is getting at.

`partition_docx()` is the only code that constructes a new `DocxPartitionerOptions` object outside of tests and it uses `.load()` to do it (as should any other callers who do so in future), so that's why that assert would never fail during operation.

PR #3313, opened 2024-06-27
rfctr: implement mongodb v2 destination connector

vangheem:
This PR provides support for V2 mongodb destination connector.

### DIFF
--- a/unstructured/ingest/v2/processes/connectors/mongodb.py
+++ b/unstructured/ingest/v2/processes/connectors/mongodb.py
@@ -1,0 +1,141 @@
+import json
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Optional
+
+from unstructured.__version__ import __version__ as unstructured_version
+from unstructured.ingest.enhanced_dataclass import enhanced_field
+from unstructured.ingest.utils.data_prep import chunk_generator
+from unstructured.ingest.v2.interfaces import (
+    AccessConfig,
+    ConnectionConfig,
+    FileData,
+    UploadContent,
+    Uploader,
+    UploaderConfig,
+    UploadStager,
+    UploadStagerConfig,
+)
+from unstructured.ingest.v2.logger import logger
+from unstructured.ingest.v2.processes.connector_registry import (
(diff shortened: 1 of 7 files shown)


potter-potter:
For some reason we are losing "Type" in the "upload_stage" document. Its in the embedded doc. I believe this is important metadata but could be wrong.

Also, the second part of the ingest test tests against a vector search. I think the file it tests again is now in the workdir so maybe try pointing it to that. I don't think we are writing to structured-output anymore for something that is going into a database.

unstructured/ingest/v2/processes/connectors/mongodb.py
```diff
@@ -0,0 +1,158 @@
...
+if TYPE_CHECKING:
+    from pymongo import MongoClient
+
+CONNECTOR_TYPE = "mongodb"
+SERVER_API_VERSION = "1"
+
+
+@dataclass
+class MongoDBAccessConfig(AccessConfig):
+    uri: Optional[str] = enhanced_field(sensitive=True, default=None)
```
rbiseck3, line 35: Don't need this to be marked sensitive since the `access_config` itself is marked. The assumption is everything in this class is sensitive.

rbiseck3:
Once you get the CI tests passing, this migration looks pretty sound

unstructured/ingest/v2/processes/connectors/mongodb.py
```diff
@@ -32,12 +32,12 @@
...
 @dataclass
 class MongoDBAccessConfig(AccessConfig):
-    uri: Optional[str] = None
+    uri: Optional[str] = enhanced_field(sensitive=True)
 
 
 @dataclass
 class MongoDBConnectionConfig(ConnectionConfig):
-    access_config: MongoDBAccessConfig = enhanced_field(sensitive=True)
+    access_config: MongoDBAccessConfig
```
potter-potter, line 40: Seems like this _is_ supposed to be enhanced_field(sensitive=True) according to the slack discussion with Roman

vangheem: Alright, put it back!

potter-potter (approved):
good migrations

PR #3315, opened 2024-06-27
Fix blocking async interfaces for v2 ingest framework

vangheem:
Right now, in a lot of places, where something does not support async, we just call the sync version of it.

This is potentially dangerous and could cause us to have blocking code.

I propose we wrap all those calls in an executor so we don't get unexpected behavior.

A big place we're missing async is in all our file IO(we have a lot of it). We could use a package like aiofiles; however, I have a feeling there is just a lot of code that won't be coverable with async interfaces. If files are small, it might be okay.

### DIFF
--- a/unstructured/utils.py
+++ b/unstructured/utils.py
@@ -12,7 +12,7 @@ import subprocess
 import tempfile
 import threading
 from datetime import datetime
-from functools import wraps
+from functools import partial, wraps
 from itertools import combinations
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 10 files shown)


unstructured/utils.py
```diff
@@ -832,3 +832,21 @@ def cleanup_file(self):
         with self.lock:
             if os.path.exists(self.file_path):
                 os.remove(self.file_path)
+
+
+RunAsyncRReturn = TypeVar("RunAsyncRReturn")
+
+
+async def run_func_async(
+    func: Callable[..., RunAsyncRReturn], *args: Any, **kwargs: Any
```
jasonbot, line 841: It'll be nice when we get to Python versions that support generics to make this type less opaque

jasonbot (approved):

PR #3317, opened 2024-06-28
fix(docx): refine file-not-found vs not-DOCX

scanny:
**Summary**
Make small refinement to distinguishing file-not-found from not-a-DOCX error message.

**Additional Context**
Conversion of a `SpooledTemporaryFile` to `io.BytesIO` must happen (when necessary) before testing whether a file-like object is a zipfile or not.

Move that check and conversion into `._validate()` so it's done before checking that a file-like object is a zip archive.


scanny:
Turned out to be unnecessary. `zipfile.is_zipfile()` works fine on a `SpooledTemporaryFile` (even though `zipfile` won't open one). So no need to convert `SpooledTemporaryFile` in `._validate()`.

PR #3320, opened 2024-06-28
feat/singlestore dest connector

rbiseck3:
### Description
Adds [SingleStore](https://www.singlestore.com/) database destination connector with associated ingest test.


unstructured/ingest/v2/processes/connectors/singlestore.py
```diff
@@ -0,0 +1,164 @@
...
+            self.upload_config.table_name,
+            ", ".join(df.columns),
+            ", ".join(["%s"] * len(df.columns)),
+        )
+        logger.debug(f"sql statement: {stmt}")
+        df.replace({np.nan: None}, inplace=True)
+        data_as_tuples = list(df.itertuples(index=False, name=None))
+        with self.connection_config.get_connection() as conn:
+            with conn.cursor() as cur:
+                for chunk in chunk_generator(
```
potter-potter, line 144: I'd love if we changed this name from chunk_generator to something else eventually. I've talked to open source users that think it is chunking here.

rbiseck3: We can just change the name? It already existed so I'm using it, but I can see the confusion. Could call it `batch_generator`?

potter-potter:
Do we still need to add brand new connectors to Makefile and MANIFEST.in

unstructured/ingest/v2/processes/connectors/singlestore.py
vangheem: `batch_generator` seems good to me

test_unstructured_ingest/dest/singlestore.sh
```diff
@@ -0,0 +1,65 @@
...
+  singlestore \
+  --host $HOST \
+  --user $USER \
+  --password $PASSWORD \
+  --database $DATABASE \
+  --port $PORT \
+  --table-name $TABLE \
+  --drop-empty-cols
+
+expected_num_elements=$(cat "$WORK_DIR"/embed/* | jq 'length')
```
potter-potter, line 61: Are you interested in adding a vector search test at the end?

rbiseck3: Leaving that as a TODO for now since I need to do some digging into how to do similarity searches on vector columns in SingleStore.

potter-potter (approved):
super cool

PR #3323, opened 2024-06-28
rfctr: Implement SQL V2 Dest Connector

vangheem:


ahmetmeleq:
Ingest test passes locally and logs include an UploadStep (indicating the use of v2)

ahmetmeleq (approved):
LGTM, 

- the test does obtain a positive number of records inside the database
- there are some todo points discovered that is out of scope for this PR

PR #3327, opened 2024-07-01
bugfix/isolate ingest v2 dependencies

rbiseck3:
### Description
This PR handles two things:
* Exposing all the connectors via the connector registries by simply importing the connector module. This should be safe assuming all connector specific dependencies themselves are imported in the methods where they are used and wrapped in `@requires_dependencies` decorator
* Remove any import that pulls from the v2 ingest.cli package

### DIFF
--- a/unstructured/ingest/v2/cli/utils.py
+++ b/unstructured/ingest/v2/cli/utils.py
@@ -3,6 +3,7 @@ import os.path
 import sys
 from dataclasses import fields, is_dataclass
 from gettext import gettext, ngettext
+from gettext import gettext as _
 from pathlib import Path
 from typing import Any, ForwardRef, Optional, Type, TypeVar, Union, get_args, get_origin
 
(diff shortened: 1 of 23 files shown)


unstructured/ingest/v2/cli/utils.py
```diff
@@ -3,6 +3,7 @@
 import sys
 from dataclasses import fields, is_dataclass
 from gettext import gettext, ngettext
+from gettext import gettext as _
```
potter-potter, line 6: could you explain this a little more. Its also imported above this line as gettext. Also `_` traditionally means unused variable.

rbiseck3: I simply copied this from the v1 file so there was no dependency on it anymore. I didn't write this code new as part of this PR.

potter-potter (approved):
works well.

PR #3328, opened 2024-07-01
fix: update slack test to point to new channel

potter-potter:
When we switched community Slack from Paid to Free we lost the CI test bot. Also if messages delete after 90 days then our expected test data will disappear.

- created a new bot in our paid company slack (test_unstructured_ingest_bot)
- added a new private channel (test-ingest)
- invited the bot to the channel
- adjusted the end datetime of the test to cover the first few messages in the channel

Still to do:
- update the CI secrets with the new bot token
- update the LastPass with the new bot token (I don't have write access.. :(.


MthwRobinson (approved):
LGTM pending follow ups in the description!

Issue #3202, opened 2024-06-13
bug/<ocr-agent> call PartitionPdf error: no ocr_agent found


ZephryLiang:
**Describe the bug**
my code:
os.environ["OCR_AGENT"] = "unstructured.partition.utils.ocr_models.paddle_ocr.OCRAgentPaddle"
elements = partition_pdf(file=f, ocr_agent=ocr_agent,strategy='ocr_only')
error :
Environment variable OCR_AGENT must be set to an existing OCR agent module, not unstructured.partition.utils.ocr_models.paddle_ocr.OCRAgentPaddle.

**Expected behavior**
i want to extract elements from pdf, how can do this?

LiangZeFenglzf:
error message in the source code :   
 def get_agent(cls) -> OCRAgent:
        """Get the configured OCRAgent instance.

        The OCR package used by the agent is determined by the `OCR_AGENT` environment variable.
        """
        ocr_agent_cls_qname = cls._get_ocr_agent_cls_qname()
        try:
            return cls.get_instance(ocr_agent_cls_qname)
        except (ImportError, AttributeError):
            raise ValueError(
                f"Environment variable OCR_AGENT must be set to an existing OCR agent module,"
                f" not {ocr_agent_cls_qname}."
            )
what agent can i use? please!

MthwRobinson:
Closing in favor of #3187. Looks like the same issue.

christinestraub:
Hi @LiangZeFenglzf, You need to install additional dependencies to use PaddleOCR. You can use the following shell script to use those dependencies:
```
#!/usr/bin/env bash

# aarch64 requires a custom build of paddlepaddle
if [ "${ARCH}" = "aarch64" ]; then
  python3 -m pip install unstructured.paddlepaddle
else
  python3 -m pip install paddlepaddle
fi
pyth
```on3 -m pip install unstructured.paddleocr

Also, you don't need to pass the `ocr_agent` param, so
```
os.environ["OCR_AGENT"] = "unstructured.partition.utils.ocr_models.paddle_ocr.OCRAgentPaddle"
elements = partition_pdf(file=f, strategy='ocr_only')
```

LiangZeFenglzf:
soryy about this.it doesn't work

LiangZeFenglzf:
> Hi @LiangZeFenglzf, You need to install additional dependencies to use PaddleOCR. You can use the following shell script to use those dependencies:
> 
> ```
[...]

PR #3335, opened 2024-07-01
fix: exception handling for OCRAgent.get_agent()

christinestraub:
The purpose of this PR is to help investigate https://github.com/Unstructured-IO/unstructured/issues/3202.


test_unstructured/partition/utils/ocr_models/test_ocr_interface.py
```diff
@@ -39,18 +41,19 @@ def it_provides_access_to_the_configured_OCR_agent(
...
 
-        _get_ocr_agent_cls_qname_.assert_called_once_with()
-        get_instance_.assert_called_once_with("Invalid.Ocr.Agent.Qname")
+        _get_ocr_agent_cls_qname_.return_value = OCR_AGENT_TESSERACT
+        for exception_cls in [ImportError, AttributeError]:
+            with patch("importlib.import_module", side_effect=exception_cls), pytest.raises(
+                RuntimeError
+            ) as exc_info:
+                OCRAgent.get_agent()
+            assert "Could not get the OCRAgent instance" in str(exc_info.value)
```
scanny, line 56: I would consider these two distinct tests, each for a different behavior. I'd suggest breaking them apart like this:
```python
    def but_it_raises_when_the_requested_agent_is_not_whitelisted(
        self, _get_ocr_agent_cls_qname_: Mock
    ):
        _get_ocr_agent_cls_qname_.return_value = "Invalid.Ocr.Agent.Qname"
        with pytest.raises(ValueError, match="must be set to a whitelisted module"):
            OCRAgent.get_agent()

    @pytest.mark.parametrize("exception_cls", [ImportError, AttributeError])
    def and_it_raises_when_the_requested_agent_cannot_be_loaded(
        self, _get_ocr_agent_cls_qname_: Mock, exception_cls: type[Exception]
    ):
        _get_ocr_agent_cls_qname_.return_value = OCR_AGENT_TESSERACT
        with patch("importlib.import_module", side_effect=exception_cls), pytest.raises(
            RuntimeError, match="Could not get the OCRAgent instance"
        ):
            OCRAgent.get_agent()
```

One fine point in the change other than splitting the two, testing the exception message (always a good idea I think) is easy to do with the `match="..."` arg of `pytest.raises()`. Saves adding an `as exc_info` clause and the need for an extra assert. Also keeps the exception detail all on one line so it's easy to grasp at a glance.

The `match` arg is a regex. I rarely find the need for that but good to know in case you want to use anchors or placeholders :)

scanny (approved):
One suggested change, otherwise LGTM.

Approving in advance :)

christinestraub:
@scanny Thanks for your suggestion!

unstructured/partition/utils/ocr_models/ocr_interface.py
```diff
@@ -31,26 +31,27 @@ def get_agent(cls) -> OCRAgent:
...
 
     @staticmethod
     @functools.lru_cache(maxsize=None)
     def get_instance(ocr_agent_module: str) -> "OCRAgent":
         module_name, class_name = ocr_agent_module.rsplit(".", 1)
-        if module_name in OCR_AGENT_MODULES_WHITELIST:
+        if module_name not in OCR_AGENT_MODULES_WHITELIST:
+            raise ValueError(
+                f"Environment variable OCR_AGENT module name {module_name} must be set to a "
+                f"whitelisted module part of {OCR_AGENT_MODULES_WHITELIST}."
```
MthwRobinson, line 43: I like the explicit list of modules, that'll make it a lot easier for people to correct.

MthwRobinson (approved):
LGTM

PR #3341, opened 2024-07-02
Tables detection f1

plutasnyy:
This pull request add table detection metrics.

One case that was considered by me:

Case: Two tables are predicted and matched with one table in ground truth
Question: Is this matching correct in both cases or just for on table

There are two subcases:
- table was predicted by OD as two sub tables (so half in two, there are two non overlapping subtables) -> in my opinion both are correct
- it is false positive from tables matching script in get_table_level_alignment -> 1 good, 1 wrong

As we don't have bounding boxes I followed the notebook calculation script and assumed pessimistic, second subcase version


pawel-kmiecik (approved):
LGTM:
- just bump versions/changelog
- see my comment for the test that needs coordination with the latter PR.

PR #3342, opened 2024-07-02
remove unused _with_spans metric

plutasnyy:
The table metrics considering spans is not used and it messes with the output thus I have cleaned the code from it. Though,  I have left table_as_cells in the source code - it still may be useful for the users


pawel-kmiecik (approved):
LGTM, just bump versions/changelog

PR #3344, opened 2024-07-02
rfctr [P6M-398]: salesforce connector v2

potter-potter:
Updates salesforce source connector to v2.


unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+Able to download Account, Case, Campaign, EmailMessage, Lead
+Salesforce returns everything as a list of json.
+This saves each entry as a separate file to be partitioned.
+Using JWT authorization
+https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_auth_key_and_cert.htm
+https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_auth_connected_app.htm
+"""
+
+import json
+import typing as t
```
rbiseck3, line 12: I've been importing these directly rather than using t. Just to keep things consistent. I found this approach a bit more cumbersome.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+        raise ValueError("private_key does not contain PEM private key or path")
+
+
+@dataclass
+class SalesforceConnectionConfig(ConnectionConfig):
+    username: str
+    access_config: SalesforceAccessConfig = enhanced_field(sensitive=True)
+
+    @requires_dependencies(["simple_salesforce"], extras="salesforce")
+    def get_client(self):
```
rbiseck3, line 105: Can we get a return type for this method and have that be part of what's included when `TYPE_CHECKING` is enabled?

unstructured/ingest/v2/cli/cmds/salesforce.py
```diff
@@ -0,0 +1,76 @@
...
+class SalesforceCliIndexerConfig(CliConfig):
+    @staticmethod
+    def get_cli_options() -> list[click.Option]:
+        possible_categories = ["Account", "Case", "Campaign", "EmailMessage", "Lead"]
+        options = [
+            click.Option(
+                ["--categories"],
+                default=None,
+                required=True,
+                type=DelimitedString(choices=possible_categories),
```
rbiseck3, line 49: Can we import the categories straight from the connector file rather than redefine them here so there's one source of truth for that.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+            consumer_key=self.access_config.consumer_key,
+            privatekey_file=pkey_value if pkey_type is Path else None,
+            privatekey=pkey_value if pkey_type is str else None,
+            version=SALESFORCE_API_VERSION,
+        )
+
+
+@dataclass
+class SalesforceIndexerConfig(IndexerConfig):
+    categories: t.List[str]
```
rbiseck3, line 121: This might be better as a set of Literals ([[Dataclass argument choices with a default option](https://stackoverflow.com/questions/61756716/dataclass-argument-choices-with-a-default-option)](https://stackoverflow.com/questions/61756716/dataclass-argument-choices-with-a-default-option))
```suggestion
    categories: set[Literal[ACCEPTED_CATEGORIES]]
```

Not sure if this works but could be a nice check on the dataclass itself.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+        appropriate format for partitioning.
+        """
+        from simple_salesforce.exceptions import SalesforceMalformedRequest
+
+        client = self.connection_config.get_client()
+
+        files_list = []
+        for record_type in self.index_config.categories:
+            if record_type not in ACCEPTED_CATEGORIES:
+                raise ValueError(f"{record_type} not currently an accepted Salesforce category")
```
rbiseck3, line 142: Might not need this if the typing can run the check for us (comment above)

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+                document = self._eml_for_record(record)
+            else:
+                document = self._xml_for_record(record)
+            download_path = self.get_download_path(file_data=file_data)
+            download_path.parent.mkdir(parents=True, exist_ok=True)
+
+            with open(download_path, "w") as page_file:
+                page_file.write(document)
+
+            return DownloadResponse(
```
rbiseck3, line 272: You'll want to use something like this to make sure to update the local file with the modified time if it exists: [onedrive.py#L202](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/v2/processes/connectors/onedrive.py#L202). I updated this in an active PR to be part of the base downloader class.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+            )
+
+        except Exception as e:
+            logger.error(
+                f"Error while downloading and saving file: {file_data.identifier}.",
+            )
+            logger.error(e)
+
+
+add_source_entry(
```
rbiseck3, line 284: This approach was changed in a recent PR. You now define just the entry and add it to the registry in the `__init__.py` file.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+            return DownloadResponse(
+                file_data=file_data,
+                path=Path(download_path),
+            )
+
+        except Exception as e:
+            logger.error(
+                f"Error while downloading and saving file: {file_data.identifier}.",
+            )
+            logger.error(e)
```
rbiseck3, line 281: We should reraise something here so that the error gets caught intentionally rather than simply logging it and letting the process swallow it up. 

```suggestion
            logger.error(
                f"Error while downloading and saving file {file_data.identifier}: {e}",
            )
            raise e
```

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+    def get_record(self, file_data: FileData) -> OrderedDict:
+        # Get record from Salesforce based on id
+        response = self._get_response(file_data)
+        logger.debug(f"response was returned for salesforce record id: {file_data.identifier}")
+        records = response["records"]
+        if not records:
+            raise ValueError(
+                f"No record found with record id {file_data.identifier}: {json.dumps(response)}"
+            )
+        record_json = records[0]
```
rbiseck3, line 255: What happens if the length of records is greater than 1?

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+                                fullpath=f"{record['attributes']['type']}/{record['Id']}",
+                            ),
+                            metadata=DataSourceMetadata(
+                                url=record["attributes"]["url"],
+                                version=record["SystemModstamp"],
+                                date_created=record["CreatedDate"],
+                                date_modified=record["LastModifiedDate"],
+                                record_locator={"id": record["Id"]},
+                            ),
+                            additional_metadata={"record_type": record["attributes"]["type"]},
```
rbiseck3, line 165: I've been adding whatever metadata comes back from the provider that isn't already being set in the metadata here, as long as it's json serializable. You never know what might need that data downstream and it won't be added to every element, so it won't bloat the partitioned output.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+            raise MissingCategoryError(
+                f"There are no categories with the name: {record_type}",
+            )
+        return extension
+
+    def get_download_path(self, file_data: FileData) -> Path:
+        record_file = file_data.identifier + self.get_file_extension(
+            file_data.additional_metadata["record_type"]
+        )
+        return Path(self.download_dir) / file_data.additional_metadata["record_type"] / record_file
```
rbiseck3, line 206: This logic should be part of the code creating the source identifiers above (setting filename), and then that should be used when creating the download path. Otherwise the `source_identifiers.filename` isn't accurate since it doesn't contain the extension.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def _xml_for_record(self, record: OrderedDict) -> str:
+        """Creates partitionable xml file from a record"""
+        import xml.etree.ElementTree as ET
+
+        def flatten_dict(data, parent, prefix=""):
```
rbiseck3, line 212: This method already exists, could that be reused? [unstructured/staging/base.py](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/staging/base.py#L224)

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+CONNECTOR_TYPE = "salesforce"
+
+if TYPE_CHECKING:
+    pass
+
+SALESFORCE_API_VERSION = "57.0"
+
+ACCEPTED_CATEGORIES = ["Account", "Case", "Campaign", "EmailMessage", "Lead"]
+
+EMAIL_TEMPLATE = Template(
```
rbiseck3, line 58: Where did this email template come from? Is it specific to saleforce? Some documentation above this would be good.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+CONNECTOR_TYPE = "salesforce"
+
+if TYPE_CHECKING:
+    pass
+
+SALESFORCE_API_VERSION = "57.0"
+
+ACCEPTED_CATEGORIES = ["Account", "Case", "Campaign", "EmailMessage", "Lead"]
```
rbiseck3, line 56: Are these all the categories possible from salesforce or just the ones we have support for at the moment? Should we add a TODO above this to expand to others if there are any?

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: @rbiseck3 This does seem to work as `Literal[ACCEPTED_CATEGORIES]` it didn't like `set[Literal[ACCEPTED_CATEGORIES]]`

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: @rbiseck3 how do we get the typing to activate for a standard user? If I run the example_salesforce.py it doesn't care if the category is legit or not without this.

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: I'm gonna leave as is. If there are multiple records with the same Id for the same user, then I think there are bigger problems

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: we choose what fields to query for this. I'd rather keep the query to a minimum of what we need. salesforce doesn't have a `select * from` for this query.

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: hmmm. Now it doesn't work anymore.

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: @rbiseck3 I can't even tell what this method does to the file. I tried running it and printing out the before and after and it looked like no change. But If I leave it out, the resulting file is way different.

I assume it is modifying the OrderedDict inplace...

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: @rbiseck3 I thought about this. there are only 2 potential extensions. .eml and .xml and they are only used to find a nice way to process the files we get from Salesforce. So it's actually not accurate to claim that a file is .xml . up for debate though.

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: we determined that the typing was not worth it.

unstructured/ingest/v2/processes/connectors/salesforce.py
rbiseck3: Was there an answer to this?

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,291 @@
...
+)
+
+
+@dataclass
+class SalesforceAccessConfig(AccessConfig):
+    consumer_key: str
+    private_key: str
+
+    @requires_dependencies(["cryptography"])
+    def get_private_key_value_and_type(self) -> Tuple[str, Type]:
```
rbiseck3, line 83: I think `tuple` is now a built in type you can use
```suggestion
    def get_private_key_value_and_type(self) -> tuple[str, Type]:
```

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,291 @@
...
+        Send them to next phase where each doc gets downloaded into the
+        appropriate format for partitioning.
+        """
+        from simple_salesforce.exceptions import SalesforceMalformedRequest
+
+        client = self.connection_config.get_client()
+
+        files_list = []
+        for record_type in self.index_config.categories:
+            if record_type not in ACCEPTED_CATEGORIES:
```
rbiseck3, line 152: Can add this as a check in the __post_init__ method of the indexer config since that's where these get set. By the time this method runs, it can assume the categories are all valid.

unstructured/ingest/v2/processes/connectors/salesforce.py
potter-potter: This is completely generic. Only used to set up an email to be partitioned. I added the comment on 57 but will add more. It's as minimal of an email template as can be made while including everything that comes in the document.

unstructured/ingest/v2/processes/connectors/salesforce.py
```diff
@@ -0,0 +1,291 @@
...
+            subject=email_json.get("Subject"),
+            from_email=email_json.get("FromAddress"),
+            to_email=email_json.get("ToAddress"),
+            textbody=email_json.get("TextBody"),
+            htmlbody=email_json.get("HtmlBody"),
+        )
+        return dedent(eml)
+
+    @SourceConnectionNetworkError.wrap
+    def _get_response(self, file_data: FileData):
```
rbiseck3, line 246: response type for method

rbiseck3 (approved):
A couple small comments but overall looks good.

PR #3349, opened 2024-07-03
build: check dependency licenses in CI

MthwRobinson:
### Summary

Adds a CI check to ensure that packages added as dependencies are appropriately licensed. All of the `.txt` files in the `requirements` directory are checked with the exception of:

- `constraints.txt`, since those are not installed and are instead conditions on the other dependency files
- `dev.txt`, since those are for local development and not shipped as part of the `unstructured` package
- `extra-pdf-image.txt` - the `extra-pdf-image.in` since checking `extra-pdf-image.txt` pulls in NVIDIA GPU related packages with an `Other/Proprietary` license type, and there's not a good way to exclude those without adding `Other/Proprietary` to the allowed licenses list.

### Testing

The new `check-licenses` job should pass in CI.


scanny (approved):
LGTM :)

PR #3218, opened 2024-06-17
rfctr(html): replace html parser

scanny:
**Summary**
Replace legacy HTML parser with recursive version that captures all content and provides flexibility to add new metadata. It's also substantially faster although that's just a happy side-effect.

**Additional Context**
The prior HTML parsing algorithm that makes up the core of HTML partitioning was buggy and very difficult to reason about because it did not conform to the inherently recursive structure of HTML. The new version retains `lxml` as the performant and reliable base library but uses `lxml`'s custom element classes to efficiently classify HTML elements by their behaviors (block-item and inline (phrasing) primarily) and give those elements the desired partitioning behaviors.

This solves a host of existing problems with content being skipped and elements (paragraphs) being divided improperly, but also provides a clear domain model for reasoning about its behavior and reliably adjusting it to suit our existing and future purposes.

The parser's operation is recursive, closely modeling the recursive structure of HTML itself. It's behaviors are based on the HTML Standard and reliably produce proper and explainable results even for novel cases.

Fixes #2325 
Fixes #2562
Fixes #2675
Fixes #3168
Fixes #3227
Fixes #3228 
Fixes #3230 
Fixes #3237 
Fixes #3245 
Fixes #3247 
Fixes #3255
Fixes #3309 

#### `emphasized_text_tags` encoding is changed:
- `<strong>` is encoded as `"b"` rather than `"strong"`.
- `<em>` is encoded as `"i"` rather than `"em"`.
- `<span>` is no longer recorded in `emphasized_text_tags` (because without the CSS we can't tell whether it's used for emphasis or if so what kind).
- nested emphasis (e.g. bold+italic) is encoded as multiple characters ("bi").
- `emphasized_text_contents` is broken on emphasis-change boundaries, like:
  ```html
   `<p>foo <b>bar <i>baz</i> bada</b> bing</p>`
  ```
  produces:
  ```json
  {
    "emphasized_text_contents": ["bar", "baz", "bada"],
    "emphasized_text_tags": ["b", "bi", "b"]
  }
  ```
   whereas previously it would have produced:
  ```json
  {
    "emphasized_text_contents": ["bar baz bada", "baz"],
    "emphasized_text_tags": ["b", "i"]
  }
  ```

#### `<pre>` text is preserved as it appears in the html
Except that a leading newline is removed if present (has to be in position 0 of text). Also, a trailing newline is stripped but only if it appears in the very last position ([-1]) of the `<pre>` text. Old parser stripped all leading and trailing whitespace.

Result is that:
[...]

PR #3351, opened 2024-07-06
rfctr(html): refine HTML parser

scanny:
**Note**
This refines the new HTML parser but _does not install it_. This is why no changes to ingest test expectations or other unit-tests are required here. Installing the new parser will happen in the next PR #3218.

**Summary**
The initial version of the parser (purposely) raised on a block element nested inside a phrasing element. While such nesting is not valid according to the HTML Standard, it is accepted by the browser and does happen in the wild.

The refinements here handle this situation similarly to how the browser does, breaking phrasing at the block element boundaries and starting it up again after the block element.

Unfortunately this adds complexity to the parser, but it makes the parser robust against pretty much any HTML we're likely to encounter and partitions it consistent with how it would be rendered in the browser.

### DIFF
--- a/test_unstructured/partition/html/test_parser.py
+++ b/test_unstructured/partition/html/test_parser.py
@@ -12,7 +12,6 @@ from lxml import etree
 
 from unstructured.documents.elements import Address, Element, ListItem, NarrativeText, Text, Title
 from unstructured.partition.html.parser import (
-    Anchor,
     Annotation,
     DefaultElement,
     Flow,
@@ -20,7 +19,10 @@ from unstructured.partition.html.parser import (
     RemovedPhrasing,
     TextSegment,
     _consolidate_annotations,
+    _ElementAccumulator,
     _normalize_text,
+    _PhraseAccumulator,
+    _PreElementAccumulator,
     html_parser,
 )
 
@@ -29,27 +31,21 @@ from unstructured.partition.html.parser import (
 # -- _consolidate_annotations() ------------------
 
 
-def it_gathers_annotations_from_text_segments():
-    text_segments = [
-        TextSegment(
-            " Ford Prefect ",
-            {
-                "link_texts": "Ford Prefect",
-                "link_url": "https://wikipedia/Ford_Prefect",
-                "emphasized_text_contents": "Ford Prefect",
-                "emphasized_text_tags": "b",
-            },
-        ),
-        TextSegment(
-            " alien  encounter",
-            {
-                "emphasized_text_contents": "alien encounter",
-                "emphasized_text_tags": "bi",
-            },
-        ),
+def it_consolidates_annotations_from_multiple_text_segments():
+    annotations = [
+        {
+            "link_texts": "Ford Prefect",
+            "link_url": "https://wikipedia/Ford_Prefect",
+            "emphasized_text_contents": "Ford Prefect",
+            "emphasized_text_tags": "b",
+        },
+        {
+            "emphasized_text_contents": "alien encounter",
+            "emphasized_text_tags": "bi",
+        },
     ]
 
-    annotations = _consolidate_annotations(text_segments)
+    annotations = _consolidate_annotations(annotations)
 
     assert annotations == {
         # -- each distinct key gets a list of values --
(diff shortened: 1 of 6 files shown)


test_unstructured/partition/html/test_parser.py
```diff
@@ -89,6 +85,263 @@ def test_normalize_text_produces_normalized_text(text: str, expected_value: str)
...
+        with pytest.raises(StopIteration):
+            next(phrase_iter)
+
+    # -- .add() -----------------------------------------------------------
+
+    def it_accumulates_text_segments(self):
+        accum = _PhraseAccumulator()
+
+        accum.add(TextSegment("Ford... you're turning ", {}))
+        accum.add(TextSegment("into a penguin.", {}))
```
MthwRobinson, line 108: 🐧

test_unstructured/partition/html/test_parser.py
```diff
@@ -89,6 +85,263 @@ def test_normalize_text_produces_normalized_text(text: str, expected_value: str)
...
+
+        with pytest.raises(StopIteration):
+            next(phrase_iter)
+
+    # -- .add() -----------------------------------------------------------
+
+    def it_accumulates_text_segments(self):
+        accum = _PhraseAccumulator()
+
+        accum.add(TextSegment("Ford... you're turning ", {}))
```
MthwRobinson, line 107: 🛻

test_unstructured/partition/html/test_parser.py
```diff
@@ -428,24 +660,574 @@ def it_generates_text_segments_for_its_text_and_children_and_tail(
...
+                "b",
+                [
+                    TextSegment(
+                        "aaa", {"emphasized_text_contents": "aaa", "emphasized_text_tags": "b"}
+                    ),
+                    TextSegment(
+                        "bbb", {"emphasized_text_contents": "bbb", "emphasized_text_tags": "b"}
+                    ),
+                    TextSegment(
+                        "ccc", {"emphasized_text_contents": "ccc", "emphasized_text_tags": "bi"}
```
MthwRobinson, line 858: @scanny - You might have told me before, but will this always be `"bi"` or would it be `"ib"` if the nesting of the tags were reversed?

test_unstructured/partition/html/test_parser.py
scanny: https://www.goodreads.com/quotes/4388-ford-you-re-turning-into-a-penguin-stop-it :)

typings/lxml/etree/_cleanup.pyi
```diff
@@ -0,0 +1,21 @@
+# pyright: reportPrivateUsage=false
+
+from __future__ import annotations
+
+from typing import Collection, overload
+
+from .._types import _ElementOrTree, _TagSelector
+
+@overload
+def strip_elements(
```
MthwRobinson, line 10: Is `strip_elements` supposed to be here twice?

unstructured/partition/html/parser.py
```diff
@@ -216,46 +371,48 @@ def _element_from_text_or_tail(
...
+    ) -> Iterator[TextSegment | Element]:
+        """Generate zero-or-more `TextSegment`s or `Element`s from text and leading phrasing.
 
-        yield ElementCls(
-            normalized_text,
-            metadata=ElementMetadata(
-                **_consolidate_annotations(text_segments), category_depth=category_depth
-            ),
-        )
+        Note that while this method is named "._iter_text_segments()", it can also generate
```
MthwRobinson, line 391: Pretty sure you told me this earlier too, but are we processing the invalid HTML and saying "quality of the results may be impacted if the HTML is invalid"?

MthwRobinson (commented):
Like all the new tests! Added a few inline comments and questions

test_unstructured/partition/html/test_parser.py
scanny: It will always be `"bi"`. Seemed like canonicalizing it was the right thing to do, just to avoid unnecessary complexity. I'm not sure we could say which came "first" when both are applied because the effect is the same from a visual perspective.

scanny: https://github.com/Unstructured-IO/unstructured/blob/2a12131b4e452b40e1597be888dcfb46c882f9aa/unstructured/partition/html/parser.py#L770-L779

typings/lxml/etree/_cleanup.pyi
scanny: Yes. This is a typing thing (signaled by the `@overload` decorator) where there are two distinct interfaces to a function and you can't represent them both with a single set of annotations. In this case it's because tag names can either be a sequence or you can specify them in variadic form, like:
```python
strip_elements(e, ["nav", "header", "footer"])
# -- OR --
strip_elements( e, "nav", "header", "footer")
```

test_unstructured/partition/html/test_parser.py
MthwRobinson: Yeah I think having a canonical representation is the way to go, just wanted to double check. Especially since, as you mentioned, it doesn't make any difference visually.

typings/lxml/etree/_cleanup.pyi
MthwRobinson: 👍 - Sounds good! Just want to double check

unstructured/partition/html/parser.py
scanny: No. I tried that approach first, but it just wasn't right. We were changing the "paragraph-ness" of the text in that approach and it was inconsistent with how the browser treated that situation.

So it was a little harder, but now we match what the browser does.

This is a pretty good example.
https://github.com/Unstructured-IO/unstructured/blob/2a12131b4e452b40e1597be888dcfb46c882f9aa/test_unstructured/partition/html/test_parser.py#L688-L700

`<strong>` is phrasing and not meant to have a block item nested in it (according to the HTML Standard). However, when it does, the browser treats the text before the block (`<p>` in this case) as phrasing and adds it to the end of the current paragraph. Then it creates a new paragraph for the block item, and then starts a new paragraph for the phrasing after the block item.

MthwRobinson: Gotcha, thanks. And yeah let's stick with what the browser does.

MthwRobinson (approved):
Comments addressed - LGTM!

PR #3360, opened 2024-07-08
Remove pytesseract dep

jake-normal:
The dependency on both `pytesseract` and `unstructured.pytesseract` can cause dependency conflicts with build tools that ensure packages are only provided once. In this PR, I remove the explicit `pytesseract` dependency, since it is also included on L17 of this file. An alternative (and probably better) approach would be to remove `unstructured.pytesseract` and depend on `pytesseract` directly, if your fork is no longer necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
-## 0.14.10-dev10
+## 0.14.10-dev11
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


mondaychen:
I think the fork is no longer necessary https://github.com/madmaze/pytesseract/releases/tag/v0.3.12

PR #3361, opened 2024-07-08
fix(CVE-2024-39705): remove nltk download

MthwRobinson:
### Summary

Addresses [CVE-2024-39705](https://nvd.nist.gov/vuln/detail/CVE-2024-39705), which highlights the risk of remote code execution when running `nltk.download` . Removes `nltk.download` in favor of a `.tgz` file with the appropriate NLTK data files and checking the SHA256 hash to validate the download. An error now raises if `nltk.download` is invoked.

The logic for determining the NLTK download directory is borrowed from `nltk`, so users can still set `NLTK_DATA` as they did previously.

### Testing

1. Create a directory called `~/tmp/nltk_test`. Set `NLTK_DATA=${HOME}/tmp/nltk_test`.
2. From a python interactive session, run:
```python
from unstructured.nlp.tokenize import download_nltk_packages

download_nltk_packages()
```
3. Run `ls /tmp/nltk_test/nltk_data`. You should see the downloaded data.


unstructured/nlp/tokenize.py
```diff
@@ -14,37 +19,121 @@
...
+        if file_hash != NLTK_DATA_SHA256:
+            os.remove(tgz_file)
+            raise ValueError(f"SHA-256 mismatch: expected {NLTK_DATA_SHA256}, got {file_hash}")
+
+        # Extract the contents
+        if not os.path.exists(nltk_data_dir):
+            os.makedirs(nltk_data_dir)
+
+        with tarfile.open(tgz_file, "r:gz") as tar:
+            tar.extractall(path=nltk_data_dir)
```
scanny, line 86: A TypeError will be raised here (L82) whenever `get_nltk_data_dir()` returns `None`, such as when running on GAE.

Need a code path for when `nltk_data_dir` is `None`.

See scratch PR #3362 for changes that allowed the type-checker to reveal this, basically cleaning up typing in this module.

unstructured/nlp/tokenize.py
```diff
@@ -14,37 +19,121 @@
...
+    raise ValueError("NLTK download disabled. See CVE-2024-39705")
+
+
+nltk.download = _raise_on_nltk_download
+
+
+# NOTE(robinson) - mimic default dir logic from NLTK
+# https://github.com/nltk/nltk/
+# 	blob/8c233dc585b91c7a0c58f96a9d99244a379740d5/nltk/downloader.py#L1046
+def get_nltk_data_dir():
```
scanny, line 36: A docstring would be welcome here. Also type-annotation (`-> str | None`) as that's a critical detail for this function.

A good docstring (in my view) states the _contract_ of the function (not how it fulfills it, like what it "does"), along with any caveats. Basically, as a "black-box", what does this function _promise_?

So in this case perhaps something like:
```python
    """The directory where `nltk` resources are located.

    `None` when the location is not writable.
    """
```

scanny (changes requested):
A few remarks, see #3362 for proposed changes you could possibly just merge in.

unstructured/nlp/tokenize.py
MthwRobinson: Docstring added.

unstructured/nlp/tokenize.py
MthwRobinson: Thanks, added.

MthwRobinson:
Thanks @scanny - your proposed typing changes are in now and I've addressed the other comments as well.

scanny (approved):
k, LGTM :)

PR #3362, opened 2024-07-08
proposed changes

scanny:
Some additional changes I recommend. Basically getting `tokenize.py` to type-check clean in strict-mode such that it reveals a core type error. Namely, `download_nltk_package()` will raise `TypeError` when `get_nltk_data_dir()` returns `None`, such as when running on Google App Engine.


MthwRobinson (approved):
Thanks @scanny ! Feel free to merge those into the branch

PR #3383, opened 2024-07-11
rfctr: move astra arg

potter-potter:
Just moving the embedding dimension from the base connector object to the write object since its not needed for an Astra source connector. Only for a destination connector.


unstructured/ingest/cli/cmds/astra.py
```diff
@@ -36,13 +36,6 @@ def get_cli_options() -> t.List[click.Option]:
...
-            click.Option(
-                ["--embedding-dimension"],
-                required=True,
-                default=384,
-                type=int,
-                help="The dimensionality of the embeddings",
-            ),
             click.Option(
                 ["--namespace"],
                 required=False,
```
ahmetmeleq, line 41: `--requested-indexing-policy` might also be a destination exclusive arg

unstructured/ingest/cli/cmds/astra.py
```diff
@@ -36,13 +36,6 @@ def get_cli_options() -> t.List[click.Option]:
                 "Note that the collection name must only include letters, "
```
ahmetmeleq, line 36: "to write into" , we can change this to connect to maybe, to indicate relationship with both use cases (src and dest)

unstructured/ingest/cli/cmds/astra.py
```diff
@@ -36,13 +36,6 @@ def get_cli_options() -> t.List[click.Option]:
...
             ),
-            click.Option(
-                ["--embedding-dimension"],
-                required=True,
-                default=384,
-                type=int,
-                help="The dimensionality of the embeddings",
-            ),
             click.Option(
                 ["--namespace"],
```
ahmetmeleq, line 40: "to write into" -> "to connect to" suggestion for --namespace as well

ahmetmeleq:
Also a suggestion to re-evaluate how the help texts for source exclusive, destination exclusive, and mutual args are worded in the docs and examples

unstructured/ingest/cli/cmds/astra.py
potter-potter: Great idea. I'll give the help strings another look.

unstructured/ingest/cli/cmds/astra.py
potter-potter: Yes. That sounds true. But lets check with @erichare

erichare: Right... these apply to the destination collection in Astra DB for sure. Both the indexing policy and the embedding dimension. So that's a very good point, these should really only apply for that. The source connector assumes the collection already exists and so these parameters are not needed.

Good catch. Thanks!

ahmetmeleq (approved):
LGTM

PR #3395, opened 2024-07-14
fix(nltk): remedy Windows temp-file problem

scanny:
**Summary**
The implementation of `tempfile.NamedTemporaryFile` on Windows Python is problematic in certain situations.

In particular, it raises `PermissionError` when attempting to access the temporary file by name rather than just by the file-descriptor returned by the context-manager.

Remedy this situation by using `tempfile.TemporaryDirectory` instead and using a file name of our choosing. The temporary directory is deleted with all its contents when the context manager closes so the effect is the same and does not produce the error on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.0-dev10
+## 0.15.0-dev11
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


MthwRobinson (approved):
LGTM - thanks for the fix!

PR #3400, opened 2024-07-15
feat: `partition_pdf()` support language specification for PaddleOCR

### LINKED ISSUE #3159: bug/language specification does not work for PaddleOCR agent
I specified the `languages` parameter with the value `["chi", "eng"]` but it did not work.
When I upload a Chinese pdf document, Unstructured still loads a English model.
I checkout the source code and found these lines in the path `unstructured\partition\utils\ocr_models\paddle_ocr.py`, where the __init__ function receives no argument for specifying language:
```
class OCRAgentPaddle(OCRAgent):
    """OCR service implementation for PaddleOCR."""

    def __init__(self):
        self.agent = self.load_agent()

    def load_agent(self, language: str = DEFAULT_PADDLE_LANG):
        """Loads the PaddleOCR agent as a global variable to ensure that we only load it once."""

        import paddle
        from unstructured_paddleocr import PaddleOCR

        # Disable signal handlers at C++ level upon failing
        # ref: https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/
        #      disable_signal_handler_en.html#disable-signal-handler
        paddle.disable_signal_handler()
        # Use paddlepaddle-gpu if there is gpu device available
        gpu_available = paddle.device.cuda.device_count() > 0
        if gpu_available:
            logger.info(f"Loading paddle with GPU on language={language}...")
        else:
            logger.info(f"Loading paddle with CPU on language={language}...")
        try:
            # Enable MKL-DNN for paddle to speed up OCR if OS supports it
            # ref: https://paddle-inference.readthedocs.io/en/master/
            #      api_reference/cxx_api_doc/Config/CPUConfig.html
            paddle_ocr = PaddleOCR(
                use_angle_cls=True,
                use_gpu=gpu_available,
                lang=language,
                enable_mkldnn=True,
                show_log=False,
            )
        except AttributeError:
            paddle_ocr = PaddleOCR(
                use_angle_cls=True,
                use_gpu=gpu_available,
                lang=language,
                enable_mkldnn=False,
                show_log=False,
            )
        return paddle_ocr
```

Is there a way to work around this?

christinestraub:
Closes #3159.

This PR extends language specification capability to `PaddleOCR` in addition to `TesseractOCR`. Users can now specify OCR languages for both OCR engines when using `partition_pdf()`.

### Testing

```
os.environ["OCR_AGENT"] = "unstructured.partition.utils.ocr_models.paddle_ocr.OCRAgentPaddle"

elements = partition_pdf(
    filename=<file_path>,
    strategy=strategy,
    languages=["chi_sim"], # chinese - simplified
    infer_table_structure=True,
)
```


unstructured/partition/lang.py
```diff
@@ -234,7 +234,14 @@ def tesseract_to_paddle_language(tesseract_language: str = "eng") -> str:
     :return: str, corresponding language code for PaddleOCR or None if not found
     """
 
-    return PYTESSERACT_TO_PADDLE_LANG_CODE_MAP.get(tesseract_language, tesseract_language)
+    lang = PYTESSERACT_TO_PADDLE_LANG_CODE_MAP.get(tesseract_language)
+    if not lang:
+        logger.warning(
+            f"{lang} is not a language supported by PaddleOCR, proceed with `en` instead."
```
scanny, line 240: As a reader would read this message, are we requesting an action from them or reporting what the system is doing?

If we need the user to fill in a parameter with "en" or something maybe we can spell that out a bit better.

But if we're just telling them we've substituted "en" then "proceeding" would be the right verb form.

Also, if `not lang` then lang will be `""` or `None`, so the message will read " is not a language supported..." or "None is not a language supported ...". Do we want to use `tesseract_language` instead or some other requested value?

scanny (approved):
A couple changes I think you'll need to make but approving in advance.

PR #3402, opened 2024-07-16
rfctr(file): improve filetype tests

scanny:
**Summary**
Improve file-detection tests in preparation for additional work and bug fixes.

**Additional Context**
- Add type annotations.
- Use mocks instead of `monkeypatch` in most cases and verify calls to mock. This revealed a dozen broken tests, broken in that the mocks weren't being called so a different code path than intended was being exercised.
- Use `example_doc_path()` instead of hard-coded paths.
- Add actual test files for cases where they were being constructed in temporary directories.
- Make test names consistent and more descriptive of behavior under test.


Coniferish (approved):
LGTM. Only thing that might be nice is breaking the tests into groups somehow and/or alphabetizing them

scanny:
@Coniferish yeah, I did that in a later commit, and it's definitely an improvement. You can see which file-types we cover well and which we don't.

But what I noticed is that the bulk of these tests have one of two exact forms and so I'm thinking collapsing most of them into a parameterized test (probably two actually) is the right move. So I'll give that a try a little later.

I want to spend some quality time with the actual implementation now that the tests are reasonably clean, and I think that's going to show up a number of cases that are not covered at all. So I'm thinking to come back to the tests after that.

PR #3405, opened 2024-07-16
feat: save file id for all fsspec connectors if present

rbiseck3:
### Description

If the id value exists in the stats response from fsspec, save it as a `file_id` field in the metadata being persisted on each element.


potter-potter:
I looked into adding this to S3 since it is based on fsspec but has it's own `get_metadata`. It didn't have an id in the `self.fs.stat` so all is good.

potter-potter (approved):
looks good

PR #3407, opened 2024-07-16
bugfix: google drive connector metadata safegaurds

rbiseck3:
### Description

At times, the google drive response doens't have some of the metadata we're grabbing to populate the `FileData` metadata. This is fine, but without the added safegaurds, this can cause a `KeyError`.


potter-potter (approved):
looks good

PR #3408, opened 2024-07-16
bugfix: conform all connectors to be added to registry

rbiseck3:
### Description

Looks like some connectors were never added to the registry explicitly since that change was introduced. All of them are now updated.


potter-potter (approved):
good catch. I was thinking about this last week when I merged salesforce

PR #3411, opened 2024-07-17
rfctr(file): make FileType enum a file-type descriptor

scanny:
**Summary**
Elaborate the `FileType` enum to be a complete descriptor of file-types. Add methods to allow `STR_TO_FILETYPE`, `EXT_TO_FILETYPE` and `FILETYPE_TO_MIMETYPE` mappings to be replaced, removing those redundant and noisy declarations.

In the process, fix some lingering file-type identification and `.metadata.filetype` errors that had been skipped in the tests.

**Additional Context**
Gathering the various attributes of a file-type into the `FileType` enum eliminates the duplication inherent in the separate `STR_TO_FILETYPE` etc. mappings and makes access to those values convenient for callers. These attributes include what MIME-type a file-type should record in metadata and what MIME-types and extensions map to that file-type. These values and others are made available as methods and properties directly on the `FileType` class and members. Because all attributes are defined in the `FileType` enum there is no risk of inconsistency across multiple locations and any changes happen in one and only one place. Further attributes and methods will be added in later commits to support other file-type related operations like mapping to a partitioner and verifying its dependencies are installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.0-dev14
+## 0.15.0-dev15
 
 ### Enhancements
 
--- a/test_unstructured/file_utils/test_exploration.py
+++ b/test_unstructured/file_utils/test_exploration.py
@@ -5,7 +5,7 @@ import pandas as pd
 import pytest
 
 from unstructured.file_utils import exploration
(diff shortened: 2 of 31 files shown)


Coniferish (approved):
Nice. LGTM

PR #3418, opened 2024-07-18
rfctr(auto): add _PartitionerLoader

scanny:
**Summary**
Replace conditional explicit import of partitioner modules in `.partition.auto` with the new `_PartitionerLoader` class. This avoids unbound variable warnings and is much less noisy.

`_PartitionerLoader` makes use of the new `FileType` property `.importable_package_dependencies` to determine whether all required packages are importable before dispatching the file to its partitioner. It uses `FileType.extra_name` to form a helpful error message when a dependency is not installed, so the caller knows which `pip install` extra to specify to remedy the error.

`PartitionerLoader` uses the `FileType` properties `.partitioner_module_qname` and `partitioner_function_name` to load
the partitioner once its dependencies are verified. Loaded partitioners are cached with module lifetime scope for efficiency.

### DIFF
--- a/test_unstructured/partition/test_auto.py
+++ b/test_unstructured/partition/test_auto.py
@@ -10,7 +10,7 @@ import sys
 import tempfile
 import warnings
 from importlib import import_module
-from typing import Callable, Iterator, cast
+from typing import Iterator, cast
 from unittest.mock import patch
 
 import pytest
@@ -27,7 +27,6 @@ from test_unstructured.unit_utils import (
     ANY,
     FixtureRequest,
     LogCaptureFixture,
-    MonkeyPatch,
     example_doc_path,
     function_mock,
     method_mock,
@@ -46,8 +45,7 @@ from unstructured.documents.elements import (
     Title,
 )
 from unstructured.file_utils.model import FileType
-from unstructured.partition import auto
-from unstructured.partition.auto import IMAGE_FILETYPES, _get_partition_with_extras, partition
+from unstructured.partition.auto import _PartitionerLoader, partition
 from unstructured.partition.utils.constants import PartitionStrategy
 from unstructured.staging.base import elements_from_json, elements_to_dicts, elements_to_json
 
@@ -570,17 +568,22 @@ def test_auto_partition_pdf_from_file(pass_metadata_filename: bool, content_type
     assert e.text.startswith("Zejiang Shen")
 
 
-def test_auto_partition_pdf_with_fast_strategy(monkeypatch: MonkeyPatch):
-    file_path = example_doc_path("pdf/layout-parser-paper-fast.pdf")
+def test_auto_partition_pdf_with_fast_strategy(request: FixtureRequest):
+    partition_pdf_ = function_mock(
+        request,
+        "unstructured.partition.pdf.partition_pdf",
+        return_value=[NarrativeText("Hello there!")],
+    )
+    partitioner_loader_get_ = method_mock(
+        request, _PartitionerLoader, "get", return_value=partition_pdf_
+    )
+    filename = example_doc_path("pdf/layout-parser-paper-fast.pdf")
 
-    mock_return = [NarrativeText("Hello there!")]
-    with patch.object(auto, "partition_pdf", return_value=mock_return) as mock_partition:
-        mock_partition_with_extras_map = {"pdf": mock_partition}
-        monkeypatch.setattr(auto, "PARTITION_WITH_EXTRAS_MAP", mock_partition_with_extras_map)
-        partition(filename=file_path, strategy=PartitionStrategy.FAST)
+    partition(filename=filename, strategy=PartitionStrategy.FAST)
 
-    mock_partition.assert_called_once_with(
-        filename=file_path,
+    partitioner_loader_get_.assert_called_once_with(ANY, FileType.PDF)
+    partition_pdf_.assert_called_once_with(
+        filename=filename,
         file=None,
         url=None,
         strategy=PartitionStrategy.FAST,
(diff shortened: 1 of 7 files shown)


unstructured/file_utils/model.py
```diff
@@ -41,8 +47,109 @@ def __lt__(self, other: FileType) -> bool:
...
+
+        `extension` must include the leading period, like `".pdf"`. Extension is suitable as a
+        secondary file-type identification method but is unreliable for primary identification..
+        """
+        # -- not super efficient but plenty fast enough for once-or-twice-per-file use and avoids
+        # -- limitations on defining a class variable on an Enum.
+        for m in cls.__members__.values():
+            if mime_type == m._canonical_mime_type or mime_type in m._alias_mime_types:
+                return m
+        return None
```
scanny, line 90: Move all the methods to the top of the class and let enum-member definitions come last. Otherwise the code is essentially split by the member definitions (which are naturally long and getting longer) and the code is likely to be read much more frequently than the member definitions.

test_unstructured/partition/test_auto.py
```diff
@@ -570,17 +568,22 @@ def test_auto_partition_pdf_from_file(pass_metadata_filename: bool, content_type
...
+def test_auto_partition_pdf_with_fast_strategy(request: FixtureRequest):
+    partition_pdf_ = function_mock(
+        request,
+        "unstructured.partition.pdf.partition_pdf",
+        return_value=[NarrativeText("Hello there!")],
+    )
+    partitioner_loader_get_ = method_mock(
+        request, _PartitionerLoader, "get", return_value=partition_pdf_
+    )
+    filename = example_doc_path("pdf/layout-parser-paper-fast.pdf")
```
Coniferish, line 580: Wouldn't this be `file_path`? I thought `filename` would be `"layout-parser-paper-fast.pdf"`

unstructured/file_utils/model.py
```diff
@@ -41,8 +57,150 @@ def __lt__(self, other: FileType) -> bool:
         """
         return self.name < other.name
 
-    BMP = ("bmp", [".bmp"], "image/bmp", cast(list[str], []))
+    @classmethod
+    def from_extension(cls, extension: str | None) -> FileType | None:
+        """Select a FileType member based on an extension.
+
+        `extension` must include the leading period, like `".pdf"`. Extension is suitable as a
+        secondary file-type identification method but is unreliable for primary identification..
```
Coniferish, line 65:
```suggestion
        secondary file-type identification method but is unreliable for primary identification.
```

unstructured/file_utils/model.py
```diff
@@ -41,8 +57,150 @@ def __lt__(self, other: FileType) -> bool:
...
+        for m in cls.__members__.values():
+            if extension in m._extensions:
+                return m
+        return None
+
+    @classmethod
+    def from_mime_type(cls, mime_type: str) -> FileType | None:
+        """Select a FileType member based on a MIME-type.
+
+        `extension` must include the leading period, like `".pdf"`. Extension is suitable as a
```
Coniferish, line 82: This docstring is talking about extensions instead of mime_types

unstructured/file_utils/model.py
```diff
@@ -41,8 +57,150 @@ def __lt__(self, other: FileType) -> bool:
...
+        secondary file-type identification method but is unreliable for primary identification..
+        """
+        # -- not super efficient but plenty fast enough for once-or-twice-per-file use and avoids
+        # -- limitations on defining a class variable on an Enum.
+        for m in cls.__members__.values():
+            if mime_type == m._canonical_mime_type or mime_type in m._alias_mime_types:
+                return m
+        return None
+
+    @property
```
Coniferish, line 92: Why use `@property` instead of `@lazyproperty` here?

unstructured/file_utils/model.py
```diff
@@ -41,8 +57,150 @@ def __lt__(self, other: FileType) -> bool:
...
+    def partitioner_function_name(self) -> str:
+        """Name of partitioner function for this file-type. Like "partition_docx".
+
+        Raises when this property is accessed on a file-type that is not partitionable. Use
+        `.is_partitionable` to avoid exceptions when partitionability is unknown.
+        """
+        # -- Raise when this property is accessed on a FileType member that has no partitioner
+        # -- shortname. This prevents a harder-to-find bug from appearing far away from this call
+        # -- when code would try to `getattr(module, None)` or whatever.
+        if (shortname := self._partitioner_shortname) is None:
```
Coniferish, line 154: where is the variable `shortname` coming from if not from `self` here?

unstructured/file_utils/model.py
```diff
@@ -41,8 +57,150 @@ def __lt__(self, other: FileType) -> bool:
...
+
+        One use is to determine whether a file-type is one of the five image types, all of which
+        are processed by `partition_image()`.
+
+        `None` for file-types that are not partitionable, although `.is_partitionable` is the
+        preferred way of discovering that.
+        """
+        return self._partitioner_shortname
+
+    BMP = (
```
Coniferish, line 191: Can you add commends for each index saying what property the position is for the first filetype?

```suggestion
    BMP = (
        "bmp",  # value
        "image",  # partitioner_shortname
        ["unstructured_inference"],  # importable_package_dependencies
        "image",  # extra_name
        [".bmp"],  # extensions
        "image/bmp",  # canonical_mime_type
        cast(list[str], []), # alias_mime_types
```

unstructured/partition/auto.py
```diff
@@ -2,131 +2,28 @@
...
-from unstructured.partition.html import partition_html
-from unstructured.partition.json import partition_json
 from unstructured.partition.lang import check_language_args
-from unstructured.partition.text import partition_text
 from unstructured.partition.utils.constants import PartitionStrategy
-from unstructured.partition.xml import partition_xml
 from unstructured.utils import dependency_exists
 
-PARTITION_WITH_EXTRAS_MAP: dict[str, Callable[..., list[Element]]] = {}
-
```
Coniferish, line 25: 👏

Coniferish (approved):
A few comments, but LGTM

test_unstructured/partition/test_auto.py
scanny: Yep, good call.

unstructured/file_utils/model.py
scanny: Fixed.

unstructured/file_utils/model.py
scanny: Yep, fixed now :)

unstructured/file_utils/model.py
scanny: Mostly I guess that the computation on these are very modest and I didn't want to reason through the implications of `@lazyproperty` on an `enum.Enum` object, which has a lot of magic of its own going on. I'm sure it would be fine to make these lazyproperties but just didn't want to test it out :)

PR #3421, opened 2024-07-19
feat: filter ingest step

rbiseck3:
### Description
Support filtering params, such as the current file glob, that can be run independently of the connector. As part of this expansion, a new filter was added to support filtering on file size, which also added this as a metadata field being persisted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.0-dev16
+## 0.15.0-dev17
 
 ### Enhancements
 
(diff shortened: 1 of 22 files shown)


cragwolfe:
Can one of the ingest tests be updated to prove this is working end to end?

potter-potter:
> Can one of the ingest tests be updated to prove this is working end to end?
Max file size is covered here:
https://github.com/Unstructured-IO/unstructured-ingest/blob/main/test_e2e/src/s3-filter.sh

File glob is covered here:
https://github.com/Unstructured-IO/unstructured-ingest/blob/031a9cf8fb68077a2efdc100508e3e3b99ca362e/test_e2e/src/local.sh

PR #3429, opened 2024-07-22
rfctr(file): refactor detect_filetype()

scanny:
**Summary**
In preparation for fixing a cluster of bugs with automatic file-type detection and paving the way for some reliability improvements, refactor `unstructured.file_utils.filetype` module and improve thoroughness of tests.

**Additional Context**
Factor type-recognition process into three distinct strategies that are attempted in sequence. Attempted in order of preference, type-recognition falls to the next strategy when the one before it is not applicable or cannot determine the file-type. This provides a clear basis for organizing the code and tests at the top level.

Consolidate the existing tests around these strategies, adding additional cases to achieve better coverage.

Several bugs were uncovered in the process. Small ones were just fixed, bigger ones will be remedied in following PRs.


test_unstructured/file_utils/test_filetype.py
```diff
@@ -85,13 +85,42 @@ def test_detect_filetype_from_filename(file_name: str, expected_value: FileType)
...
+        ("fake-doc.rtf", FileType.RTF),
+        ("spring-weather.html.json", FileType.JSON),
+        ("fake.odt", FileType.ODT),
+        ("fake-incomplete-json.txt", FileType.TXT),
+    ],
+)
+def test_detect_filetype_from_filename_extension_when_file_does_not_exist(
+    file_name: str, expected_value: FileType, monkeypatch: MonkeyPatch
+):
+    """File-type is detected from extension when file does not exist."""
```
scanny: These are not new test cases, just separating `test_detect_filetype_from_filename_with_extension()` into the part that stays (works when libmagic not available) and the part that will be removed (derives `FileType` from extension when file does not exist at `file_path`).

unstructured/file_utils/filetype.py
```diff
@@ -40,58 +40,25 @@ def detect_filetype(
...
-            mime_type = ft.guess_mime(file.read(4096))
-        if mime_type is None:
-            logger.warning(
-                "libmagic is unavailable but assists in filetype detection on file-like objects. "
-                "Please consider installing libmagic for better results.",
-            )
-            return FileType.from_extension(extension) or FileType.UNK
-
-    else:
-        raise ValueError("No filename, file, nor file_filename were specified.")
```
scanny: The work of these 40 lines is now provided by `_FileTypeDetectionContext.mime_type`, which is why they are removed here as `_FileTypeDetectionContext` is installed.

test_unstructured/metrics/test_element_type.py
```diff
@@ -34,8 +37,8 @@
         ),
     ],
 )
-def test_get_element_type_frequency(filename, frequency):
-    elements = partition(filename=f"example-docs/{filename}")
+def test_get_element_type_frequency(filename: str, frequency: dict[tuple[str, int | None], int]):
+    elements = partition(example_doc_path(filename))
```
scanny, line 41: The changes in this file were required because the new `detect_filetype()` recognizes `"fake-email.txt` as an `EML` file, even though it has a `.txt` extension. So the results are correct, which they weren't before when it was parsed as plain text.

unstructured/partition/auto.py
```diff
@@ -471,10 +471,11 @@ def file_and_type_from_url(
...
 
-    content_type = (
-        content_type or response.headers.get("Content-Type", "").split(";")[0].strip().lower()
-    )
-    encoding = response.headers.get("Content-Encoding", "utf-8")
+    if content_type := content_type or response.headers.get("Content-Type", None):
+        content_type = content_type.split(";")[0].strip().lower()
+
+    # -- non-None when response is textual --
+    encoding = response.encoding
```
scanny, line 478: This was just plain wrong before. Not sure exactly why the test used to pass. The `Content-Encoding:` header describes the compression scheme used for downloading the HTTP response (`"gzip" in this case) so was not the right thing at all. `encoding` is how to decode the bytes of a textual file-format into a `str`, so like "utf-8" or similar.

test_unstructured/file_utils/test_model.py
```diff
@@ -10,6 +10,12 @@
 class DescribeFileType:
     """Unit-test suite for `unstructured.file_utils.model.Filetype`."""
 
+    # -- .__lt__() ----------------------------------------------
+
+    def it_is_a_collection_ordered_by_name_and_can_be_sorted(self):
```
Coniferish, line 15: This is ordered alphabetically, right?

scanny: Magic methods precede other methods, just like in the class. Is that what you're asking?

Coniferish: No, I mean the ordering of FileType on name (below--I commented on the wrong line)

test_unstructured/file_utils/test_filetype.py
```diff
@@ -548,3 +550,260 @@ def magic_from_buffer_(request: FixtureRequest):
...
+
+        assert isinstance(head, bytes)
+        assert len(head) == 4096
+        assert head.startswith(b"Iwan Roberts\nRoberts celebrating after")
+
+    # -- .file_path ---------------------------------------------
+
+    @pytest.mark.parametrize("file_path", [None, "a/b/c.pdf"])
+    def it_knows_the_file_path_provided_by_the_caller(self, file_path: str | None):
+        assert _FileTypeDetectionContext(file_path=file_path).file_path == file_path
```
Coniferish, line 680: Should we have a test for if a file object is used instead of a file_path?

test_unstructured/file_utils/test_filetype.py
```diff
@@ -548,3 +550,260 @@ def magic_from_buffer_(request: FixtureRequest):
...
+            assert ctx.mime_type == "text/plain"
+
+    def and_it_provides_the_MIME_type_from_file_using_filetype_lib_when_magic_is_unavailable(self):
+        with patch("unstructured.file_utils.filetype.LIBMAGIC_AVAILABLE", False):
+            file_path = example_doc_path("simple.doc")
+            with open(file_path, "rb") as f:
+                ctx = _FileTypeDetectionContext(file=f)
+                assert ctx.mime_type == "application/msword"
+
+    # -- .open() ------------------------------------------------
```
Coniferish, line 752: Should there be any tests showing the file is closed appropriately?

test_unstructured/file_utils/test_filetype.py
```diff
@@ -548,3 +550,260 @@ def magic_from_buffer_(request: FixtureRequest):
...
+    def and_it_uses_character_detection_to_correct_a_wrong_encoding_arg_for_file_path(self):
+        ctx = _FileTypeDetectionContext(
+            file_path=example_doc_path("norwich-city.txt"), encoding="utf_32_be"
+        )
+
+        text_head = ctx.text_head
+
+        assert isinstance(text_head, str)
+        assert len(text_head) == 4096
+        assert text_head.startswith("Iwan Roberts\nRoberts celebrating after")
```
Coniferish, line 786: Should there be an assert using `ctx.encoding` here?

test_unstructured/file_utils/test_model.py
scanny: Ah, `.extra_name` was out of order. fixed that ;)

test_unstructured/file_utils/test_filetype.py
scanny: Hmm, good question. From a unit-testing perspective, the computation of `.is_zipfile` does not vary between file-provided-as-path and file-provide-as-file-like-object. So no, from that unit-testing perspective.

What _does_ vary depending on file-source-type is how `.open()` works, so we rely on the unit tests for that to make sure both cases produce a working file-descriptor (open file) for `.is_zipfile` to use.

As far as integration tests are concerned, both those cases are exercised dozens of times, so we rely on those to verify `.open()` and `.is_zipfile` work as a composition.

unstructured/file_utils/filetype.py
```diff
@@ -107,24 +107,51 @@ def file_type(cls, ctx: _FileTypeDetectionContext) -> FileType:
     @property
     def _file_type(self) -> FileType:
```
Coniferish, line 136: Why are these properties instead of lazyproperties?

test_unstructured/file_utils/test_filetype.py
scanny: We only want it to be closed when it's from a file-path, otherwise we need to leave it open.

But I can put a line to do that in the `"..._when_it_is_a_file_path()"` case.

test_unstructured/file_utils/test_filetype.py
scanny: No, I'd say not; the behavior this test is checking is more like "it can still read the text file, at least in this case, when the encoding asserted by the caller is wrong". We could have instead mocked `detect_file_encoding()` but I thought it was better to retain that integration in the test while we exercise the `UnicodeDecodeError` code-path. 

How the code deeper down gets that done is not under test here. I do think it's worth testing over there but I assume we have some tests for that already and anyway that's outside the scope of this work.

unstructured/file_utils/filetype.py
scanny: They could be I suppose. No compelling reason why not, but they're only called at most once so I went with keeping things simple. There's no code path from calling `_FileTypeDetector.file_type(ctx)` that would ever execute any of these more than once. Slightly simplifies the stack trace to use `@property`, but any reason to make them lazy would override that small benefit. Those reasons would be:
- computationally expensive and called more than once
- not idempotent, like it constructs a collaborator object you only want the one instance of.

It's a borderline call either way. If these were already lazyproperties and I had occasion to refactor, that wouldn't be something I'd be inclined to change unless one of the mentioned conditions applied.

It's also cheap and easy to change in the future if we need too. The interface is unchanged by it and it's a one line change.

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.15.1-dev0
+## 0.15.1-dev1
```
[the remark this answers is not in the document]
scanny: Hmm, good idea. Let me see if I can remember what they were :)

unstructured/file_utils/filetype.py
```diff
@@ -581,13 +581,13 @@ def file_type(self) -> FileType | None:
             # .docx/.xlsx file by looking for expected filenames within the zip file.
             filenames = [f.filename for f in zip.filelist]
 
-            if all(f in filenames for f in ("docProps/core.xml", "word/document.xml")):
+            if all(f in filenames for f in ("word/document.xml",)):
```
Coniferish, line 584: Am I reading this correctly that "word/document.xml" and the other files in this series are example documents we're using to compare the partitioned document against for the presence of specific files? If so, can we add a comment here saying as much?

unstructured/metrics/element_type.py
```diff
@@ -1,10 +1,23 @@
...
+from typing_extensions import TypeAlias
+
+FrequencyDict: TypeAlias = "dict[tuple[str, int | None], int]"
+"""Like:
+    {
+        ("ListItem", 0): 2,
+        ("NarrativeText", None): 2,
+        ("Title", 0): 5,
+        ("UncategorizedText", None): 6,
+    }
```
Coniferish, line 14: Why is there an integer in the key? Wouldn't you just need the string for the key and the int/frequency as the value?

unstructured/file_utils/filetype.py
scanny: `filenames` is a bit of a misnomer here, but it's consistent with the interface attribute `f.filename` for `zip.filelist` two lines above.

A DOCX file is a Zip file and a Zip contains a collection of "members". Each member has a name (that looks like a path but isn't exactly) and some content as bytes. I would have used `members` instead of `filenames` normally, but this seemed easier to follow so I left it.

A DOCX file will always contain a member with the name `"word/document.xml"` and that's what we're checking for here. Same idea with PPTX and XLSX.

Turns out that a DOCX file does not need to contain a member with the name "docProps/core.xml". Those produced by Microsoft Word always do, but a DOCX can come from a number of sources, including converting a PDF to a DOCX and those have unpredictable behaviors. The spec does not require that member.

Any Office 2007+ (OpenXml) file _can_ have that member, so it doesn't prove it's a DOCX file if it's present and doesn't prove it's _not_ a DOCX file if it's absent. So it's of no use to us in this check and I ditched it.

Normally I wouldn't have bothered, but one of the (perfectly valid) PPTX metrics files we use doesn't have a "docProps/core.xml" member so the test was failing and this fixed it. But it's the right thing to do anyway.

In fact there are more reliable ways to tell, but that involves parsing one of the small XML files in the Zip and I didn't want to bump scope on this PR.

unstructured/metrics/element_type.py
scanny: I didn't invent this data structure, but from what I gather, the `int | None` member of the tuple is to record the `category_level` value for the element. So like how many top-level headings do we have separate from how many second and third-level headings. Something in the metrics wants to make sure we detect the right level for headings and list items is my take.

scanny: Btw, the reason I had to touch this was because of a "bug fix" or at least an improvement along the way.

The file `fake-email.txt` was previously identified as TXT and partitioned quite differently because all the transport headers and so forth were included, as garbage text I'm sure.

But the refactored version correctly identified it as EML and partitions it as a user would intend. That resulted in fewer elements, partly because it's a very short example email.

test_unstructured/file_utils/test_model.py
scanny: But yes, the ordering is alphabetical by name, the `.name` attribute of an `enum.Enum` member being the way you access it from the `Enum` class, like the name of `FileType.DOCX` is `"DOCX"`.

Coniferish (approved):
LGTM

unstructured/file_utils/filetype.py
Coniferish: Okay, thanks for the explanation. The removal of `docProps/core.xml` was also something I was implicitly wondering about.

unstructured/metrics/element_type.py
Coniferish: Okay. Didn't realize this type was used elsewhere.

PR #3431, opened 2024-07-23
feat/pdf-page-limit-in-hi-res

mackurzawa:
# Description:
Passing `max_pages` argument allows rejecting pdf files which exceeds this page number limit while `high_res` strategy is chosen. By default it will allow parsing pdf files with unlimited number of pages.

# Testing:
```python

elements = partition(filename="unstructured/example-docs/pdf/reliance.pdf", strategy='hi_res')  # should pass
elements = partition(filename="unstructured/example-docs/pdf/reliance.pdf", strategy='hi_res', max_pages=4)  # should pass
elements = partition(filename="unstructured/example-docs/pdf/reliance.pdf", strategy='hi_res', max_pages=2)  # should raise PdfMaxPagesExceededError
```


test_unstructured/partition/pdf_image/test_pdf.py
```diff
@@ -1362,3 +1363,27 @@ def test_analysis_artifacts_saved():
         for el in expected_layouts:
             for page in expected_pages:
                 assert bboxes_dir / f"page{page}_layout_{el}.png" in bboxes_files
+
+
+@pytest.mark.parametrize(
+    ("max_pages", "expected_result"),
```
pawel-kmiecik, line 1369: I think `expected_result` could be just `expected_error` boolean value.

test_unstructured/partition/pdf_image/test_pdf.py
```diff
@@ -1362,3 +1363,27 @@ def test_analysis_artifacts_saved():
...
+    [
+        (3, "expected_successful"),
+        (2, "expected_error"),
+    ],
+)
+def test_max_pages_argument(max_pages, expected_result):
+    if expected_result == "expected_successful":
+        pdf.partition_pdf_or_image(
+            filename=example_doc_path("pdf/reliance.pdf"),
```
pawel-kmiecik, line 1378: one more would be nice and `filename` added as a separate `@pytest.mark.parametrize` variable, e.g. `example-docs/layout-parser-paper-with-empty-pages.pdf`

unstructured/errors.py
```diff
@@ -0,0 +1,8 @@
+class PdfMaxPagesExceededError(ValueError):
+    """Error raised, when number of PDF pages exceeds max_pages limit
+    and HI_RES strategy is chosen.
+    """
+
+    def __init__(self, message):
+        self.message = message
```
pawel-kmiecik, line 7: we don't need keeping the message in this class. So the whole `__init__` method is not needed.

unstructured/partition/pdf.py
```diff
@@ -506,6 +507,33 @@ def _process_pdfminer_pages(
     return elements
 
 
+def _get_PDF_page_number(
```
pawel-kmiecik, line 510: PDF -> lowercase

unstructured/partition/pdf.py
```diff
@@ -506,6 +507,33 @@ def _process_pdfminer_pages(
...
+    filename: str = "",
+    file: Optional[bytes | IO[bytes]] = None,
+) -> int:
+    from pypdf import PdfReader
```
pawel-kmiecik, line 514: I think we don't need local import in this function so this can got to  top imports.

unstructured/partition/pdf.py
```diff
@@ -506,6 +507,33 @@ def _process_pdfminer_pages(
...
+    else:
+        ValueError("Either 'file' or 'filename' must be provided.")
+    return number_of_pages
+
+
+def _is_max_pages_exceeded(
+    filename: str = "", file: Optional[bytes | IO[bytes]] = None, max_pages: int = None
+) -> bool:
+    """Checks whether PDF exceeds max_pages limit."""
+    if max_pages:
```
pawel-kmiecik, line 530:
```python
if max_pages:
        return _get_PDF_page_number(filename=filename, file=file) > max_pages
return False
```

unstructured/partition/pdf.py
```diff
@@ -530,6 +558,7 @@ def _partition_pdf_or_image_local(
     starting_page_number: int = 1,
     extract_forms: bool = False,
     form_extraction_skip_tables: bool = True,
+    max_pages: int = None,
```
pawel-kmiecik, line 561:
```suggestion
    max_pages: Optional[int] = None,
```

unstructured/partition/pdf.py
```diff
@@ -544,6 +573,12 @@ def _partition_pdf_or_image_local(
         process_file_with_pdfminer,
     )
 
+    if _is_max_pages_exceeded(filename=filename, file=file, max_pages=max_pages):
+        raise PdfMaxPagesExceededError(
+            f"Maximum number of PDF file pages exceeded - "
+            f"pages={_get_PDF_page_number(filename=filename, file=file)}, maximum={max_pages}."
```
pawel-kmiecik, line 579: Hmm we'd need to avoid opening and reading the file twice - so we need to rethink the design here.

I know I've asked for `_is_max_pages_exceeded -> bool` but to have the error with nice message we'd need to introduce only one function e.g. `check_max_pages_exceeded` that won't return any value but might raise the `PdfMaxPagesExceededError`.

Sorry for the inconvenience - I wasn't conscious the number of pages won't be available here.

pawel-kmiecik (changes requested):

unstructured/partition/pdf.py
pawel-kmiecik: No longer valid comment

pawel-kmiecik (commented):
LGTM now but let's wait for some other engineer to take a look as it introduces a new param and new exception

awalker4:
This feels better suited for the API to me. A library user probably doesn't need this param, and if they do want to self limit their page count, they can implement this logic before in the caller. IMO this is breaking the abstraction that partition accepts a file input and just processes it.

pawel-kmiecik:
> This feels better suited for the API to me. A library user probably doesn't need this param, and if they do want to self limit their page count, they can implement this logic before in the caller. IMO this is breaking the abstraction that partition accepts a file input and just processes it.

But then we'd need to duplicate the logic that resolves `auto` strategy in the API code and load the PDF file (including the big ones) twice - in API and inside `unstructured`.

awalker4:
That's fair, and I guess it would be nice for each filetype to have its own page counting logic.

unstructured/errors.py
```diff
@@ -0,0 +1,4 @@
+class PdfMaxPagesExceededError(ValueError):
```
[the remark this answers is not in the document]
MthwRobinson: Yeah agree with @awalker4 on that

unstructured/errors.py
```diff
@@ -0,0 +1,4 @@
...
+    """Error raised, when number of PDF pages exceeds max_pages limit
```
MthwRobinson, line 2: And if we make the above change, we can change the verbiage down here too to make it more generic.

MthwRobinson (approved):
Looks good from my end pending the error comment.

unstructured/errors.py
pawel-kmiecik: Additionally, @mackurzawa proposed to add some custom fields to this error class there but I stated it's enough to init with just a message.
Now I'm thinking if adding 2 fields:
- `document_pages: int`
- `max_pages: int`
would be helpful forming some custom messages in the downstream try-catch logic?

MthwRobinson: Yeah I think that would be a good addition, since then the user knows how to split up the document to resolve this issue.

scanny:
> > This feels better suited for the API to me. A library user probably doesn't need this param, and if they do want to self limit their page count, they can implement this logic before in the caller. IMO this is breaking the abstraction that partition accepts a file input and just processes it.
> 
> But then we'd need to duplicate the logic that resolves `auto` strategy in the API code and load the PDF file (including the big ones) twice - in API and inside `unstructured` - just to count the pages.
> 
> Additionally - this can serve as a "memory-and-time-keeper", not letting loading too big files.

I'm inclined to think Austin's initial thought here is quite right. This belongs in the API and not the open-source library.

We are already doing PDF file detection in the API for page-splitting purposes. File-type detection, at least of that type (like is it a PDF or TIFF) is trivial and may very well be required in the API, as it already is.

The big thing that `partition()` does is the _dispatch_ of the partition call. No one is suggesting that be moved to the API. File-type detection is provided by `detect_filetype()` which can easily be called from there if absolutely necessary and in fact may need to be for other purposes.

But identifying a PDF is very easy to do very reliably.

scanny (dismissed):
I'd like to see this happen in API (or possibly SDK or both), not in open-source, for reasons stated in comment.

scanny:
@pawel-kmiecik can you provide a link to the "auto-strategy-resolver" code you're talking about?

scanny:
Re: loading twice argument, we're already loading PDFs in the API to split pages, not sure I see the problem there.

mackurzawa:
Correct me, if I'm wrong @pawel-kmiecik, but the issue here is that we want to reject pdfs that exceeds page limit only when `hi_res` strategy is chosen at the end. Ofc if api calls provides `hi_res` as the strategy we can easily take care of potential rejection on api level, but if the `auto` strategy have been provided  `unstructured` deals with determining which strategy to choose, so to check it on api level we need to duplicate `determine_pdf_or_image_strategy` from `unstructured/partition/strategies.py` in api.

pawel-kmiecik:
> @pawel-kmiecik can you provide a link to the "auto-strategy-resolver" code you're talking about?

Here the strategy is resolved:
https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L290

scanny:
What's the rest of the plan for adding a new parameter? It would need to go in API (form-model), SDK, and documentation for both as well.

I'm less concerned with any performance impact for two reasons:
1. We can implement it such that there is no performance impact unless the API caller specifies a `max_pages` argument.
2. Counting pages, even in a large PDF is in the tens of milliseconds range. `example-docs/DA-1p.pdf` for example counts `619` in `0.071` seconds.

I expect the logic would look something like this:

```python
def reject_for_too_many_pages() -> None:
    """Raises when PDF document exceeds caller-specified `max_pages` in non-fast mode."""
    # -- no limit when no max_pages not specified --
    if max_pages_arg is None:
        return
    # -- no limit on fast strategy --
    if strategy == "fast":
        return
    # -- only applies to PDFs --
    if not is_pdf:
        return
    # -- shorter PDFs are just fine --
    if pdf_page_count <= max_pages:
        return
    # -- non-fast strategy that resolves to fast is okay too --
    if detect_strategy() == "fast":
        strategy = "fast"
        return

    raise PdfTooLongError( ... )    
```

Thinking through this you can see that the strategy-determination logic would never be invoked twice because:
- it is only invoked here when the page-count is exceeded
- test will fail if strategy resolves to non-fast
- if it does resolve to fast we can change the strategy to fast and then detection logic won't be re-executed.

Just so you know where I'm coming from:
- I'm strongly inclined to avoid adding new parameters to the open-source unless they clearly apply there. We already have a large number (50+) and that degrades the new-developer experience.
- Violating responsibility boundaries for short-term optimization benefits is very expensive in the long run. It produces couplings across system boundaries that make reasoning about the assembled system, an already challenging job, more difficult, and in this case would add new concepts to partitioning as a whole, which are unnecessary for most partitioners. Those new concepts make partitioning incrementally more complex and we pay that incrementally increased maintenance cost indefinitely.

pawel-kmiecik:
> What's the rest of the plan for adding a new parameter? It would need to go in API (form-model), SDK, and documentation for both as well.
> 
> I'm less concerned with any performance impact for two reasons:
> 
> 1. We can implement it such that there is no performance impact unless the API caller specifies a `max_pages` argument.
> 2. Counting pages, even in a large PDF is in the tens of milliseconds range. `example-docs/DA-1p.pdf` for example counts `619` in `0.071` seconds.
> 
> I expect the logic would look something like this:
> 
> ```python
> def reject_for_too_many_pages() -> None:
>     """Raises when PDF document exceeds caller-specified `max_pages` in non-fast mode."""
>     # -- no limit when no max_pages not specified --
>     if max_pages_arg is None:
>         return
>     # -- no limit on fast strategy --
>     if strategy == "fast":
>         return
>     # -- only applies to PDFs --
>     if not is_pdf:
>         return
>     # -- shorter PDFs are just fine --
>     if pdf_page_count <= max_pages:
>         return
>     # -- non-fast strategy that resolves to fast is okay too --
>     if detect_strategy() == "fast":
>         strategy = "fast"
>         return
> 
>     raise PdfTooLongError( ... )    
> ```
> 
> Thinking through this you can see that the strategy-determination logic would never be invoked twice because:
> 
> * it is only invoked here when the page-count is exceeded
> * test will fail if strategy resolves to non-fast
> * if it does resolve to fast we can change the strategy to fast and then detection logic won't be re-executed.
> 
> Just so you know where I'm coming from:
> 
> * I'm strongly inclined to avoid adding new parameters to the open-source unless they clearly apply there. We already have a large number (50+) and that degrades the new-developer experience.
> * Violating responsibility boundaries for short-term optimization benefits is very expensive in the long run. It produces couplings across system boundaries that make reasoning about the assembled system, an already challenging job, more difficult, and in this case would add new concepts to partitioning as a whole, which are unnecessary for most partitioners. Those new concepts make partitioning incrementally more complex and we pay that incrementally increased maintenance cost indefinitely.

That was ofc the initial design - to add it just in API. And I agree that's more API responsibility to "protect" its instance :)
I just don't agree with that:
> Thinking through this you can see that the strategy-determination logic would never be invoked twice because:
If the strategy will resolve to `hi_res` and page count will not be exceeded, then will do the same operations (open with `pdfminer`) twice - but if it's an acceptable overhead then let's reject this PR and open a new one in `unstructured-api` @mackurzawa

mackurzawa:
Thank you @scanny for your suggestions, but I still don't fully agree with you.

Just to make sure we're on the same page - we want to reject a pdf, when number of pages exceeds some limit and **exactly hi_res** strategy is chosen - no matter if the user provides it when calling API (here we could easily handle it in API) or user provides `auto` strategy which transforms to `hi_res`.

Regarding your solution, I still can see a potential issue here:
>  if detect_strategy() == "fast":

(As I said before it should be checked if `hi_res` is chosen, but it doesn't matter now)

Your `detect_strategy` function here should work just like `determine_pdf_or_image_strategy` from `unstructured/partition/strategies.py`, but calling this function outside of unstructured is cumbersome:

`determine_pdf_or_image_strategy`  have few parameters:
```
def determine_pdf_or_image_strategy(
    strategy: str,
    is_image: bool = False,
    pdf_text_extractable: bool = False,
    infer_table_structure: bool = False,
    extract_images_in_pdf: bool = False,
    extract_image_block_types: Optional[List[str]] = None,
):
```
To get the same results as in unstructured for code running in API I should pass all the arguments as it is done in `unstructured/partition/pdf.py`. The problem is the complexity of the arguments passed - they are the result of other functions in unstructured, whose arguments are also the result of other functions, and so on, e.g. `pdf_text_extractable` depends on `extracted_elements`, which is the result of the function `extractable_elements`, whose arguments also depend on others...

This means that I would have to copy all the logic behind the arguments of the determine_pdf_or_image_strategy function including all the functions that ultimately decide what those arguments look like, which in my opinion will add huge amounts of duplicate code. 

I think making unstructured a little bit developer friendly here will result in making unstrutured-api a lot less readable. So, at the end my question is should I still duplicate all the logic that decides on ultimate strategy from unstructured to unstructured-api?

pawel-kmiecik:
> Thank you @scanny for your suggestions, but I still don't fully agree with you.
> 
> Just to make sure we're on the same page - we want to reject a pdf, when number of pages exceeds some limit and **exactly hi_res** strategy is chosen - no matter if the user provides it when calling API (here we could easily handle it in API) or user provides `auto` strategy which transforms to `hi_res`.
> 
> Regarding your solution, I still can see a potential issue here:
> 
> > if detect_strategy() == "fast":
> 
> (As I said before it should be checked if `hi_res` is chosen, but it doesn't matter now)
> 
> Your `detect_strategy` function here should work just like `determine_pdf_or_image_strategy` from `unstructured/partition/strategies.py`, but calling this function outside of unstructured is cumbersome:
> 
> `determine_pdf_or_image_strategy` have few parameters:
> 
> ```
> def determine_pdf_or_image_strategy(
>     strategy: str,
>     is_image: bool = False,
>     pdf_text_extractable: bool = False,
>     infer_table_structure: bool = False,
>     extract_images_in_pdf: bool = False,
>     extract_image_block_types: Optional[List[str]] = None,
> ):
> ```
> 
> To get the same results as in unstructured for code running in API I should pass all the arguments as it is done in `unstructured/partition/pdf.py`. The problem is the complexity of the arguments passed - they are the result of other functions in unstructured, whose arguments are also the result of other functions, and so on, e.g. `pdf_text_extractable` depends on `extracted_elements`, which is the result of the function `extractable_elements`, whose arguments also depend on others...
> 
> This means that I would have to copy all the logic behind the arguments of the determine_pdf_or_image_strategy function including all the functions that ultimately decide what those arguments look like, which in my opinion will add huge amounts of duplicate code.
> 
> I think making unstructured a little bit developer friendly here will result in making unstrutured-api a lot less readable. So, at the end my question is should I still duplicate all the logic that decides on ultimate strategy from unstructured to unstructured-api?

@scanny I must agree with @mackurzawa - the effort to introduce it in API only is too huge for me (which maybe wasn't visible on the first look).
And well - we are not exposing this new parameter - it's only added to a protected (hidden) `_partition_pdf_or_image_local` function, like many other parameters - so it's not explicitly set in main public functions.

And I find it useful for users of the library as well - PDF and hi_res are not like other file types and may introduce various issues for very big files, so users can start protecting their applications with this parameter from timeouts, hangs, etc.
Maybe one thing I would change to be more precise is renaming `max_pages` param to `hi_res_max_pages`

scanny:
Ahh, okay, now I think I understand better why you wanted to put it inside the PDF partitioning boundary.

One solution to the tangly relocation problem would be to add a simpler function inside `unstructured.partition.pdf` that encapsulated whatever calls were necessary to work out `is_image` and `pdf_text_extractable`. That would keep the tangly parts inside the boundary but expose a simple interface to the outside.

But whatever, I don't want to delay progress further. I drop my opposition to implementing it inside open-source.

awalker4 (approved):

CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
-## 0.15.1-dev5
+## 0.15.1-dev6
 
 ### Enhancements
 
 ### Features
 
+* **Add `max_pages` argument for partitioning, which allows rejecting PDF files that exceed this page number limit, when the `high_res` strategy is chosen.** By default, it will allow parsing PDF files with an unlimited number of pages.
```
pawel-kmiecik, line 7: I'd rename this parameter to `pdf_hi_res_max_pages` everywhere

pawel-kmiecik (approved):
Just a nit left

PR #3434, opened 2024-07-24
fix(file): confirm or correct asserted DOCX, PPTX, and XLSX content types

scanny:
**Summary**
The `content_type` argument received by `partition()` from the API is sometimes unreliable for MS-Office 2007+ MIME-types. What we've observed is that it gets the MS-Office bit right but falls down on distinguishing PPTX from DOCX or XLSX.

Confirmation of these types is simple, fast, and reliable. Confirm all MS-Office `content_type` argument values asserted by callers of `detect_filetype()` and correct swapped values.

### DIFF
--- a/test_unstructured/file_utils/test_filetype.py
+++ b/test_unstructured/file_utils/test_filetype.py
@@ -153,6 +153,38 @@ def test_it_detects_correct_file_type_from_file_no_name_with_correct_asserted_co
     assert file_type is expected_value
 
 
+@pytest.mark.parametrize(
+    ("expected_value", "file_name"),
+    [
+        (FileType.DOCX, "simple.docx"),
+        (FileType.PPTX, "fake-power-point.pptx"),
+        (FileType.XLSX, "stanley-cups.xlsx"),
+    ],
+)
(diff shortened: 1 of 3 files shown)


test_unstructured/file_utils/test_filetype.py
```diff
@@ -664,13 +696,13 @@ def and_it_returns_the_empty_string_as_the_extension_when_there_are_no_file_name
...
 
-    def it_grabs_the_first_4k_bytes_of_the_file_for_use_by_magic(self):
+    def it_grabs_the_first_8k_bytes_of_the_file_for_use_by_magic(self):
         ctx = _FileTypeDetectionContext(file_path=example_doc_path("norwich-city.txt"))
 
         head = ctx.file_head
 
         assert isinstance(head, bytes)
-        assert len(head) == 4096
+        assert len(head) == 8192
```
scanny, line 705: This bump of `.file_head` size to `8192` has nothing specific to do with fixing the problem remedied by this PR, it was just too small to put in it's own PR so came along for the ride.

Coniferish (approved):
LGTM

PR #3437, opened 2024-07-24
fix(file): fix OLE-based file-type auto-detection

### LINKED ISSUE #3364: bug(doc): DOC, PPT, and XLS file-like-objects misidentified as MSG file
**Summary**
A DOC, PPT, or XLS file sent to `partition()` as a file-like object is misidentified as a MSG file and raises an exception in `python-oxmsg` (which is used to process MSG files).

**To Reproduce**
```python
>>> with open(example_doc_path("simple.doc"), "rb") as f:
...    partition(file=f)
KeyError: '__properties_version1.0'
.../lib/python3.9/site-packages/oxmsg/storage.py:83: KeyError
```

**Additional Information**
- `magic` identifies the file MIME-type as "application/x-ole-storage". This MIME-type is "correct" for both MSG and DOC files (as well as PPT and XLS), it's just not specific enough.
- Assuming this is the best `libmagic` can do, we will need to disambiguate. OLE files are easily and performantly readable by the Python `olefile` package. OLE files or CFBF (Compound File Binary Format) are vaguely like Zip files in that they contain a "filesystem" of "directory" and "file" objects (called "storage" and "stream" in CFBF parlance). By looking for expected items inside the file we should be able to disambiguate "application/x-ole-storage" into MSG, DOC, XLS, and PPT formats, and possibly other Microsoft formats.

scanny:
**Summary**
A DOC, PPT, or XLS file sent to partition() as a file-like object is misidentified as a MSG file and raises an exception in python-oxmsg (which is used to process MSG files).

**Fix**
DOC, PPT, XLS, and MSG are all Microsoft OLE-based files, aka. Compound File Binary Format (CFBF). These can be reliably distinguished by inspecting magic bytes in certain locations. `libmagic` is unreliable at this or doesn't try, reporting the generic `"application/x-ole-storage"` which corresponds to the "container" CFBF format (vaguely like a Microsoft Zip format) that all these document types are stored in.

Unconditionally use `filetype.guess_mime()` provided by the `filetype` package that is part of the base unstructured install. Unlike `libmagic`, this package reliably detects the distinguished MIME-type (e.g. `"application/msword"`) for OLE file subtypes.

Fixes #3364

### DIFF
--- a/test_unstructured/file_utils/test_filetype.py
+++ b/test_unstructured/file_utils/test_filetype.py
@@ -23,6 +23,7 @@ from unstructured.file_utils.filetype import (
     _TextFileDifferentiator,
     _ZipFileDifferentiator,
     detect_filetype,
+    is_json_processable,
 )
 from unstructured.file_utils.model import FileType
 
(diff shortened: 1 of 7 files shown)


test_unstructured/file_utils/test_filetype.py
```diff
@@ -575,6 +637,41 @@ def test_it_detect_CSV_from_path_and_file_when_content_contains_escaped_commas()
...
+def it_affirms_JSON_is_array_of_objects_from_text():
+    with open(example_doc_path("simple.json")) as f:
+        text = f.read()
+    assert is_json_processable(file_text=text) is True
+
+
+def and_it_affirms_JSON_is_NOT_an_array_of_objects_from_text():
+    with open(example_doc_path("not-unstructured-payload.json")) as f:
+        text = f.read()
+    assert is_json_processable(file_text=text) is False
```
scanny, line 672: These new tests have nothing to do with this fix per se, there were just no tests for this interface function in the test suite.

unstructured/file_utils/filetype.py
```diff
@@ -112,12 +112,12 @@ def is_json_processable(
...
+
     if file_text is None:
-        file_text = _read_file_start_for_type_check(
-            file=file,
-            filename=filename,
-            encoding=encoding,
-        )
+        file_text = _FileTypeDetectionContext.new(
+            file_path=filename, file=file, encoding=encoding
+        ).text_head
```
scanny, line 119: Same with this change subsuming `_read_file_start_for_file_check()`, this just got left out of the prior refactoring PR.

unstructured/file_utils/filetype.py
```diff
@@ -351,19 +358,19 @@ def mime_type(self) -> str | None:
 
         A `str` return value is always in lower-case.
         """
+        file_path = self.file_path
```
Coniferish, line 361: why declare this instead of using `self.file_path` down below?

Coniferish (approved):
LGTM

PR #3443, opened 2024-07-29
feat: Add deprecation warning on import of any ingest code

rbiseck3:
### Description
Any time `unstructed.ingest` is imported, this deprecation warning gets emitted:
```
DeprecationWarning: unstructured.ingest will be removed in a future version
```

### DIFF
--- a/unstructured/ingest/__init__.py
+++ b/unstructured/ingest/__init__.py
@@ -1,1 +1,7 @@
 from __future__ import annotations
+
+import warnings
+
+warnings.warn(
+    "unstructured.ingest will be removed in a future version", DeprecationWarning, stacklevel=2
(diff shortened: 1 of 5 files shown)


unstructured/ingest/__init__.py
```diff
@@ -1 +1,7 @@
 from __future__ import annotations
+
+import warnings
```
MthwRobinson, line 3: Should we use the logger from `unstructurued.logger` for this?

rbiseck3: From what I read, this is the recommended way to mark a warning.

MthwRobinson (approved):
LGTM!

PR #3444, opened 2024-07-30
feat: msg and email metadata

Coniferish:
Update partition_eml and partition_msg to capture cc, bcc, and message id fields.

Docs PR: https://github.com/Unstructured-IO/docs/pull/135/files

Testing
```
from test_unstructured.unit_utils import example_doc_path

elements = partition_email(filename=example_doc_path("eml/fake-email-header.eml"), include_headers=True)
print(elements)
elements[0].metadata.to_dict()
```

Note to reviewers:
Tests in `test_unstructured/partition/test_email.py` were refactored and rearranged to group similar tests together, so it will be easiest to review those changes commit by commit.


test_unstructured/partition/test_msg.py
```diff
@@ -379,6 +379,24 @@ def it_adds_email_specific_fields_to_the_msg_element_metadata(self, opts_args: d
...
+            "john-ctr@unstructured.io",
+            "steve@unstructured.io",
+            "hello@unstructured.io",
+        ]
+
+    def it_captures_message_id_element_metadata(self, opts_args: dict[str, Any]):
+        opts_args["file_path"] = example_doc_path("fake-email-with-cc-and-bcc.msg")
+        opts = MsgPartitionerOptions(**opts_args)
+
+        assert opts.msg_metadata.message_id == "<14DDEF33-2BA7-4CDD-A4D8-E7C5873B37F2@gmail.com>"
```
Coniferish, line 398: Should the angle brackets stay or be removed for `message_id`?

MthwRobinson: Based on this from the doc, I'd say remove the angle brackets

> Message-ID consists of a long string of characters (unique identifier of this message on the mail server) ending with the FQDN of the mail server (e.g. 25eb7e48-2252-48ea-80ce-cf2f6119a8e3@ENV95-E2013-1.domain95.lab).

Coniferish: Done :)

CHANGELOG.md
```diff
@@ -1,11 +1,11 @@
-## 0.15.1-dev7
+## 0.15.1-dev8
 
 ### Enhancements
 
 ### Features
 
+* **Update partition_eml and partition_msg to capture cc, bcc, and message_id fields** Cc, bcc, and message_id information is captured in element medatada for both msg and email partitioning and `Recipient` elements are generated for cc and bcc when `include_headers=True` for email partitioning.
```
MthwRobinson, line 7: Small typo: `medatada` -> `metadata`.

test_unstructured/partition/test_email.py
```diff
@@ -540,135 +650,29 @@ def test_partition_email_can_process_min_max_with_attachments(
             assert len(element.text) >= 6
 
 
-def test_partition_msg_raises_with_no_partitioner(
```
MthwRobinson, line 543: Did these tests get removed or is that just the diff being weird?

unstructured/partition/email.py
```diff
@@ -131,21 +166,29 @@ def build_email_metadata(
...
+    bcc_recipient = parse_recipients(header_dict.get("Bcc"))
+    cc_recipient = parse_recipients(header_dict.get("Cc"))
+    message_id = header_dict.get("Message-ID")
+    if message_id:
+        message_id = _strip_angle_brackets(message_id)
 
     element_metadata = ElementMetadata(
+        bcc_recipient=bcc_recipient,
+        cc_recipient=cc_recipient,
+        message_id=message_id,
```
MthwRobinson, line 191: @scanny - What do you think about `email_message_id` vs `message_id` for this? We also have our own ID, was thinking we might want to differentiate.

test_unstructured/partition/test_email.py
Coniferish: The diff being weird from my refactoring and grouping similar tests together in the second commit

MthwRobinson: Gotcha, thanks!

unstructured/partition/email.py
```diff
@@ -83,6 +83,21 @@ def _parse_received_data(data: str) -> list[Element]:
 
 
 def _parse_email_address(data: str) -> tuple[str, str]:
+    """Extract and return the name and email address from the given email header.
+
+    This function assumes that the email address is enclosed in angle brackets and follows
+    the format specified by the PATTERN regex.
```
scanny, line 89: I'm not sure this is quite a safe assumption, is it? An email address can also appear as:
```
From: steve@unstructured.io
```
can't it?

Also, I'm pretty sure that strings with spaces in them need to be quoted by themselves, like the "John Doe" nickname in `"John Doe" <john.doe@example.com>`.

I think we're going to want a separate parameterized unit test for `_parse_email_address()` that has test cases for all the varieties we can think of. The variations would be with and without quotes, with and without nickname+<mailbox>, and maybe a few others.

scanny (changes requested):
I think we need to have better tests for mailbox parsing in From: To: etc. header fields. I know you didn't add that parsing in this commit but I think we need to fill the gap while we're in here. Also I think there are valid cases the regex doesn't catch.

unstructured/partition/email.py
scanny: Btw, parsing this is probably a little too much to ask of regular expressions. I'd look into this stdlib utility method as an alternative: https://docs.python.org/3/library/email.utils.html#email.utils.parseaddr

I'm sure they've given it more thought and testing than we'd like to invest ourselves :)

unstructured/partition/email.py
scanny: Yeah, I'd say so. This is in metadata, right? So folks will be encountering that field when email is far from their mind, so good to qualify it I'd say. Like we do `image_base64` and `image_mime_type`.

`message_id` is already established though, right? So not sure how to go about changing something like that. I suppose we either assume no on is using it and just go for it, or we populate both for a while and remove the old one later.

unstructured/partition/email.py
Coniferish: Whether or not we can assume angle brackets and the likes depends on the standard for email headers I think. This function is only called by `partition_email_header` when iterating through `Message.raw_items()`, which has seemed like it will always have the angle brackets, but I'll look into it

unstructured/partition/email.py
Coniferish: No, it was added as a metadata field in this, so there shouldn't be any problem with changing it.

unstructured/documents/elements.py
```diff
@@ -502,6 +513,7 @@ def field_consolidation_strategies(cls) -> dict[str, ConsolidationStrategy]:
             "link_urls": cls.LIST_CONCATENATE,
             "link_start_indexes": cls.DROP,
             "links": cls.DROP,  # -- deprecated field --
+            "message_id": cls.FIRST,
```
Coniferish, line 516: @scanny, I wanted to double-check that this was the correct consolidation strategy to use.

scanny: We would expect the value to be the same for all elements in the element stream, right? This is basically what the `FIRST` strategy is for. It accommodates a situation in which some elements might have `None` for this value, but once it encounters a non-`None` value, the first one is as good as any other (because they're all the same).

Coniferish: Yah, they would all be the same

unstructured/partition/email.py
Coniferish: @MthwRobinson and @scanny, I figured out how to use `email` to get addresses more reliably. In doing so, I noticed that we are returning the full header value (display_name and email address) for `From` and `To` and am wondering if we want to make these just the email address or keep them as they are. We've been talking about just returning the email address for the `Cc` and `Bcc` fields, so I'm wondering if we want to treat all the addresses the same way or not.

```
E       AssertionError: assert {'email_messa.../rfc822', ...} == {'email_messa.../rfc822', ...}
E
E         Omitting 7 identical items, use -vv to show
E         Differing items:
E         {'sent_from': 'mrobinson@unstructured.io'} != {'sent_from': ['Matthew Robinson <mrobinson@unstructured.io>']}
```

MthwRobinson: `Cc` and `Bcc` should be consistent with the current behavior for `From` and `To`

scanny: I'd be inclined to include the nickname when present. I suppose the rationale for me is:
- removing it would be "destructive"
- removing it would make the field less human-readable

I'd say whatever we take out though ought to be round-trippable. Like still a valid RFC822 email address such that you can give that string to `email.parse` or whatever and it can reliably parse it as an email address.

unstructured/partition/email.py
```diff
@@ -82,40 +83,59 @@ def _parse_received_data(data: str) -> list[Element]:
...
+    'another>test'
+    >>> _strip_angle_brackets("<<edge>>")
+    '<edge>'
+    """
+    return re.sub(r"^<|>$", "", data)
+
+
+def partition_email_header(msg: EmailMessage) -> list[Element]:
 
+    def append_address_header_elements(header: AddressHeader, element_type: Type[Element]):
```
Coniferish, line 105: Some context for changing the `msg` type to `EmailMessage`:
"The central class in the [email](https://docs.python.org/3/library/email.html#module-email) package is the [EmailMessage](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage) class, imported from the [email.message](https://docs.python.org/3/library/email.message.html#module-email.message) module. It is the base class for the [email](https://docs.python.org/3/library/email.html#module-email) object model. [EmailMessage](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage) provides the core functionality for setting and querying header fields, for accessing message bodies, and for creating or modifying structured messages."

"There is a default policy used by all classes in the email package. For all of the [`parser`](https://docs.python.org/3/library/email.parser.html#module-email.parser "email.parser: Parse flat text email messages to produce a message object structure.") classes and the related convenience functions, and for the [`Message`](https://docs.python.org/3/library/email.compat32-message.html#email.message.Message "email.message.Message") class, this is the [`Compat32`](https://docs.python.org/3/library/email.policy.html#email.policy.Compat32 "email.policy.Compat32") policy, via its corresponding pre-defined instance [`compat32`](https://docs.python.org/3/library/email.policy.html#email.policy.compat32 "email.policy.compat32"). This policy provides for complete backward compatibility (in some cases, including bug compatibility) with the pre-Python3.3 version of the email package. This default value for the _policy_ keyword to [`EmailMessage`](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage "email.message.EmailMessage") is the [`EmailPolicy`](https://docs.python.org/3/library/email.policy.html#email.policy.EmailPolicy "email.policy.EmailPolicy") policy, via its pre-defined instance [`default`](https://docs.python.org/3/library/email.policy.html#email.policy.default "email.policy.default")."

"The default value for the _policy_ keyword for the [`email.parser`](https://docs.python.org/3/library/email.parser.html#module-email.parser "email.parser: Parse flat text email messages to produce a message object structure.") classes and the parser convenience functions **will be changing** in a future version of Python. Therefore you should **always specify explicitly which policy you want to use** when calling any of the classes and functions described in the [`parser`](https://docs.python.org/3/library/email.parser.html#module-email.parser "email.parser: Parse flat text email messages to produce a message object structure.") module."

test_unstructured/partition/test_email.py
```diff
@@ -110,24 +119,17 @@
 
 
 def test_partition_email_from_filename():
-    filename = os.path.join(EXAMPLE_DOCS_DIRECTORY, "fake-email.eml")
-    elements = partition_email(filename=filename)
+    elements = partition_email(filename=example_doc_path("eml/fake-email.eml"))
```
scanny, line 122: Just for future reference, these are more compact and easier to scan if you remove the `filename=` bit. `filename` is always the first parameter and can be provided positionally. Since this form appears so often in tests I think it improves readability if it's reduced to "partition(this)" :)

scanny (approved):
LGTM. Could remove the `filename=` bit I mentioned if you have another CI run.

test_unstructured/partition/test_email.py
Coniferish: Thanks. Didn't think about that, but will try to remember to do so :)

PR #3458, opened 2024-08-01
chore: rename astra to astradb

potter-potter:
DataStax wanted all references to be astradb instead of astra. As per @erichare

We'll also have to do the same in unstructured-ingest :)


ahmetmeleq:
I think this needs an update to the fixtures folder name:
https://github.com/Unstructured-IO/unstructured/tree/potter/update-astradb-naming/test_unstructured_ingest/expected-structured-output/astra/ingest_test_src

ahmetmeleq (approved):
Looks good! Should be good to merge after the tests pass.

(IMO) Optional points:

In case comments are relevant:
- https://github.com/Unstructured-IO/unstructured/blob/b749b891a7d8508ec835277653613260fb2976ed/test_unstructured_ingest/dest/mongodb.sh#L39
- https://github.com/Unstructured-IO/unstructured/blob/b749b891a7d8508ec835277653613260fb2976ed/test_unstructured_ingest/src/mongodb.sh#L25

If past Changelog items are relevant:
- https://github.com/Unstructured-IO/unstructured/blob/main/CHANGELOG.md#fixes-1

If "AstraDB" is preferred rather than "Astra DB":
image

PR #3473, opened 2024-08-02
fix: disabled checking max pages for images

mackurzawa:
Added fix related to https://github.com/Unstructured-IO/unstructured/pull/3431, which disables checking max pages for images


pawel-kmiecik (approved):
LGTM!

Issue #3381, opened 2024-07-10
bug(pptx): invalid MIME-type `image/jpg` not accommodated


scanny:
**Summary**
`image/jpg` is not a MIME-type. `image/jpeg` is. Certain file converters, possibly Adobe Online PDF Converter (https://www.adobe.com/acrobat/online/pdf-to-ppt.html) incorrectly map JPEG images in the PDF content to the (non-)MIME-type `image/jpg` in the PPTX output.

PPTX files with this invalid MIME-type do not extract images and raise an exception when submitted via the API with `strategy` other than "fast".
```python
AttributeError: 'Part' object has no attribute 'image'
```
This file exhibits this behavior:
[TestPPTX_PADI.pptx](https://github.com/user-attachments/files/16169360/TestPPTX_PADI.pptx)

PPTX files processed locally do not raise an exception (because images are ignored).

**Proposed solution**
Update `python-pptx` upstream to accommodate `image/jpg` as an alias for `image/jpeg`.

Probably related to this upstream issue:
https://github.com/scanny/python-pptx/issues/929

PR #3475, opened 2024-08-03
fix(pptx): accommodate invalid image/jpg MIME-type

scanny:
As described in #3381, some clients, perhaps including Adobe PDF Converter, map JPEG images to the invalid `image/jpg` MIME-type. Prior to v1.0.0, `python-pptx` would not load these images, which caused image extraction to fail.

Update the `python-pptx` dependency to `v1.0.1` or above to ensure this upstream fix is always available.

Fixes: #3381


scanny:
Hi Matt, not sure what the process is for this these days. Should I run `make pip-compile` on this PR or just wait for this to get picked up by the next scheduled weekly run?

MthwRobinson:
@scanny - You can run `make pip-compile` on this PR, and we can skip the dedicated dependency bump PR this week since this will cover it.

MthwRobinson (approved):
LGTM assuming tests pass

PR #3480, opened 2024-08-05
chore: small edits

Coniferish:
Add comments and fix decorators on some tests.


test_unstructured/staging/test_weaviate.py
```diff
@@ -53,6 +54,7 @@ def test_stage_for_weaviate(filename="example-docs/layout-parser-paper-fast.pdf"
     }
 
 
+@pytest.mark.skipif(skip_outside_ci, reason="Skipping test run outside of CI")
```
scanny, line 57: I know there are a lot like this, but this reason is not a reason, it just restates the "skip_outside_ci" label. A reason would answer the question "Why do we skip this when we're not running in the CI environment?"

test_unstructured/staging/test_weaviate.py
```diff
@@ -17,9 +17,10 @@
 )
 
 is_in_docker = os.path.exists("/.dockerenv")
+skip_outside_ci = os.getenv("CI", "").lower() in {"", "false", "f", "0"}
```
scanny, line 20: This is a predicate (resolves to True or False) but is named like a function.

I'd say something like "is_in_ci" would be a better name, flip the logic to "not in" and then use `not is_in_ci` in the `skipif` expression.

requirements/deps/constraints.txt
```diff
@@ -1,7 +1,8 @@
 ####################################################################################################
 # This file can house global constraints that aren't *direct* requirements of the package or any
 # extras. Putting a dependency here will only affect dependency sets that contain them -- in other
-# words, if something does not require a constraint, it will not be installed.
+# words, if something does not require a constraint, it will not be installed. Refer to the "Adding
+# a new Dependency Group" notes in Notion for more info about managing requirements.
```
scanny, line 5: Open-source users aren't going to have access to the internal Notion. Not sure exactly what to do about that but something to think about.

scanny (changes requested):
A few suggested changes.

requirements/deps/constraints.txt
Coniferish: That did cross my mind and I was wondering if we should just add a README or something with the contents of that Notion page. It's just general info that can be a good reference, even for potential contributors. Lmk if you have any more thoughts or if we should bring someone else to weigh in

scanny: I'd ask Matt and let him make the call.

test_unstructured/partition/test_api.py
```diff
@@ -102,7 +102,7 @@ def test_partition_via_api_raises_with_bad_response(request: FixtureRequest):
     partition_mock_.assert_called_once()
 
 
-@pytest.mark.skipif(skip_outside_ci, reason="Skipping test run outside of CI")
+@pytest.mark.skipif(not is_in_ci, reason="Skipping test run outside of CI")
```
scanny, line 105: Are the rest of these skipped for the same reason?

Coniferish: I don't know. I didn't look into it

requirements/deps/constraints.txt
Coniferish: @MthwRobinson ?

MthwRobinson: I'd leave references to our internal Notion out of the open source repo.

scanny (approved):
k, good enough for me. Maybe ping Matt about those non-reason reasons and we can clarify those later.

Issue #2616, opened 2024-03-06
Unable to parse single-column csv files


cwang:
**Describe the bug**
Given a single column csv file (see one example as attached), it fails to parse it because of the failure of determining the delimiter. 
See https://github.com/Unstructured-IO/unstructured/blob/4096a38371bae062832b976dc7ebff4184b7991f/unstructured/partition/csv.py#L109 for where the issue is. 

**To Reproduce**
Use the file attach to be parsed. 

**Expected behavior**
It should be parsed successfully and treated as a single-column spreadsheet.

**Screenshots**

**Environment Info**
unstructured.io 0.11.8 
Python 3.10

**Additional context**
Here's a 16yo Python bug that's relevant to the `sniff` function used here for the expected behaviour, https://bugs.python.org/issue2078:
[test_single_column.csv](https://github.com/Unstructured-IO/unstructured/files/14512026/test_single_column.csv)

MthwRobinson:
Thanks @cwang - will take a look as soon as we are able

cbrendanprice:
just ran into this today. seems like a pretty simple fix; just need to catch the error on [invocation of `get_delimiter`](https://github.com/Unstructured-IO/unstructured/blob/0.15.0/unstructured/partition/csv.py#L90) and conditionally skip passing the `sep` argument on failure. happy to create a pr for this if that's preferable.

scanny:
@cbrendanprice can you post the full stack trace?

cbrendanprice:
sure. im also creating a pr for this presently. let me know if i should stop.

```
Traceback (most recent call last):
  File "/home/notebook-user/.local/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py", line 399, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/notebook-user/.local/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 70, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/notebook-user/.local/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
[...]

PR #3483, opened 2024-08-05
rfctr(csv): accommodate single column CSV files

scanny:
**Summary**
Improve factoring, type-annotation, and tests for `partition_csv()` and accommodate single-column CSV files.

Fixes: #2616


test_unstructured/partition/test_csv.py
```diff
@@ -263,15 +253,158 @@ def test_partition_csv_respects_languages_arg():
...
+        get_last_modified_date_from_file_.return_value = "2024-08-04T13:17:47"
+        file = io.BytesIO(b"abcdefg")
+        ctx = _CsvPartitioningContext(file=file, date_from_file_object=True)
+
+        last_modified = ctx.last_modified
+
+        get_last_modified_date_from_file_.assert_called_once_with(file)
+        assert last_modified == "2024-08-04T13:17:47"
+
+    def but_it_falls_back_to_None_for_the_last_modified_date_when_date_from_file_object_is_False(
```
Coniferish, line 362: Is there still a test case for "long lines" like the test that was removed:
```
def test_partition_csv_detects_the_right_csv_delimiter():
    assert get_delimiter("example-docs/csv-with-long-lines.csv") == ","
```

typings/lxml/_types.pyi
```diff
@@ -23,6 +25,9 @@ _TagName: TypeAlias = str
 
 _TagSelector: TypeAlias = _TagName | Callable[..., _Element]
 
+# String argument also support QName in various places
+_TextArg: TypeAlias = str | bytes | QName
```
Coniferish, line 29: I'm not understanding this comment

Coniferish (approved):
a few comments, but lgtm

test_unstructured/partition/test_csv.py
scanny: Yeah that one moved into the unit tests as a test-case.

typings/lxml/_types.pyi
scanny: That comment just came along for the ride from `lxml-stubs`. But I believe the idea is a qualified XML Schema name. So an argument name like `r:rId` that has a namespace ("r:") as well as a name ("rId").

I doubt it will ever come into play for us.

Issue #3449, opened 2024-07-30
bug/Unable to parse EML files with MSIP Labels


TaylorN15:
**Describe the bug**
When processing an e-mail in EML format using `partition_email` no content is extracted when there are MSIP (Microsoft Information Protection) Labels in the header.

**To Reproduce**
[ExampleEmail.eml.txt](https://github.com/user-attachments/files/16433935/ExampleEmail.eml.txt)
```
from unstructured.partition.email import partition_email
elements = partition_email(filename="ExampleEmail.eml")
print(elements)
```

I'm not really sure why this is occuring, but when I remove the `msip_labels` from the file, it get's parsed properly. Otherwise, it seems to only extract `multipart/mixed`, `multipart/related`, `multipart/alternative` into the `content_map` in `email.py` which is not in `VALID_CONTENT_SOURCES`

TaylorN15:
Some debugging I have done, I added the below to line 361 in [email.py](https://github.com/Unstructured-IO/unstructured/blob/8fd216cc9fbb25704281e50f498a92f169b462b6/unstructured/partition/email.py#L361)

```
    for part in msg.walk():
        print(f"Content Disposition Type: {part.get_content_disposition()}")
        print(f"Content Type: {part.get_content_type()}")
        print(f"Content Main Type: {part.get_content_maintype()}")
        print(part.get_payload())
```

Which produces the following output:

```
Content Disposition Type: None
Content Type: multipart/mixed
Content Main Type: multipart
[<email.message.Message object at 0x000001FC5317F5D0>]
Content Disposition Type: None
Content Type: multipart/related
Content Main Type: multipart
[<email.message.Message object at 0x000001FC5317D5D0>]
Content Disposition Type: None
Content Type: multipart/alternative
Content Main Type: multipart
[<email.message.Message object at 0x000001FC5317F8D0>]
Content Disposition Type: inline
Content Type: text/plain
Content Main Type: text
Hi

=20

Please find attached a project proposal.

Please let us know if you have any questions or would like to discuss=
 anything at this stage.

=20

Kind regards

User

=20
```

But it doesn't add the `text/plain` part to the content_map here `content_map[content_type] = part.get_payload()`.

scanny:
@TaylorN15 I expect this is a limitation in the Python email parser.

You can see the partitioning code starts off with parsing the email using the Python stdlib `email` module:
[...]

PR #3489, opened 2024-08-07
Added 'inline' to content disposition check

TaylorN15:
See discussion [here](https://github.com/Unstructured-IO/unstructured/issues/3449). This adds `inline` to the allowed list of Content Disposition values when parsing EML files.

### DIFF
--- a/unstructured/partition/email.py
+++ b/unstructured/partition/email.py
@@ -387,9 +387,9 @@ def partition_email(
     is_encrypted = False
     content_map: dict[str, str] = {}
     for part in msg.walk():
-        # NOTE(robinson) - content dispostiion is None for the content of the email itself.
+        # content dispostiion is None/inline for the content of the email itself.
         # Other dispositions include "attachment" for attachments
-        if part.get_content_disposition() is not None:
+        if part.get_content_disposition() not in (None, "inline"):
             continue
         content_type = part.get_content_type()
 


cragwolfe:
CC @MthwRobinson :)

TaylorN15:
@scanny @cragwolfe - one of the tests is failing but unrelated to my change. I’ll leave it for you guys to fix?

MthwRobinson:
@TaylorN15 - Is there a test you could add for this?

TaylorN15:
> @TaylorN15 - Is there a test you could add for this?

I checked and there is already a test EML file with inline and None for content disposition within the test data 😃

MthwRobinson:
Awesome! Would we have expected that test to fail prior to the fix?

TaylorN15:
No, it would have passed except the inline content would not have been returned. I thought that was a bit strange that it was in the test file but not catered for in the code. 

¯\_(ツ)_/¯

MthwRobinson:
In that case, could we add a test that asserts that the inline content is included now?

TaylorN15:
> In that case, could we add a test that asserts that the inline content is included now?

It seems the test already has an assertion for content disposition of `inline` and `None` and the test passes in the current and this PR version of the code...

scanny:
@TaylorN15 here's what we need: A test that fails before your change and passes afterward.

If the current test can be modified to suit, that's fine. I can see the assertions there are pretty weak, maybe you can add something there.

But what I would do is this:
- take your original problematic email with the MSIP labels, make a copy of it and anonymize it as necessary but keep all the parts.
- write a test that does what you were trying to do and fails when run on `main`, just like the failure you were seeing, like no text content or whatever it was.
- add that test in this PR and demonstrate that it passes (passing CI is this demonstration).

Basic idea is, if we don't have a test that fails when we take out your change, how do we know your change fixed it?

Also it gives us a basis for reasoning about whether your change fixed the general-case problem or only an isolated use case.

TaylorN15:
@scanny I understand, I'm just not so great at writing tests :)
I need to work out what the difference is between the EML files to understand why one works when the other does not. I am guessing it's something to do with setting `process_attachments` as I don't do that with my version, but the code that processes attachments includes content disposition of `inline` and `attachment`

scanny:
@TaylorN15 why don't you post an anonymized version of the problematic message and I'll take a look. I expect I can whip up an appropriate test in a minute or two :)

TaylorN15:
> @TaylorN15 why don't you post an anonymized version of the problematic message and I'll take a look. I expect I can whip up an appropriate test in a minute or two :)

The original email I linked should suffice. I have just confirmed though that the issue is whether `process_attachments` is set or not. If true, all the content with a content disposition (including inline) is processed, however in my application I am not processing attachments which is how I first discovered this. The text in my email with the content disposition of `inline` isn't an attachment though, so I think the assumption that all message body/text has no content disposition is false.

scanny:
k, this test should be plenty:
```python
def test_partition_reads_message_part_with_inline_content_disposition():
    elements = partition_email(
        example_doc_path("eml/text-part-marked-inline.eml"), process_attachments=False
    )

    assert len(elements) == 1
    e = elements[0]
    assert e.text.startswith("Hi   Please find attached a project proposal.")
    assert e.text.endswith("Kind regards User  ")
```

Just name your original file `text-part-marked-inline.eml` and add it to `example-docs/eml/`.

PR #3504, opened 2024-08-09
feat(chunk): split tables on even row boundaries

scanny:
**Summary**
Use more sophisticated algorithm for splitting oversized `Table` elements into `TableChunk` elements during chunking to ensure element text and HTML are "synchronized" and HTML is always parseable.

**Additional Context**
Table splitting now has the following characteristics:
- `TableChunk.metadata.text_as_html` is always a parseable HTML `<table>` subtree.
- `TableChunk.text` is always the text in the HTML version of the table fragment in `.metadata.text_as_html`. Text and HTML are "synchronized".
- The table is divided at a whole-row boundary whenever possible.
- A row is broken at an even-cell boundary when a single row is larger than the chunking window.
- A cell is broken at an even-word boundary when a single cell is larger than the chunking window.
- `.text_as_html` is "minified", removing all extraneous whitespace and unneeded elements or attributes. This maximizes the semantic "density" of each chunk.


test_unstructured/common/test_html_table.py
```diff
@@ -33,3 +41,163 @@ def test_htmlify_matrix_handles_multiple_rows_and_cells(self):
...
+            "<table><tbody><tr><td>foobar</td></tr></tbody><tfoot></tfoot></table>",
+        ],
+    )
+    def it_removes_any_thead_tbody_or_tfoot_elements_present_within_the_table_element(
+        self, html_text: str
+    ):
+        html_table = HtmlTable.from_html_text(html_text)
+        assert html_table.html == "<table><tr><td>foobar</td></tr></table>"
+
+    def it_changes_any_th_elements_to_td_elements_for_cell_element_uniformity(self):
```
Coniferish, line 93: This doesn't need to change, but can you explain why this behavior is desired?

scanny: Basically, the `.text_as_html` we receive is not uniform. There are seven different partitioners that produce `.text_as_html` and they all have their own way of doing it. Some use `<td>` only, others sometimes use `<th>` etc.

Someone programming to `.text_as_html` shouldn't have to concern themselves with which partitioner originally produced the HTML, it should just be consistent. So this makes all the `.text_as_html` coming out of chunking look the same, except for the content and the number of rows and columns.

In a stacked PR I make the sources more consistent as well, but this acts as a backstop so will be belts and suspenders at that point. Also, PDF/Image is a bit of a wildcard since that's done by a model, so this will be something we'll want to leave in for the foreseeable future.

test_unstructured/chunking/test_base.py
```diff
@@ -1199,6 +1202,96 @@ def it_strips_whitespace_around_the_split(self, separators: Sequence[str]):
...
+        assert html == "<table><tr><td>abcde fghij klmno</td><td>pqrst uvwxy z</td></tr></table>"
+        assert accum._cells == []
+
+    def but_it_does_not_generate_a_TextAndHtml_pair_when_empty(self):
+        accum = _CellAccumulator(maxlen=100)
+
+        with pytest.raises(StopIteration):
+            next(accum.flush())
+
+
```
Coniferish, line 1471: very nice tests

test_unstructured/chunking/test_base.py
```diff
@@ -1102,6 +1103,168 @@ def it_knows_the_concatenated_text_of_the_pre_chunk_to_help(
 # ================================================================================================
 
 
+class Describe_TableSplitter:
+    """Unit-test suite for `unstructured.chunking.base._TableSplitter`."""
+
+    def it_splits_an_HTML_table_on_even_rows_when_possible(self):
```
Coniferish, line 1122: "even" potentially sounds like even numbered rows (0,2,4...)

Coniferish (approved):
LGTM

PR #3510, opened 2024-08-11
rfctr(chunk): prep for adding TableSplitter

scanny:
**Summary**
Mechanical refactoring in preparation for adding (pre-chunk) `TableSplitter` in a PR stacked on this one.

### DIFF
--- a/typings/lxml/html/__init__.pyi
+++ b/typings/lxml/html/__init__.pyi
@@ -1,0 +1,8 @@
+from __future__ import annotations
+
+from ._element import (
+    HtmlElement as HtmlElement,
+)
+from ._parse import (
+    fragment_fromstring as fragment_fromstring,
+)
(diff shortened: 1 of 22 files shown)


typings/lxml/html/__init__.pyi
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+from ._element import (
+    HtmlElement as HtmlElement,
```
Coniferish, line 4: Why use the `import/as` syntax here if you're not changing the name?

Coniferish (approved):
LGTM

typings/lxml/html/__init__.pyi
scanny: It's commonly done that way in type-stubs where the file is just "republishing" an import from somewhere else. It avoids an "unused import" error, which I suppose is why the original author does that and definitely why I keep it that way.

The other way I know of to avoid the unused-import error is to use `__all__ = ["HtmlElement"]` for example, but this is more compact and a straight copy-paste from the stub packages these come from.

PR #3524, opened 2024-08-15
build: update to latest base image

MthwRobinson:
### Summary

Updates to the latest `wolfi-base` base image to pull in more recent package version. A notable update is that upgrading to `libreoffice==24.2.5.2` resolves several CVEs.


tabossert (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/unstructured-io/base-images:wolfi-base-e48da6b@sha256:8ad3479e5dc87a86e4794350cca6385c01c6d110902c5b292d1a62e231be711b as base
+FROM quay.io/unstructured-io/base-images:wolfi-base-e48da6b@sha256:f495978575bab94c2e20d09f286d73af55ae8ea83a255f5eb8794121b70f30fb as base
```
christinestraub, line 1: What about using `FROM quay.io/unstructured-io/base-images:wolfi-base-latest as base`​ like in https://github.com/Unstructured-IO/core-product/blob/main/unstructured-api/Dockerfile#L5? So we don't have to update the base image every time whenever it's updated?

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.15.5
```
christinestraub, line 1: is it possible to change version to `0.15.5-dev0`? I want to include a PR to fix issue in publishing `0.15.4` to PyPI in this release `0.15.5`. Details are in https://unstructuredai.slack.com/archives/C049XRFGR1R/p1723662226499499?thread_ts=1723658194.037349&cid=C049XRFGR1R

MthwRobinson: Ahh okay cool! Updated to `-dev0`

Dockerfile
MthwRobinson: Yeah that's a good call. Just updated to `latest` with no hash. And if `latest` ever breaks the build for whatever reason, we can alway pin to an older version.

christinestraub (approved):

tabossert (approved):

PR #3512, opened 2024-08-12
fix(CVE-2024-39705): update to latest `nltk` version

MthwRobinson:
### Summary

Addresses [CVE-2024-39705](https://nvd.nist.gov/vuln/detail/CVE-2024-39705) by updating to `nltk==3.8.2` and closes #3511. This CVE had previously been mitigated in #3361.

PR #3527, opened 2024-08-15
build: downgrade `nltk` version

christinestraub:
This PR aims to roll back `nltk` to `3.8.1` which bumped to `3.8.2` in https://github.com/Unstructured-IO/unstructured/pull/3512 because `3.8.2` is no longer available in PyPI due to some issues(https://github.com/nltk/nltk/issues/3301)

### DIFF
--- a/unstructured/nlp/tokenize.py
+++ b/unstructured/nlp/tokenize.py
@@ -16,9 +16,9 @@ from nltk import word_tokenize as _word_tokenize
 
 CACHE_MAX_SIZE: Final[int] = 128
 
-NLTK_DATA_FILENAME = "nltk_data_3.8.2.tar.gz"
+NLTK_DATA_FILENAME = "nltk_data.tgz"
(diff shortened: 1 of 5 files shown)


MthwRobinson (approved):
LGTM when tests pass

unstructured/nlp/tokenize.py
```diff
@@ -114,10 +114,10 @@ def _download_nltk_packages_if_not_present():
...
         package_category="taggers",
-        package_name="averaged_perceptron_tagger_eng",
+        package_name="averaged_perceptron_tagger",
     )
     tokenizer_available = check_for_nltk_package(
-        package_category="tokenizers", package_name="punkt_tab"
+        package_category="tokenizers", package_name="punkt"
     )
 
     if not (tokenizer_available and tagger_available):
```
maddosaientisuto, line 123: Hi,

The condition on line 120 should be 'or', otherwise it fails to download package when one of the taggers or tokenizers is missing,
    `if not (tokenizer_available or tagger_available):`

PR #3528, opened 2024-08-16
feat: replace `pytesseract` with `unstructured.pytesseract` fork

christinestraub:
This PR reverts `pytesseract` dependency to `unstructured.pytesseract` fork due to the unavailability of some recent release versions of `pytesseract` on PyPI. 

This PR also addresses an issue encountered during the publication of `unstructured==0.15.4` to PyPI. The error was due to the fact that PyPI does not allow direct dependencies from Version Control System URLs like GitHub in the `install_requires` or `extras_require` sections of the `setup.py` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.5-dev1
+## 0.15.5
 
 ### Enhancements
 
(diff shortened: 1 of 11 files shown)


MthwRobinson (approved):
LGTM

PR #3545, opened 2024-08-20
rfctr(docx): DOCX emits std minified .text_as_html

scanny:
**Summary**
Eliminate historical "idiosyncracies" of `table.metadata.text_as_html` HTML introduced by `partition_docx()`. Produce minified `.text_as_html` consistent with that formed by chunking.

**Additional Context**
- nested tables appear as their extracted text in the parent cell (no nested `<table>` elements in `.text_as_html`).
- DOCX `.text_as_html` is minified (no extra whitespace or thead, tbody, tfoot elements).

### DIFF
--- a/unstructured/partition/docx.py
+++ b/unstructured/partition/docx.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import html
 import io
 import itertools
 import os
@@ -23,11 +22,11 @@ from docx.text.hyperlink import Hyperlink
 from docx.text.pagebreak import RenderedPageBreak
 from docx.text.paragraph import Paragraph
 from docx.text.run import Run
-from tabulate import tabulate
 from typing_extensions import TypeAlias
 
 from unstructured.chunking import add_chunking_strategy
 from unstructured.cleaners.core import clean_bullets
+from unstructured.common.html_table import htmlify_matrix_of_cell_texts
 from unstructured.documents.elements import (
     Address,
     Element,
@@ -498,7 +497,7 @@ class _DocxPartitioner:
         # NOTE(scanny) - if all that fails we give it the default `Text` element-type
         yield Text(text, metadata=metadata, detection_origin=DETECTION_ORIGIN)
 
-    def _convert_table_to_html(self, table: DocxTable, is_nested: bool = False) -> str:
+    def _convert_table_to_html(self, table: DocxTable) -> str:
         """HTML string version of `table`.
 
         Example:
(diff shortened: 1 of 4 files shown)


unstructured/partition/docx.py
```diff
@@ -520,44 +519,39 @@ def _convert_table_to_html(self, table: DocxTable, is_nested: bool = False) -> s
...
+                if isinstance(paragraph := block_item, Paragraph):
                     # -- all docx content is ultimately in a paragraph; a nested table contributes
                     # -- structure only
-                    yield f"{html.escape(block_item.text)}"
-                elif isinstance(  # pyright: ignore[reportUnnecessaryIsInstance]
-                    block_item, DocxTable
-                ):
-                    yield self._convert_table_to_html(block_item, is_nested=True)
+                    # yield f"{html.escape(block_item.text)}"
+                    yield paragraph.text
```
Coniferish, line 529: Why did this yield change? And should the commented code line above it be removed?

unstructured/partition/docx.py
```diff
@@ -520,44 +519,39 @@ def _convert_table_to_html(self, table: DocxTable, is_nested: bool = False) -> s
...
 
         def iter_row_cells_as_text(row: _Row) -> Iterator[str]:
-            """Generate the text of each cell in `row` as a separate string.
+            """Generate the normalized text of each cell in `row` as a separate string.
 
-            The text of each paragraph within a cell is separated from the next by a newline
-            (`"\n"`). A table nested in a cell is first converted to HTML and then included as a
-            string, also separated by a newline.
+            The text of each paragraph within a cell is not separated. A table nested in a cell is
+            converted to to a normalized string of its contents and combined with the text of the
```
Coniferish, line 538:
```suggestion
            converted to a normalized string of its contents and combined with the text of the
```

unstructured/partition/docx.py
```diff
@@ -520,44 +519,39 @@ def _convert_table_to_html(self, table: DocxTable, is_nested: bool = False) -> s
...
 
             for cell in row.cells:
-                yield "\n".join(iter_cell_block_items(cell))
+                cell_text = " ".join(iter_cell_block_items(cell))
+                yield " ".join(cell_text.split())
 
-            # -- each omitted cell at the end of the row (also rare) gets the empty string --
+            # -- Each omitted cell at the end of the row (also rare) gets the empty string. --
             for _ in range(row.grid_cols_after):
                 yield ""
```
Coniferish, line 552: Do we need to cover the scenario where there is a row with a gap between cells of data? It seems like this just handles if there are cells before and/or after the data, but not in between. I'm not totally clear on what these scenarios are as it is, so disregard if it's not applicable. 
Also, just wanted to double-check that these paragraphs and tables should be joined with " " and not "\n"

Coniferish (approved):
Couple of comments, but lgtm in general

unstructured/partition/docx.py
scanny: Good catch, the comment on line 528 needs to go, I've taken that out.

Other than that, this change doesn't change overall behavior, just clarifies. Two things:
1. Use the walrus operator to give `block_item` a differentiated name based on the differentiated type it turns out to be. If we're in the first branch we know it's a paragraph, so call it that such that we don't have to trace back to remember what we're operating on.
2. Remove the `html.escape()` wrapper and leave it to `htmlify_matrix_of_cell_texts()` to take care of that for all text you give it. This avoids repetition and accidental omissions and simplifies use of `htmlify_matrix_of_cell_texts()`. Basically placing all responsibility for producing valid HTML into `htmlify_...()`.

unstructured/partition/docx.py
scanny: Yep, done :)

unstructured/partition/docx.py
scanny: There is no such thing as an omitted cell in the middle of a row, not in DOCX at least. Even omitted cells at the beginning and end is weird and DOCX is the only format that supports that as far as I know.

There can be _empty_ cells in the middle of a row, and those resolve to the empty string. But they still occupy a position in the matrix.

---

We _could_ separate blocks within a cell with `"\n"`, but that would be reduced to " " in the HTML during the process of whitespace normalization. There's no such thing as a `"\n"` in HTML as far as semantic content is concerned; whitespace between words is whitespace and is all resolved to a single space when the HTML is rendered.

What we _could_ do is place a `<br/>` element between paragraphs, or a `<br/><br/>`, but there would be no corresponding line breaks in the text representation of the cell so the two would be inconsistent and there are other problems as well I'd have to think to remember.

This is just a reflection of a broader conundrum with tables and representing a table as a single element.

Non-Table elements are each a paragraph. So there is no occurrence of "multiple paragraphs per element" for other elements. And how we convert a table, which is inherently multiple paragraphs, into a textual representation hasn't been carefully specified and has organically developed into something of a mixed bag.

In this and following commits I make this representation at least consistent, choosing the simplest thing that could possibly work as the baseline (clean-concatenated text (CCT)).

At some point perhaps the ML or SE folks will weigh in with a point-of-view on how that can be elaborated for better recognition or recall or whatever. At that point we could take another look and refine this.

unstructured/partition/docx.py
Coniferish: Sounds good. I was mostly looking at getting rid of `html.escape()`, but that separation of responsibilities makes sense :)

unstructured/partition/docx.py
Coniferish: > At some point perhaps the ML or SE folks will weigh in with a point-of-view on how that can be elaborated for better recognition or recall or whatever. At that point we could take another look and refine this.

Do we want to get feedback on this now?

PR #3546, opened 2024-08-20
fix: `nltk` data download path to prevent redundant nested directories

### LINKED ISSUE #3543: bug/partition_pdf-returns-empty-lists
**Describe the bug**
I am trying to partition a simple pdf file with `unstructured==0.16.1` (and `nltk==3.9.1`). 

If I run `partition_pdf` with `strategy="fast"`, this works, but returns an empty list. It is also very slow.

If I run it with the default strategy, the script is stuck forever with no output (or at least it takes so long that I give up on it before it finishes)

**To Reproduce**

Use the attached [sample.pdf](https://github.com/user-attachments/files/16678117/sample.pdf) file.

To reproduce the behaviour with `strategy="fast"`:

```
chunks = partition_pdf("sample.pdf", strategy="fast")
print(chunks)
```
To reproduce it with the default strategy:

```

# these two lines are suggested by an error when I run this in a fresh virtual environment 
import nltk
nltk.download('punkt_tab')
[sample.pdf](https://github.com/user-attachments/files/16678018/sample.pdf)


chunks = partition_pdf("sample.pdf")
print(chunks)
```

**Expected behavior**
This should result in a list of chunks from the PDF document.

**Environment Info**
I don't know where to get `scripts/collect_env.py` but the Python version is `3.12.4`

and the output of pip list (from a virutal environment this is running in) is attached as 
[piplist.txt](https://github.com/user-attachments/files/16678021/piplist.txt)

**Additional context**

christinestraub:
Closes #3543.

### Summary
This PR addresses an issue with the NLTK data download process. Previously, when downloading NLTK data, a nested "nltk_data" directory was created within the parent "nltk_data" directory if the parent directory already existed. This redundant directory structure led to two significant problems:
- errors in checking if data had already been downloaded, potentially causing redundant downloads in subsequent calls.
- failures in loading models from the downloaded NLTK data due to incorrect path resolution.

This fix modifies the NLTK data download logic to prevent creation of unnecessary nested directories. If the download path ends with "nltk_data" and that directory already exists, we now use the existing directory instead of creating a new nested one.

### Testing
CI should pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.15.7
+
+### Enhancements
+
+### Features
+
+### Fixes
+
+* **Fix NLTK data download path**. Modified the NLTK data download process to use the parent directory when the current path ends with "nltk_data". This change prevents redundant directory creation and ensures proper data storage location.
+
 ## 0.15.6
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


MthwRobinson (approved):
LGTM - thanks!

PR #3550, opened 2024-08-21
chore: remove scipy and packaging pins

Coniferish:
Remove scipy and packaging constraint pins


MthwRobinson (approved):
🎉 - LGTM if CI passes!

PR #3558, opened 2024-08-22
fix(xlsx): XLSX emits std minified .text_as_html

scanny:
**Summary**
Eliminate historical "idiosyncracies" of `table.metadata.text_as_html` HTML introduced by `partition_xlsx()`. Produce minified `.text_as_html` consistent with that formed by chunking.

**Additional Context**
- XLSX `.text_as_html` is minified (no extra whitespace or thead, tbody, tfoot elements).
- `table.text` is clean-concatenated-text (CCT) of table.


test_unstructured/partition/test_xlsx.py
```diff
@@ -388,6 +349,63 @@ def test_partition_xlsx_with_more_than_1k_cells():
...
+            "Stanley Cups Team Location Stanley Cups Blues STL 1 Flyers PHI 2 Maple Leafs TOR 13"
+        ),
+        Table(
+            "Stanley Cups Since 67 Team Location Stanley Cups Blues STL 1 Flyers PHI 2 Maple"
+            " Leafs TOR 0"
+        ),
+    ]
+
+
+def test_partition_xlsx_with_find_subtables_False_and_infer_table_structure_False_works():
```
Coniferish, line 373: Do we need a test case for `find_subtable=True, infer_table_structure=False`?

scanny: I'd actually be inclined not to have any tests like this :) This test boils down to "it doesn't raise for this arg combination", which is an impoverished test at best.

But I found a bug while I was in there on the html-minification and needed a test that triggered that bug so I could confirm my fix and defend against regression.

I think the broader remedy would be to break up `partition_xlsx()` a bit, like into a `_XlsxPartitioner` that we could write isolated unit tests for, but I didn't want to get into the extra scope.

Anyway, as it turns out, that particular case is handled by `test_partition_xlsx_infer_table_structure()` on line 95.

Coniferish: Sounds good to me then. I had to look up what these parameters both did and wasn't positive on their interaction. It did cross my mind to add `find_subtable` to the docstring, but I assumed you would be against that. The docs might need to be updated to reflect that xlsx uses `infer_table_structure`, though, because I did query the ai and it said "The infer_table_structure parameter is not directly applicable to XLSX files in Unstructured."

scanny: I think both those have a place in the docstring. Definitely `find_subtable`, since it's XLSX-specific. `infer_table_structure` too because it's used in the partitioner code.

I've added `find_subtable`. `infer_table_structure` was already there. Sounds like the AI got that wrong.

Issue #3390, opened 2024-07-12
PermissionError: [Errno 13] Permission denied: 


yufengaotianya:
Problem description The system is Windows 11. When reading a txt document, PermissionError: [Errno 13] Permission denied: appears. The code is as follows:
```python
from unstructured.partition.auto import partition
partition("data_input\cureus\cureus-0015-00000040274.txt")
```
The problem code is located as follows:
```python
tokenize.py
with tempfile.NamedTemporaryFile() as tmp_file:
    tgz_file = tmp_file.name
urllib.request.urlretrieve(NLTK_DATA_URL, tgz_file)
```

I tested the following usage and it is also incorrect, and the same problem occurs;
```python
with tempfile.NamedTemporaryFile() as tmp_file:
    tfp = open(tmp_file.name, 'wb')
```
After checking the document, I found that the following code is normal on the linux system, but not supported on the windows system. So how to use it correctly on the windwos system partition("data_input\cureus\cureus-0015-00000040274.txt")

scanny:
@yufengaotianya please provide a full stack trace with line numbers so we can see where the problem is occurring.

Also, if you want us to understand the text please translate it to English for us :) I adjusted this one using Google Translate: https://translate.google.com/?sl=auto&tl=en&op=translate

@Coniferish probably something for you to look at while you're in there on the `txt` partitioner. I think what we're going to want to do is use `with tempfile.TemporaryDirectory() as tmpdir:` rather than `NamedTemporaryFile` which has known problems under Windows.

Seanmjeong:
[errno13.txt](https://github.com/user-attachments/files/16230022/errno13.txt)
Hi, I've recently ran into this issue as well. 

I've attached the trace. Hopefully it will help.

scanny:
Okay, thanks for this @seanmj :)

I think I've found the problem. Should be pushing the fix tomorrow some time.

In the meantime if you revert to v0.14.9 that should avoid this this problem for the time being. Let us know how you go :)

ecarpentieri:
Hi, 
I have the same problem when trying to read a docx!

I will try the last proposed solution of downgrading to v0.14.9

ecarpentieri:
> ```shell
> pip install -U unstructured==0.14.9
> ```

this solution works! ty

hamitguner:
> ```shell
> $ pip install -U unstructured==0.14.9
> ```

It worked for me too.

Coniferish:
[...]

PR #3562, opened 2024-08-23
bug: fix file_conversion disk leak

Coniferish:
Fix disk space leaks and Windows errors when accessing file.name on a NamedTemporaryFile

Uses of `NamedTemporaryFile(..., delete=False)` and/or uses of `file.name` of NamedTemporaryFile have been replaced with TemporaryFileDirectory to avoid a known issue: 
- https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile 
- https://github.com/Unstructured-IO/unstructured/issues/3390

The first 7 commits each address an individual occurrence of the issue if reviewers want to review commit-by-commit.

### DIFF
--- a/unstructured/file_utils/file_conversion.py
+++ b/unstructured/file_utils/file_conversion.py
@@ -56,7 +56,7 @@ def convert_file_to_html_text_using_pandoc(
 
     if file is not None:
         with tempfile.TemporaryDirectory() as temp_dir_path:
-            tmp_file_path = os.path.join(temp_dir_path, "tmp_file")
+            tmp_file_path = os.path.join(temp_dir_path, f"tmp_file.{source_format}")
             with open(tmp_file_path, "wb") as tmp_file:
                 tmp_file.write(file.read())
             return convert_file_to_text(
--- a/unstructured/file_utils/file_conversion.py
+++ b/unstructured/file_utils/file_conversion.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import tempfile
 from typing import IO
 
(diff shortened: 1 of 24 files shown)


unstructured/file_utils/file_conversion.py
```diff
@@ -54,11 +55,12 @@ def convert_file_to_html_text_using_pandoc(
     exactly_one(filename=filename, file=file)
 
     if file is not None:
-        with tempfile.NamedTemporaryFile() as tmp:
-            tmp.write(file.read())
-            tmp.flush()
+        with tempfile.TemporaryDirectory() as temp_dir_path:
+            tmp_file_path = os.path.join(temp_dir_path, "tmp_file")
```
scanny, line 59: It would be nice if we could do a little better than "tmp_file" as a name. Not sure what `source_format` is, but if it's an extension or something acceptable as one, maybe `f"document.{source_format}"` to produce a name like `document.doc` or `document.epub` or similar if we happen to be looking around in there during debugging or something.

unstructured/partition/pdf_image/pdf_image_utils.py
```diff
@@ -125,7 +125,7 @@ def save_elements(
     element_category_to_save: str,
     pdf_image_dpi: int,
     filename: str = "",
-    file: Optional[Union[bytes, BinaryIO]] = None,
+    file: Optional[Union[bytes, BinaryIO, BufferedReader]] = None,
```
scanny, line 128: Hmm, what's up with needing `BufferedReader`. I'd be inclined to make the `file` parameter always `IO[bytes]` as it already is in most of the partitioners. Anything special about this case why that wouldn't work here?

scanny (approved):
Curious about a couple things but otherwise LGTM, approving in advance :)

unstructured/partition/pdf_image/pdf_image_utils.py
scanny: A couple things:
- I would try `file: bytes | IO[bytes] | None = None`. There are only three possible types, raw-data, a file open for reading bytes, and `None`. `BinaryIO` is a weaker form of `IO[bytes]` which is problematic for historical Python reasons so best avoided.
- A "typing-type" won't work with `isinstance()`, it has to be a concrete type. "Typing"-types (`Protocol` types) are only known at type-checking time and `isinstance()` can only recognize runtime (concrete) types. I think what I would try is:
```python
if isinstance(file, bytes):
    file_data = file
else:
    file.seek(0)
    file_data = file.read()
```

PR #3564, opened 2024-08-24
feat: Support encoding parameter in partition_csv

awalker4:
See added test file. Added support for the encoding parameter, which can be passed directly to `pd.read_csv`.


Coniferish:
In `typings/pandas/io/parsers/readers.pyi`, can you add `encoding: str | None = ...,` to `read_csv`? This is a stub file for the type hints and `dataframe = pd.read_csv(file, header=ctx.header, sep=ctx.delimiter, encoding=encoding)` is raising linting errors without that being added to the stub file.

unstructured/partition/csv.py
```diff
@@ -110,6 +114,7 @@ def __init__(
         self,
         file_path: str | None = None,
         file: IO[bytes] | None = None,
+        encoding: str | None = None,
         metadata_file_path: str | None = None,
         metadata_last_modified: str | None = None,
         include_header: bool = False,
```
Coniferish, line 120: @scanny, is there a reason why these parameters are not in alphabetical order?

Coniferish (approved):
One change requested for typing and a non-blocking question for @scanny, but lgtm.

unstructured/partition/csv.py
scanny: `encoding` is very closely related to `file_path | file` for textual document formats, so I would place it in exactly this position. I'd like to see `metadata_file_path` and `metadata_last_modified` go away and be handled by the post-processing decorator, as they may already be. And I'd like to see `date_from_file_object` go away because I think we should always assume `False` for that. Once those three are gone, all that are left are `include_header` and `infer_table_structure` which are already in alphabetical order.

The reason I think we should assume `date_from_file_object` is always `False` is because all it does is try to get the file path for the open file and then consult the filesystem for the last-modified of that file. I can't see that being useful in the general case. If the user has a path to the file, why wouldn't they give us that instead of taking the extra step to open the file? So I'm thinking that's a noise parameter.

The position of the "source-document" args should be consistent across partitioners and up-front/top, IMO. `file_path` definitely needs to come first in partitioners so it can be provided positionally in the easy "partition-this" case.

Not sure if that completely answers your question, but basically:
- source-document related params up top, by convention, and common across partitioners.
- alphabetical order when we don't have a more pressing ordering; that ordering is inherently consistent.
- wouldn't it be nice to get rid of some unnecessary partitioner params so the list is generally so short that it doesn't need to be scanned vertically by eye to pick out a particular item.

PR #3572, opened 2024-08-27
feat/remove all traces of ingest code

rbiseck3:
### Description
Now that the ingest code has been emitting deprecation warnings for several versions, this PR removes all traces of ingest from the open source unstructured repo.


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 0.15.9-dev0
```
MthwRobinson, line 1: Could we bump the version to `0.16.0` to reflect the major change?

unstructured/embed/huggingface.py
```diff
@@ -66,15 +65,3 @@ def _add_embeddings_to_elements(self, elements, embeddings) -> List[Element]:
             element.embeddings = embeddings[i]
             elements_w_embedding.append(element)
         return elements
-
-    @EmbeddingEncoderConnectionError.wrap
```
MthwRobinson, line 70: Is this okay to totally remove?

rbiseck3: This logic should be moved to the ingest repo anyways.

MthwRobinson: Got it, thanks!

MthwRobinson (approved):
LGTM as long as tests pass!

cragwolfe (changes requested):
ah, wait a second. these tests have also been valuable for unstructured, not just ingest. can we keep CI intact by pip installing unstructured-ingest and keeping the "fixtures" (expected outputs)?

christinestraub (changes requested):
I believe ingest codes are used by CI ingest test jobs in both the `open-source` and `core-product` repositories.

cragwolfe:
> @cragwolfe if we want to keep some tests in this repo, how do we maintain parity as new connectors are added and new tests for these connectors are added to the new repo? Will we need to update this repo to have those tests as well? This would only persist the tests that existed here before we moved the ingest work to the new repo.

yes, my main concern is not about adding connectors in the future, it is preserving the tests we have (which are basically two tests in one: output correctness and ingest smoke test). in practice, this means that new ingest tests which relied on new connectors wouldn't immediately be added to this repo, at least not until a pypi package was available with the new connector (or merged to main if we just want to clone the repo in tests). or, if we feel like the separate unstructured-ingest repo tests are sufficient end-to-end for the new connectors, we can skip adding them to this repo.

rbiseck3:
I advocate for using the ingest tests that have been moved over to the new ingest repo as a valid smoke test since the version of unstructured is pinned there for testing. This alleviated the burden of matching new patterns and connectors in both places as they get added to the new repo. If for any reason, the ingest tests break due to a bump in the unstructured repo, we can work to remedy that.  This will also make the CI process in this repo more streamline for developers given the overhead running the ingest tests add.

PR #3574, opened 2024-08-28
feat: make analysis drawing more flexible

pawel-kmiecik:
This PR changes the way the analysis tools can be used:
- by default if `analysis` is set to `True` in `partition_pdf` and the strategy is resolved to `hi_res`:
  - for each file 4 layout dumps are produced and saved as JSON files (`object_detection`, `extracted`, `ocr`, `final`) - similar way to the current `object_detection` dump
  - the drawing functions/classes now accept these dumps accordingly instead of the internal classes instances (like `TextRegion`, `DocumentLayout`
  - it makes it possible to use the lightweight JSON files to render the bboxes of a given file after the partition is done
- `_partition_pdf_or_image_local` has been refactored and most of the analysis code is now encapsulated in `save_analysis_artifiacts` function
- to do this, helper function `render_bboxes_for_file` is added
Screenshot 2024-08-28 at 14 37 56


unstructured/partition/pdf_image/analysis/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+
+    Args:
+        layout_dumpers: The layout dumpers to save and use for bboxes rendering
+        filename: The filename of the sources analyzed file (pdf/image)
+        analyzed_image_output_dir_path: The directory to save the analysis artifacts
+    """
+    if not filename:
+        filename = _generate_filename(is_image)
+    skip_bboxes = env_config.ANALYSIS_BBOX_SKIP
+    skip_dump_od = env_config.ANALYSIS_DUMP_OD_SKIP
```
amadeusz-ds, line 76: now you can't use this function without `env_config`, so it's not pure

unstructured/partition/pdf_image/analysis/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+            format=env_config.ANALYSIS_BBOX_FORMAT,
+        )
+
+        for layout_dumper in layout_dumpers:
+            drawer = _get_drawer_for_dumper(layout_dumper)
+            analysis_drawer.add_drawer(drawer)
+        analysis_drawer.process()
+
+
+def render_bboxes_for_file(
```
amadeusz-ds, line 109: I don't see it used anywhere

unstructured/partition/pdf_image/analysis/bbox_visualisation.py
```diff
@@ -390,51 +385,74 @@ def draw_bbox_on_image(
...
+            return page_image
+        if len(layout_pages) < page_num:
+            print(f"Error! Page {page_num} not found in layout (pages: {len(layout_pages)})")
+            return page_image
+        image_draw = ImageDraw.ImageDraw(page_image)
+        page_layout_dump = layout_pages[page_num - 1]
+        if dump_page_num := page_layout_dump.get("number"):
+            assert (
+                dump_page_num == page_num
+            ), f"Requested page num {page_num} differs from dump {page_layout_dump.get('number')}"
```
amadeusz-ds, line 407: You shouldn't use assert in runtime

amadeusz-ds: https://snyk.io/blog/the-dangers-of-assert-in-python/

unstructured/partition/pdf.py
```diff
@@ -624,19 +620,15 @@ def _partition_pdf_or_image_local(
...
-                if not skip_dump_od:
+                if not skip_analysis_dump:
                     od_model_layout_dumper = ObjectDetectionLayoutDumper(
                         layout=inferred_document_layout,
                         model_name=hi_res_model_name,
                     )
+                    extracted_layout_dumper = ExtractedLayoutDumper(
+                        layout=extracted_layout,
+                    )
+                    ocr_layout_dumper = OCRLayoutDumper()
```
amadeusz-ds, line 631: Why do some dumpers expect the layout during the init and some don't?

amadeusz-ds (approved):
Just some comments

unstructured/partition/pdf_image/analysis/__init__.py
pawel-kmiecik: hmm right - I'll parametrize this and pass from the main partition function.

unstructured/partition/pdf.py
pawel-kmiecik: OCR is the special one as we cannot collect the layout for all the pages but only one-by-one. I could pass some special collector object (or just a list reference) but I didn't want to complicate too much.

PR #3581, opened 2024-08-30
fix: check ole storage content to differentiate filetypes

MthwRobinson:
### Summary

Updates the file detection logic for OLE files to check the storage content of the file to more reliable differentiate between DOC, PPT, XLS and MSG files. This corrects a bug that caused file type detection to be incorrect in cases where the `filetype` library guessed and incorrect MIME type, such as `'application/vnd.ms-excel'` for a `.msg` file.

As part of this work, the `"msg"` extra was removed because the `python-oxmsg` package is now a base dependency.

### Testing

Using a test `.msg` file that returns `'application/vnd.ms-excel'` from `filetype.guess_mime`.

```python

filename = "test-file.msg"
detect_filetype(filename=filename) # result should be FileType.MSG
```


test_unstructured/file_utils/test_filetype.py
```diff
@@ -530,21 +545,6 @@ def test_it_falls_back_to_extension_strategy_when_prior_strategies_fail(
...
 
 
-@pytest.mark.parametrize(
-    ("metadata_file_path", "expected_value"),
-    [
-        ("fake-email.msg", FileType.MSG),
-        ("fake-email.msg.outlook", FileType.UNK),
-    ],
-)
-def test_it_can_only_detect_MSG_format_by_extension(
```
MthwRobinson, line 540: Test was removed because the extension is no longer required to detect MSG files.

scanny (approved):
Sweet! LGTM :)

PR #3591, opened 2024-09-02
feat: utility script to process large PDFs through the API by script

cragwolfe:
Adds the bash script `process-pdf-parallel-through-api.sh` that allows splitting up a PDF into smaller parts (splits) to be processed through the API concurrently, and is re-entrant. If any of the parts splits fail to process, one can attempt reprocessing those split(s) by rerunning the script.

Note: requires the `qpdf` command line utility.

The below command line output shows the scenario where just one split had to be reprocessed through the API to create the final `layout-parser-paper_combined.json` output.

```
$ BATCH_SIZE=20 PDF_SPLIT_PAGE_SIZE=6 STRATEGY=hi_res \
  ./scripts/user/process-pdf-parallel-through-api.sh example-docs/pdf/layout-parser-paper.pdf
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 Skipping processing for /Users/cragwolfe/tmp/pdf-splits/layout-parser-paper-output-8a76cb6228e109450992bc097dbd1a51_split-6_strat-hi_res/layout-pars\
er-paper_pages_1_to_6.json as it already exists.
Skipping processing for /Users/cragwolfe/tmp/pdf-splits/layout-parser-paper-output-8a76cb6228e109450992bc097dbd1a51_split-6_strat-hi_res/layout-parser-paper_pages_7_to_12.json as it already exists.
Valid JSON output created: /Users/cragwolfe/tmp/pdf-splits/layout-parser-paper-output-8a76cb6228e109450992bc097dbd1a51_split-6_strat-hi_res/layout-parser-paper_pages_13_to_16.json
Processing complete. Combined JSON saved to /Users/cragwolfe/tmp/pdf-splits/layout-parser-paper-output-8a76cb6228e109450992bc097dbd1a51_split-6_strat-hi_res/layout-parser-paper_combined.json
```

Bonus change to `unstructured-get-json.sh` to point to the standard hosted Serverless API, but allow using the Free API with --freemium.

### DIFF
--- a/scripts/user/process-pdf-parallel-through-api.sh
+++ b/scripts/user/process-pdf-parallel-through-api.sh
@@ -42,7 +42,7 @@ MD5_SUM=$(md5sum "$PDF_FILE" | awk '{ print $1 }')
 PDF_DIR="$PDF_SPLITS_DIR/$PDF_NAME-${MD5_SUM}_split-${SPLIT_SIZE}"
 PDF_OUTPUT_DIR="$PDF_SPLITS_DIR/${PDF_NAME}-output-${MD5_SUM}_split-${SPLIT_SIZE}_strat-${STRATEGY}"
 
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # Check if PDF parts directory exists
 if [ ! -d "$PDF_DIR" ]; then
@@ -52,7 +52,7 @@ fi
 # Create output directory if it does not exist
 mkdir -p "$PDF_OUTPUT_DIR"
 
-incomplete=0  # Flag to track incomplete processing
+incomplete=0 # Flag to track incomplete processing
 
 # Function to process a single PDF part file
 process_file_part() {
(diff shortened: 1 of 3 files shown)


MKhalusova (approved):
This would be an alternative to ingest with a local source for a single large file, correct? 

For those who don't want to use a python library (ingest), I'm assuming. Is performance similar to ingest in this use case?

cragwolfe:
> This would be an alternative to ingest with a local source for a single large file, correct?
> 
> For those who don't want to use a python library (ingest), I'm assuming. Is performance similar to ingest in this use case?

Correct. Performance is similar in the best case. In the worst case when there is a failure for one or more splits (that is not a "real" 500 server error), it would be much, much faster.

cragwolfe:
> For those who don't want to use a python library (ingest), I'm assuming. Is performance similar to ingest in this use case?

Yes, with the caveats:
1/ it only does the partition step (not embedding or chunking)
2/ API params other than `strategy` would have to be updated directly in the curl command [here](https://github.com/Unstructured-IO/unstructured/blob/e9690b2/scripts/user/process-pdf-parallel-through-api.sh#L70-L78), e.g. if one wanted to `extract_image_block_types` or to skip table extraction with `skip_infer_table_types=["pdf"]`, or any of the other [documented parameters](https://docs.unstructured.io/api-reference/api-services/api-parameters).

Issue #3604, opened 2024-09-07
Partition_email with process_attachments=True for nested eml


S1M0N38:
**Description**
When try to process attachment in partition_email, the following error is raised:
```
Traceback (most recent call last):
  File "eml_to_elements.py", line 74, in <module>
    process_email(eml_file)
  File ".../scripts/preprocessing/eml_to_elements.py", line 58, in process_email
    elements = partition_email(
               ^^^^^^^^^^^^^^^^
  File ".../unstructured/unstructured/documents/elements.py", line 605, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File ".../unstructured/unstructured/file_utils/filetype.py", line 731, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File ".../unstructured/unstructured/file_utils/filetype.py", line 687, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File ".../unstructured/unstructured/chunking/dispatch.py", line 74, in wrapper
    elements = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
  File ".../unstructured/unstructured/partition/email.py", line 527, in partition_email
    extract_attachment_info(msg, tmpdir)
  File ".../unstructured/unstructured/partition/email.py", line 232, in extract_attachment_info
    f.write(attachment["payload"])
TypeError: a bytes-like object is required, not 'NoneType'
```

**Reproduce**
Try to partition an email containing other emails using `partition_email`, e.g.
```python
import functools
from unstructured.partition.auto import partition
from unstructured.partition.email import partition_email

attachment_partitioner = functools.partial(
    partition,
    max_partition=None,
    include_headers=True,
    process_attachments=True,
    strategy="hi_res",
    hi_res_model_name="yolox",
)

def process_email(document_path: Path):
    print("document_path", document_path)
    elements = partition_email(
        filename=str(document_path),
        max_partition=None,
        process_attachments=True,
        attachment_partitioner=attachment_partitioner,
    )

```

**Debug**
The reason of this bug is that the method `get_payload` return `None` if the `part` is a multipart and `decode=True`. Then, the function try to same this `None` value.

https://github.com/Unstructured-IO/unstructured/blob/c0604670182fb6e4b27268fe264fadda3388f06d/unstructured/partition/email.py#L201-L223

[...]

PR #3605, opened 2024-09-07
fix: process attchments in partitioning nested emails (#3604)

S1M0N38:


tc360950:
@S1M0N38 Will this not break existing code of the users? Should `attachment_partitioner` be optional?

S1M0N38:
#3671 `attachment_partitioner` is no more a valid param of partion_msg. This PR is outdated

Issue #3430, opened 2024-07-23
bug/ partition_pdf omits extractable text


JMTarraga:
partition_pdf() correctly detects the table, but it does not include within the element the numbers as text in the table. These numbers are indeed detected by the **process_file_with_pdfminer()** function.

![image](https://github.com/user-attachments/assets/092079fb-7c35-4e60-83b7-bbf776efaba5)

This information dissapears with the clean_pdfminer_inner_elements() and clean_pdfminer_duplicate_image_elements() functions.

Element text: 

'Affectations to houses and public buildings Quantity Houses affected Houses flooded Houses destroyed Other buildings affected Collapsed walls'

**To Reproduce**
elements = partition_pdf(filename=pdf_path, 
                        include_page_breaks=True, 
                        strategy="hi_res",  
                        extract_image_block_types=['Image', 'Table'],
                        model_name="yolox",
                        infer_table_structure=False, 
                        extract_image_block_to_payload=True)

**Expected behavior**
Output the table numbers in the unstructured element text. 

[18466.pdf](https://github.com/user-attachments/files/16345111/18466.pdf)

christinestraub:
Addressed in https://github.com/Unstructured-IO/unstructured/pull/3618

PR #3618, opened 2024-09-10
feat: improve pdfminer element processing

christinestraub:
This PR implements splitting of `pdfminer` elements (`groups of text chunks`) into smaller bounding boxes (`text lines`). This implementation prevents loss of information from the object detection model and facilitates more effective removal of duplicated `pdfminer` text. This PR also addresses #3430.

### DIFF
--- a/test_unstructured/partition/pdf_image/test_pdfminer_processing.py
+++ b/test_unstructured/partition/pdf_image/test_pdfminer_processing.py
@@ -2,7 +2,7 @@ import numpy as np
 import pytest
 from PIL import Image
 from unstructured_inference.constants import Source as InferenceSource
-from unstructured_inference.inference.elements import Rectangle, TextRegion
+from unstructured_inference.inference.elements import EmbeddedTextRegion, Rectangle, TextRegion
 from unstructured_inference.inference.layout import DocumentLayout, LayoutElement, PageLayout
 
 from unstructured.partition.pdf_image.pdfminer_processing import (
@@ -11,6 +11,7 @@ from unstructured.partition.pdf_image.pdfminer_processing import (
     boxes_self_iou,
     clean_pdfminer_duplicate_image_elements,
     clean_pdfminer_inner_elements,
+    remove_duplicate_embedded_text,
 )
 from unstructured.partition.utils.constants import Source
 
@@ -209,3 +210,22 @@ def test_bboxes1_is_almost_subregion_of_bboxes2(coords1, coords2, expected):
 def test_boxes_self_iou(coords, threshold, expected):
     bboxes = [Rectangle(*row) for row in coords]
     np.testing.assert_array_equal(boxes_self_iou(bboxes, threshold), expected)
+
+
+def test_remove_duplicate_embedded_text():
+    sample_elements = [
+        EmbeddedTextRegion(bbox=Rectangle(0, 0, 10, 10), text="Text 1"),
+        EmbeddedTextRegion(bbox=Rectangle(0, 0, 10, 10), text="Text 2"),
+        EmbeddedTextRegion(bbox=Rectangle(20, 20, 30, 30), text="Text 3"),
+    ]
+
+    result = remove_duplicate_embedded_text(sample_elements)
+
+    # Check that duplicates were removed and only 2 unique elements remain
+    assert len(result) == 2
+    assert result[0].text == "Text 2"
+    assert result[1].text == "Text 3"
+
+    # Ensure the duplicate was removed by checking that result contains no redundant bboxes
+    assert result[0].bbox == Rectangle(0, 0, 10, 10)
+    assert result[1].bbox == Rectangle(20, 20, 30, 30)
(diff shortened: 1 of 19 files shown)


unstructured/partition/pdf_image/pdfminer_utils.py
```diff
@@ -34,6 +34,19 @@ def extract_image_objects(parent_object: LTItem) -> List[LTImage]:
...
+    """Recursively extracts text objects from a given parent object in a PDF document."""
+    objects = []
+
+    if isinstance(parent_object, LTTextLine):
+        objects.append(parent_object)
+    elif isinstance(parent_object, LTContainer):
+        for child in parent_object:
+            objects.extend(extract_text_objects(child))
+
+    return objects
```
badGarnet, line 47: what's the compute cost of this function for a pdf page with lots of elements? This recursion could get expensive when there are deeply nested structures.

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -292,6 +303,38 @@ def clean_pdfminer_duplicate_image_elements(document: "DocumentLayout") -> "Docu
...
 
 
+@requires_dependencies("unstructured_inference")
+def remove_duplicate_embedded_text(elements: list["TextRegion"]) -> list["TextRegion"]:
+    """Removes duplicate text elements extracted by PDFMiner from a document layout."""
+    from unstructured_inference.inference.elements import EmbeddedTextRegion
+
+    bboxes = []
+    texts = []
+    bbox_to_iou_mapping = {}
```
pawel-kmiecik, line 313: Wouldn't it be slightly easier if we gathered text elements and image (other) elements in two separate lists in `process_data_with_pdfminer`? We already has a "switch case" there so it would be easier and we'd avoid introducing this mapping. So it could look like:
```python
text_layout = []
image_layout = []
if hasattr(obj, "get_text"):
    # ....
    text_layout.append(text_region)
# ...
clean_text_layout = remove_duplicate_embedded_text(text_layout)
layout = [*clean_text_layout, *image_layout]
layout = sort_text_regions(layout, SORT_MODE_BASIC)
# ....
```
(ofc naming could be corrected :) )

pawel-kmiecik (commented):
Generally looks good but I'd consider simplifying `remove_duplicate_embedded_text` function - so I'm not requesting changes to not block the PR :)

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -57,15 +58,23 @@ def process_data_with_pdfminer(
...
+                    text_region = _create_text_region(
+                        new_x1,
+                        new_y1,
+                        new_x2,
+                        new_y2,
+                        coef,
+                        _text,
+                        Source.PDFMINER,
+                        EmbeddedTextRegion,
+                    )
```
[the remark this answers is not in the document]
badGarnet: nit nit 😬 
can just do `*rect_to_bbox(inner_obj.bbox, height)`

another question about this: do we need to create those as text regions? Or can we just have them aggregated together as numpy array and that would save us time to unpack them in dedup function. Then we can just create the boxes after dedup to make the return compatible to downstream.

test_unstructured_ingest/expected-structured-output/s3/recalibrating-risk-report.pdf.json
```diff
@@ -660,9 +704,383 @@
...
+        },
+        "date_created": "1676196572.0",
+        "date_modified": "1676196572.0"
+      }
+    }
+  },
+  {
+    "type": "Title",
+    "element_id": "f4df01bee1b8ffb973ac8539649c5189",
+    "text": "W",
```
badGarnet, line 753: just a note that those vertical text are now broken into each letter per element; not sure what the impact would be but just a note

test_unstructured/partition/pdf_image/test_pdfminer_processing.py
```diff
@@ -209,3 +210,22 @@ def test_bboxes1_is_almost_subregion_of_bboxes2(coords1, coords2, expected):
...
+        EmbeddedTextRegion(bbox=Rectangle(0, 0, 10, 10), text="Text 1"),
+        EmbeddedTextRegion(bbox=Rectangle(0, 0, 10, 10), text="Text 2"),
+        EmbeddedTextRegion(bbox=Rectangle(20, 20, 30, 30), text="Text 3"),
+    ]
+
+    result = remove_duplicate_embedded_text(sample_elements)
+
+    # Check that duplicates were removed and only 2 unique elements remain
+    assert len(result) == 2
+    assert result[0].text == "Text 2"
```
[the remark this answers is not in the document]
christinestraub: Yes, it's a pdfminer issue (on some PDFs) that are duplicated elements with the same text.

badGarnet (approved):
vertical text results probably can use a follow up investigation (on its impact with chunking etc)

PR #3625, opened 2024-09-13
build(deps): bump protobuf pin

Coniferish:
Bumps max version of `protobuf<5.0` and sets min version of `chromadb>0.4.14` in `requirements/ingest/chroma.in`. Also fixes some type hints in `unstructured/ingest/v2/processes/connectors/chroma.py`


MthwRobinson (approved):
LGTM

PR #3629, opened 2024-09-13
Fix/reduce memory usage

badGarnet:
This PR fixes the high memory usage when computing intersection areas.

- it now converts the coordinates into half precision floating point numbers instead of double
- removes some intermediate variables to free up memory usage

## test

Using a memory profiler like `memory_profiler` in `ipython`:

```ipython
## cell 1
from unstructured.partition.pdf_image.pdfminer_processing import areas_of_boxes_and_intersection_area
import numpy as np
%load_ext memory_profiler

## cell 2
%%memit
coords = np.random.rand(40000).reshape((10000,4)).astype(np.float16)

## cell 3
%%memit
inter_area, boxa_area, boxb_area = areas_of_boxes_and_intersection_area(coords, coords)
```

The peak memory and incremental memory from cell 3 should be close to 
```
peak memory: 730.55 MiB, increment: 573.22 MiB
```

On main branch the `coords` is double precision and running the same code with 

```
coords = np.random.rand(40000).reshape((10000,4)).astype(np.float64)
```

would result in peak memory usage more than 4GiB


christinestraub (changes requested):
There are some unusual changes in the ingest test fixtures. I'm investigating this issue.

christinestraub (approved):
LGTM!

PR #3641, opened 2024-09-17
add requirements files to ingest cache hash key

badGarnet:
This PR adds the requirement files for base and extras for the ingest cache's hash key.

- The current workflow uses only the ingest requirements to generate hash key for the gitaction cache
- Sometimes only base or extra requirements (like extra-pdf.txt) updated but not any ingest requirements -> this would mean the ingest test would fetch a cache with outdated non-ingest dependencies
- When we generate new ingest cache we actually do check first base and extra requirements and generate a base env before layer on top the ingest dependencies.
- This PR allows the ingest step to recognize changes to non-ingest dependency changes and trigger new cache generation when either ingest or base/extra requirement files changes.

This PR also bumps the setup python action version in cache actions; it also adds installation of `virtualenv` for the ingest cache action to avoid errors like https://github.com/Unstructured-IO/unstructured/actions/runs/10905551870/job/30265057515?pr=3641#step:3:111

### DIFF
--- a/.github/actions/base-cache/action.yml
+++ b/.github/actions/base-cache/action.yml
@@ -29,13 +29,13 @@ runs:
       if: steps.virtualenv-cache-restore.outputs.cache-hit != 'true'
       shell: bash
       run: |
-        python3 -m pip install --upgrade virtualenv
-        python3 -m venv .venv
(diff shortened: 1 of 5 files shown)


christinestraub (approved):
LGTM!

PR #3645, opened 2024-09-18
chore(file): remove dead code

scanny:
**Summary**
Remove dead code in `unstructured.file_utils`.

**Additional Context**
These modules were added in 12/2022 and 1/2023 and are not referenced by any code. Removing to reduce unnecessary complexity. These can of course be recovered from Git history if we decide we want them again in future.


MthwRobinson (approved):
🧹  LGTM!

PR #3650, opened 2024-09-20
rfctr(part): add new decorator to replace four

scanny:
**Summary**
In preparation for pluggable auto-partitioners, add a new metadata decorator to replace the four existing ones.

**Additional Context**
"Global" metadata items, those applied to all element on all partitioners, are applied using a decorator.

Currently there are four decorators where there only needs to be one. Consolidate those into a single metadata decorator.
One or two additional behaviors of the new decorator will allow us to remove decorators from delegating partitioners which is a prerequisite for pluggable auto-partitioners.


Coniferish (approved):
One question, but approving in advance.

PR #3655, opened 2024-09-23
rfctr(part): prepare for pluggable auto-partitioners 1

scanny:
**Summary**
In preparation for pluggable auto-partitioners simplify metadata as discussed.

**Additional Context**
- Pluggable auto-partitioners requires partitioners to have a consistent call signature. An arbitrary partitioner provided at runtime needs to have a call signature that is known and consistent. Basically `partition_x(filename, *, file, **kwargs)`.
- The current `auto.partition()` is highly coupled to each distinct file-type partitioner, deciding which arguments to forward to each.
- This is driven by the existence of "delegating" partitioners, those that convert their file-type and then call a second partitioner to do the actual partitioning. Both the delegating and proxy partitioners are decorated with metadata-post-processing decorators and those decorators are not idempotent. We call the situation where those decorators would run twice "double-decorating". For example, EPUB converts to HTML and calls `partition_html()` and both `partition_epub()` and `partition_html()` are decorated.
- The way double-decorating has been avoided in the past is to avoid sending the arguments the metadata decorators are sensitive to to the proxy partitioner. This is very obscure, complex to reason about, error-prone, and just overall not a viable strategy. The better solution is to not decorate delegating partitioners and let the proxy partitioner handle all the metadata.
- This first step in preparation for that is part of simplifying the metadata processing by removing unused or unwanted legacy parameters.
- `date_from_file_object` is a misnomer because a file-object never contains last-modified data.
- It can never produce useful results in the API where last-modified information must be provided by `metadata_last_modified`.
- It is an undocumented parameter so not in use.
- Using it can produce incorrect metadata.


test_unstructured/partition/test_auto.py
```diff
@@ -381,16 +381,13 @@ def test_auto_partitioned_json_output_maintains_consistency_with_fixture_element
         expected_result = json.load(json_f)
 
     partitioning_result = json.loads(
-        cast(
```
MthwRobinson, line 384: Was this `cast` just unnecessary?

unstructured/partition/text.py
```diff
@@ -60,7 +57,6 @@ def partition_text(
     chunking_strategy: Optional[str] = None,
```
MthwRobinson, line 57: Should this kwarg go away like it did for `pptx`?

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.15.14-dev0
+## 0.15.14-dev1
```
MthwRobinson, line 1: Reminder to describe the changes in here

MthwRobinson (commented):
Couple small questions. Could you also add some detail to the PR summary describing the metadata changes in the PR? Otherwise looking good from my end.

test_unstructured/partition/test_auto.py
scanny: Yes. During type-checking it flagged up as unnecessary. I believe what happened is that sometime after that was added I got to `elements_to_json()` and annotated its types correctly, so then this call could tell the result was going to be `(list[Element]) -> str` without needing a cast.

unstructured/partition/text.py
scanny: Ah, good catch, yes, I'll do that ...

MthwRobinson (approved):
LGTM!

PR #3656, opened 2024-09-23
fix: update python SDK syntax for forward compatibility

awalker4:
Wrap the `shared.PartitionParameters` usage with `operations.PartitionRequest`. This syntax has been deprecated since v0.23.0 of the SDK, and will be unsupported in v0.26.0.


MthwRobinson (approved):
LGTM!

PR #3658, opened 2024-09-24
fix: fix occasional key error when mapping parent id

badGarnet:
This PR fixes an occasional `KeyError` when calling `assign_and_map_hash_ids`. 

- This happens when the input `elements` has duplicated element instances or metadata.
- When there are duplications the logic to iterate through all elements and map their parent ids will raise an error when an already mapped parent id is up for mapping.
- The fix adds a logic to check if the parent id exists in `old_to_new_mapping` and if it doesn't we skip mapping it

## test

This PR adds a unit test on this case and the test would fail without the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.14-dev1
+## 0.15.14-dev2
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


christinestraub (approved):
LGTM!

PR #3661, opened 2024-09-24
rfctr(part): prepare for pluggable auto-partitioners 3

scanny:
**Summary**
Remove unused `include_metadata` parameter.

**Additional Context**
- The `include_metadata` parameter was originally added circa v0.7.12 as a mechanism for avoiding the "double-decorating" problem on delegating partitioners.
- It turns out it doesn't fully address that problem, is now unused, and is unnecessary for the solution we'll be adding as part of pluggable partitioners.
- Remove the unnecessary complexity introduced by this unused parameter.


MthwRobinson (approved):
🧹  - LGTM!

PR #3667, opened 2024-09-26
rfctr(meta): refine @apply_metadata() decorator

scanny:
**Summary**
Refine `@apply_metadata()` replacement decorator. Note it has not been installed yet.

- Apply `metadata_last_modified` arg with the `@apply_metadata()` decorator. No need for redundant processing in each partitioner.
- Add "unique-ify" step to fix any cases where the same `Element` or `ElementMetadata` instance was used more than once in the element stream. This prevents unexpected "multi-mutation" in downstream processes.
- Apply "global" metadata items before computing hash-ids. In particular, `.metadata.filename` is used in the hash computation and will produce different results if that's not already settled.
- Compute hash-ids _before_ computing `.metadata.parent_id`. This removes the need for mapping UUID element-ids to their hash counterpart and doing a fixup of `.parent_id` after applying hash-ids to elements.

**Additional Context**
- The `@apply_metadata()` decorator replaces the four metadata-related decorators: `@process_metadata()`, `@add_metadata_with_filetype()`, `@add_metadata()`, and `@add_filetype()`.
- It will be installed on each partitioner in a series of following PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.14-dev5
+## 0.15.14-dev6
 
 ### Enhancements
 
--- a/test_unstructured/partition/common/test_metadata.py
+++ b/test_unstructured/partition/common/test_metadata.py
@@ -1,7 +1,10 @@
 """Test-suite for `unstructured.partition.common.metadata` module."""
 
+# pyright: reportPrivateUsage=false
+
 from __future__ import annotations
 
+import copy
 import datetime as dt
 import os
 import pathlib
@@ -22,6 +25,7 @@ from unstructured.documents.elements import (
 )
 from unstructured.file_utils.model import FileType
 from unstructured.partition.common.metadata import (
+    _assign_hash_ids,
     apply_metadata,
     get_last_modified_date,
     set_element_hierarchy,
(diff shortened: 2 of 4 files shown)


Coniferish:
Question stemming from looking at `_assign_hash_ids` in `unstructured/partition/common/metadata.py`.
There's a similar util function in `test_unstructured/unit_utils.py`, but it just uses the element idx instead of `seq_on_page_counter`:
```
def assign_hash_ids(elements: list[Element]) -> list[Element]:
    """Updates the `id` attribute of each element to a hash."""
    for idx, element in enumerate(elements):
        element.id_to_hash(idx)
    return elements
```

Is it okay for these to have slightly different algorithms?

scanny:
> Question stemming from looking at `_assign_hash_ids` in `unstructured/partition/common/metadata.py`. There's a similar util function in `test_unstructured/unit_utils.py`, but it just uses the element idx instead of `seq_on_page_counter`:
> 
> ```
> def assign_hash_ids(elements: list[Element]) -> list[Element]:
>     """Updates the `id` attribute of each element to a hash."""
>     for idx, element in enumerate(elements):
>         element.id_to_hash(idx)
>     return elements
> ```
> 
> Is it okay for these to have slightly different algorithms?

Hmm, looks like that was added by Michael (DeepSense) but is now unused. I expect he used it in some test but it's not used anymore so should come out. I'll get rid of that.

Coniferish:
LGTM

scanny:
Okay, turns out that `assign_hash_ids()` utility function is indeed used by staging tests, so I put it back in.

The algorithm is essentially the same, basically call `.id_to_hash()` on every element, so no harm. And it's better than using the internal function in `common.metadata`, so I'm inclined to just leave it alone.

Coniferish (approved):

Issue #3674, opened 2024-09-28
bug/Auto partition fails on text files which are empty or contain only whitespaces


tc360950:
**Describe the bug**
Inference of .txt file type fails if the file has only whitespaces.

**To Reproduce**
```
from tempfile import NamedTemporaryFile

from unstructured.partition.auto import partition

with NamedTemporaryFile(mode="w", suffix=".txt") as f:
    f.write("   \n")
    f.seek(0)
    elements = partition(filename=f.name)
```
Raises IndexError
**Expected behavior**
The file should be properly partitioned.

**Environment Info**
OS version:  Linux-6.8.0-45-generic-x86_64-with-glibc2.35
Python version:  3.10.12
unstructured version:  0.15.13
unstructured-inference version:  0.7.36
pytesseract is not installed
Torch version:  2.4.1
Detectron2 is not installed
PaddleOCR is not installed
Libmagic version: file-5.41
magic file from /etc/magic:/usr/share/misc/magic
LibreOffice version:  LibreOffice 7.3.7.2 30(Build:2)

Prateek-coder19:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

scanny:
Fixed by #3675

PR #3675, opened 2024-09-28
Fix bug - Auto partition fails on text files which are empty or contain only whitespaces

tc360950:
This is a fix for this [bug](https://github.com/Unstructured-IO/unstructured/issues/3674).

Inference of .txt file type fails if the file has only whitespaces.

To Reproduce:

```
from tempfile import NamedTemporaryFile

with NamedTemporaryFile(mode="w", suffix=".txt") as f:
    f.write("   \n")
    f.seek(0)
    elements = partition(filename=f.name)
```


cragwolfe (approved):

cragwolfe:
CI passed in https://github.com/Unstructured-IO/unstructured/pull/3676 (needed for extra creds), so merged.

cragwolfe:
thanks for the contribution, @tc360950!

PR #3685, opened 2024-10-01
rfctr(part): remove double-decoration 1

scanny:
**Summary**
Install new `@apply_metadata()` on CSV and DOCX and remove decoration from DOC and ODT.

**Additional Context**
- Working in alphabetical order and keeping PR size manageable, replace use of `@process_metadata()` and `@add_metadata_with_filetype()` decorators with `@apply_metadata()` on principal partitioners (those that do not delegate to other partitioners.
- Remove all decorators from delegating partitioners (DOC and ODT in this case); this removes the "double-decorating".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.14-dev6
+## 0.15.14-dev7
 
 ### Enhancements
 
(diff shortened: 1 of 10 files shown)


unstructured/partition/common/metadata.py
```diff
@@ -201,8 +217,70 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> list[Element]:
...
 
+            # ------------------------------------------------------------------------------------
+            # compute hash ids (when so requestsd)
+            # ------------------------------------------------------------------------------------
+
+            # -- Compute and apply hash-ids if the user does not want UUIDs. Note this mutates the
+            # -- elements themselves, not their metadata.
+            unique_element_ids: bool = call_args.get("unique_element_ids", False)
+            if unique_element_ids is False:
+                elements = _assign_hash_ids(elements)
```
Coniferish, line 228: Looking at `_assign_hash_ids` and `assign_and_map_hash_ids` from `unstructured/documents/elements.py`, I'm wondering if that function needs to be used since it takes into account `parent_id`. Maybe that doesn't need to be used here for some reason I'm not thinking of, but wanted to check.

scanny: In the prior "multiple decorator" situation, assigning hash-ids and assigning `.parent_id` happened in different decorators. Further, `assign_and_map_hash_ids()` would always execute _after_ `.metadata.parent_id` was computed. That meant that the `.parent_id` value was a UUID and it needed to be mapped to the new SHA1 hash id. That's what the "map" part was about.

But now that they are not in separate decorators, we can just compute `.metadata.parent_id` _after_ we (optionally) assign hash-ids, so it's no longer necessary to do a mapping. That's a good thing not only for performance reasons because we would get bugs cropping up from time-to-time right in that spot.

Coniferish (approved):
LGTM

PR #3686, opened 2024-10-01
rfctr(part): remove double-decoration 2

scanny:
**Summary**
Install new `@apply_metadata()` on PPTX, TSV, XLSX, and XML and remove decoration from PPT.

**Additional Context**
- Alphabetical order turns out to be hard, so this is the remaining "easy" delegating partitioner and the remaining principal partitioners.
- Replace use of `@process_metadata()` and `@add_metadata_with_filetype()` decorators with `@apply_metadata()` on principal partitioners (those that do not delegate to other partitioners.
- Remove all decorators from delegating partitioners (PPT in this case); this removes the "double-decorating".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.14-dev7
+## 0.15.14-dev8
 
 ### Enhancements
 
@@ -13,6 +13,7 @@
 * **Fix occasional `KeyError` when mapping parent ids to hash ids.** Occasionally the input elements into `assign_and_map_hash_ids` can contain duplicated element instances, which lead to error when mapping parent id.
 * **Allow empty text files.** Fixes an issue where text files with only white space would fail to be partitioned.
 * **Remove double-decoration for CSV, DOC, ODT partitioners.** Refactor these partitioners to use the new `@apply_metadata()` decorator and only decorate the principal partitioner (CSV and DOCX in this case); remove decoration from delegating partitioners.
+* **Remove double-decoration for PPT, PPTX, TSV, XLSX, and XML partitioners.** Refactor these partitioners to use the new `@apply_metadata()` decorator and only decorate the principal partitioner; remove decoration from delegating partitioners.
 
 ## 0.15.13
 
--- a/test_unstructured/partition/test_xlsx.py
+++ b/test_unstructured/partition/test_xlsx.py
@@ -322,15 +322,6 @@ class Describe_XlsxPartitionerOptions:
     """Unit-test suite for `unstructured.partition.xlsx._XlsxPartitionerOptions` objects."""
 
     @pytest.mark.parametrize("arg_value", [True, False])
-    def it_knows_whether_to_detect_language_for_each_element_individually(
-        self, arg_value: bool, opts_args: dict[str, Any]
-    ):
-        opts_args["detect_language_per_element"] = arg_value
-        opts = _XlsxPartitionerOptions(**opts_args)
-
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * **Fix occasional `KeyError` when mapping parent ids to hash ids.** Occasionally the input elements into `assign_and_map_hash_ids` can contain duplicated element instances, which lead to error when mapping parent id.
 * **Allow empty text files.** Fixes an issue where text files with only white space would fail to be partitioned.
 * **Remove double-decoration for CSV, DOC, ODT partitioners.** Refactor these partitioners to use the new `@apply_metadata()` decorator and only decorate the principal partitioner (CSV and DOCX in this case); remove decoration from delegating partitioners.
+* **Remove double-decoration for PPT, PPTX, TSV, XLSX, and XML partitioners.** Refactor these partitioners to use the new `@apply_metadata()` decorator and only decorate the principal partitioner; remove decoration from delegating partitioners.
```
Coniferish, line 16: Something I remembered--doesn't this also remove support for one or two kwargs that are unused? Even if they're not intended to be used by users, I think we should mention those here (and in the bullet above this one).

unstructured/partition/pptx.py
```diff
@@ -417,7 +384,9 @@ def increment_page_number(self) -> Iterator[PageBreak]:
             yield PageBreak(
                 "",
                 detection_origin=DETECTION_ORIGIN,
-                metadata=ElementMetadata(last_modified=self.last_modified),
+                metadata=ElementMetadata(
+                    last_modified=self.last_modified, page_number=self.page_number - 1
```
Coniferish, line 388: Is this a bug fix (the `page_number` for `PageBreak` elements)?

test_unstructured/partition/test_xlsx.py
```diff
@@ -321,15 +321,6 @@ def test_partition_xlsx_with_more_than_1k_cells():
 class Describe_XlsxPartitionerOptions:
     """Unit-test suite for `unstructured.partition.xlsx._XlsxPartitionerOptions` objects."""
 
-    @pytest.mark.parametrize("arg_value", [True, False])
-    def it_knows_whether_to_detect_language_for_each_element_individually(
```
Coniferish, line 325: Isn't this test still relevant?

Coniferish (approved):

unstructured/partition/pptx.py
scanny: Yes, I'd say so. No one has ever complained about it as far as I know, but the idea is that hash element-ids need to be unique within a single document and this was causing all PageBreak elements to get the same hash-id.

The hash-id is based on text, filename, page-number, and sequence-number within that page. When `PageBreak` did not have `.metadata.page_number`:
- text was always "" as it is for any `PageBreak` element.
- filename was the same for all elements in the file
- page_number was always the same (`None`)
- A missing page-number causes the page-sequence counter to reset, so page_seq_num would always be 1

Consequently the hash for a `PageBreak` would always be sha1(("", filename, None, 1)) which is constant for a given file.

test_unstructured/partition/test_xlsx.py
scanny: It is for `partition_xlsx()` and is still there for that case.

But `XlsxPartitionerOptions` no longer takes a `languages` arg and doesn't provide access to that value or any values derived from it, so this unit test goes away.

PR #3687, opened 2024-10-02
rfctr(part): remove double-decoration 3

scanny:
**Summary**
Install new `@apply_metadata()` on HTML and remove decorators from delegating partitioners EPUB, MD, ORG, RST, and RTF.

**Additional Context**
- All five of these delegating partitioners delegate to `partition_html()` so they're something of a matched set. EML and MSG also partially delegate to HTML but that's a harder problem (they also delegate to all other partitioners for attachments)  that we'll address a couple PRs later .
- Replace use of `@process_metadata()` and `@add_metadata_with_filetype()` decorators with `@apply_metadata()` on `partition_html()`.
- Remove all decorators from delegating partitioners; this removes the "double-decorating".


Coniferish (approved):
LGTM

PR #3690, opened 2024-10-02
rfctr(part): remove double-decoration 4

scanny:
**Summary**
Install new `@apply_metadata()` on TXT.

**Additional Context**
- Both EML and MSG delegate to both HTML and TXT to partition the message-body, depending on which MIME-part body payload is selected (`text/plain` or `text/html`). This PR prepares the way to remove decorators from EML and MSG in the next PR.


Coniferish (approved):
lgtm

PR #3692, opened 2024-10-03
rfctr(part): remove double-decoration 5

scanny:
**Summary**
Remove double-decoration from EML and MSG.

**Additional Context**
- These needed to wait to the end because `partition_email()` and `partition_msg()` can use any other partitioner for one of their attachments.


Coniferish (approved):
LGTM

PR #3694, opened 2024-10-03
rfctr(email): eml partitioner rewrite

### LINKED ISSUE #3671: bug/partition_msg halts for attachmentes with UNK type
Currently, the `attachment_partitioner` is hardcoded to `partition` in the following file:

https://github.com/Unstructured-IO/unstructured/blob/50d75c47d346413de8a6dbbcf72009871f7ecd56/unstructured/partition/msg.py#L259-L277

However, according to the [official documentation](https://docs.unstructured.io/open-source/core-functionality/partitioning#partition-msg), the `partition-msg` function accepts `attachment_partitioner` as an argument.

scanny:
**Summary**
Initial attempts to incrementally refactor `partition_email()` into shape to allow pluggable partitioning quickly became too complex for ready code-review. Prepare separate rewritten module and tests and swap them out whole.

**Additional Context**
- Uses the modern stdlib `email` module to reliably accomplish several manual decoding steps in the legacy code.
- Remove obsolete email-specific element-types which were replaced 18 months or so ago with email-specific metadata fields for things like Cc: addresses, subject, etc.
- Remove accepting an email as `text: str` because MIME-email is inherently a binary format which can and often does contain multiple and contradictory character-encodings.
- Remove `encoding` parameters as it is now unused. An email file is not a text file and as such does not have a single overall encoding. Character encoding is specified individually for each MIME-part within the message and often varies from one part to another in the same message.
- Remove the need for a caller to specify `attachment_partitioner`. There is only one reasonable choice for this which is `auto.partition()`, consistent with the same interface and operation in `partition_msg()`.
- Fixes #3671 along the way by silently skipping attachments with a file-type for which there is no partitioner.
- Substantially extend the test-suite to cover multiple transport-encoding/charset combinations.


test_unstructured/partition/test_new_email.py
```diff
@@ -0,0 +1,592 @@
...
+
+def test_partition_email_from_filename_can_partition_an_RFC_822_email():
+    assert partition_email(example_doc_path("eml/simple-rfc-822.eml")) == [
+        NarrativeText("This is an RFC 822 email message."),
+        NarrativeText(
+            "An RFC 822 message is characterized by its simple, text-based format, which includes"
+            ' a header and a body. The header contains structured fields such as "From", "To",'
+            ' "Date", and "Subject", each followed by a colon and the corresponding information.'
+            " The body follows the header, separated by a blank line, and contains the main"
+            " content of the email."
```
Coniferish: Can you add an assert on the element metadata showing the additional email info that's mentioned here?

test_unstructured/partition/test_auto.py
```diff
@@ -200,14 +201,6 @@ def test_auto_partition_email_from_file():
     assert elements == EXPECTED_EMAIL_OUTPUT
 
 
-def test_auto_partition_eml_add_signature_to_metadata():
-    elements = partition(example_doc_path("eml/signed-doc.p7s"))
-
-    assert len(elements) == 1
-    assert elements[0].text == "This is a test"
-    assert elements[0].metadata.signature == "<SIGNATURE>\n"
```
Coniferish, line 208: Is `metadata.signature` becoming not supported with these changes?

unstructured/partition/email.py
```diff
@@ -1,510 +1,429 @@
...
+                **self._ctx.partitioning_kwargs,
+            )
+        except UnsupportedFileFormatError:
+            # -- indicates `auto.partition()` has no partitioner for this file-format;
+            # -- silently skip the attachment
+            return
+
+        for e in elements:
+            e.metadata.attached_to_filename = self._attached_to_filename
+            yield e
```
Coniferish, line 405: Just noticing that you define `e.metadata.attached_to_filename` directly here instead of using the `.update()` method. Doesn't need to change, but any reason why?

test_unstructured/partition/test_email.py
```diff
@@ -54,565 +35,578 @@
...
+        partition_email(example_doc_path("eml/mime-no-body.eml"), process_attachments=False) == []
+    )
 
 
-def test_partition_email_from_text_work_with_empty_string():
-    assert partition_email(text="") == []
+def test_partition_email_does_not_break_on_an_encrypted_message():
+    assert (
+        partition_email(example_doc_path("eml/fake-encrypted.eml"), process_attachments=False) == []
+    )
```
Coniferish, line 147: Is a warning logged, or does this fail silently? I believe it used to log a message.

Coniferish (approved):
a couple of questions and comments, but lgtm

test_unstructured/partition/test_auto.py
scanny: Right. Matt and I decided that should go as it is large-ish, would be repeated on each element, and is of no direct use without a public key.

We did talk about possibly adding `.metadata.is_signed` as a bool field but that goes on the backlog.

unstructured/partition/email.py
scanny: `.update()` requires a whole `ElementMetadata()` object to be passed, like "update yourself from this other `ElementMetadata` instance". That can be handy when you're updating multiple fields or collecting the fields to update from multiple places, but isn't necessary here.

test_unstructured/partition/test_email.py
scanny: Well, it doesn't fail, it just produces no elements for the encrypted portions of the email.

Users don't have access to the logs so the only real option for communicating this is to fail with an exception. That would produce an HTTP 4xx and that could have a message. But failing is not appropriate here as far as I know.

Coniferish: Don't open source users have access to the logs?

test_unstructured/partition/test_new_email.py
scanny: We could I suppose, but this `elements == [Element("..."), ...]` testing is consistent with what we do for other email variants in the tests that follow it.

I purposely separated out the metadata tests so the failures are more specific and it's easier to tell what tests exercise what behavior. If we were doing strict black-box testing we would have a bunch of cases like N subtypes X M behaviors, but I don't see that paying off here. We happen to know that all email-specific metadata is applied on a particular line, not "subvariant" specific, So the one test (further down) is fine I think. Especially when you add the unit tests for each email-specific metadata value.

PR #3700, opened 2024-10-04
add tests describing the behavior of set_element_hierarchy

Coniferish:
Small pr adding tests describing the behavior of `set_element_hierarchy`. No tests were changed, just added.


test_unstructured/partition/common/test_metadata.py
```diff
@@ -62,73 +62,139 @@ def but_it_returns_None_when_there_is_no_file_at_that_path(self, tmp_path: pathl
...
+        ]
+
+        result = set_element_hierarchy(elements)
+
+        assert result[0].metadata.parent_id is None
+        assert result[1].metadata.parent_id == "0"  # Text0 is under Title0
+        assert result[2].metadata.parent_id is None  # Header0 is higher than Title0
+        assert result[3].metadata.parent_id == "2"  # Text1 is under Header0
+        assert result[4].metadata.parent_id == "2"  # Title1 is under Header0
+        assert result[5].metadata.parent_id == "4"  # Text2 is under Title1, which is under Header0
```
scanny, line 84: Hmm, interesting. So we don't actually need a `.metadata.category_level` value for this to work? I'm assuming that's only necessary when the title or whatever is not at the top level; that makes sense now that I think of it.

The clarity of this test is good; makes details like that pop out :)

test_unstructured/partition/common/test_metadata.py
```diff
@@ -62,73 +62,139 @@ def but_it_returns_None_when_there_is_no_file_at_that_path(self, tmp_path: pathl
...
+            Title(element_id="0", text="Title", metadata=ElementMetadata(category_depth=2)),
+            Text(element_id="1", text="Text", metadata=ElementMetadata(category_depth=1)),
+            Header(element_id="2", text="Header", metadata=ElementMetadata(category_depth=2)),
+            Text(element_id="3", text="Text", metadata=ElementMetadata(category_depth=1)),
+        ]
+
+        result = set_element_hierarchy(elements)
+
+        assert result[1].metadata.parent_id == "0"  # Text is under Title despite category_depth
+        assert result[3].metadata.parent_id == "2"  # Text is under Header despite category_depth
```
scanny, line 97: Maybe the name can more clearly describe the behavior. Are we saying something like `it_ignores_category_depth_except_on_top_level_elements` or something like that?

A separate behavior might be "it_does_not_mix_hierarchies_between_rulesets" or something. Like if you had `Header cat_lvl=1` (which actually wouldn't make any sense since headers don't have levels like section-headings do) followed by `Title cat_lvl=0`, would the `Title` use the `Header` as its parent or become a top-level (parentless) element?

Maybe you could describe the behavior you're looking to verify here in as many words as it takes, then we could look for a short way to say it. The docstring can also be used to elaborate on the behavior.

test_unstructured/partition/common/test_metadata.py
```diff
@@ -62,73 +62,139 @@ def but_it_returns_None_when_there_is_no_file_at_that_path(self, tmp_path: pathl
...
+            Text(element_id="2", text="Text1", metadata=ElementMetadata(category_depth=1)),
+            Text(element_id="3", text="Text2", metadata=ElementMetadata(category_depth=0)),
+        ]
+
+        result = set_element_hierarchy(elements)
+
+        assert result[0].metadata.parent_id is None
+        assert result[1].metadata.parent_id == "0"
+        assert result[2].metadata.parent_id == "1"
+        assert result[3].metadata.parent_id == "0"
```
scanny, line 112: Hmm, this behavior seems a little odd. I'm wondering when we would even use this sort of thing.

It strikes me that not just any old element should be able to be a parent. Like why would we make one `Text` element a child of another?

Strikes me this would be the distinction between nodes and leaves, maybe container-nodes and leaf-nodes in the hierarchy. `Title` can definitely be a parent because a document-section "contains" paragraphs. But I would expect `NarrativeText` for example to be a leaf-node.

More of a discussion topic I suppose :)

test_unstructured/partition/common/test_metadata.py
```diff
@@ -62,73 +62,139 @@ def but_it_returns_None_when_there_is_no_file_at_that_path(self, tmp_path: pathl
...
+
+        result = set_element_hierarchy(elements)
+
+        # Parent ID should not change and element is skipped in figuring out other elements' parents
+        assert result[0].metadata.parent_id == "10"
+        assert result[1].metadata.parent_id is None
+        assert result[2].metadata.parent_id == "1"
+
+    def test_set_element_hierarchy(self):
+        elements_to_set = [
```
scanny, line 129: Probably needs a parallel name, maybe `it_sets_parent_id_for_each_element_in_elements()` or something like that.

scanny (approved):
A few remarks, but looks like a definite improvement to me :)

Approving in advance.

test_unstructured/partition/common/test_metadata.py
Coniferish: I think this applies mostly (or only) to `ListItem`s, but I just used `Text` for copy/paste reasons. I can change it to `ListItem` if you think that'd be better.

test_unstructured/partition/common/test_metadata.py
Coniferish: Sounds good. This was a test I didn't touch since it was a little unclear what all it was intending to test.

test_unstructured/partition/common/test_metadata.py
scanny: I would. It's helpful to the reader to see typical usage as much as possible I'd say. One of the benefits of reading the tests should be seeing examples of how one might want to use the units under test.

test_unstructured/partition/common/test_metadata.py
Coniferish: So, your example is what I'm trying to illustrate. I was expecting the `category_level` to supersede the element `category`, but it doesn't, it's ignored. The `category_level` is ignored when it comes to comparing different _types_ of elements. In the test, I set the `category_depth` of the `Title` elements to be "under" the `Text` elements (setting them to `"2"`), but that is ignored and the only important factor is the element's `category` (`Title`/`Text`).

Coniferish: I feel like the semantics are fine, but the vocabulary might be off--I was trying to reuse the terms used in the file: "ruleset", "category", "category_depth" etc. It's a little confusing to have "category" for the element type and have "category_depth"

scanny: Oooh, okay, that _is_ weird!

Okay, I've had a look at the code and I don't think I can fully understand it without incrementally executing a bunch of cases. If I did take the time to do that I'm sure it would result in me wanting to make changes.

So I'm thinking time to cut bait unless there's a clear problem we're trying to diagnose :)

I'm fine with this one just the way it is :)

Coniferish: I just realized we have `category_depth` kind of doing double-duty. It's used internally to determine parent, but we also make it accessible to users--I have to factor that into my partitioning of onenote...

scanny: Okay, I should have known I wouldn't be able to help myself, so after a further look, here are some possible insights:

Q. What is a _rule-set_?
A. A rule-set is is a `{parent: list[permissible_children]}` mapping, where both parent and permitted-child are expressed as an `element.category` (str/name) value.

Inferences:
1. Only an element-type that appears as a key (parent) in the rule-set mapping can be the target of a `.metadata.parent_id` reference.

Loop behavior: for each element:
- the current "parent" is the top element on the stack.
- the current element gets the current parent-id when the current element meets either of these criteria:
    - it is the same kind as the parent but a lower level, like both are `Title` but the current element is level 2 instead of 1
    - it's not the same kind but the current element is one of the permissible children of the current parent
- Otherwise, remove parents from the stack one at a time until you either:
    - find a suitable parent, or
    - run out of parents on the stack

In the later case (run out of parents) the the element gets `parent_id=None`.

There are more subtleties this doesn't cover, like how does the first parent get there and so on, but I'm already feeling like rewriting it, so I'd better stop here :)

PR #3701, opened 2024-10-04
rfctr(ppt): remove double-decoration

scanny:
Somehow this slipped through the earlier PR removing double-decoration from PPTX.

Remove the decorators from PPT (because it is a delegating partitioner) and let the decorators on the proxy partitioner (`partition_pptx()`) do the needful.


Coniferish (commented):
Glynn

Coniferish (approved):

PR #986, opened 2023-07-27
build(deps): split up dependencies by document type

MthwRobinson:
### Summary

Closes #611. Updates the dependencies to create a slim base installation + extras for each document type. The `all-docs` extra includes the dependencies for all doc types. The `local-inference` extra still remains as an alias for `all-docs` since that's the extra many currently users use to install all of the dependencies.

As a follow on, the PDF and image requirements could be broken down further by strategy.

### Testing

- In a fresh virtual environment, run `make install-base` and `pip install -r requirements/test.txt` and `pip install -e .`. The following tests should pass:
    - `pytest test_unstructured/partition/test_text.py`
    - `pytest test_unstructured/partition/test_email.py`
    - `pytest test_unstructured/partition/test_html_partition.py`
    - `pytest test_unstructured/partition/test_text.py`
    - `pytest test_unstructured/partition/test_xml_partition.py`
- Run `make install-csv`. The following tests should pass:
    - `pytest test_unstructured/partition/test_csv.py`
    - `pytest test_unstructured/partition/test_tsv.py`
- Run `make install-docx`. The following tests should pass:
    - `pytest test_unstructured/partition/test_doc.py`
    - `pytest test_unstructured/partition/test_docx.py`
- Run `make install-odt`. The following tests should pass:
    - `pytest test_unstructured/partition/test_odt.py`
- Run `make install-pypandoc`. The following tests should pass:
    - `pytest test_unstructured/partition/test_epub.py`
    - `pytest test_unstructured/partition/test_org.py`
    - `pytest test_unstructured/partition/test_rst.py`
    - `pytest test_unstructured/partition/test_rtf.py`
- Run `make install-markdown`. The following tests should pass:
    - `pytest test_unstructured/partition/test_md.py`
- Run `make install-msg`. The following tests should pass:
    - `pytest test_unstructured/partition/test_msg.py`
- Run `make install-pdf-image`. The following tests should pass:
    - `pytest test_unstructured/partition/test_pdf.py`
    - `pytest test_unstructured/partition/test_image.py`
- Run `make install-pptx`. The following tests should pass:
    - `pytest test_unstructured/partition/test_ppt.py`
    - `pytest test_unstructured/partition/test_pptx.py`
- Run `make install-xlsx`. The following tests should pass:
    - `pytest test_unstructured/partition/test_xlsx.py`
- Run `make install-ci` and `pip install argilla weaviate-client`. `make test` should pass

[...]

PR #3712, opened 2024-10-09
build: Fix build reproducibility.

jsirois:
Without going the extra length of exporting `PYTHONHASHSEED=0` or
similar, the built sdist and wheel for a given version of the project
would have inconsistent random ordering of extras_require entries
between builds.

This was observed attempting to lock a VCS requirement on this project
here: https://github.com/pantsbuild/pants/discussions/21145
On the bright side, it led to Pex fixing its building and locking code
to be robust to this sort of (unintended) bad behavior:
https://github.com/pex-tool/pex/pull/2554

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ rst_reqs = load_requirements("requirements/extra-pandoc.in")
 tsv_reqs = load_requirements("requirements/extra-csv.in")
 xlsx_reqs = load_requirements("requirements/extra-xlsx.in")
 
-all_doc_reqs = list(
+all_doc_reqs = sorted(
     set(
         csv_reqs
         + docx_reqs


setup.py
```diff
@@ -56,7 +56,7 @@ def load_requirements(file_list: Optional[Union[str, List[str]]] = None) -> List
 tsv_reqs = load_requirements("requirements/extra-csv.in")
 xlsx_reqs = load_requirements("requirements/extra-xlsx.in")
 
-all_doc_reqs = list(
+all_doc_reqs = sorted(
```
jsirois, line 59: Although I added no reproducibility test to this repo, the Pex repo now has one inspired directly by the prior `list(set(...))` construction. That test failed prior to Pex plumbing `PYTHONHASHSEED=0` to all process where distributions are built.

jsirois:
@MthwRobinson it looks like you may be most appropriate to review. The problematic use of set appears to stem directly from #986.

PR #3716, opened 2024-10-12
Fix typing issue in inference_utils.py

cckolon:


christinestraub:
What versions of `unstructured` and `unstructured-inference` libraries are you using?

bohaiwangzi:
Any updates?

PR #3724, opened 2024-10-16
feat: expose retry params in partition via api

pawel-kmiecik:
This PR:
- adds parameters to control the retry-mechanism behaviour for `partition_via_api`:
```
    retries_initial_interval: [int] = None,
    retries_max_interval: Optional[int] = None,
    retries_exponent: Optional[float] = None,
    retries_max_elapsed_time: Optional[int] = None,
    retries_connection_errors: Optional[bool] = None,
```
- adds tests that check using them according to defaults

### DIFF
--- a/unstructured/partition/api.py
+++ b/unstructured/partition/api.py
@@ -6,11 +6,19 @@ from typing import IO, Any, Optional, Sequence
 import requests
 from unstructured_client import UnstructuredClient
 from unstructured_client.models import operations, shared
+from unstructured_client.utils import retries
 
 from unstructured.documents.elements import Element
 from unstructured.logger import logger
 from unstructured.partition.common.common import exactly_one
 from unstructured.staging.base import elements_from_dicts, elements_from_json
+
+# Default retry configuration taken from the client code
+DEFAULT_RETRIES_INITIAL_INTERVAL_SEC = 3000
+DEFAULT_RETRIES_MAX_INTERVAL_SEC = 720000
+DEFAULT_RETRIES_EXPONENT = 1.5
+DEFAULT_RETRIES_MAX_ELAPSED_TIME_SEC = 1800000
+DEFAULT_RETRIES_CONNECTION_ERRORS = True
 
 
 def partition_via_api(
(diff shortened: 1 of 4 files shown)


unstructured/partition/api.py
```diff
@@ -97,6 +136,94 @@ def partition_via_api(
...
+            retries_exponent,
+            retries_max_elapsed_time,
+            retries_connection_errors,
+        )
+    ):
+        default_retries_initial_interval = (
+            default_retries_config.backoff.initial_interval
+            if default_retries_config and default_retries_config.backoff.initial_interval
+            else DEFAULT_RETRIES_INITIAL_INTERVAL_SEC
+        )
```
badGarnet, line 187: small nit; this pattern repeated enough times for us to consider refactor like
```suggestion
        def _config_or_default(attr: str, default_val: int | float):
            return (getattr(default_retries_config.backoff, attr) if default_retries_config else None) or default_val
        default_retries_initial_interval = _config_or_default("initial_interval", DEFAULT_RETRIES_INITIAL_INTERVAL_SEC)
        # and those below
```

badGarnet: and one more note this logic means if the config value is 0 we will be using default; is that intended or should we change the second condition to `default_retries_config.backoff.initial_interval is not None` (I got burned by this recently 😬 )

badGarnet (commented):
this is effectively create a wrapper for the `unstructured-client` interface so my question for the design is why do we need to add this to the document processing project. This creates ambiguous boundary of the projects (is this just document processing or does it also include client interfaces to a remote service, that under the hood runs the document processing code). 
The client part depends on the document processing function being complete. So putting the client function in the document processing project creates a recursion where client needs client to be complete to complete.

unstructured/partition/api.py
pawel-kmiecik: I've started the implementation to check for `None` (as you suggested) but in this specific case all these settings should have values greater than 0 (zero is incorrect) - so this simplified check should be ok.
I'll mention it in the comments.

badGarnet (approved):

PR #3728, opened 2024-10-17
Remove unsupported chipper model

vangheem:
The chipper model is no longer


badGarnet (commented):
please also bump version; otherwise lgtm

badGarnet (approved):

PR #3732, opened 2024-10-18
Add parsing HTML to unstructured elements

plutasnyy:
> This is POC change; not everything is working correctly and code quality could be improved significantly 

This ticket add parsing HTML to unstructured element and back. How is it working? 

HTML has a tree structure, Unstructured Elements is a list.
HTML structure is traversed in DFS order, creating Elements and adding them to list. So the reading order from HTML is preserved. To be able to compose tree again all elements has IDs, and metadata.parent_id is leveraged 

How html is preserved if there are 'layout' without text, or there are deeply nested HTMLs that are just text from the point of view of Unstructured Element? 
Each element is parsed back to HTML using metadata.text_as_html field. For layout elements only html_tag are there, for long text elements there is everything required to recreate HTML - you can see examples in unit tests or .json file I attached. 

Pros of solution:
- Nothing had to be changed in element types 

Cons:
- There are elements without Text which may be confusing (they could be replaced by some special type) 

Core transformation logic can be found in 2 functions in `unstructured/documents/transformations.py` 

Knowns bugs (they are minor):
- sometimes html tag is changed incorrectly
- metadata.category_depth and metadata.page_number are not set
- page break is not added between pages 

How to test. Generate HTML:
```python3
from pathlib import Path

from vlm_partitioner.src.partition import partition

if __name__ == "__main__":
    doc_dir = Path("out_dir")
    file_path = Path("example_doc.pdf")
    partition(str(file_path), provider="anthropic", output_dir=str(doc_dir))
```

Then parse to unstructured elements and back to html
```python3
from pathlib import Path

from unstructured.documents.html_utils import indent_html
from unstructured.documents.transformations import parse_html_to_ontology, ontology_to_unstructured_elements, \
    unstructured_elements_to_ontology
from unstructured.staging.base import elements_to_json

if __name__ == "__main__":
    output_dir = Path("out_dir/")
    output_dir.mkdir(exist_ok=True, parents=True)

    doc_path = Path("out_dir/example_doc.html")

    html_content = doc_path.read_text()

    ontology = parse_html_to_ontology(html_content)
    unstructured_elements = ontology_to_unstructured_elements(ontology)

    elements_to_json(unstructured_elements, str(output_dir / f"{doc_path.stem}_unstr.json"))

    parsed_ontology = unstructured_elements_to_ontology(unstructured_elements)
    html_to_save = indent_html(parsed_ontology.to_html())

    Path(output_dir / f"{doc_path.stem}_parsed_unstr.html").write_text(html_to_save)
```

I attached example doc before and after running these scripts
[outputs.zip](https://github.com/user-attachments/files/17438673/outputs.zip)

### DIFF
--- a/test_unstructured/documents/test_ontology_to_unstructured_parsing.py
+++ b/test_unstructured/documents/test_ontology_to_unstructured_parsing.py
@@ -139,8 +139,11 @@ def test_embeddings_are_applied_on_elements(mocker):
     ],
 )
 def test_ingest(html_file_path, json_file_path):
-    html_code = Path(html_file_path).read_text()
-    expected_json_elements = elements_from_json(json_file_path)
+    html_file_path = Path(__file__).parent / html_file_path
+    json_file_path = Path(__file__).parent / json_file_path
+
+    html_code = html_file_path.read_text()
+    expected_json_elements = elements_from_json(str(json_file_path))
 
     ontology = parse_html_to_ontology(html_code)
     unstructured_elements = ontology_to_unstructured_elements(ontology)
@@ -149,9 +152,11 @@ def test_ingest(html_file_path, json_file_path):
 
 @pytest.mark.parametrize("json_file_path", ["unstructured_json_output/example.json"])
 def test_parsed_ontology_can_be_serialized_from_json(json_file_path):
-    expected_json_elements = elements_from_json(json_file_path)
+    json_file_path = Path(__file__).parent / json_file_path
 
-    json_elements_text = Path(json_file_path).read_text()
+    expected_json_elements = elements_from_json(str(json_file_path))
+
+    json_elements_text = json_file_path.read_text()
     elements = partition_json(text=json_elements_text)
 
     assert len(elements) == len(expected_json_elements)
@@ -159,7 +164,7 @@ def test_parsed_ontology_can_be_serialized_from_json(json_file_path):
         assert elements[i] == expected_json_elements[i]
         # The partitioning output comes from PDF file, so only stem is compared
         # as the suffix is different .pdf != .json
-        assert Path(elements[i].metadata.filename).stem == Path(json_file_path).stem
+        assert Path(elements[i].metadata.filename).stem == json_file_path.stem
 
 
 @pytest.mark.parametrize(
@@ -169,8 +174,11 @@ def test_parsed_ontology_can_be_serialized_from_json(json_file_path):
     ],
 )
 def test_parsed_ontology_can_be_serialized_from_html(html_file_path, json_file_path):
-    expected_json_elements = elements_from_json(json_file_path)
-    html_code = Path(html_file_path).read_text()
+    html_file_path = Path(__file__).parent / html_file_path
+    json_file_path = Path(__file__).parent / json_file_path
+
+    expected_json_elements = elements_from_json(str(json_file_path))
+    html_code = html_file_path.read_text()
 
     predicted_elements = partition_html(text=html_code, html_parser_version="v2")
     assert len(expected_json_elements) == len(predicted_elements)


test_unstructured/documents/structured_jsons/example.json
```diff
@@ -0,0 +1,121 @@
+[
+    {
+        "element_id": "897a8a47-377c-4ad6-aab8-39a929879537",
+        "metadata": {
+            "category_depth": 0,
+            "text_as_html": "<body class=\"Document\" id=\"897a8a47-377c-4ad6-aab8-39a929879537\" />"
```
cragwolfe, line 6: we should not include `class="Document"`'s in Unstructured Elements output.
by convention, a list of Unstructured elements represents a document, so this is not adding any information (and gets in the way if combining multiple pages

test_unstructured/documents/structured_jsons/example.json
```diff
@@ -0,0 +1,121 @@
...
+            "parent_id": "45b3d005-3468-484b-a1c7-b53998115412",
+            "text_as_html": "<h1 class=\"Title\" id=\"c95473e8-a370-4fc2-b418-697f9fddb27b\">Header </h1>"
+        },
+        "text": "Header",
+        "type": "Title"
+    },
+    {
+        "element_id": "379cbfdc-16d4-4bd6-a59e-6cfabe6438d5",
+        "metadata": {
+            "category_depth": 3,
```
cragwolfe, line 47: just noting that this has a`"category_depth": 3,` which is strictly true as UncategorizedText, though maybe not particularly useful.

test_unstructured/documents/structured_jsons/example.json
```diff
@@ -0,0 +1,121 @@
...
+            "category_depth": 1,
+            "page_number": 1,
+            "parent_id": "897a8a47-377c-4ad6-aab8-39a929879537",
+            "text_as_html": "<div class=\"Page\" data-page-number=\"1\" id=\"3a6b156a-8176-4e17-be12-8264241f8136\" />"
+        },
+        "text": "",
+        "type": "UncategorizedText"
+    },
+    {
+        "element_id": "45b3d005-3468-484b-a1c7-b53998115412",
```
cragwolfe, line 23: we have done a lot of work to always format element_id's as a hash, which is pretty much a built in expectation.
e.g.:
``` 
 "element_id": "782cf07be8b3ab8f05188e479edb7f61",
 ```
. however, fortunately it appears there is a super quick fix, just remove the `-`'s in element_id (which still maps 1-1 with html `id`s)

File: test_unstructured/documents/structured_jsons/example.json [hunk 4]
```diff
@@ -0,0 +1,121 @@
...
+        "type": "UncategorizedText"
+    },
+    {
+        "element_id": "dc3792d4-422e-444f-9087-6b56d0cfb20d",
+        "metadata": {
+            "category_depth": 3,
+            "page_number": 1,
+            "parent_id": "59242237-3ed7-41b6-8a07-7e2003f8ed81",
+            "text_as_html": "<table class=\"Table\" id=\"dc3792d4-422e-444f-9087-6b56d0cfb20d\"> <thead class=\"TableHeader\" id=\"50a5548a-87e8-4024-af59-0b3d2830d140\"> <tr class=\"TableRow\" id=\"5e473d77-4247-4412-be72-dc4e2c45bd4a\"> <th class=\"TableCellHeader\" id=\"01800309-aa42-411c-98ae-30f85b23f399\">Description </th><th class=\"TableCellHeader\" id=\"c2765b63-d089-46a2-8519-55e79e301de4\">Row header </th></tr></thead><tbody class=\"TableBody\" id=\"e0a9a8ff-dd71-48ad-8b4a-274b073d340a\"> <tr class=\"TableRow\" id=\"77e82997-4632-4551-9133-0b0b8545d1e3\"> <td class=\"TableCell\" id=\"7fee12d4-c555-4b7d-a778-d6f8fdec8a57\">Value description </td><td class=\"TableCell\" id=\"5a7a33b0-c57b-4eb8-81a3-5bce9f87c831\"> <span class=\"Currency\" id=\"87220f9d-62c3-482e-92e7-de72a26869cd\">50 $ </span><span class=\"Measurement\" id=\"0095b9ef-b90a-4cca-991e-73547c7165f1\">(1.32 %) </span></td></tr></tbody></table>"
```
cragwolfe, line 83: **non-blocking comment for this PR**, something to follow up on later: these `id`'s and `class`s per cell is adding a lot of bloat to the table. it should be safe to drop `id`s, maybe also `class`s? (for consistency with other table `text_as_html`

test_unstructured/documents/structured_jsons/example.json
```diff
@@ -0,0 +1,121 @@
...
+            "parent_id": "1032242a-f75c-4b37-984e-a7fea9aac74c",
+            "text_as_html": "<h2 class=\"Subtitle\" id=\"2a4e2c4a-689f-4f9a-8c18-0b6b521e45c3\">2. Subtitle </h2>"
+        },
+        "text": "2. Subtitle",
+        "type": "Title"
+    },
+    {
+        "element_id": "5591f7a4-df01-447e-8251-5ce45f686fbe",
+        "metadata": {
+            "category_depth": 3,
```
cragwolfe, line 113: this `category_depth` should be `1` because there are no `NarrativeText` blocks that this `NarrativeText` block is nested under (which generally may never happen)

cragwolfe (commented):
Looks very good, some nits/requests:

1. Please rename 
`test_unstructured/documents/structured_jsons/example.json`
to something like
`test_unstructured/documents/unstructured_json_output/example.json`
to make it clear this is Unstructured Element json to begin with.
2. Please add a test like https://github.com/Unstructured-IO/unstructured/blob/a11ad22609ca016f8080646c20e6da9ad5e7dc08/test_unstructured/partition/test_json.py#L152 for `example.json` which tests that `partition_json` deserializes `example.json` correctly (gives back the original elements)

unstructured/documents/mappings.py
```diff
@@ -0,0 +1,152 @@
+from collections import defaultdict
```
cragwolfe, line 1: please add docstring, what is this module for?

test_unstructured/documents/test_ontology_to_unstructured_parsing.py
```diff
@@ -0,0 +1,146 @@
...
+        ("html_files/example.html", "structured_jsons/example.json"),
+    ],
+)
+def test_ingest(html_file_path, json_file_path):
+    html_code = Path(html_file_path).read_text()
+    expected_json_elements = elements_from_json(str(Path(json_file_path)))
+
+    ontology = parse_html_to_ontology(html_code)
+    unstructured_elements = ontology_to_unstructured_elements(ontology)
+    assert unstructured_elements == expected_json_elements
```
cragwolfe, line 146: yay!

test_unstructured/documents/structured_jsons/example.json
cragwolfe: *this does not have to be a blocker for this PR, can be tracked as follow on

test_unstructured/documents/structured_jsons/example.json
cragwolfe: more detail: this may seem nit-picky, but the reason i feel strongly we should drop `Document` is it violates the standard pattern for Unstructured output for all other documents that are partitioned with `partition()`, i.e. the result is a set of Unstructured Elements that immediately begins with document content - there are no elements reserved for the Document itself.

in the past, we have discussed introducing something like a `DocumentMeta` element `type` in the past, and that could still be on the table.

all of that said, i don't think this strictly needs to be a blocker at this point.

test_unstructured/documents/test_ontology_to_unstructured_parsing.py
```diff
@@ -0,0 +1,146 @@
...
+    [
+        ("html_files/example.html", "structured_jsons/example.json"),
+    ],
+)
+def test_ingest(html_file_path, json_file_path):
+    html_code = Path(html_file_path).read_text()
+    expected_json_elements = elements_from_json(str(Path(json_file_path)))
+
+    ontology = parse_html_to_ontology(html_code)
+    unstructured_elements = ontology_to_unstructured_elements(ontology)
```
cragwolfe, line 145: for consumers that just want unstructured elements, these two functions calls combined `parse_html_to_ontology()` and `ontology_to_unstructured_elements()`, i.e. html -> unstructured elements should accessible to consumers using [paritition_html()](https://github.com/Unstructured-IO/unstructured/blob/a11ad22609ca016f8080646c20e6da9ad5e7dc08/unstructured/partition/html/partition.py#L23) (or plain old [partition()](https://github.com/Unstructured-IO/unstructured/blob/a11ad22609ca016f8080646c20e6da9ad5e7dc08/unstructured/partition/auto.py#L25) which may call `partition_html()`)  with an additional flag that indicates whether to parse with the prior or new parsing method.

unstructured/partition/html/partition.py
```diff
@@ -31,6 +35,7 @@ def partition_html(
     ssl_verify: bool = True,
     skip_headers_and_footers: bool = False,
     detection_origin: Optional[str] = None,
+    contains_ontology_schema: bool = False,
```
cragwolfe, line 38: let's rename this parameter, and instead a boolean let's allow a string argument where None (or v1?) is the "classic" html parser. this will allow for additional "parsers" in the future without having to introduce a new parameter.

so to support the new parsing method, how about:

```
html_parser=v2
```

test_unstructured/documents/structured_jsons/example.json
cragwolfe: `category_depth` only counts the nested depth of the _same_ type of tags or elements and ignores all others in between -- the word `category` in `category_depth` is the cue for "same category" (or type/tag).
"layout" elements don't really matter here, i think.
in reality, there are not many cases where this matters in HTML. mainly list items, heading tags, and once in awhile tables in HTML.

mariannaparzych (approved):
lgtm

PR #3747, opened 2024-10-24
Ml 384/whitespaces in cct

mariannaparzych:
This ticket  ensures that CCT metric will not be sensitive to differences in whitespace (including newline).
All whitespaces in string are changed to single space `" "` in both GT and PRED before the metric is computed.

Additional changes in CHANGELOG due to auto-formatting.


pawel-kmiecik (approved):
LGTM

PR #3749, opened 2024-10-24
Ml 415/merge inline elements

plutasnyy:


unstructured/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.16.3-dev1"  # pragma: no cover
+__version__ = "0.16.3-dev2"  # pragma: no cover
```
cragwolfe, line 1: the version should either be `0.16.4-dev0` if not intending to cut a new release, or `0.16.4` if this should be a new release.

cragwolfe (approved):

PR #3779, opened 2024-11-11
Add text as html to orig elements chunks

plutasnyy:
This simplest solution doesn't drop HTML from metadata when merging Elements from HTML input. We still need to address how to handle nested elements, and if we want to have `LayoutElements` in the metadata of Composite Elements, a unit test showing the current behavior.
Note: metadata still contains `orig_elements` which has all the metadata.

### DIFF
--- a/test_unstructured/chunking/test_html_output.py
+++ b/test_unstructured/chunking/test_html_output.py
@@ -41,8 +41,8 @@ def test_combining_html_metadata_with_nested_relationship_between_elements(chunk
     <Document>
         <Page>
             <Section>
-                <p>First </p>
-                <p>Second </p>
+                <p>First</p>
+                <p>Second</p>
             </Section>
         </Page>
     </Document>


unstructured/chunking/base.py
```diff
@@ -774,6 +774,8 @@ def iter_kwarg_pairs() -> Iterator[tuple[str, Any]]:
                     # -- Python 3.7+ maintains dict insertion order --
                     ordered_unique_keys = {key: None for val_list in values for key in val_list}
                     yield field_name, list(ordered_unique_keys.keys())
+                elif strategy is CS.STRING_CONCATENATE:
+                    yield field_name, "".join(values)
```
pawel-kmiecik, line 778: Wouldn't be better `" ".join(val.strip for val in values)`?

test_unstructured/chunking/test_html_output.py
```diff
@@ -0,0 +1,90 @@
...
+    assert chunks[0].metadata.text_as_html == combined_metadata
+
+
+def test_combining_html_metadata_with_nested_relationship_between_elements(chunking_fn):
+    """
+    Ground truth
+    <Document>
+        <Page>
+            <Section>
+                <p>First </p>
```
pawel-kmiecik, line 44: Are the spaces here (before `</p>`) intended?

pawel-kmiecik (approved):
LGTM!

PR #3786, opened 2024-11-19
Prefer using provided filename over detection from file.name

framp:
In some situations files uploaded with SpooledTemporaryFiles have a file.name of type int.

In those cases unstructured fails parsing the file name (splittext can't handle the int).

User provided metadata_filename is not being used because it's just a fallback

I think if the user provided a metadata_filename that should take precedence for extension detection

```
    return os.path.splitext(file.name)[1].lower()

  File "<frozen posixpath>", line 117, in splitext
TypeError: expected str, bytes or os.PathLike object, not int
```


srisudarsan:
Hello Team,
Is there a progress on this, we are hitting the same scenario when a specific file is sent as byte stream, we get the error as stated above
@scanny  can you help us on this ?

Issue #3781, opened 2024-11-13
bug/check for magic library availability doesn't appear to be correct


aaronsteers:
**Describe the bug**

Currently, it appears that libmagic is detected with this line:

```py
LIBMAGIC_AVAILABLE = bool(importlib.util.find_spec("magic"))
```

This finds whether `magic` is installed, but that's not the correct check, I think. Since `python-magic` in included in the `unstructured` dependencies, the library will be installed at the same time as Unstructured. However, if the `magiclib` c library is not available, those implementations will fail.

**To Reproduce**

On my Mac, I'm running this code:

```py
        try:
            file_type = detect_filetype(
                file_path=remote_file.uri,
            )
        except Exception:
            # Path doesn't exist locally. Fall back to using an extension-based detection.
            file_type: FileType | None = FileType.from_extension(
                remote_file.uri.split(".")[-1],
            )
```

**Expected behavior**

Expected to fail over to other means of detection if the magiclib c library is not available.

**Screenshots**

**Environment Info**
Please run `python scripts/collect_env.py` and paste the output here. 
This will help us understand more about the environment in which the bug occurred.

**Additional context**

Log is here:

```
 File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 131, in file_type
    return cls(ctx)._file_type
  File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 141, in _file_type
    if file_type := self._file_type_from_guessed_mime_type:
  File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 181, in _file_type_from_guessed_mime_type
    mime_type = self._ctx.mime_type
  File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/unstructured/utils.py", line 154, in __get__
    value = self._fget(obj)
  File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 362, in mime_type
    import magic
  File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/magic/__init__.py", line 209, in <module>
    libmagic = loader.load_lib()
  File "/Users/ajsteers/Source/my-project/.venv/lib/python3.10/site-packages/magic/loader.py", line 49, in load_lib
[...]

PR #3790, opened 2024-11-21
fix(filetype): handle missing libmagic library

metadaddy:
As reported in #3781, the check for availability of the `libmagic` library is not correct. The existing code checks whether the `magic` module is available, but the attempt to `import magic` fails if the`libmagic` library is not also available. On the Mac, `libmagic` is not installed by default; the user must install it manually, typically via `brew install libmagic`.

This PR detects whether `libmagic` is installed by importing the `magic` module in a `try` block, setting `LIBMAGIC_AVAILABLE` accordingly. `MAGIC_AVAILABLE` is set to true if the `magic` module is installed so that an appropriate warning can be displayed if the fallback `filetype` module returns `None` for a mime type. Tests that rely on `libmagic` being installed are skipped if it is not.

`pytest -v test_unstructured/file_utils` succeeds:

```console
(.venv) ppatterson@MBP-W7FQ7Y97F0 unstructured % pytest -v test_unstructured/file_utils
============================================== test session starts ==============================================
platform darwin -- Python 3.10.15, pytest-8.3.3, pluggy-1.5.0 -- /Users/ppatterson/src/unstructured/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /Users/ppatterson/src/unstructured
configfile: setup.cfg
plugins: cov-5.0.0, mock-3.14.0, anyio-4.6.2.post1, requests-mock-1.12.1
collected 440 items      
...
================================== 411 passed, 28 skipped, 1 xfailed in 3.68s ===================================
```

### DIFF
--- a/unstructured/file_utils/filetype.py
+++ b/unstructured/file_utils/filetype.py
@@ -51,7 +51,17 @@ from unstructured.partition.common.common import add_element_metadata, exactly_o
 from unstructured.partition.common.metadata import set_element_hierarchy
 from unstructured.utils import get_call_args_applying_defaults, lazyproperty
 
-LIBMAGIC_AVAILABLE = bool(importlib.util.find_spec("magic"))
+# Is the magic *module* available?
+MAGIC_AVAILABLE = bool(importlib.util.find_spec("magic"))
+
+# Is the libmagic *library* also available?
+LIBMAGIC_AVAILABLE = False
+if MAGIC_AVAILABLE:
+    try:
+        import magic
+        LIBMAGIC_AVAILABLE = True
+    except ImportError:
+        pass
 
 
 def detect_filetype(
@@ -359,8 +369,6 @@ class _FileTypeDetectionContext:
         file_path = self.file_path
 
         if LIBMAGIC_AVAILABLE:
-            import magic
-
             mime_type = (
                 magic.from_file(file_path, mime=True)
                 if file_path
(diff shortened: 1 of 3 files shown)


metadaddy:
Rebased to `main`. `pytest -v test_unstructured/file_utils` still succeeds.

unstructured/file_utils/filetype.py
```diff
@@ -51,7 +51,17 @@
...
+MAGIC_AVAILABLE = bool(importlib.util.find_spec("magic"))
+
+# Is the libmagic *library* also available?
+LIBMAGIC_AVAILABLE = False
+if MAGIC_AVAILABLE:
+    try:
+        import magic
+        LIBMAGIC_AVAILABLE = True
+    except ImportError:
+        pass
```
aaronsteers, line 64: It might be safe and simpler to keep the original constant name with the new logic. something like...


```suggestion
MAGIC_AVAILABLE: bool
if bool(importlib.util.find_spec("magic")):
    try:
        # This should fail if libmagic not installed:
        import magic
    except ImportError:
        MAGIC_AVAILABLE = False
    else:
        # No failure. Magic is available.
        MAGIC_AVAILABLE = True
else:
    MAGIC_AVAILABLE = False
```

aaronsteers (approved):
For my part at least, looks good! 👍

metadaddy:
Closing, since this was apparently fixed by #3828

PR #3792, opened 2024-11-21
Add timeout to scarf telemetry requests

### LINKED ISSUE #3791: Telemetry request has no configured timeout
The requests in `scarf_analytics` have no timeout set, which can cause an application using the `unstructured` client to hang.

https://github.com/Unstructured-IO/unstructured/blob/3b9b01c502cf5f319fd8bb2427232af96af5c637/unstructured/utils.py#L281C1-L309C18

Right now the only work around is to set `SCARF_NO_ANALYTICS` or `DO_NOT_TRACK`, but that opts out of analytics completely.  A better approach is to set the `timeout` parameter to a reasonable value to avoid deadlocking the application.

Rojuinex:
Resolves #3791 by setting a default timeout of 10 seconds.


cragwolfe:
good call

cragwolfe (approved):

PR #3799, opened 2024-11-26
add script to render html from unstructured elements

mariannaparzych:
Script to render HTML from unstructured elements.

NOTE: This script is not intended to be used as a module.
NOTE: This script is only intended to be used with outputs with non-empty `metadata.text_as_html`.

TODO: It was noted that unstructured_elements_to_ontology func always returns a single page
This script is using helper functions to handle multiple pages. I am not sure if this was intended, or it is a bug - if it is a bug it would require bit longer debugging - to make it usable fast I used workarounds.

Usage: test with any outputs with non-empty `metadata.text_as_html`.
Example files attached.
`[Example-Bill-of-Lading-Waste.docx.pdf.json](https://github.com/user-attachments/files/17922898/Example-Bill-of-Lading-Waste.docx.pdf.json)`

[Breast_Cancer1-5.pdf.json](https://github.com/user-attachments/files/17922899/Breast_Cancer1-5.pdf.json)


unstructured/partition/html/rendered_html_from_elements.py
```diff
@@ -0,0 +1,104 @@
+# pyright: reportPrivateUsage=false
+
+"""
+Script to render HTML from unstructured elements.
+NOTE: This script is not intended to be used as a module.
+NOTE: This script is only intended to be used with "platinum" strategy of the partitioning pipeline.
```
cragwolfe, line 6:
```suggestion
NOTE: For now script is only intended to be used with elements generated with `partition_html(html_parser_version=v2)`

**Comment by cragwolfe** [comment]
since this is not intended to be used as a module, lets move it to a new directory under `/scripts`:

```
/scripts/htmlv2/
```

cragwolfe (approved):

PR #3808, opened 2024-12-04
fix: ipv4 address regex

praktiskt:
I noticed the ipv4 regex is wrong (it only capture one or two-digit octets, e.g. `n.nn.n.nn`). Here's a correction and a bumped test for it.

If you wish I can break out the ipv4 test to its own case, so we don't interfere with the existing `EMAIL_META_DATA_INPUT` ipv6 extraction test.

Side note: The comment at `unstructured/nlp/patterns.py#95` includes a bad ipv4 address example (last octet is wrongfully left-padded with a zero). I left it as it is because I'm not sure if the intention is to include "non-conventional" ipv4 addresses, like octal or hexadecimal octets.


praktiskt:
@cragwolfe Is there anything else you need from me? (Pinging you since you seem to be an active contributor for this project - apologies if I'm mistaken.)

cragwolfe (approved):
Looks good, however please update the version to be `0.16.11-dev0`.
Thanks for the contribution, @praktiskt !

praktiskt:
@cragwolfe Done! Thanks.

PR #3811, opened 2024-12-05
fix doctype parsing error

tbs17:
- per [ticket](https://unstructured-ai.atlassian.net/browse/ML-551), there is a bug in the `unstructured` lib under metrics/evaluate.py that incorrectly retrieves the file extension before the conversion to cct file from paths like '*.pdf.txt' . (see below screenshot)
    - the current status is in the top example
    - we should have the correct version in the bottom example of the screenshot.
   
image

- in addition, i also observe the doctype returned are not aligned, some returning '.*'  and some are returning without the dot. 
  - therefore, i just aligned them to be output into the same version which is '.*".


tbs17:
> Thanks for the fix! and for adding tests 🙇 ❤️
> 
> I have 2 small requests:
> 
> * Could you update **version** file and CHANGELOG.md?
> * Could you fix/update your pre-commit in a way that all the artifacts are not modified?

thank you for the quick review! I just realized the changelog.md. So nowadays, do we still do dev version? i will see how can i change pre-commit~

ajjimeno (approved):
LGTM@

badGarnet (commented):
I want to call out that this is implicitly relying on file name follows a certain pattern. Which the metric functions themselves have no control. In other words, if any code change in the ingestion process that would modify the output filename pattern the (current and previous) method for finding doc type can fail. 
if we know the metric function is only dealing with txt files for example, we can add a protection there by first check if the `suffix` is `.txt` then it is safer to assume `suffixes[-2]` is the original file's suffix.
but really there should be an explicit contract between ingestion functions and metric eval (e.g., passing in the index of the file type suffix in the `suffixes` attribute).

So with the reasoning above I would like to request changing the code so that the suffix index is parameter that can be passed in and it is fine to set the default to whatever works for the existing ingestion code (-2) for now.

But given this is boarder line a nit I won't request changes as an action but just leave a comment here.

PR #3812, opened 2024-12-06
chunk: relax table segregation during chunking

scanny:
**Summary**
Relax table-segregation rule applied during chunking such that a `Table` and `Text`-subtype elements can be combined into a single chunk when the chunking window allows.

**Additional Context**
Until now, `Table` elements have always been segregated during chunking, i.e. a chunk that contained a table would never contain any other element. In certain scenarios, especially when a large chunking window of say 2000 characters is used, this behavior can reduce retrieval effectiveness by isolating the table from surrounding context.


badGarnet (approved):

alexyoung23jj:
@scanny If there are ever multiple Table elements that are combined into a CompositeElement as part of this change, which Table will be referenced by the `text_as_html` field?

scanny:
@alexyoung23jj  It is (now) possible to have more than one `Table` element combined into a single `CompositeElement`. The `.metadata.text_as_html` of that `CompositeElement` would then contain two or more HTML `<table>...</table>` elements.

It should _not_ include a container `<html>` or `<body>` element so should be parseable as an HTML fragment. The HTML parsers I know of accomplish that fine, either by default or with a separate call like `.parse_fragment()`.

If there are any challenges you encounter with this new behavior, we're grateful for you reaching out. A change like this can lead to unexpected consequences and we'd like to hear about any of those sooner than later :)

Thanks for asking the question :)

alexyoung23jj:
@scanny Thanks for the quick reply. So to clarify, is this following scenario correct?

I have a document with 2 small tables and a couple headers/other text that appear around them. These are all combined into a single CompositeElement within my chunk size limit from the output of the chunking. This CompositeElement will only have one `text_as_html` field, but this field will contain two <table></table> groups (concatenated together).

Is that right?

What im basically trying to accomplish is a function that can take a CompositeElement and return the contents of that chunk but with all "OCR'd" table text replaced by the html tables you've defined. I'm currently using the `elements_from_base64_gzipped_json` function to extract the original elements from the CompositeElement and replacing a Table element directly with the `text_as_html` field. This seems to work if I have one Table in that CompositeElement, but I'm unclear on how i should be handling it in the case of arbitrary numbers of Tables inside CompositeElements. Should I just be parsing the <table> groups out and assuming they are in the same order as the original elements?

Would have been awesome if this was included as a breaking change, as I've had to rewrite a large portion of my RAG logic due to the "tables are always their own chunks" assumption changing.

scanny:
@alexyoung23jj If you're running locally, then yes. `.metadata.text_as_html` should have exactly two tables in it. Certain partitioning methods may add HTML renditions of other elements, now and in future, but not so far in this repo. So you should expect to find something like this in `.metadata.text_as_html` for that chunk:
```python
<table>...</table> <table>...</table>
```
all on one line for compactness I believe.

---

Regarding interleaving HTML and text: I would have a strong preference for the `.orig_elements` method I think, at least for handling the text, because it would otherwise be tricky to separate the `Table` element text from that of the other elements.

Chunk text is formed by simple joining, divided by a "blank line" (`"\n\n"`). So I'd be thinking it easier to do that over with special behaviors rather than try to reverse it.

---

Note that another change made in the same PR was to determine chunk length solely on the length of `element.text` and no longer consider the `max(len(text), len(text_as_html))`. Of course `text_as_html` always won that comparison when present for tables. Anyway, there's no guarantee that the chunk text will still fit within your chunking window if you use the HTML as the text for tables (as I'm understanding what you propose to do).

---

Apologies for not anticipating the breakage. I suppose that would have made it release `0.17.0` instead of `0.16.11`. Would that have made a difference for your process? You can still pin to `<0.16.11` can't you? Or am I missing your point there?

PR #3816, opened 2024-12-07
release 0.16.10

tbs17:
release 0.16.10 so that competitor-eval can install and take advantage of the latest change in the metric calculation


badGarnet (approved):

tbs17:
opened another one that has cleaner commit history

Issue #3815, opened 2024-12-06
bug/Unstructured unnecessarily "repairs" then falls back to OCR on extremely large documents


dhdaines:
TL;DR this is because Unstructured.io applies a ["fix" to pdfminer.six](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/patches/pdfminer.py) which actually [introduces another bug](https://github.com/pdfminer/pdfminer.six/issues/1025).

The fix (which I have tested here) is, until the bug is fixed in pdfminer.six, to *not* apply the patch, and to simply avoid the following versions of pdfminer.six which contain the bug: 20231228 and 20240706

The source of the problem is this change to pdfminer.six (which I think was submitted by unstructured?): https://github.com/pdfminer/pdfminer.six/pull/885

This broke the parsing of long content streams, because due to the complex and not particularly useful buffering done by pdfminer.six's "parser" (actually a lexer), it is frequently the case that keywords are split across a buffer boundary.  In my particular testcase (cannot share, sorry) this results in the "Detected invalid dictionary construct for PDFminer" error reported by Unstructured.io, because a dictionary ends up with an odd number of items as `false` is getting split into `fals` and `e`.

"Repairing" the PDF will not help this problem because there isn't anything wrong with the PDF, it just has some long content streams.  So what happens is that (on my document of 1500+pages) unstructured falls back to OCR, which as you can imagine is ... suboptimal.

dhdaines:
If you *must* patch pdfminer.six, then please apply this fix, which fixes both the original problem you patched and the problem you introduce by patching it: https://github.com/pdfminer/pdfminer.six/pull/1030

PhorstenkampFuzzy:
+1

PhorstenkampFuzzy:
Any update on this?

PR #3822, opened 2024-12-12
fix: Correctly patch pdfminer to avoid unnecessarily and unsuccessfully repairing PDFs with long content streams, causing needless and endless OCR

dhdaines:
Fixes: #3815 

Verified on my very large documents that it doesn't unnecessarily and unsuccessfully "repair" them.

You may or may not wish to keep the version check in `patch_psparser`.  Since ~you're pinning the version of pdfminer.six and since it isn't guaranteed that the bug in question will be fixed in the next pdfminer.six release (but it is rather serious, so I should hope so), then perhaps you just want to unconditionally patch it.~ it seems like pinning of versions is only operative when running from Docker (good!) so never mind!  Keep that version check!

Also corrected an import so that if you do feel like using a newer version of pdfminer.six, it won't break on you.


PhorstenkampFuzzy:
Any update on this?

qued (approved):
LGTM!

Here's what I checked:
- Ran the tests with this branch to verify that `invalid-pdf-structure-pdfminer-entire-doc.pdf` is being repaired while `invalid-pdf-structure-pdfminer-one-page.pdf` is not.
- Processed `invalid-pdf-structure-pdfminer-one-page.pdf` (unrepaired) using `fast` strategy with the new code and looked over the results to ensure the doc is being processed properly.
- Processed `invalid-pdf-structure-pdfminer-entire-doc.pdf` (repaired) using `fast` strategy with the new code and looked over the results to ensure the doc is being processed properly.
- Processed linked doc [here](https://github.com/pdfminer/pdfminer.six/issues/1025) with this branch and `main` branch to verify that the doc fails before the fix and succeeds afterwards.

qued:
@dhdaines Thanks for fixing this!

PR #3828, opened 2024-12-14
fix: refine filetype detection

### LINKED ISSUE #3596: bug/file type detection fallback strategy not working
**Describe the bug**
The following line tried to detect whether `libmagic` has been installed, but it actually only detects the installation of python-magic package instead of the true `libmagic` dependency.

https://github.com/Unstructured-IO/unstructured/blob/acd070c5d5064bcf6b98106e335e8145a5f183a3/unstructured/file_utils/filetype.py#L58

Therefore, as long as python-magic is installed, LIBMAGIC_AVAILABLE will always be True, and if libmagic is not installed, import magic will cause an error.

### LINKED ISSUE #3781: bug/check for magic library availability doesn't appear to be correct

scanny:
**Summary**
Fixes a bug where a CSV file with asserted content-type `application/vnd.ms-excel` was incorrectly identified as an XLS file and failed partitioning.

**Additional Context**
The `content_type` argument to partitioning is often authored by the client system (e.g. Unstructured SDK) and is both unreliable and outside the control of the user. In this case the `.csv -> XLS` mapping is correct for certain purposes (Excel is often used to load and edit CSV files) but not for partitioning, and the user has no readily available way to override the mapping.

XLS files as well as seven other common binary file types can be efficiently detected 100% of the time (at least 99.999%) using code we already have in the file detector.

- Promote this direct-inspection strategy to be tried first.
- When DOC, DOCX, EPUB, ODT, PPT, PPTX, XLS, or XLSX is detected, use that file-type.
- When one of those types is NOT detected, clear the asserted `content_type` when it matches any of those types. This prevents the problem seen in the bug where the asserted content type was used to determine the file-type.
- The remaining content_type, guess MIME-type, and filename-extension mapping strategies are tried, in that order, only when direct inspection fails. This is largely the same as it was before.
- Fix #3781 while we were in the neighborhood.
- Fix #3596 as well, essentially an earlier report of #3781.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.16.12-dev2
+## 0.16.12-dev3
 
 ### Enhancements
 
@@ -9,6 +9,7 @@
 ### Fixes
 
 - **Upgrade ruff to latest.** Previously the ruff version was pinned to <0.5. Remove that pin and fix the handful of lint items that resulted.
+- **CSV with asserted XLS content-type is correctly identified as CSV.** Resolves a bug where a CSV file with an asserted content-type of `application/vnd.ms-excel` was incorrectly identified as an XLS file.
 
 ## 0.16.11
 
--- a/test_unstructured/file_utils/test_filetype.py
+++ b/test_unstructured/file_utils/test_filetype.py
@@ -14,15 +14,14 @@ from test_unstructured.unit_utils import (
     LogCaptureFixture,
     Mock,
     example_doc_path,
-    function_mock,
     patch,
     property_mock,
 )
 from unstructured.file_utils.filetype import (
     _FileTypeDetectionContext,
-    _OleFileDifferentiator,
+    _OleFileDetector,
     _TextFileDifferentiator,
-    _ZipFileDifferentiator,
+    _ZipFileDetector,
     detect_filetype,
     is_json_processable,
 )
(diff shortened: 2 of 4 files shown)


cragwolfe:
worth adding a test for the specific bug, "CSV file with asserted content-type application/vnd.ms-excel was incorrectly identified as an XLS file"?

scanny:
> worth adding a test for the specific bug, "CSV file with asserted content-type application/vnd.ms-excel was incorrectly identified as an XLS file"?

@cragwolfe Oh yes, absolutely, that was the first one I added :) it's this one, named a little differently as `test_it_ignores_asserted_XLS_content_type_when_file_is_CSV()`:
https://github.com/Unstructured-IO/unstructured/pull/3828/commits/2d15feafe874bbe4447e247529d7f49c91e96ef7#diff-d88874605ea6403a7c07fd5ebb3bf5d795888e7fb2ecd90f15e60ef2300901a5R364

Line 364 in `‎test_unstructured/file_utils/test_filetype.py` if the link doesn't work quite right.

mrogers23458 (approved):

PR #3835, opened 2024-12-17
Fix issue #3815

PhorstenkampFuzzy:
PR https://github.com/pdfminer/pdfminer.six/pull/885 is merged.
The fix leads to problems with strange behavior of psparser when unstructrued isn't used


PhorstenkampFuzzy:
@cragwolfe What kind of entry do you want / need for your changelog?

cragwolfe:
> @cragwolfe What kind of entry do you want / need for your changelog?

Actually, we are stuck on an older version of https://github.com/Unstructured-IO/unstructured/blob/340a07f/requirements/extra-pdf-image.txt#L148-L153 which is basically due to the dependency on [layoutparser](https://github.com/Unstructured-IO/unstructured-inference/blob/main/requirements/base.in#L2) that should be removed from unstructured-inference.

Issue #2467, opened 2024-01-29
Add possibility to deactivate OCR


thomascerbelaud:
**Is your feature request related to a problem? Please describe.**
I am working on large text-based PDF files, and would like to parse them as fats as possible, while keeping a high resolution (`strategy="hi_res"`). I am interested in extracting tables, and tables from pictures, however I would like to deactivate OCR for images detected as such. The first obvious reason is speed. But also images do not matter much. 

**Describe the solution you'd like**
A keyword argument that would enable or disable OCR would be the most easy thing to code I guess and would be a nice additional feature, especially if it can differentiate between tables and images. Another nice feature would be to not perform OCR on tables for text-based regions, in order to speed the partition process.

**Describe alternatives you've considered**
Adding a new `OCRMode.NO_OCR`

**Additional context**
I am not interested in images s.a. graphs or photos.

dhdaines:
Yes, not only is OCR super slow, but table contents are frequently mangled because `infer_table_structure` uses OCR rather than any existing text.

This seems like an important problem for people who care about the quality of their data (so, I guess, not anyone using Generative AI, but...)

leogout:
I would like this feature very much. Do you need something else from the author ? May I help in any way to make this a part of your next release  ? Thanks.

leogout:
For anyone interested in doing this, here is how I did : 
```python
def detect_layout_only(image_path: Path, confidence_threshold: float = 0.5):
    """
    Detect layout using ONLY the detection model - no OCR.

    This runs detectron2 for layout detection but does NOT extract text.

    Args:
        image_path: Path to the image file
        confidence_threshold: Minimum confidence for detections (default: 0.5, library default is 0.8)
    """

    detection_model: UnstructuredDetectronONNXModel = get_model("detectron2_onnx")
    detection_model.confidence_threshold = confidence_threshold

    # Load and prepare image
    image = Image.open(image_path)
    if image.mode != 'RGB':
        image = image.convert('RGB')

    # Run detection model directly
    elements_list: list[LayoutElement] = detection_model.predict(image)

    # Convert list to LayoutElements for deduplication

    elements = LayoutElements.from_list(elements_list)
    # Deduplicate overlapping elements
    elements = detection_model.deduplicate_detected_elements(elements)

    # Extract layout boxes from elements
    layout_boxes = []
    for element in elements.iter_elements():
        layout_boxes.append({
            "type": element.type,
            "bbox": {
[...]

PR #3839, opened 2024-12-17
feat: Allow deactivating OCR entirely with hi_res strategy

dhdaines:
Fixes: #2467 by adding `OCRMode.NONE` with a value of `"none"` (I think this is clearer and easier to type than `no_ocr` but you may have other opinions!)

Note that this does *not* permit `infer_table_structure` to work without OCR as that is considerably more complicated.  I have a pretty good idea how to do it but I think it should be a separate PR.


cragwolfe (approved):

leogout:
@dhdaines thank you very much for your work. May I suggest to add a line in the CHANGELOG.md about this new feature so that the CI passes ?

dhdaines:
> @dhdaines thank you very much for your work. May I suggest to add a line in the CHANGELOG.md about this new feature so that the CI passes ?

Oops!  Yes, good idea :smile: I'll make a note to do this asap.

PR #3851, opened 2024-12-23
add ability to contain nltk data on docker image

christinestraub:


Dockerfile
```diff
@@ -18,12 +18,16 @@ RUN chown -R notebook-user:notebook-user /app && \
...
-  python3.11 -c "from unstructured.nlp.tokenize import download_nltk_packages; download_nltk_packages()" && \
-  python3.11 -c "from unstructured.partition.model_init import initialize; initialize()" && \
-  python3.11 -c "from unstructured_inference.models.tables import UnstructuredTableTransformerModel; model = UnstructuredTableTransformerModel(); model.initialize('microsoft/table-transformer-structure-recognition')"
+RUN find requirements/ -type f -name "*.txt" -exec pip3.11 install --no-cache-dir --user -r '{}' ';'
+
+RUN python3.11 -c "import os; os.makedirs('/home/notebook-user/nltk_data', exist_ok=True)" && \
+    python3.11 -c "from nltk.downloader import download; download('punkt_tab'); download('averaged_perceptron_tagger_eng')"
+
+RUN python3.11 -c "from unstructured.partition.model_init import initialize; initialize()" && \
+    python3.11 -c "from unstructured_inference.models.tables import UnstructuredTableTransformerModel; model = UnstructuredTableTransformerModel(); model.initialize('microsoft/table-transformer-structure-recognition')"
```
cragwolfe, line 27: is there a reason to move these into 3 separate docker layers (with the 3 different `RUN`s)? (generally fewer is better)

Dockerfile
```diff
@@ -18,12 +18,16 @@ RUN chown -R notebook-user:notebook-user /app && \
...
 
 ENV PATH="${PATH}:/home/notebook-user/.local/bin"
 ENV TESSDATA_PREFIX=/usr/local/share/tessdata
+ENV NLTK_DATA=/home/notebook-user/nltk_data
```
cragwolfe, line 31: should this be before the relevant RUN commands?

PR #3853, opened 2024-12-26
Feat/contain nltk assets in docker image

christinestraub:
This pull request adds NLTK data to the Docker image by pre-packaging the data to ensure a more reliable and efficient deployment process, as the required NLTK resources are readily available within the container.

**Current updated solution:**
- Dockerfile Update: Integrated NLTK data directly into the Docker image, ensuring that the API can operate independently of external - data sources. The data is stored at /home/notebook-user/nltk_data.
- Environment Variable Setup: Configured the NLTK_PATH environment variable, enabling Python scripts to automatically locate and use the embedded NLTK data. This eliminates the need for manual configuration in deployment environments.
- Code Cleanup: Removed outdated code in tokenize.py and related scripts that previously downloaded NLTK data from S3. This streamlines the codebase and removes unnecessary dependencies.
- Script Updates: Updated tokenize.py and test_tokenize.py to utilize the NLTK_PATH variable, ensuring consistent access to the embedded data across all environments.
- Dependency Elimination: Fully eliminated reliance on the S3 bucket for NLTK data, mitigating risks from network failures or access changes.
- Improved System Reliability: By embedding assets within the Docker image, the API now has a self-contained setup that ensures consistent behavior regardless of deployment location.
- Updated the Dockerfile to copy the local NLTK data to the appropriate directory within the container.
- Adjusted the application setup to verify the presence of NLTK assets during the container build process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 ### Fixes
 
 - **Fix NLTK Download** to use nltk assets in docker image
-
+- removed the ability to automatically download nltk package if missing
+  
 ## 0.16.12
 
 ### Enhancements


Dockerfile
```diff
@@ -9,21 +9,26 @@ COPY unstructured unstructured
...
+    fc-cache -fv && \
+    [ -e /usr/bin/python3 ] || ln -s /usr/bin/python3.11 /usr/bin/python3
 
 USER notebook-user
 
-RUN find requirements/ -type f -name "*.txt" -exec pip3.11 install --no-cache-dir --user -r '{}' ';' && \
-  python3.11 -c "from unstructured.nlp.tokenize import download_nltk_packages; download_nltk_packages()" && \
-  python3.11 -c "from unstructured.partition.model_init import initialize; initialize()" && \
-  python3.11 -c "from unstructured_inference.models.tables import UnstructuredTableTransformerModel; model = UnstructuredTableTransformerModel(); model.initialize('microsoft/table-transformer-structure-recognition')"
+RUN find requirements/ -type f -name "*.txt" -exec pip3.11 install --no-cache-dir --user -r '{}' ';'
```
badGarnet, line 22: lets define args `PYTHON` and `PIP` then reuse that definition throughout the dockerfile

nltk_data/taggers/averaged_perceptron_tagger_eng/averaged_perceptron_tagger_eng.classes.json
badGarnet, line 1: gitignore? I don't think it is the intention to have nltk data in the repo, right?

unstructured/nlp/tokenize.py
```diff
@@ -11,66 +11,60 @@
 
 CACHE_MAX_SIZE: Final[int] = 128
 
+NLTK_DATA_PATH = os.getenv("NLTK_DATA", "/home/notebook-user/nltk_data")
+nltk.data.path.append(NLTK_DATA_PATH)
```
badGarnet, line 15: doesn't `NLTK_DATA` set the location of data for `nltk` out of the box? https://www.nltk.org/data.html#interactive-installer

unstructured/nlp/tokenize.py
```diff
@@ -11,66 +11,60 @@
...
 
 
-# We cache this because we do not want to attempt
-# downloading the packages multiple times
-@lru_cache()
-def _download_nltk_packages_if_not_present():
-    """If required NLTK packages are not available, download them."""
+# Ensure NLTK data exists in the specified path (pre-baked in Docker)
+def validate_nltk_assets():
+    """Validate that required NLTK packages are preloaded in the image."""
```
badGarnet, line 34:
```suggestion
    """Validate that required NLTK packages are preloaded in the environment."""
```

unstructured/nlp/tokenize.py
```diff
@@ -11,66 +11,60 @@
...
+    required_assets = [
+        ("punkt_tab", "tokenizers"),
+        ("averaged_perceptron_tagger_eng", "taggers"),
+    ]
+    for package_name, category in required_assets:
+        if not check_for_nltk_package(package_name, category):
+            raise RuntimeError(
+                f"Required NLTK package '{package_name}' is missing. "
+                f"Ensure it is baked into the Docker image at '{NLTK_DATA_PATH}'."
```
badGarnet, line 43: this message is specific to a docker image use case but user can get here when just using unstructured library, right? So maybe update this message to prompt user to download nltk data first.

Dockerfile
```diff
@@ -9,21 +9,26 @@ COPY unstructured unstructured
 COPY test_unstructured test_unstructured
 COPY example-docs example-docs
 
+# Copy the downloaded NLTK data folder to your local environment.s
+COPY ./nltk_data /home/notebook-user/nltk_data
```
badGarnet, line 13: maybe better to parametrize the source path instead of assuming it will always be in the same directory as the dockerfile?

test_unstructured_ingest/test-ingest-src.sh
```diff
@@ -40,8 +40,8 @@ all_tests=(
   'against-api.sh'
   'gcs.sh'
   'kafka-local.sh'
-  'onedrive.sh'
-  'outlook.sh'
+  #'onedrive.sh'
+  #'outlook.sh'
```
badGarnet, line 44: are those intended as part of the PR?

unstructured/nlp/tokenize.py
```diff
@@ -33,44 +33,45 @@ def check_for_nltk_package(package_name: str, package_category: str) -> bool:
...
+        package_name="punkt_tab",
     )
 
-    if (not tokenizer_available) or (not tagger_available):
-        download_nltk_packages()
+    if not tagger_available or not tokenizer_available:
+        raise RuntimeError(
+            "Required NLTK packages are not available. "
+            "Ensure the assets are pre-baked into the image."
+        )
```
badGarnet, line 53: wouldn't nltk raise an error with similar message about missing assets without us implementing this? This function was useful when it was also tasked to download the packages when missing. But if we are going to disable that behavior this function really is not needed at all

badGarnet (approved):

ajjimeno (approved):
LGTM!

cragwolfe (approved):

badGarnet (approved):

davidlmorton:
I agree @NolanTrem, it was surprising for us that a bugfix version bump would break our tests.  Downloading the NLTK stuff at container build time is fine, removing automatic downloading if found missing is not backwards compatible for a lot of folks.

PR #3860, opened 2025-01-06
Character confidence threshold

plutasnyy:
This change adds the ability to filter out characters predicted by Tesseract with low confidence scores.

Some notes:
- I intentionally disabled it by default; I think some low score(like 0.9-0.95 for Tesseract) could be a safe choice though 
- I wanted to use character bboxes and combine them into word bbox later. However, a bug in Tesseract in some specific scenarios returns incorrect character bboxes (unit tests caught it 🥳 ). More in comment in the code

### DIFF
--- a/unstructured/partition/utils/ocr_models/tesseract_ocr.py
+++ b/unstructured/partition/utils/ocr_models/tesseract_ocr.py
@@ -127,12 +127,17 @@ class OCRAgentTesseract(OCRAgent):
                     {
                         "left": left,
                         "top": top,
-                        "width": right - left,
-                        "height": bottom - top,
+                        "right": right,
+                        "bottom": bottom,
                         "text": text,
                     }
                 )
-        ocr_df = pd.DataFrame(df_entries, columns=["left", "top", "width", "height", "text"])
+        ocr_df = pd.DataFrame(df_entries, columns=["left", "top", "right", "bottom", "text"])
+
+        ocr_df["width"] = ocr_df["right"] - ocr_df["left"]
+        ocr_df["height"] = ocr_df["bottom"] - ocr_df["top"]
+
+        ocr_df = ocr_df.drop(columns=["right", "bottom"])
         return ocr_df
 
     @staticmethod


unstructured/partition/utils/config.py
```diff
@@ -96,6 +96,11 @@ def TESSERACT_OPTIMUM_TEXT_HEIGHT(self) -> int:
         """optimum text height for tesseract OCR"""
         return self._get_int("TESSERACT_OPTIMUM_TEXT_HEIGHT", 20)
 
+    @property
+    def TESSERACT_CHARACTER_CONFIDENCE_THRESHOLD(self) -> int:
+        """Tesseract predictions with confidence below this threshold are ignored"""
+        return self._get_float("TESSERACT_CHARACTER_CONFIDENCE_THRESHOLD", 0.0)
```
MaksOpp, line 102: I wonder, maybe we'd like to have some really low default threshold, i.e. 0.1, just to filter out complete garbage chars?

unstructured/partition/utils/ocr_models/tesseract_ocr.py
```diff
@@ -76,17 +77,89 @@ def get_layout_from_image(self, image: PILImage.Image) -> List[TextRegion]:
...
         ocr_regions = self.parse_data(ocr_df, zoom=zoom)
 
         return ocr_regions
 
+    def image_to_data_with_character_confidence_filter(
+        self,
+        image: np.ndarray,
+        lang: str = "eng",
+        config: str = "",
+        character_confidence_threshold: float = 0.5,
```
MaksOpp, line 95: Here we are adding some default, so maybe let's also keep it in config?

MaksOpp: I see below we again have 0.5 as a default in `hocr_to_dataframe`, so either way, I would unify those

MaksOpp (approved):
LGTM!

unstructured/partition/utils/config.py
badGarnet: I am ok with 0; the default behavior is no filter at all so this PR should just keep that for now. We can use followups to change this value.

unstructured/partition/utils/ocr_models/tesseract_ocr.py
```diff
@@ -76,17 +77,89 @@ def get_layout_from_image(self, image: PILImage.Image) -> List[TextRegion]:
...
+        hocr: str = unstructured_pytesseract.image_to_pdf_or_hocr(
+            image,
+            lang=lang,
+            config="-c hocr_char_boxes=1 " + config,
+            extension="hocr",
+        )
+        ocr_df = self.hocr_to_dataframe(hocr, character_confidence_threshold)
+        return ocr_df
+
+    def hocr_to_dataframe(
```
badGarnet, line 106: what's the compute performance with this code? We essentially were relying on tesseract internal cpp code to parse results but here we do it in python.

unstructured/partition/utils/ocr_models/tesseract_ocr.py
```diff
@@ -76,17 +77,89 @@ def get_layout_from_image(self, image: PILImage.Image) -> List[TextRegion]:
...
+            )
+            if text and bbox_match:
+                word_bbox = list(map(int, bbox_match.groups()))
+                left, top, right, bottom = word_bbox
+                df_entries.append(
+                    {
+                        "left": left,
+                        "top": top,
+                        "width": right - left,
+                        "height": bottom - top,
```
badGarnet, line 131: small nit on performance we can create df using bbox first then use vector ops to compute width and height (and overwrite the data for right and bottom).

PR #3869, opened 2025-01-16
add post chunking strategy

tbs17:
this draft PR in unstructured intends to do below:
- modified dispatch.py for the chunk() to include contextual_chunking_spec to fetch chunker func chunk_by_contextual in core-product


scanny (commented):
Will not review.

PR #3876, opened 2025-01-18
Fix password

Coniferish:


Coniferish:
@pprados, I went ahead and rebased your branch and made some linting fixes for your branch. It looks like the tests that are failing right now stem from not having your PR in `unstructured_inference` not merged yet. So it seems like we just need to get your changes in that repo and then we can get this merged. For future reference, though, you can can run most of the tests that our CI runs with `make test` and `make tidy`.

pprados:
@Coniferish

The [PR392](https://github.com/Unstructured-IO/unstructured-inference/pull/392) in unstructured-inference MUST be validated before this one.

can you take a look?

PR #3880, opened 2025-01-20
update unstructured-inference lib

tbs17:
update unstructured-inference to 0.8.6 in requirements in extra-pdf-image.in

0.8.6 has pdfminer=20240706 (newer version)

### DIFF
--- a/.github/workflows/ingest-test-fixtures-update-pr.yml
+++ b/.github/workflows/ingest-test-fixtures-update-pr.yml
@@ -109,6 +109,7 @@ jobs:
           sudo apt-get install -y tesseract-ocr-kor
           sudo apt-get install diffstat
           tesseract --version
+          python -m nltk.downloader punkt_tab averaged_perceptron_tagger_eng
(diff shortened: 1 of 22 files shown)


tbs17:
> * pdfminer is not bumped to the version you would like it:
> 
> ```
> pdfminer-six==20231228
>     # via
>     #   -r ./extra-pdf-image.in
>     #   pdfplumber
> ```
> 
> * `sources` related changes were introduced in [Feat/add more attributes to layoutelements unstructured-inference#404](https://github.com/Unstructured-IO/unstructured-inference/pull/404)

i did first try to change it to 20240706 the newer version, however, it introduced pycrypto package which caused compatibility issue in CI so i reverted it back. I had all the checks passed when there was no update on pdfminer for once. 

But now even if i reverted them back, it didn't pass. working on to see if there's anything else...

ajjimeno (approved):
Looks like bumping a few libraries, which seems good.

PR #3883, opened 2025-01-22
added auto_download logic to download data runtime

christinestraub:
- **Add auto-download for NLTK for Python Enviroment** When user import `tokenize`, It will automatically download nltk data. 

- Added `AUTO_DOWNLOAD_NLTK` flag in `tokenize.py` to download `NLTK_DATA`


setup.py
```diff
@@ -128,4 +128,9 @@ def load_requirements(file_list: Optional[Union[str, List[str]]] = None) -> List
     },
     package_dir={"unstructured": "unstructured"},
     package_data={"unstructured": ["nlp/*.txt", "py.typed"]},
+    entry_points={
+        "console_scripts": [
+            "post_install_unstructured=unstructured.nlp.tokenize:download_nltk_packages"
```
badGarnet, line 133: this solution doesn't work for installing from `.whl` distribution though

cragwolfe (approved):

CHANGELOG.md
```diff
@@ -6,12 +6,13 @@
 - **Vectorize layout (inferred, extracted, and OCR) data structure** Using `np.ndarray` to store a group of layout elements or text regions instead of using a list of objects. This improves the memory efficiency and compute speed around layout merging and deduplication.
 
 ### Fixes
+- **Add auto-download for NLTK for Python Enviroment** When user install python library without image. It will automatic download nltk data from `tokenize.py` file
 - **Correctly patch pdfminer to avoid PDF repair**. The patch applied to pdfminer's parser caused it to occasionally split tokens in content streams, throwing `PDFSyntaxError`.  Repairing these PDFs sometimes failed (since they were not actually invalid) resulting in unnecessary OCR fallback.
 
+
```
cragwolfe, line 12: extra newline

unstructured/nlp/tokenize.py
```diff
@@ -32,6 +27,30 @@ def check_for_nltk_package(package_name: str, package_category: str) -> bool:
         return False
 
 
+# We cache this because we do not want to attempt
+# downloading the packages multiple times
+@lru_cache()
```
badGarnet, line 32: Based on the logic below we don't need cache. Further more (this is kind of pathological edge case but...) caching means we could ignore situations where between checks the nltk data is removed.

badGarnet (approved):
probably need a version bump as well?

cragwolfe (approved):

PR #3886, opened 2025-01-23
bugfix/drop use of ndjson dep, use local code

rbiseck3:
### Description
Avoid using the ndjson dependency due to the limiting license that exists on it

### DIFF
--- a/unstructured/file_utils/ndjson.py
+++ b/unstructured/file_utils/ndjson.py
@@ -1,24 +1,62 @@
 import json
+from typing import IO, Any
 
 
-def dumps(obj: list[dict], **kwargs) -> str:
-    lines = []
-    for each in obj:
-        lines.append(json.dumps(each, **kwargs))
-    return "\n".join(lines)
+def dumps(obj: list[dict[str, Any]], **kwargs) -> str:
+    """
+    Converts the list of dictionaries into string representation
+
+    Args:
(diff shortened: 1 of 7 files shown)


unstructured/file_utils/ndjson.py
```diff
@@ -0,0 +1,24 @@
+import json
```
scanny, line 1: In my view every module should have a docstring answering the question "What would I want to know about this module if I were encountering for the first time?"

I can think of a few topics:
- Why do we need `ndjson`?
- Is it something we partition?
- Who's using it now?
- How robust is this?
- Why didn't we use a standard library for this?

Perhaps some others will occur to you.

It's up to each of us to keep the code-base understandable for all those who end up here and using prose to expose the non-obvious aspects is an important part of that in my view.

unstructured/file_utils/ndjson.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def dumps(obj: list[dict], **kwargs) -> str:
+    lines = []
+    for each in obj:
+        lines.append(json.dumps(each, **kwargs))
+    return "\n".join(lines)
```
scanny, line 8: A matter of style, but this can be:
```python
return "\n".join(json.dumps(each, **kwargs) for each in obj)

Which is both more concise and more efficient.

scanny (approved):
LGTM :)

Issue #3823, opened 2024-12-12
Hard upper bound on Python version considered harmful


dhdaines:
Unstructured's `setup.py` contains this line:

    python_requires=">=3.9.0,<3.13",

which causes no end of problems when people want to use Unstructured as a library, and also want to use other libraries at the same time.  It also wastes a lot of bandwidth.  This is particularly problematic when using Poetry.

For why you should not do this: https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special

> And, if you use Poetry, as soon as someone caps the Python version, every Poetry project that uses it must also cap, even if you believe it is a detestable practice and confusing to users. It is also wrong unless you fully pin the dependency that forced the cap - if the dependency drops it in a patch release or something else you support, you no longer would need the cap. Even worse, if someone adds a cap or tightens a cap, unless they yank every single older release, a locking solver like Poetry or PDM will backsolve to the last versions without the cap so that the lock file it creates will be “valid” on all the Python versions you are requesting!

lucasgadams:
Agreed, I just hit this issue with poetry. I want to upgrade my application to python3.13 to deal with some vulnerability warnings, and unstructured is the only library preventing me from doing this. Is there any reason why unstructured can't be used with python3.13?

PR #3893, opened 2025-01-29
fix: allow users to install package with python3.13 or higher

cragwolfe:
Although, python3.13 is not officially supported or tested in CI just yet.


cragwolfe:
Addresses https://github.com/Unstructured-IO/unstructured/issues/3823

PR #3898, opened 2025-01-31
Fix/fix table id checking logic

badGarnet:
- there is a bug in deciding if a page has tables before performing table extraction. This logic checks if the id associated with Table type element is True
- however, it should be checking if the id is `None` because sometimes the id can be 0 (the first type of element in the page)
- the fix updates the logic
- adds a unit test for this specific case


unstructured/partition/pdf_image/ocr.py
```diff
@@ -276,7 +276,7 @@ def supplement_element_with_table_extraction(
     from unstructured_inference.models.tables import cells_to_html
 
     table_id = {v: k for k, v in elements.element_class_id_map.items()}.get(ElementType.TABLE)
-    if not table_id:
+    if table_id is None:
```
ryannikolaidis, line 279: 🤦

ryannikolaidis (approved):

PR #3900, opened 2025-02-04
Feat/vectorize layout merging

badGarnet:
This PR rewrites the logic in `unstructured_inference` that merges extracted with inferred layout using vectorized operations. The goal is to:
- vectorize the operation to improve memory and cpu efficiency
- apply logic equally without order being a factor (the `unstructured_inference` version uses loops and modifies the content of the inner loop on the fly -> order of the out loop, which is the order of extracted elements becomes a factor) determining the merging results
- rewrite the loop into clear steps with clear rules
- setup stage for followup improvements

While this PR aim to reproduce the existing behavior as much as possible it is not an exact replica of the looped version. Because order is not a factor any more some extracted elements that used to be not considered part of a larger inferred element (due to processing order being not optimum) are now properly merged. This lead to changes in one ingest test. For example, the change shows that now we properly merge the section numerical number with the section title as the full title element. 

## Test:

Since the goal of this refactor is to preserve as much existing behavior as possible we rely on existing tests. As mentioned above the one file that changed output during ingest test is a net positive change.


pawel-kmiecik (approved):
LGTM!
Just a nit: I'd add type hints in all functions and some docstrings, especially for `_merge` and `_mark` functions

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -278,17 +589,12 @@ def merge_inferred_with_extracted_layout(
 ) -> "DocumentLayout":
     """Merge an inferred layout with an extracted layout"""
 
-    from unstructured_inference.inference.layoutelement import LayoutElements
-    from unstructured_inference.inference.layoutelement import (
-        merge_inferred_layout_with_extracted_layout as merge_inferred_with_extracted_page,
-    )
```
tbs17, line 284: i see previously we are importing the func `merge_inferred_layout_with_extracted_layout` as `merge_inferred_with_extracted_page`, that means it relies on `unstructured-inference` package. 

now here, after removing this import, it seems to decouple the `unstructured-inference` and unstructured, will there be updates in `unstructured-inference` to sync? the affected use case would be: some other repo is relying on `unstructured-inference` for that function instead of importing from `unstructured`

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -50,6 +52,304 @@ def _validate_bbox(bbox: list[int | float]) -> bool:
...
+            )
+        )
+
+    if len(inferred_layout_to_proc) == 0:
+        return extracted_layout.slice(np.concatenate((image_indices_to_keep, text_element_indices)))
+
+    extracted_text_layouts = extracted_layout.slice(text_element_indices)
+    # ==== RULE 2. if there is a inferred region almost the same as the extracted text-region ->
+    # keep inferred and removed extracted region; here we put more trust in OD model more than
+    # pdfminer for bounding box
```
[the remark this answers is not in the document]
badGarnet: Not sure to be honest. For the moment we generally observe OD model has pretty good bound boxes and definitely richer ontology of element types so here it is more a statement that: "we keep OD model's element type"

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -50,6 +52,304 @@ def _validate_bbox(bbox: list[int | float]) -> bool:
...
+                threshold=FULL_PAGE_REGION_THRESHOLD,
+            )
+            .sum(axis=1)
+            .astype(bool)
+        )
+        image_indices_to_keep = image_indices_to_keep[~full_page_image_mask]
+
+    # ==== RULE 1: any inferred box that is almost the same as an extracted image box, inferred is
+    # removed
+    # NOTE (yao): what if od model detects table but pdfminer says image -> we would lose the table
```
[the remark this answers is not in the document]
badGarnet: right; if text is extractable in general this is not a problem. It is only a problem when table is in a container that pdfminer marks as image, which can happen.

unstructured/partition/pdf_image/pdfminer_processing.py
badGarnet: I think long term goal is to fold inference into unstructured as we see currently the setup is slowing down development. And as this refactor suggests many functions really need both repos together to be properly defined (e.g., types defined in inference)

test_unstructured_ingest/expected-structured-output/local-single-file-with-pdf-infer-table-structure/layout-parser-paper.pdf.json
```diff
@@ -2376,9 +2376,53 @@
       }
     }
   },
+  {
+    "type": "FigureCaption",
```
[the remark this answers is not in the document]
badGarnet: great catch; this is actually a bug in refactoring; fixing that

test_unstructured_ingest/expected-structured-output/google-drive/recalibrating-risk-report.pdf.json
```diff
@@ -4773,9 +4773,71 @@
       }
     }
   },
+  {
+    "type": "Image",
```
badGarnet, line 4777: now the two side by side photos are recognized as two images instead of just one

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -50,6 +52,322 @@ def _validate_bbox(bbox: list[int | float]) -> bool:
...
+    extracted_to_proc: np.ndarray,
+    inferred_to_proc: np.ndarray,
+) -> LayoutElements:
+    """merged extracted elements that are subregions of inferrred elements into those inferred
+    elements: the inferred elements' bounding boxes expands, if needed, to include those subregion
+    elements. Returns the modified inferred layout where some of its elements' bounding boxes may
+    have expanded due to merging.
+    """
+    # in theory one extracted __should__ only match at most one inferred region, given inferred
+    # region can not overlap; so first match here __should__ also be the only match
```
tbs17, line 149: maybe it's rare to see inferred regions being overlapped, but if that does happen, how much of the below code still would hold? maybe a test on such edge case?

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -50,6 +52,322 @@ def _validate_bbox(bbox: list[int | float]) -> bool:
...
+        # it is not clear which one is overall faster so might worth profiling in the future
+        extracted_to_iter[matches] = False
+        inferred_to_iter[inferred_index] = False
+        # then expand inferred box by all the extracted boxes
+        # FIXME (yao): this part is broken at the moment
+        inferred_layout.element_coords[[inferred_index]] = _minimum_containing_coords(
+            inferred_layout.slice([inferred_index]),
+            *[extracted_layout.slice([match]) for match in matches],
+        )
+    inferred_to_proc[inferred_to_proc] = inferred_to_iter
```
tbs17, line 166: are the use of inferred_to_iter and extracted_to_iter necessary? i got confused when i first saw `inferred_to_proc`  and `extracted_to_proc` didn't get returned. 

would below work?

```
def _merge_extracted_that_are_subregion_of_inferred_text(
    extracted_layout,
    inferred_layout,
    extracted_is_subregion_of_inferred,
    extracted_to_proc,
    inferred_to_proc,
):
    """Merge extracted regions that are subregions of inferred text regions."""
    for inferred_index, inferred_row in enumerate(extracted_is_subregion_of_inferred.T):
        matches = np.where(inferred_row)[0]
        if not matches.size:
            continue
        # Mark matched regions as processed
        extracted_to_proc[matches] = False
        inferred_to_proc[inferred_index] = False
        # Expand inferred bounding box to encompass matched extracted regions
        inferred_layout.element_coords[[inferred_index]] = _minimum_containing_coords(
            inferred_layout.slice([inferred_index]),
            *[extracted_layout.slice([match]) for match in matches],
        )
    return inferred_layout
```

unstructured/partition/pdf_image/pdfminer_processing.py
```diff
@@ -50,6 +52,304 @@ def _validate_bbox(bbox: list[int | float]) -> bool:
...
+    for inferred_index, inferred_row in enumerate(extracted_is_subregion_of_inferred.T):
+        matches = np.where(inferred_row)[0]
+        if not matches.size:
+            continue
+        # Technically those two lines below can be vectorized but this loop would still run anyway;
+        # it is not clear which one is overall faster so might worth profiling in the future
+        extracted_to_iter[matches] = False
+        inferred_to_iter[inferred_index] = False
+        # then expand inferred box by all the extracted boxes
+        # FIXME (yao): this part is broken at the moment
```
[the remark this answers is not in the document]
tbs17: is `this part` still broken based on the comment?

PR #3905, opened 2025-02-05
bugfix/fix ndjson detection

rbiseck3:
### Description
NDJSON files were being detected as JSON due to having the same mime-type. This adds additional logic to skip mime-type based detection if extension is `.ndjson`


jordan-homan (approved):

unstructured/file_utils/filetype.py
```diff
@@ -172,7 +172,8 @@ def _file_type(self) -> FileType:
             return file_type
 
         # -- strategy 3: guess MIME-type using libmagic and use that --
-        if file_type := self._file_type_from_guessed_mime_type:
+        if file_type := self._file_type_from_guessed_mime_type and self._ctx.extension != ".ndjson":
+            # ndjson gets detected as json mime type, will be incorrectly classified as JSON
```
scanny, line 176: I would put this logic _within_ `._file_type_from_guessed_mime_type` as it's very similar to other ambiguous MIME-type guesses like this on line 231:
```python
        if mime_type.endswith("xml"):
            return FileType.HTML if extension in (".html", ".htm") else FileType.XML
```
so something like:
```python
        if mime_type.endswith("json"):  # -- or whatever will reliably catch JSON MIME-type --
            return FileType.NDJSON if extension == ".ndjson") else FileType.JSON
```
Rationale being the special-casing can happen at a lower level and there are other parallel examples there. So it doesn't encourage having a bunch of special-case if-else branches growing at the top level.

scanny (changes requested):
I propose moving this lower down, otherwise LGTM.

Let me know and I can approve it then.

scanny:
Also @rbiseck3 I'd like to see a test for this. Good chance to fill in for the other possible cases, like when it comes with an asserted content-type.

scanny:
The test suite is well developed, you should be able to do most of it with just adding a case to parameterized tests.

unstructured/file_utils/filetype.py
rbiseck3: Good call, it's been moved there.

scanny (approved):
LGTM :)

PR #3916, opened 2025-02-10
fix/Prepare for breaking change in upcoming unstructured-client

awalker4:
In the unreleased 0.30.0 of unstructured-client, handling of the server_url is changing. Instead of passing a url to the instance of the client, it's preferred that we set the server_url per endpoint. We can make this change against existing versions of the client in order to be compatible when 0.30.0 is released.

### DIFF
--- a/unstructured/partition/api.py
+++ b/unstructured/partition/api.py
@@ -94,7 +94,7 @@ def partition_via_api(
     # Note(austin) - the sdk takes the base url, but we have the full api_url
     # For consistency, just strip off the path when it's given
     base_url = api_url[:-19] if "/general/v0/general" in api_url else api_url
-    sdk = UnstructuredClient(api_key_auth=api_key, server_url=base_url)
+    sdk = UnstructuredClient(api_key_auth=api_key)
 
     if filename is not None:
         with open(filename, "rb") as f:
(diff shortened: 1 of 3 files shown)


unstructured/partition/api.py
```diff
@@ -94,7 +94,7 @@ def partition_via_api(
     # Note(austin) - the sdk takes the base url, but we have the full api_url
     # For consistency, just strip off the path when it's given
     base_url = api_url[:-19] if "/general/v0/general" in api_url else api_url
-    sdk = UnstructuredClient(api_key_auth=api_key, server_url=base_url)
+    sdk = UnstructuredClient(api_key_auth=api_key)
```
cragwolfe, line 97: can we inspect the version of UnstructuredClient to do the right thing?

cragwolfe (changes requested):

unstructured/partition/api.py
awalker4: Ah, will do! I took it for granted that passing `server_url` in the method has been backwards compatible, but it looks like `unstructured` pins the client at a much older version and this does break CI. Actually, I'll close this for now. We constrain the client to below 0.26.0, so this won't block us publishing the new SDK. Looks like we have some tech debt to get the tests updated for later versions.

PR #3925, opened 2025-02-18
chore: update dependencies

ahmetmeleq:

PR #3928, opened 2025-02-18
Luke/CVE bump

luke-kucing:
bumping dependancies and updated the tokenizer constraint


requirements/deps/constraints.txt
```diff
@@ -18,3 +18,5 @@ botocore<1.34.132
 importlib-metadata>=8.5.0
 # (austin): Versions below this have a different interface for passing parameters
 unstructured-client>=0.23.0,<0.26.0
+# (luke): conflicting versions installed. pinned to 5.3.0 for now.
+lxml==5.3.0
```
six5532one, line 22: Looks like @ahmetmeleq also opened a PR to address CVEs but it doesn't include this pin. Would the changes in https://github.com/Unstructured-IO/unstructured/pull/3925 resolve the CVE?

luke-kucing: looks like that PR would resolve the cryptography CVE and label-studio sdk but would not bump transformers from 4.442 to >=4.48.0

luke-kucing: tested removing the constraint and it doesnt seem like anything required 5.3.0 so that might have just been an error on my part

tabossert:
If the image seems like it's working, the changes seem good to me

tabossert (approved):

six5532one (approved):

PR #3945, opened 2025-03-05
feat: include images when partitioning html

ryannikolaidis:
Currently we [filter img tags](https://github.com/Unstructured-IO/unstructured/blob/2addb19473ba9e27af995291f57d35fb50bec4b0/unstructured/partition/html/partition.py#L226-L229) before tags are converted to Elements by the html partitioner. More importantly we also don’t currently have a defined “block” / mapping to support these. This adds these mappings and logic to process.

It also respects `extract_image_block_types` and `extract_image_block_to_payload` (as we do with pdfs) to determine whether base64 is included in the metadata.

The partitioned Image Elements sets the text to the img tag’s alt text if available.

The partitioned Image Elements include the [url in the metadata](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/documents/elements.py#L209) (rather than image_base64) if the img tag src is a url.

## Testing

unit tests have been added for explicit coverage.
existing integration tests and other unit test fixtures have been updated to account for `Image` elements now present


unstructured/partition/html/partition.py
```diff
@@ -202,6 +215,10 @@ def _iter_elements(self) -> Iterator[Element]:
         for e in elements_iter:
             e.metadata.last_modified = self._opts.last_modified
             e.metadata.detection_origin = self._opts.detection_origin
+
+            # -- remove <image_base64> if not requested --
+            if not self._should_include_image_base64(e):
+                e.metadata.image_base64 = None
```
ryannikolaidis, line 221: not my favorite pattern, but this seems like the cleanest solution without refactoring how we parse downstream

cragwolfe (approved):

PR #3946, opened 2025-03-06
Enable dynamic file type registration

vangheem:
The purpose of this PR is to enable registering new file types dynamically.

The PR enables this through 2 primary functions:

1. `unstructured.file_utils.model.create_file_type` This registers the new `FileType` enum which enables the rest of unstructured to understand a new type of file
2. `unstructured.file_utils.model.register_partitioner` Decorator that enables registering a partitioner function to run for a file type.

### DIFF
--- a/unstructured/file_utils/model.py
+++ b/unstructured/file_utils/model.py
@@ -66,7 +66,7 @@ class FileType(enum.Enum):
     _alias_mime_types: tuple[str, ...]
     """MIME-types accepted as identifying this file-type."""
 
-    _partitioner_full_module_path: str | None
+    _partitioner_full_module_path: Optional[str]
     """Fully-qualified name of module providing partitioner for this file-type."""
 
     def __new__(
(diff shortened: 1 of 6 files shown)


badGarnet (approved):
probably bump version numbers as well?

unstructured/file_utils/model.py
```diff
@@ -30,6 +66,9 @@ class FileType(enum.Enum):
     _alias_mime_types: tuple[str, ...]
     """MIME-types accepted as identifying this file-type."""
 
+    _partitioner_full_module_path: str | None
```
rbiseck3, line 69: This repo still supports 3.9 and I don't think this syntax exists in that version. 
```suggestion
    _partitioner_full_module_path: Optional[str]
```

vangheem: All the existing fields already used that style so I think we're okay

PR #3960, opened 2025-03-17
feat: use lxml instead of bs4 to parse hOCR data

badGarnet:
- `lxml` is a much faster library than `bs4` when the input data is regular
- since the hOCR data is guaranteed to be regular (programmatically generated) we don't need `bs4` here to parse the data
- `lxml` improves parsing speed by about 10x

Example runtime profiling locally using the same `hocr` data from 1 page pdf, where `agent.hocr_to_dataframe_bs4` is the current method on main and `agent.hocr_to_dataframe` is the PR's method.

Screenshot 2025-03-17 at 12 14 59 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
-## 0.17.1-dev0
+## 0.17.1-dev1
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


ryannikolaidis (approved):
Amazing! Yes, please!

PR #3962, opened 2025-03-19
Image within div or span with no text is annotated as Image

ajjimeno:
Ticket: https://unstructured-ai.atlassian.net/browse/ML-942

The following uncompressed HTML document can be used to test the transformation using the `partition_html` function from the VLM partitioner.

[recalibrating-risk-report.pdf.json.html.zip](https://github.com/user-attachments/files/19330528/recalibrating-risk-report.pdf.json.html.zip)

### DIFF
--- a/unstructured/partition/html/transformations.py
+++ b/unstructured/partition/html/transformations.py
@@ -437,7 +437,7 @@ def extract_tag_and_ontology_class_from_tag(
         html_tag = "span"
         element_class = ontology.UncategorizedText
 
-    # Scenario 5: Image with no text a ontology.UncategorizedText element_class
+    # Scenario 5: UncategorizedText has image and no text
     # Typically, this happens with a span or div tag with an image inside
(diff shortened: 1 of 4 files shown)


ryannikolaidis:
worth a quick test somewhere just to validate / prevent breaking in the future?

unstructured/partition/html/transformations.py
```diff
@@ -437,6 +437,12 @@ def extract_tag_and_ontology_class_from_tag(
         html_tag = "span"
         element_class = ontology.UncategorizedText
 
+    # Scenario 5: Image with no text a ontology.UncategorizedText element_class
```
ryannikolaidis, line 440: wording here is a bit confusing. 
```suggestion
    # Scenario 5: UncategorizedText has image and no text
```

maybe?

ryannikolaidis (approved):
lgtm!

PR #3969, opened 2025-03-27
build: remove test and dev deps from docker image

qued:
Removed the dependencies contained in `test.txt`, `dev.txt`, and `constraints.txt` from the things that get installed in the docker image. In order to keep testing the image (running the tests), I added a step to the `docker-test` make target to install `test.txt` and `dev.txt`. Thus we presumably get a smaller image (probably not much smaller), reduce the dependency chain or our images, and have less exposure to vulnerabilities while still testing as robustly as before.

Incidentally, I removed the `Dockerfile` for our ubuntu image, since it made reference to non-existent make targets, which tells me it's stale and wasn't being used.

### Review:
- Reviewer should ensure the dev and test dependencies are not being installed in the docker image. One way to check is to check the logs in CI, and note, e.g. that [this](https://github.com/Unstructured-IO/unstructured/actions/runs/14112971425/job/39536304012#step:3:1700) is the first reference to `pytest` in the docker build and test logs, after the image build is completed.
- Reviewer should ensure docker image is still being tested in CI and is passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 ## 0.17.5
 
 ### Enhancements
-- **Remove test and dev dependencies from docker image** This reduces the docker image size slightly and reduces potential security vulnerabilities.
+- **Remove test and dev dependencies from docker image.** This reduces the docker image size slightly and reduces potential security vulnerabilities.
 
 ### Features
 
 ### Fixes
+- **Removed out of date ubuntu Dockerfile.** The Dockerfile was out of date and non-functional.
 
 ## 0.17.4
 


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.17.5
+
+### Enhancements
+- **Remove test and dev dependencies from docker image** This reduces the docker image size slightly and reduces potential security vulnerabilities.
```
badGarnet, line 4: also removed ubuntu-22 (which is out of date anyway but still good to document it)

badGarnet (approved):

CHANGELOG.md
qued: Agreed, done.

PR #3974, opened 2025-04-02
Update pdfminer_utils.py

Nathan-GoSupply:
Fix for 'PSSyntaxError' import error:
"cannot import name 'PSSyntaxError' from 'pdfminer.pdfparser'"

Latest pdfminer-six doesn't import PSSyntaxError into `pdfminer.pdfparser` anymore. It must now be directly imported from its source (`pdfminer.psexceptions`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,30 +1,3 @@
-## 0.17.7
-
-### Enhancements
-
-### Features
-
-### Fixes
-Fix for 'PSSyntaxError' import error:
-"cannot import name 'PSSyntaxError' from 'pdfminer.pdfparser'"
-
-Latest pdfminer-six doesn't import PSSyntaxError into pdfminer.pdfparser anymore. It must now be directly imported from its source (pdfminer.psexceptions)
-
-This change will also work on the older version. In the older version, `pdfminer.pdfparser` imports PSSyntaxError from  `pdfminer.psexceptions`.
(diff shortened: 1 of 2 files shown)


cragwolfe (commented):
Thanks for the patch. However, I think the version of pdfminer should be checked to determine whether to import the older or newer version for backwards compat.

Nathan-GoSupply:
This change will also work on the older version. In the older version, `pdfminer.pdfparser` imports PSSyntaxError from  `pdfminer.psexceptions`.

However they have since removed the PSSyntaxError import from `pdfminer.pdfparser`.

Therefore, for the new pdfminer version we must change to directly import from `pdfminer.psexceptions`.
So instead of
`pdfminer_utils.py` -> `pdfminer.pdfparser` ->`pdfminer.psexceptions`

We can do
`pdfminer_utils.py` -> `pdfminer.psexceptions`

PSSyntaxError is defined in `pdfminer.psexceptions` in both the old and new versions of pdfminer, so we will still get backward compatibility.

[Here is the commit](https://github.com/pdfminer/pdfminer.six/commit/b9b75ff85877b7cd373539c79014cbde39508969#diff-a0da8fc41f6e21cd3ab62f914df6ce72d433d3222168af458993d911a5dac37c) for the change on pdfminer.

cragwolfe (approved):

PR #3978, opened 2025-04-03
Fix sort_page_element. ensures that sorting is stable and not random.

pprados:
The sort_page_element() use the element id to sort the elements.
Two executions of the same code, on the same file, produce different results. The order of the elements is random.
This makes it impossible to write stable unit tests, for example, or to obtain reproducible results.

### DIFF
--- a/test_unstructured/partition/pdf_image/test_pdf.py
+++ b/test_unstructured/partition/pdf_image/test_pdf.py
@@ -823,8 +823,8 @@ def test_partition_categorization_backup():
             example_doc_path("pdf/layout-parser-paper-fast.pdf"),
             strategy=PartitionStrategy.HI_RES,
         )
-        # Should have changed the element class from Text to Title
-        assert isinstance(elements[0], Title)
+        # Should NOT have changed the element class from Text to Title
+        assert isinstance(elements[0], Text)
         assert elements[0].text == text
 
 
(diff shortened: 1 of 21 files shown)


badGarnet:
Hi @pprados could you please write up a unit test that:
- reproduces the issue you discovered without the fix
- test passes with the fix
Thanks!

test_unstructured/partition/pdf_image/test_pdf.py
```diff
@@ -823,8 +823,8 @@ def test_partition_categorization_backup():
             example_doc_path("pdf/layout-parser-paper-fast.pdf"),
             strategy=PartitionStrategy.HI_RES,
         )
-        # Should NOT have changed the element class from Text to Title
-        assert isinstance(elements[0], Text)
+        # Should have changed the element class from Text to Title
+        assert isinstance(elements[0], Title)
```
badGarnet, line 827: why is this change necessary? This PR https://github.com/Unstructured-IO/unstructured/pull/3975 intentionally changed the behavior so the class doesn't change to Title (to avoid clutter in the layout).

pprados:
done.
You can remove the fix in `sorting.py` to see the bug.

badGarnet:
> done. You can remove the fix in `sorting.py` to see the bug.

approved. Thanks for the contribution!

PR #3984, opened 2025-04-09
Feature/remove unnessary re for table ele in pdf

JIAQIA:
RE_MULTISPACE_INCLUDING_NEWLINES is applied to all elements of the Text category after partitioning PDF files. The relevant code is shown below:

```python
out_elements = []
for el in elements:
    if isinstance(el, PageBreak) and not include_page_breaks:
        continue

    if isinstance(el, Image):
        out_elements.append(cast(Element, el))
    # NOTE(crag): this is probably always a Text object, but check for the sake of typing
    elif isinstance(el, Text):
        el.text = re.sub(
            RE_MULTISPACE_INCLUDING_NEWLINES,
            " ",
            el.text or "",
        ).strip()
        if el.text or isinstance(el, PageBreak):
            out_elements.append(cast(Element, el))
```

Newlines will not be removed from Table or TableChunk elements now.


qued:
As written, when an element is an instance of `Table` or `TableChunk` it won't be added to `out_elements` at all, which doesn't seem like desired behavior.

PR #3995, opened 2025-05-05
fix: properly handle the case when an element's text is None

badGarnet:
Some elements, like `Image`, can have `None` as its `text` attribute's value. In that case current chunking logic fails because it expects the field to always have a length or can be split. The fix is to update the logic as `element.text or ""` for checking length and add flow control to early exit to avoid calling split on `None`.


ryannikolaidis (approved):
yes please!

PR #3998, opened 2025-05-08
fix: Fix for Pillow error when extracting PNG images

awalker4:
When I tried to partition a PNG file and extract images, I got an error from Pillow:

```
WARNING  unstructured:pdf_image_utils.py:230 Image Extraction Error: Skipping the failed image
  File "/Users/austin/.pyenv/versions/unstructured/lib/python3.10/site-packages/PIL/JpegImagePlugin.py", line 666, in _save
    rawmode = RAWMODE[im.mode]
KeyError: 'RGBA'
```

The issue is that a PNG has an additional layer that cannot be saved off in jpeg format. We can fix this with a quick conversion. I added a png test case that is now passing with this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.17.7-dev0
+
+### Enhancements
+
+### Features
+
+### Fixes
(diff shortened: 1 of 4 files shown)


badGarnet (approved):

ajjimeno (commented):
LGTM

PR #4004, opened 2025-05-21
Config for VoyageAI's v3.5 embedding models

fzowl:
Setting the default batch size for VoyageAI's new v3.5 models


SudSampath (changes requested):
I believe this should be in unstructured-ingest but

fzowl:
@SudSampath Hm, not sure. I mean, the voyageai integration code is in this repo, it is just a slight modification.

PR #4012, opened 2025-06-05
Remove IDs from HTML code

plutasnyy:
In this pull request parent-child relationship for elements generated with v2 parser is based on actual element IDs instead of IDs baked somewhere in the HTML script. 
With some extra bug fixing it allowed for significantly simplifying json -> HTML script


unstructured/documents/mappings.py
```diff
@@ -134,9 +136,7 @@ def get_ontology_to_unstructured_type_mapping() -> dict[str, Element]:
     for tag in element_type().allowed_tags
 }
 CSS_CLASS_TO_ELEMENT_TYPE_MAP: Dict[str, Type[ontology.OntologyElement]] = {
-    element_type().css_class_name: element_type
-    for element_type in ALL_ONTOLOGY_ELEMENT_TYPES
-    for tag in element_type().allowed_tags
```
badGarnet, line 139: how was this line of code working before?

plutasnyy: This was copy-paste form the mapping above (which is more specific), here we just did more loops, but the 'tag' is not used anywhere. After the change, the mapping is the same, just unnecessary loops are removed:
```python
CSS_CLASS_TO_ELEMENT_TYPE_MAP_BEFORE: Dict[str, Type[ontology.OntologyElement]] = {
    element_type().css_class_name: element_type
    for element_type in ALL_ONTOLOGY_ELEMENT_TYPES
    for tag in element_type().allowed_tags
}

CSS_CLASS_TO_ELEMENT_TYPE_MAP_AFTER: Dict[str, Type[ontology.OntologyElement]] = {
    element_type().css_class_name: element_type for element_type in ALL_ONTOLOGY_ELEMENT_TYPES
}
print(CSS_CLASS_TO_ELEMENT_TYPE_MAP_BEFORE == CSS_CLASS_TO_ELEMENT_TYPE_MAP_AFTER)
# Output: True
```

This is because some CSS classes can have multiple HTML tags assigned eg.
```python
class Heading(OntologyElement):
    description: str = Field("Section headings (levels 1-6)", frozen=True)
    elementType: ElementTypeEnum = Field(ElementTypeEnum.text, frozen=True)
    allowed_tags: List[str] = Field(["h1", "h2", "h3", "h4", "h5", "h6"], frozen=True)
```
In this case, we were creating a 6 times entry: `"Heading": OntologyElement.Heading`

plutasnyy:
Removed `detection_origin` from unit tests for 2 reasons:
- This field is intentionally excluded from `__eq__` anyway 
- It doesn't work? Or work in a different way than I would expect. This field is excluded from `to_dict()`, but this code:
```
from unstructured.documents.elements import NarrativeText, ElementMetadata
metadata = ElementMetadata(
    text_as_html='<p class="NarrativeText">DEALER ONLY</p>',
)
metadata.detection_origin = "vlm_partitioner"
element = NarrativeText(
        text="DEALER ONLY",
        detection_origin="Test",
        metadata=metadata,
    )
print(element.metadata.detection_origin)
```
Print `None` as [this line](https://github.com/Unstructured-IO/unstructured/blob/37d2f021a39b1b5682a5306349e2d36a6a1076a3/unstructured/documents/elements.py#L324) guards assigning debug field 

For some reason, it was assigned in the unit test (even though I got None = None locally), possibly due to some version mismatch. 

I think it needs another debugging session and is not a blocker for this ticket (this change didn't touch detection origin)

plutasnyy:
I have verified if the script produces the same outputs, and the only difference is in the order of classes in CSS + `br` is generated with a space (which I assume is fine, as it is rendered in the same way)
image

unstructured/partition/html/transformations.py
```diff
@@ -235,34 +237,39 @@ def unstructured_elements_to_ontology(
...
         )
+        element_id = element.id
+        parent_id = element.metadata.parent_id
+
+        if parent_id is None:
+            # Make sure that no element is lost
+            parent_id = root_element_id
+
+        if parent_id not in id_to_element_mapping:
+            raise ValueError(f"Parent element {parent_id} not found in id_to_element_mapping")
```
badGarnet: maybe add a unit test for the new exception?

plutasnyy: I tried, but I noticed that they make no sense. parent_id is always set before (if isn't it is generated) and ontology always returns some element so I just removed raising exception part

badGarnet:
hold on while I bump our base image to fix the dockerfile cve

PR #4016, opened 2025-06-11
fix [NEX-28]: file_type is None for result_file_type in chunker partition json

yuming-long:
### Summary
`'NoneType' object has no attribute 'partitioner_shortname'` due to `result_file_type = self._disambiguate_json_file_type` could return None for file type

### Bug Invest Notes
to fix the error `'NoneType' object has no attribute 'partitioner_shortname'` in chucker i found the ES logs under same org:
Screenshot 2025-06-11 at 7 02 01 PM
stack trace from the `utic-api` ES log doc:
Screenshot 2025-06-11 at 7 09 38 PM
and note that the content-type got changed into `"content_type": "application/x-ole-storage",` in unstructured code (doc type for partitioner ?) even though we are making an API call with the correct content type `"GeneralFormParams(coordinates=False, encoding=None, content_type='application/json'`
Screenshot 2025-06-11 at 7 24 09 PM
so maybe
Screenshot 2025-06-12 at 11 23 34 AM
we are passing the msg file here not json?


yuming-long:
this is not the root cause of the bug - will close this one for now

yuming-long:
~~reopen to add some None type protection~~ Moved to another one

PR #4022, opened 2025-06-13
fix [NEX-28]: file_type is None for result_file_type in chunker partition json

yuming-long:
### Summary
`'NoneType' object has no attribute 'partitioner_shortname'` due to `result_file_type = self._disambiguate_json_file_type` could return None for file type


yuming-long:
@claude pr review

cragwolfe (approved):

PR #4024, opened 2025-06-13
Parallelize tests

jiajun-unstructured:


test_unstructured/partition/pdf_image/test_ocr.py
```diff
@@ -625,8 +625,16 @@ def mock_page(mock_ocr_layout, mock_layout):
 def test_supplement_layout_with_ocr(mocker, mock_page):
     from unstructured.partition.pdf_image.ocr import OCRAgent
 
+    def mock_get_instance(ocr_agent_module, language):
```
PastelStorm: This mock is created 3 times in different tests, so it should be a fixture in the conftest I think?

jiajun-unstructured: thanks for the catch. done

PR #4034, opened 2025-06-26
fix: xml processing not escaped

jiajun-unstructured:
`<?xml version="1.0"?>` does not get escaped when converting to html, in a code block like this in the markdown file
````
<?xml version="1.0"?>
<sparql xmlns="http://www.w3.org/2005/sparql-results#">
  <head></head>
  <boolean>true</boolean>
</sparql>
````
which causes the parser to throw error like 

> AttributeError: 'lxml.etree._ProcessingInstruction' object has no attribute 'is_phrasing'.

This PR processes the original md file and add indentation to `<?xml version="1.0"?>` to force the xml code to be escaped when being converted to html

https://github.com/Unstructured-IO/unstructured/issues/3935

### CODE REVIEW

File: unstructured/partition/md.py [hunk 1]
```diff
@@ -21,6 +22,24 @@ def optional_decode(contents: str | bytes) -> str:
 DETECTION_ORIGIN: str = "md"
 
 
+def _preprocess_markdown_code_blocks(text: str) -> str:
+    """Pre-process markdown to ensure code blocks with XML are properly formatted.
+
+    The markdown library can fail to properly escape XML processing instructions
+    in code blocks if they're not formatted correctly. e.g. <?xml version="1.0"?>
+    """
+    code_block_pattern = r"```\s*\n(<\?xml[^>]*\?>.*?)\n```"
```
qued: I found a case that still fails under this pattern. Here's an example:
```python
  <?xml version="1.0"?>
  <sparql xmlns="http://www.w3.org/2005/sparql-results#">
  <head></head>
  <boolean>true</boolean>
</sparql>
```
Any time there are 1-3 spaces indenting the `<?xml?>` tag, I still get the error. I think the pattern can be modified to deal with those edge cases though.

qued: Also, I'm able to get the same error using other processing instructions than xml, like `<?php ...>`, `<?ruby ...>`, or even `<?asdf ...>`.

qued (approved):
LGTM! The CHANGELOG needs to be fixed since it changed upstream, but once that conflict is resolved this should be good to merge.

PR #4037, opened 2025-06-30
fix: update md to reads umlauts on non-utf-8 files

Klaijan:
This PR updates the `partition_md` to reads files with non-utf8 encodings without fail.

Closes issue https://github.com/Unstructured-IO/unstructured-api/issues/489


qued (approved):
LGTM! Don't forget to bump the version before merging.

PR #4043, opened 2025-07-03
update ingest version

jiajun-unstructured:


qued:
This was an experiment to see if a certain issue was related to the ingest version, and updating the ingest version didn't help.

PR #4046, opened 2025-07-07
fix(partition, csv): increase csv field limit

ds-filipknefel:
Increase the csv field limit to support partitioning of files with large data in fields.

### DIFF
--- a/unstructured/partition/csv.py
+++ b/unstructured/partition/csv.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import contextlib
 import csv
-import ctypes
 from typing import IO, Any, Iterator
 
 import pandas as pd
@@ -15,6 +14,7 @@ from unstructured.partition.common.metadata import apply_metadata, get_last_modi
 from unstructured.utils import is_temp_file_path, lazyproperty
 
 DETECTION_ORIGIN: str = "csv"
+CSV_FIELD_LIMIT = 10 * 1048576  # 10MiB
 
 
 @apply_metadata(FileType.CSV)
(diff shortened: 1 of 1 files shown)


unstructured/partition/csv.py
```diff
@@ -54,6 +55,8 @@ def partition_csv(
         infer_table_structure=infer_table_structure,
     )
 
+    # NOTE: Set maximum possible field size limit (LONG_MAX)
+    csv.field_size_limit(int(ctypes.c_ulong(-1).value // 2))
```
mateuszkuprowski: As cpp dev I never thought python would make something more complicated than C++, but here it is, obtaining LONG_MAX value is so cryptic 🥇 
On the serious note: I don't think it's safe to allow fields that big. I don't think it's reasonable to allow for fields that can eat up to 8 exabytes of RAM just in case anybody actually has very large fields that can crash the system at best or eat all resources in the cluster at worst. Let's maybe allow up to 100MB or 1GB, but let's stay reasonable.

mateuszkuprowski (changes requested):
Please check my suggestion about using some reasonable limit afterall.

unstructured/partition/csv.py
ds-filipknefel: I can get behind that, it's rather an edge case already that such a large field is present so we could try with a significantly larger but not really a large limit and see whether any further reports show up.

mateuszkuprowski: Looks good to me now!

mateuszkuprowski (approved):
LGTM now!

PR #4050, opened 2025-07-10
feat: keep img tag's class attr

badGarnet:
This change affects partition html.

Previously when there is a table in the html, we clean any tags inside the table of their `class` and `id` attributes. However, sometimes there are images, `img` tags, present in a table and its `class` attribute identifies some important information about the image. This change preserves the `class` attribute for `img` tags inside a table. This change is reflected in a table element's `metadata.text_as_html` attribute.


badGarnet:
@claude review

PR #4052, opened 2025-07-11
feat: added new exception type to epub conversions

mateuszkuprowski:
Added UnprocessableEpubError to better handle the case when incoming epub file is actually damanged which makes pandoc lib crash with exit code 64.


unstructured/errors.py
```diff
@@ -9,3 +9,9 @@ def __init__(self, document_pages: int, pdf_hi_res_max_pages: int):
             f"pages={document_pages}, maximum={pdf_hi_res_max_pages}."
         )
         super().__init__(self.message)
+
+
+class UnprocessableEpubError(Exception):
+    """Error raised when the epub file is not valid."""
+
+    pass
```
mpolomdeepsense: Could we maybe have it subclass from a more general `UnprocessableFileError`, just to have something more general in case we want to have multiple Unprocessable errors for different file types.

mateuszkuprowski: Absolutely, good call! Modified.

mpolomdeepsense (approved):
LGTM

mateuszkuprowski:
Dockerfile CI is not related to the cahnge I made hence I'm merging this.

PR #4060, opened 2025-07-15
chore: switch to charset normalizer

qued:
Closes [SPI-44](https://linear.app/unstructured/issue/SPI-44/spike-replace-chardet-with-charset-normalizer-if-possible).

Removes `chardet` as a dependency, standardizing on `charset-normalizer`.

This involved:
- Changing `chardet` to `charset-normalizer` in our base dependency file
- Updating the code (in only one place) where `chardet` was used
- pip-compiling to update our published dependency tree
- Updating one test... `charset-normalizer` misdiagnosed the encoding of a file used as a test fixture. My guess is that the ~10 characters in the file were not enough for `charset-normalizer` to do a proper inference, so I re-encoded another slightly longer file that's also used for encoding testing, and it got that one.
- Updating an ingest test fixture.
- Updating the ingest test fixture update workflow to also update the expected markdown results (this was a task I missed when adding the markdown ingest tests)


unstructured/file_utils/encoding.py
```diff
@@ -70,11 +70,11 @@ def detect_file_encoding(
     else:
         raise FileNotFoundError("No filename nor file were specified")
 
-    result = chardet.detect(byte_data)
+    result = detect(byte_data)
     encoding = result["encoding"]
     confidence = result["confidence"]
 
-    if encoding is None or confidence < ENCODE_REC_THRESHOLD:
+    if encoding is None or (confidence is not None and confidence < ENCODE_REC_THRESHOLD):
```
ahmetmeleq:
```suggestion
    if encoding is None or confidence is None or confidence < ENCODE_REC_THRESHOLD:
```

ahmetmeleq: We might want to fallback when confidence is `None` too

example-docs/umlauts-non-utf8.md
```diff
@@ -1 +1,5 @@
-Umlauts: ����
\ No newline at end of file
+## k�nnen
+
+k�nnen
```
ahmetmeleq, line 3: For curiosity reasons, what's the story with these being added

ahmetmeleq (approved):
Surprisingly simple!

example-docs/umlauts-non-utf8.md
qued: In the unit tests on the old file, `charset_normalizer` misdiagnosed it (`chardet` had previously gotten it right). So I added some stuff by re-encoding `umlauts-utf8.md` to give it more content on which to make an inference.

qued: As far as the content of `umlauts-utf8.md`, I think it was suggested by a user in a github issue.

awalker4 (approved):
LGTM!

PR #4062, opened 2025-07-16
fix: add empty string case for language metadata

shreyanid:
Add an empty string edge case for when the element text field is None or not a string.

most of the diff is `make tidy`

### DIFF
--- a/unstructured/partition/common/lang.py
+++ b/unstructured/partition/common/lang.py
@@ -487,7 +487,7 @@ def apply_lang_metadata(
     if not isinstance(elements, list):
         elements = list(elements)
 
-    full_text = " ".join(e.text or "" for e in elements if hasattr(e, "text"))
+    full_text = " ".join(str(e.text) for e in elements if hasattr(e, "text") and e.text)
     detected_languages = detect_languages(text=full_text, languages=languages)
     if (
         detected_languages is not None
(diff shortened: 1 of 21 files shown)


yuming-long:
@claude looks good right

unstructured/partition/common/lang.py
```diff
@@ -501,7 +501,7 @@ def apply_lang_metadata(
     else:
         for e in elements:
             if hasattr(e, "text"):
-                e.metadata.languages = detect_languages(e.text)
+                e.metadata.languages = detect_languages(e.text or "")
```
yuming-long: do we need to call ` detect_languages(e.text or "")`  -> should be no lanugaga for empty str imo

unstructured/partition/common/lang.py
```diff
@@ -487,7 +487,7 @@ def apply_lang_metadata(
     if not isinstance(elements, list):
         elements = list(elements)
 
-    full_text = " ".join(e.text for e in elements if hasattr(e, "text"))
+    full_text = " ".join(e.text or "" for e in elements if hasattr(e, "text"))
```
yuming-long: full_text = " ".join(e.text  for e in elements if (hasattr(e, "text") and e.text))
 maybe this is easier to read?

unstructured/partition/common/lang.py
shreyanid: if e.text is not a string (ex. a numerical value) or a None value, we'll use an empty string to safely trigger the logic that already exists in the helper function to set the language field accordingly. I'll add some string casting

unstructured/partition/common/lang.py
shreyanid: Other falsy values like 0 could get caught into the e.text condition, but for the purpose of language detection that doesn't really matter :) I can do something like this!

PR #4064, opened 2025-07-16
feat: keep input tag's class attr in table

badGarnet:
This change affects partition html.

Previously when there is a table in the html, we clean any tags inside the table of their class and id attributes except for the class attribute for `img` tags. This change also preserves the class attribute for `input` tags inside a table. This change is reflected in a table element's metadata.text_as_html attribute.


badGarnet:
@claude review

PR #4065, opened 2025-07-17
build: Update CodeQL GHA to v3

qued:
We were using CodeQL v2, which has been [deprecated since January](https://github.blog/changelog/2025-01-10-code-scanning-codeql-action-v2-is-now-deprecated/).


ahmetmeleq (approved):
We shouldn't need lib version updates to register CI file updates. But approving in case there's another intention / if I'm wrong

PR #4071, opened 2025-07-25
fix: replace UnicodeDecodeError to prevent large payload logging

CyMule:
Replace UnicodeDecodeError with UnprocessableEntityError in encoding detection to avoid logging entire file contents. UnicodeDecodeError.object automatically stores complete input data, causing memory issues with large files in logging and error reporting systems.

### DIFF
--- a/unstructured/file_utils/encoding.py
+++ b/unstructured/file_utils/encoding.py
@@ -2,8 +2,8 @@ from typing import IO, Optional, Tuple, Union
 
 from charset_normalizer import detect
 
-from unstructured.partition.common.common import convert_to_bytes
 from unstructured.errors import UnprocessableEntityError
+from unstructured.partition.common.common import convert_to_bytes
 
 ENCODE_REC_THRESHOLD = 0.8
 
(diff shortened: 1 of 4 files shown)


unstructured/file_utils/encoding.py
```diff
@@ -88,17 +89,26 @@ def detect_file_encoding(
...
+        # UnicodeDecodeError automatically stores the complete input data in its
+        # 'object' attribute, which can cause issues with large files in logging
+        # and error reporting systems.
+        try:
+            file_text = byte_data.decode(encoding)
+        except (UnicodeDecodeError, UnicodeError):
+            raise UnprocessableEntityError(
+                f"File encoding detection failed: detected '{encoding}' but decode failed. "
+                f"File may be corrupted or in an unsupported format."
+            ) from None
```
yuming-long, line 111: why from None

yuming-long:
@claude review this?

mateuszkuprowski (approved):
LGTM! 
Love what you've done here!

unstructured/file_utils/encoding.py
CyMule: APM would end up capturing the original exception which would contain the full bytes we tried to decode. The `None` prevents the exception chaining and importantly prevents the bytes from being passed to APM

yuming-long: gotcha thanks!

yuming-long (approved):
🎉

PR #4072, opened 2025-08-08
email date format flexibility

potter-potter:
we are seeing some .eml files come through the VLM partitioner. Which then downgrades to hi-res i believe.

For some reason they have a date format that is not standard email format. But it is still legitimate.

This uses a more robust date package to parse the date. This package is already installed.


potter-potter:
@claude please review

unstructured/partition/email.py
```diff
@@ -279,7 +281,11 @@ def _sent_date(self) -> str | None:
         date_str = self.msg.get("Date")
         if not date_str:
             return None
-        sent_date = email.utils.parsedate_to_datetime(date_str)
+        try:
+            sent_date = parser.parse(date_str)
+        except (parser.ParserError, TypeError, ValueError):
+            return None
```
badGarnet, line 287: maybe add a test for this route?

badGarnet (approved):

PR #4083, opened 2025-08-22
fix: None text attribute when normalizing Picture to Image element

ishahroz:
This PR fixes this [issue](https://github.com/Unstructured-IO/unstructured/issues/4084). Issue description is below: 

When using yolox as the Hi-res model for loading outputs / annotations, it annotates with bbox dimensions but missing text for complex images, resulting in text being set to None. But later on printing or accessing the Image element (__str__ method), it should be returning string instead of None.

### DIFF
--- a/unstructured/partition/common/common.py
+++ b/unstructured/partition/common/common.py
@@ -50,7 +50,7 @@ def normalize_layout_element(
     else:
         layout_dict = layout_element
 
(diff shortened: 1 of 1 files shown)


ishahroz:
This PR should also fix this issue: https://github.com/langchain-ai/langchain-community/issues/310

PR #4085, opened 2025-08-25
manual fix for open CVEs

luke-kucing:


cragwolfe (changes requested):
still a nit on changelog, looks like test-out.sh was duplicated

tabossert:
Why are we including a Mac only script in here?

cragwolfe (approved):

PR #4167, opened 2026-01-07
Eliminate cleaners/core import time bottleneck

aseembits93:
Previously the code was going through a million unicode characters and for each of them checking if the character was unicode punctuation or not. This was heavy and took about 350ms during import time. 

After processing the number of punctuation was only about ~750 so we just hard coded the unicode punctuation characters in source code itself. This way we eliminate the runtime calculation.

To ensure correctness, we added a test.
Before
Screenshot 2026-01-06 at 19 39 43
After
Screenshot 2026-01-06 at 19 38 41


qued:
I don't think the gains here (1/3 second at import time) are worth putting data in a code file.

misrasaurabh1:
@qued ok, there are a couple of other approaches - storing the list in a separate data file or reducing the sys.maxunicode range to a group of smaller ranges and checking for punctuation membership within that. Let me know what approach works better

qued:
@misrasaurabh1 My first instinct is to use a data file, but does the file IO to load it erode all the gains from not checking all the unicode characters?

misrasaurabh1:
we can experiment with the other approaches and check

PR #4169, opened 2026-01-07
Fix: partially filled inferred layout mark as not extracted

badGarnet:
This PR fixes an issue where elements with partially filled extracted text is marked as extracted.

## bug scenario
This PR adds a new unit test to show case a scenario:
- during merging inferred and extracted layout the function `aggregate_embedded_text_by_block` aggregates extracted text that falls into an inferred element; and if all text has the flag `is_extracted` being `"true"` the inferred element is marked as such as well
- however, there can be a case where the extracted text only partially fills the inferred element. There might be text in the inferred element region that are not present as extracted text (i.e., require OCR). But the current logic would still mark this inferred element as `is_extracted = "true"`

## Fix
The fix adds another check in the function `aggregate_embedded_text_by_block`: if the intersect over union of between the source regions and target region cross a given threshold. This new check correctly identifies the case in the unit test that the inferred element should be be marked a `is_extracted = "false"`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Optimize `_PartitionerLoader._load_partitioner` (codeflash)
 - Optimize `detect_languages` (codeflash)
 - Optimize `contains_verb` (codeflash)
+- Optimize `get_bbox_thickness` (codeflash)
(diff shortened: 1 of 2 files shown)


badGarnet:
@claude review

PR #4183, opened 2026-01-09
Test ci caching

badGarnet:


badGarnet:
purge pip cache worked as well and it freed a much larger amount of space

PR #4190, opened 2026-01-14
feat: consider rotated text as low fidelityfeat: consider rotated text

badGarnet:
This PR updates the function `is_text_embedded`:
- now considers both if chars are invisible or rotated (as a result includes some refactoring of variable names)
- rotated text elements can have wrong character order compared to natural reading order -> if feed into downstream applications like embedding text the element loses its semantic meaning
- as a result this update flags texts with too many rotated characters as only partially embedded: its source is technically embedded but it may need post processing to be useful


badGarnet:
@claude review

cragwolfe (approved):

PR #4192, opened 2026-01-14
Luke/update dockerfile

luke-kucing:


Dockerfile
```diff
@@ -11,20 +11,53 @@ COPY ./requirements requirements/
 COPY unstructured unstructured
 COPY test_unstructured test_unstructured
 COPY example-docs example-docs
+COPY ./docker/packages/*.apk /tmp/packages/
 
-RUN chown -R notebook-user:notebook-user /app && \
```
qued, line 15: I don't see where we're doing this. The tests are passing so maybe we don't need it? Have we confirmed the /app folder ends up with the right ownership?

Dockerfile
```diff
@@ -11,20 +11,53 @@ COPY ./requirements requirements/
...
+ARG NB_UID=1000
+ARG NB_USER=notebook-user
+RUN addgroup --gid ${NB_UID} ${NB_USER} && \
+    adduser --disabled-password --gecos "" --uid ${NB_UID} -G ${NB_USER} ${NB_USER}
+
+ENV USER=${NB_USER}
+ENV HOME=/home/${NB_USER}
+COPY --chown=${NB_USER} scripts/initialize-libreoffice.sh ${HOME}/initialize-libreoffice.sh
+
 RUN rm -rf /usr/lib/python3.10 && \
```
qued, line 49:
```suggestion
# Remove unused Python versions
RUN rm -rf /usr/lib/python3.10 && \
```

PR #4195, opened 2026-01-15
fix: NameError: LayoutElements not defined in paddle_ocr.py

### LINKED ISSUE #3931: bug/NameError: name 'LayoutElements' is not defined
Hello unstructured team, 

I ran into a bug where the module: unstructured/partition/utils/ocr_models/paddle_ocr.py

Would return the bug described above.

Here was my fix:

in the document where I call partition_pdf()

I set the namespace like so:

```import unstructured.partition.utils.ocr_models.paddle_ocr as paddle_ocr_modue
from unstructure_inference.inference.layoutelement import LayoutElements

paddle_ocr_module.LayoutElements = LayoutElements```

I am not sure if this is a sustainable solution, and I am not sure why the module itself is not able to find the definition of LayoutElements itself without me manually setting the namespace.

I made sure to have TYPING dependency, and at some point I even removed the condition, neither of which worked. 

Please let me know if I missed something during installation that is causing this. 

Thanks!

mohansinghi:
Fixes #3931

Adds runtime import of `LayoutElements` inside `get_layout_elements_from_image()` method to resolve `NameError` when the method is called. The import was previously only available under `TYPE_CHECKING`, but the class is instantiated at runtime.

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=158349177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
-## 0.18.30-dev2
+## 0.18.30-dev5
 
 ### Enhancement
 - `is_text_embedded` now considers rotated text as low fidelity and and elements with no trivial amount of it are considered not embedded
 - Replace `pdf2image` with PyPDFium2 for PDF rendering
+- Optimize `_get_optimal_value_for_bbox` (codeflash)
+- Optimize `_DocxPartitioner._style_based_element_type` (codeflash)
 
 ### Fixes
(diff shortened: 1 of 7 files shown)


0xsid0703:
@badGarnet Could you please check my first pr?

0xsid0703:
@badGarnet Could you please check my first pr?

0xsid0703:
@badGarnet Could you please check my first pr?

PR #4197, opened 2026-01-15
feat: chunking by character and title now isolates tables

badGarnet:
This PR introduces a new chunking option `isolate_tables`.
- default to `False` and keeps current chunking behavior
- when `True` tables are not chunked with any other element when using `chunk_elements` and `chunk_by_title`


badGarnet:
@claude review

badGarnet:
@claude review

PR #4201, opened 2026-01-21
perf: speed up standardize_quotes with str.translate()

KRRT7:
## Summary

Speed up `standardize_quotes` in `unstructured/metrics/text_extraction.py` by replacing the loop-based character replacement with a single `str.translate()` call using a pre-computed translation table.

**Before:** Two loops iterate through ~40 quote types, calling `unicode_to_char()` on each to parse the Unicode escape string, then checking membership and calling `str.replace()` — O(n × m) with repeated string allocations.

**After:** A module-level translation table maps all Unicode quote codepoints directly to their ASCII equivalents. The function body becomes a single `text.translate(_TRANSLATION_TABLE)` call — O(n) single pass at C level.

All 47 existing tests pass.

## Bug fix

The original `standardize_quotes()` has a duplicate-key bug. The `double_quotes` and `single_quotes` dicts use literal quote characters as keys — but `"`, `"`, and `"` all encode as byte `0x22` (ASCII double quote), and `'`, `'`, `'` all encode as `0x27`. Python silently deduplicates them, so only the last value survives in each case. This means `U+201C` (left double smart quote) and `U+2018` (left single smart quote) were **never** being normalized — on `main`, not just in this PR.

This PR restructures the mappings as tuples of `\uXXXX` escape sequences, which eliminates the duplicate-key problem entirely. Left smart quotes are now correctly normalized.

### Azure Standard_D8s_v5 — 8 vCPU Intel Xeon Platinum 8473C, 32 GiB RAM

| | Min | Median | Mean | OPS | Rounds |
|:---|---:|---:|---:|---:|---:|
| `origin/main` (base) | 161.25μs | 199.57μs | 200.72μs | 4.98 Kops/s | 5,461 |
| `codeflash/op` (head) | 99.17μs | 126.40μs | 127.86μs | 7.82 Kops/s | 10,581 |
| **Speedup** | **🟢 1.63x** | **🟢 1.58x** | **🟢 1.57x** | **🟢 1.57x** | |

| Function | base (μs) | head (μs) | Improvement | Speedup |
|:---|---:|---:|:---|---:|
| `standardize_quotes` | 128.60μs | 53.86μs | `██████░░░░` +58% | 🟢 2.39x |
| `run_standardize_quotes` | - | 172.99μs | - | - |

---
*Generated by codeflash optimization agent*

<details>
<summary>Reproduce with codeflash</summary>

```bash
uv add codeflash
codeflash compare main codeflash/optimize-standardize_quotes-mklcp188
```

</details>

### DIFF
--- a/unstructured/__version__.py
+++ b/unstructured/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.22.12"  # pragma: no cover
+__version__ = "0.22.13"  # pragma: no cover


KRRT7:
Closing to reopen from upstream branch (trying to avoid slack bot issue).

PR #4206, opened 2026-01-22
fix: filter coordinates kwargs to prevent TypeError in hi_res PDF processing

### LINKED ISSUE #4126: Bug: coordinates=True kwarg causes TypeError in hi_res PDF processing
Bug: coordinates=True kwarg causes TypeError in hi_res PDF processing

  Description

  Passing coordinates=True as a keyword argument to partition() causes a TypeError when processing PDFs with the hi_res strategy. The error occurs because the kwarg is forwarded to add_element_metadata(), which expects coordinates to be actual coordinate data (tuple[tuple[float, float], ...]), not a boolean.

  Error Message

  TypeError: unstructured.partition.common.common.add_element_metadata() got multiple values for keyword argument 'coordinates'

  Environment

  - unstructured: 0.18.21 (also confirmed in 0.18.22)
  - unstructured-inference: 1.1.2
  - Python: 3.10.12
  - Platform: Linux (Docker with CUDA 12.8)

  Reproduction

  # This works fine
  elements = partition(
      filename="scanned_document.pdf",
      strategy="hi_res",
      hi_res_model_name="yolox"
  )

  # This raises TypeError
  elements = partition(
      filename="scanned_document.pdf",
      strategy="hi_res",
      hi_res_model_name="yolox",
      coordinates=True  # <-- causes the error
  )

  Analysis

  1. partition() accepts **kwargs and forwards them through the call chain
  2. When using hi_res strategy on PDFs, the inference pipeline already computes and attaches coordinates to elements
  3. The coordinates=True boolean gets passed down to add_element_metadata() via kwargs
  4. add_element_metadata() has a named parameter coordinates: Optional[tuple[tuple[float, float], ...]] expecting coordinate data
  5. The conflict between the kwarg and the internally-computed coordinates causes the "multiple values" error

  Expected Behavior

  Either:
  1. The coordinates boolean kwarg should be filtered out before reaching add_element_metadata(), or
  2. Documentation should clarify that coordinates=True is not a valid parameter (since hi_res already includes coordinates by default)

  Workaround

  Don't pass coordinates=True to partition(). Coordinates are included automatically when using hi_res strategy:

  elements = partition(
      filename="scanned_document.pdf",
      strategy="hi_res",
      hi_res_model_name="yolox"
  )

  # Coordinates are already present
  print(elements[0].metadata.coordinates)  # Works

MkDev11:
## Summary

This PR fixes a bug where passing `coordinates=True` to `partition()` causes a TypeError when processing PDFs with the hi_res strategy.

## Problem

When users call `partition()` with `coordinates=True`, the boolean value flows through kwargs and eventually reaches `add_element_metadata()`. However, this function already receives computed coordinate data as an explicit parameter. Python then raises:

```
TypeError: add_element_metadata() got multiple values for keyword argument 'coordinates'
```

This is confusing because users reasonably expect `coordinates=True` to enable coordinate output, not realizing that hi_res strategy already computes and includes coordinates automatically.

## Solution

Filter out `coordinates` and `coordinate_system` from kwargs before passing them to `add_element_metadata()`. This prevents the conflict while preserving the internally-computed coordinate data.

The fix is minimal and targeted - just 3 lines of code that filter the problematic kwargs.

## Changes

- `unstructured/partition/pdf.py`: Added filtering for `coordinates` and `coordinate_system` kwargs
- `CHANGELOG.md`: Added entry for this fix

Fixes #4126

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN apk update && \
     apk add --no-cache font-ubuntu fontconfig && \
     apk upgrade --no-cache py3.12-pip && \
     fc-cache -fv && \
-    [ -e /usr/bin/python3 ] || ln -s /usr/bin/$PYTHON /usr/bin/python3
+    ln -sf /usr/bin/$PYTHON /usr/bin/python3
 
 ARG NB_UID=1000
 ARG NB_USER=notebook-user


mkdev5:
@badGarnet could you please take a look at the PR I mentioned earlier? really appreciate your time and feedback.

badGarnet:
> @badGarnet could you please take a look at the PR I mentioned earlier? really appreciate your time and feedback.

please add a unit test and remember to bump version number

mkdev5:
@badGarnet updated again! could you please have a look at the changes and let me know ther result? appeciate for your time!

mkdev5:
The CI's docker-build step fails because the Dockerfile tries to create a symlink python3 but it already exists (Python 3.12 installation created it).
This is an upstream infrastructure issue - it affects the main Dockerfile, not my PR code. 
@badGarnet can you retry to merge?

Issue #4036, opened 2025-06-30
Future Warning: Passing bytes to `read_excel` is deprecated


EdwinWenink:
Hi, 

When using unstructured in a test, I get the following deprecation warning.

```
[redacted]/site-packages/unstructured/partition/xlsx.py:197: FutureWarning: Passing bytes to 'read_excel' is deprecated and will be removed in a future version. To read from a byte string, wrap it in a `BytesIO` object.
    return pd.read_excel(f, sheet_name=None, header=self.header_row_idx)
```

See: https://github.com/Unstructured-IO/unstructured/blob/c04235c16842e5be504dc7509ad2f12ea6adb43c/unstructured/partition/xlsx.py#L197

Deprecation is logged here: https://pandas.pydata.org/docs/dev/reference/api/pandas.read_excel.html

Fix should be simple: wrap f in an BytesIO object.

jbbqqf:
Hi! Triaging older issues — this `FutureWarning` was fixed about a year ago: `pd.read_excel` now receives an `io.BytesIO` wrapper, exactly as you suggested.

Evidence:
- Current source at [`unstructured/partition/xlsx.py`](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/xlsx.py#L200-L202) reads `return pd.read_excel(io.BytesIO(self._file_bytes), sheet_name=None, header=self.header_row_idx)` — the `BytesIO` wrap you proposed is in place.
- The change shipped in PR #4057 ("fix: properly handle password protected xlsx", merged 2025-07-16), which refactored the `sheets` property to always go through `self._file_bytes` wrapped in `io.BytesIO`.
- A current `unstructured` release no longer emits the `FutureWarning` from `xlsx.py:197`.

If you still see the warning against a current release (`0.18.31`+), would you mind reopening with the version string? Otherwise this looks closeable.

---
*Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the references above were verified manually.*

PR #4213, opened 2026-01-27
Fix FutureWarning: Add test to verify bytes are wrapped in BytesIO for read_excel

Achieve3318:
This PR addresses issue #4036 by ensuring bytes are properly wrapped in BytesIO before passing to `pd.read_excel()`.

**Implementation:**
The code at line 203 in `unstructured/partition/xlsx.py` already wraps bytes in `BytesIO` before calling `pd.read_excel()`, which is the correct solution to avoid the FutureWarning deprecation.

**Changes made:**
1. Added a comment (lines 200-201) explaining why bytes are wrapped in BytesIO
2. Added a test (`test_partition_xlsx_no_future_warning_for_bytes`) to verify no FutureWarning is raised

This ensures the fix is documented and tested, preventing future regressions.

### DIFF
--- a/unstructured/partition/xlsx.py
+++ b/unstructured/partition/xlsx.py
@@ -197,6 +197,8 @@ class _XlsxPartitionerOptions:
         if office_file.is_encrypted():
             raise UnprocessableEntityError("XLSX file is password protected.")
 
+        # Wrap bytes in BytesIO to avoid FutureWarning from pd.read_excel
+        # See: https://github.com/Unstructured-IO/unstructured/issues/4036
         return pd.read_excel(
             io.BytesIO(self._file_bytes), sheet_name=None, header=self.header_row_idx
         )


Achieve3318:
Hi @badGarnet  I'd love to get your review and approval
Thank you for your time

test_unstructured/partition/test_xlsx.py
```diff
@@ -169,6 +169,42 @@ def test_partition_xlsx_from_file_with_header():
...
+    This test ensures that bytes are properly wrapped in BytesIO before being passed to
+    pd.read_excel(), preventing the deprecation warning.
+    """
+    import warnings
+
+    with open("example-docs/stanley-cups.xlsx", "rb") as f:
+        file_bytes = f.read()
+
+    # Create a BytesIO object from bytes to simulate the scenario
+    file_like = io.BytesIO(file_bytes)
```
badGarnet, line 184: this points to a good solution to the issue #4036 : internally in `pd.read_excel` [line](https://github.com/Unstructured-IO/unstructured/blob/c04235c16842e5be504dc7509ad2f12ea6adb43c/unstructured/partition/xlsx.py#L191) in partition excel function, we should open file and wrap in BytesIO.

badGarnet: so I would suggest this PR to actually implement that solution instead of adding this test

Achieve3318: > so I would suggest this PR to actually implement that solution instead of adding this test
So, can you let me know what do I have to do?

Achieve3318:
Hi, @badGarnet  #4036
this issue is to wrap funcion in BytesIO. but that function already wapped by BytesIO.
So I added test case.
Please check my PR. I'd like to solve this issue and merge this PR

test_unstructured/partition/test_xlsx.py
badGarnet: the root of the issue is this code path: https://github.com/Unstructured-IO/unstructured/blob/c04235c16842e5be504dc7509ad2f12ea6adb43c/unstructured/partition/xlsx.py#L190-L191
where it tries to open a filename using `read_excel` function.
a solution would be update that path of the code to first load the file into a `BytesIO` stream then use `read_excel` to read the stream instead.

Achieve3318: Thank you for your reply
As you can see in 200~201 lines on main branch, BytesIO was implemented.
So I added test case cuz this issue was opened.

PR #4217, opened 2026-01-29
Fix: make pdf image dpi consistent

badGarnet:
This PR fixes an issue when using hi-res strategy the dpi for render pdf is not consistent between the unstructured config and what's passed into `unstructured-inference`. 

## problem

In `unstructured.partition.pdf._partition_pdf_or_image_local` we set `pdf_image_dpi` to 200 if not set. However, in the env config we set the default value to 350 and this value could be changed by a env variable. Those are not reflected in the `pdf_image_dpi` parameter inside `_partition_pdf_or_image_local`, causing inconsistencies. This shows up when one tries to extract image from elements:
- the image renders using the env config with 350 default value
- the bounding boxes are extracted in `unstructured-inference` using the passed in `pdf_image_dpi` which is 200 default

## solution

Set the `pdf_image_dpi` to the env config value so they are consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - **Add `group_elements_by_parent_id` utility function**: Groups elements by their `parent_id` metadata field for easier document hierarchy traversal (fixes #1489)
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


badGarnet:
@claude review

PR #4233, opened 2026-02-11
Fix ARM64 image issues

PastelStorm:
> [!NOTE]
> **Medium Risk**
> Touches release CD for multi-arch Docker publishing and alters Docker build dependency installation behavior; failures could block image releases or mask install issues despite limited runtime code impact.
> 
> **Overview**
> Improves the Docker publish GitHub Actions workflow by adding a manual `workflow_dispatch` trigger, switching to native runners for both `amd64` and `arm64`, and setting `fail-fast: false` so one-arch flakiness doesn’t cancel the other.
> 
> Reduces CD image testing to a consistent lightweight target (`test_unstructured/partition/test_text.py` plus smoke tests) and hardens Docker builds by adding 3-attempt retry logic around `apk update/add` to mitigate transient Chainguard mirror failures. Bumps version to `0.19.3` and updates the changelog accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 559244ddc4603fcb362ebf9ab3e2f152c1db3076. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.19.2
+## 0.19.3
 
 ### Enhancements
 - Use native ARM64 runner (`ubuntu-24.04-arm`) for Docker ARM64 builds in CD, replacing QEMU emulation for faster and more reliable builds
--- a/unstructured/__version__.py
+++ b/unstructured/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.19.2"  # pragma: no cover
+__version__ = "0.19.3"  # pragma: no cover


PastelStorm:
> I thought we were doing builds on the self-hosted runners so we could use the buildx builder that has both arm64 and amd64 backends configured (to get a nice manifest build in one step). At least that's what I was thinking when first making those builder instances... Does this technique produce a multi-arch manifest?

The problem with running those using QEMU is that the emulation layer slows down CD approximately 10 to 20 times. We're still producing a single manifest after both build are done:

1. build-images runs two parallel jobs (amd64 on opensource-linux-8core, arm64 on ubuntu-24.04-arm), each building natively and pushing its arch-specific image to the build repository
2. publish-images pulls both arch-specific images, then creates and pushes multi-arch manifests for latest, $SHORT_SHA, and $VERSION using docker manifest create

ctrahey:
The buildx builders specifically avoid qemu. There is a deployment running on native ARM nodes, and one running on x86. Then a "builder" is conifgured with both of them as backends. So any build that uses that builder will get native builds without emulation.

ctrahey:
Ah! I just realized we're chatting on the open source unstructured repo, so the self-hosted runners I was discussing are irrelevant! Scratch that, and love the change! 😆

PR #4240, opened 2026-02-15
fix: pin numba to >=0.63.0 for Python 3.12 compatibility

### LINKED ISSUE #4216: bug/Dependency resolution discrepancy for numba
**Describe the bug**
When installing `unstructured==0.18.31`, different package managers resolve the numba dependency to different versions:

`pip`: Resolves to `numba==0.63.1` (compatible with Python 3.12)
`uv`: Resolves to `numba==0.53.1` (requires Python <3.10, incompatible with Python 3.12)
The package declares unstructured as supporting Python 3.10, 3.11, and 3.12 (per classifiers and Requires-Python: >=3.10.0), but the unpinned `numba` dependency in the package metadata allows resolver tools to select incompatible versions.

**To Reproduce**
Run `uv pip install unstructured==0.18.31` in a new/clean virtual environment.

```
      RuntimeError: Cannot install on Python version 3.12.12; only versions >=3.6,<3.10 are supported.

      hint: This usually indicates a problem with the package or the build environment.
  help: `llvmlite` (v0.36.0) was included because `unstructured` (v0.18.31) depends on `numba` (v0.53.1) which depends on `llvmlite`
```

**Expected behavior**
Installation should complete without error

**Suggested fix**
Pin `numba` to `>=0.63.0` in `constraints.txt`

aviu16:
## Summary
Fixes #4216

This PR resolves inconsistent dependency resolution for numba across different package managers (pip vs uv) when installing on Python 3.12.

## The Problem
The package declares support for Python 3.10-3.13, but the previous constraint `numba>=0.60.0` allowed package managers to resolve to incompatible versions:

- **pip**: Correctly selects `numba==0.63.1` (Python 3.12 compatible)
- **uv**: Selects `numba==0.53.1` (requires Python <3.10, **incompatible** with 3.12)

This causes installation failures on Python 3.12 with uv:
```
Cannot install on Python version 3.12.12; only versions >=3.6,<3.10 are supported
```

## The Solution
Pin numba to `>=0.63.0`, which is the first version with Python 3.12 support. This ensures all package managers resolve to a compatible version.

## Changes
- `pyproject.toml`: Updated numba constraint from `>=0.60.0` to `>=0.63.0`

## Testing
This change aligns with numba's Python version support matrix and ensures consistent behavior across pip, uv, and poetry.


aviu16:
closing this for now, gonna revisit later

PR #4241, opened 2026-02-16
fix: accept any IO[bytes] object in convert_to_bytes()

bittoby:
## Problem

When a user opens a file from inside a zip archive and passes it directly into `partition()`,
the library crashes with `ValueError: Invalid file-like object type`.

Uploading a text file directly works fine. Uploading a zip containing that same text file
fails every time, even though the file inside the zip is perfectly readable.

## Root Cause

The `convert_to_bytes()` function only accepted a fixed list of known file types.
Anything not on that list was immediately rejected with an error — no attempt to read it,
no fallback, just a crash.

The file object returned when opening a file from a zip archive is not on that list,
so it was always rejected. This was a flaw in the design of the function: it checked
what the file *was* instead of checking what it could *do*.

## Fix

Replaced the rigid type-checking approach with a simple capability check.
Before giving up, the function now asks: does this object support reading?
If yes, read it. This makes the function behave correctly for zip archive files
and any other standard readable file object that was previously unrecognised,
without changing how any of the existing accepted types are handled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
-## 0.21.8
+## 0.21.9
+
 ### Fixes
 - **Fix `ValueError` when partitioning a text file loaded from a zip archive**: `convert_to_bytes()`
   rejected any file-like object not in its hardcoded type whitelist (`BytesIO`, `BufferedReader`,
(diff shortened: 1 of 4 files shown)


Crystora:
@badGarnet Could you please review this PR? thank you

bittoby:
@badGarnet Sorry for tagging again. I'd appreciate your feedback.

Crystora:
@badGarnet please give me any feedback! thanks

Crystora:
@cragwolfe Could you please review this PR? 
thank you

Crystora:
These testing fails are not relate my changes. Please review my PR @badGarnet @qued

PR #4243, opened 2026-02-18
fix: gracefully handle invalide html string during chunking

badGarnet:
This PR fixes an issue where an invalid `text_as_html` input into html based table chunking logic can lead to chunking failing. Like the following stack trace shows:

```
    |   File "/app/unstructured/unstructured/chunking/base.py", line 594, in iter_chunks
    |     yield from _TableChunker.iter_chunks(
    |   File "/app/unstructured/unstructured/chunking/base.py", line 837, in _iter_chunks
    |     html_size = measure(self._html) if self._html else 0
    |                                        ^^^^^^^^^^
    |   File "/app/unstructured/unstructured/utils.py", line 154, in __get__
    |     value = self._fget(obj)
    |             ^^^^^^^^^^^^^^^
    |   File "/app/unstructured/unstructured/chunking/base.py", line 866, in _html
    |     if not (html_table := self._html_table):
    |                           ^^^^^^^^^^^^^^^^
    |   File "/app/unstructured/unstructured/utils.py", line 154, in __get__
    |     value = self._fget(obj)
    |             ^^^^^^^^^^^^^^^
    |   File "/app/unstructured/unstructured/chunking/base.py", line 884, in _html_table
    |     return HtmlTable.from_html_text(text_as_html)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/app/unstructured/unstructured/common/html_table.py", line 61, in from_html_text
    |     root = fragment_fromstring(html_text)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/app/.venv/lib/python3.12/site-packages/lxml/html/__init__.py", line 810, in fragment_fromstring
    |     elements = fragments_fromstring(
    |                ^^^^^^^^^^^^^^^^^^^^^
    |   File "/app/.venv/lib/python3.12/site-packages/lxml/html/__init__.py", line 780, in fragments_fromstring
    |     raise etree.ParserError(
    | lxml.etree.ParserError: There is leading text: '```html\n'
```

The solution is to catch the parser error and return a `None` instead in `unstructured/chunking/base.py` in `_html_table`. This way we fallback to text based chunking for this element with a warning log.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,10 +193,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Set up Docker
       uses: docker/setup-docker-action@v4
-    - name: Setup docker-compose
-      uses: KengoTODA/actions-setup-docker-compose@v1
-      with:
-        version: '2.22.0'
+    - name: Set up Docker Compose
+      uses: docker/setup-compose-action@v1
     - name: Test (end-to-end)
       env:
         AIRTABLE_PERSONAL_ACCESS_TOKEN: ${{ secrets.AIRTABLE_PERSONAL_ACCESS_TOKEN }}
--- a/.github/workflows/ingest-test-fixtures-update-pr.yml
+++ b/.github/workflows/ingest-test-fixtures-update-pr.yml
@@ -29,10 +29,8 @@ jobs:
       - uses: ./.github/actions/base-cache
         with:
           python-version: ${{ env.PYTHON_VERSION }}
-      - name: Setup docker-compose
-        uses: KengoTODA/actions-setup-docker-compose@v1
-        with:
-          version: "2.22.0"
+      - name: Set up Docker Compose
+        uses: docker/setup-compose-action@v1
       - name: Update test fixtures
         env:
           AIRTABLE_PERSONAL_ACCESS_TOKEN: ${{ secrets.AIRTABLE_PERSONAL_ACCESS_TOKEN }}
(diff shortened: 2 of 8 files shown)


unstructured/chunking/base.py
```diff
@@ -881,7 +885,15 @@ def _html_table(self) -> HtmlTable | None:
         if not text_as_html:  # pragma: no cover
             return None
 
-        return HtmlTable.from_html_text(text_as_html)
+        try:
+            return HtmlTable.from_html_text(text_as_html)
+        except ParserError:
```
PastelStorm: We should also catch `ValueError` here and a test to cover this new path.

test_unstructured/chunking/test_base.py
```diff
@@ -1322,6 +1322,27 @@ def it_handles_table_with_none_text_without_error(self):
...
+
+        # -- should not raise ParserError --
+        chunks = list(_TableChunker.iter_chunks(table, "", ChunkingOptions()))
+
+        # -- falls back to text-only: a single Table chunk with no .text_as_html --
+        assert len(chunks) == 1
+        chunk = chunks[0]
+        assert isinstance(chunk, Table)
+        assert chunk.metadata.text_as_html is None
+
```
PastelStorm, line 1375: we're using a standard logger, from what I see added in `base.py`, so we can add a `caplog` assertion to validate the `ParserError, ValueError` warning here.

PastelStorm (approved):

PR #4245, opened 2026-02-19
fix: remap parent id after hashing

badGarnet:
This PR addresses an issue where hashing element id loses the reference for parent id. 
This happens when calling `partition_html` where the partition process already assigned parent ids for elements based on html structure before `apply_metadata` is called, i.e., before element id hashing happens. 
This fix ensures that the parent references stay unchanged after hashing.


badGarnet:
@claude review

PR #4246, opened 2026-02-19
Adds an action for detecting a regression in partion 

aadland6:


scripts/performance/benchmark_partition.py
```diff
@@ -0,0 +1,110 @@
...
+
+def _measure(filepath: str, strategy: str, iterations: int) -> float:
+    """Return the average wall-clock seconds for partitioning *filepath*.
+
+    Identical logic to time_partition.measure_execution_time().
+    """
+
+    total = 0.0
+    for _ in range(iterations):
+        t0 = time.time()
```
badGarnet, line 72: let's use time.perf_counter instead of time.time

PR #4262, opened 2026-02-23
fix: relax lower bound for pdfminer.six

badGarnet:
The current lower bound for pdfminer.six is still too new for some commonly used file parsing tools like `pdfplumber`. This PR lowers this bound so that `unstructured` is compatible with those tools.


lawrence-u10d:
@claude does this intro any CVEs?

lawrence-u10d:
@claude when did that performance bump get added? Which PR?

PR #4266, opened 2026-02-26
fix: avoid O(N²) re-scanning in _patch_current_chars_with_render_mode

KRRT7:
## Problem

`_patch_current_chars_with_render_mode` is called on every `do_TJ`/`do_Tj` text operator during PDF parsing. The original implementation re-scans the entire `cur_item._objs` list each time, checking `hasattr(item, "rendermode")` to skip already-patched items. For a page with N characters across M text operations, this is O(N*M) -- effectively quadratic.

Memray profiling showed this function as the #1 allocator: 17.57 GB total across 549M allocations in a session processing just 4 files.

## Fix

Track the last-patched index so each call only processes newly-added `LTChar` objects. Reset automatically when `cur_item` changes (new page or figure).

**Before:** O(N^2) per page -- re-scans all accumulated objects on every text operator
**After:** O(N) per page -- each object visited exactly once

### test_benchmark_patch_current_chars (100 do_TJ calls, 2000 chars)

| | Min | Median | Mean | OPS | Rounds |
|:---|---:|---:|---:|---:|---:|
| `031b0cf8e777` (base) | 1,853.4us | 2,155.1us | 2,311.3us | 432.7 ops/s | 279 |
| `aef3bc4d9af0` (head) | 230.3us | 270.1us | 444.2us | 2,251.1 ops/s | 2,326 |
| **Speedup** | **8.05x** | **7.98x** | **5.20x** | **5.20x** | |

### test_benchmark_patch_current_chars_dense (200 do_TJ calls, 5000 chars)

| | Min | Median | Mean | OPS | Rounds |
|:---|---:|---:|---:|---:|---:|
| `031b0cf8e777` (base) | 9,197.5us | 10,372.4us | 12,055.4us | 83.0 ops/s | 90 |
| `aef3bc4d9af0` (head) | 594.7us | 736.7us | 1,954.2us | 511.7 ops/s | 543 |
| **Speedup** | **15.47x** | **14.08x** | **6.17x** | **6.17x** | |

The superlinear speedup on the dense test confirms the O(N^2) to O(N) complexity change -- doubling the input size more than doubles the improvement.

---
*Generated by pytest-benchmark on Azure VM (codeflash compare worktree bug prevented automated comparison; ran manually at each ref)*

<details>
<summary><b>Reproduce the benchmark locally</b></summary>

```bash
# Base (before fix):
git checkout 031b0cf8e777
mkdir -p test_unstructured/benchmarks
# Copy benchmark file (see source below)
uv run pytest test_unstructured/benchmarks/test_benchmark_patch_current_chars.py \
  --benchmark-only --benchmark-json=/tmp/bench_base.json

# Head (after fix):
git checkout aef3bc4d9af0
# Copy benchmark file (see source below)
uv run pytest test_unstructured/benchmarks/test_benchmark_patch_current_chars.py \
  --benchmark-only --benchmark-json=/tmp/bench_head.json
```

</details>

<details>
<summary><b>Benchmark test source</b></summary>

```python
"""Benchmark for _patch_current_chars_with_render_mode optimization."""

import inspect
from unittest.mock import MagicMock
from pdfminer.layout import LTChar


class LightLTChar(LTChar):
    """Lightweight LTChar subclass that bypasses the heavy __init__."""
    def __init__(self):
        pass


class FakeCurItem:
    """Mimics device.cur_item with a list of layout objects."""
    def __init__(self):
        self._objs = []


def _make_interpreter():
    """Build a minimal CustomPDFPageInterpreter for benchmarking."""
    from unstructured.partition.pdf_image.pdfminer_utils import CustomPDFPageInterpreter
    interp = object.__new__(CustomPDFPageInterpreter)
    interp.device = MagicMock()
    cur_item = FakeCurItem()
    interp.device.cur_item = cur_item
    textstate = MagicMock()
    textstate.render = 0
    interp.textstate = textstate
    sig = inspect.signature(interp._patch_current_chars_with_render_mode)
    has_start = "start" in sig.parameters
    return interp, cur_item, has_start


def test_benchmark_patch_current_chars(benchmark):
    """Benchmark: 100 do_TJ calls, 2000 chars total."""
    interp, cur_item, has_start = _make_interpreter()
    chars_per_op = 20
    n_ops = 100

    def run_page():
        cur_item._objs.clear()
        for _ in range(n_ops):
            start = len(cur_item._objs)
            for _ in range(chars_per_op):
                cur_item._objs.append(LightLTChar())
            if has_start:
                interp._patch_current_chars_with_render_mode(start)
            else:
                interp._patch_current_chars_with_render_mode()

    benchmark(run_page)


def test_benchmark_patch_current_chars_dense(benchmark):
    """Benchmark dense: 200 do_TJ calls, 5000 chars total."""
    interp, cur_item, has_start = _make_interpreter()
    chars_per_op = 25
    n_ops = 200

    def run_page():
        cur_item._objs.clear()
        for _ in range(n_ops):
            start = len(cur_item._objs)
            for _ in range(chars_per_op):
                cur_item._objs.append(LightLTChar())
            if has_start:
                interp._patch_current_chars_with_render_mode(start)
            else:
                interp._patch_current_chars_with_render_mode()

    benchmark(run_page)
```

</details>

## Test plan
- [x] Benchmarked with pytest-benchmark on Azure VM (Standard_D8s_v5)
- [x] 7.98x median speedup (2000 chars), 14.08x median speedup (5000 chars)
- [x] Superlinear scaling confirms O(N^2) to O(N) complexity change
- [x] Existing unit tests pass

### DIFF
--- a/unstructured/partition/pdf_image/pdfminer_utils.py
+++ b/unstructured/partition/pdf_image/pdfminer_utils.py
@@ -18,29 +18,20 @@ class CustomPDFPageInterpreter(PDFPageInterpreter):
     """a custom pdfminer page interpreter that adds character render mode information to LTChar
     object as `rendermode` attribute. This is intended to be used to detect invisible text."""
 
-    def _patch_current_chars_with_render_mode(self):
-        """Add render_mode to recently created LTChar objects"""
+    def _patch_current_chars_with_render_mode(self, start: int):
+        """Add render_mode to LTChar objects added since index `start`."""
         cur_item = getattr(self.device, "cur_item", None)
         if not cur_item:
             return
-        objs = getattr(cur_item, "_objs", ())
         render_mode = self.textstate.render
-        # Reset index when cur_item changes (new page or figure)
-        if getattr(self, "_patched_cur_item", None) is not cur_item:
-            self._last_patched_idx = 0
-            self._patched_cur_item = cur_item
-        for obj in objs[self._last_patched_idx:]:
+        for obj in getattr(cur_item, "_objs", ())[start:]:
             if isinstance(obj, LTChar):
                 obj.rendermode = render_mode
-        self._last_patched_idx = len(objs)
 
     def do_TJ(self, seq):
+        start = len(getattr(getattr(self.device, "cur_item", None), "_objs", ()))
         super().do_TJ(seq)
-        self._patch_current_chars_with_render_mode()
-
-    def do_Tj(self, s):
-        super().do_Tj(s)
-        self._patch_current_chars_with_render_mode()
+        self._patch_current_chars_with_render_mode(start)
 
 
 class PDFMinerConfig(BaseModel):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.21.8
+
+### Enhancements
+- **Optimize PDF render mode patching performance**: Optimized `_patch_current_chars_with_render_mode` in `CustomPDFPageInterpreter` to avoid O(N²) re-scanning by tracking the last-patched index, so each `do_TJ`/`do_Tj` call only processes newly-added characters.
+
+
 ## 0.21.7
 
 ### Enhancements
--- a/test_unstructured/partition/pdf_image/test_pdfminer_utils.py
+++ b/test_unstructured/partition/pdf_image/test_pdfminer_utils.py
@@ -1,17 +1,128 @@
 from importlib import reload
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
-from pdfminer.layout import LTChar, LTContainer, LTTextLine
+from pdfminer.layout import LTChar, LTContainer, LTFigure, LTLayoutContainer, LTTextLine
 
 from test_unstructured.unit_utils import example_doc_path
 from unstructured.partition.pdf import partition_pdf
 from unstructured.partition.pdf_image.pdfminer_utils import (
+    CustomPDFPageInterpreter,
     _is_duplicate_char,
     deduplicate_chars_in_text_line,
     extract_text_objects,
     get_text_with_deduplication,
 )
 from unstructured.partition.utils import config as partition_config
+
+
+def _make_char():
+    return LTChar(
+        matrix=(1, 0, 0, 1, 0, 0),
+        font=MagicMock(),
+        fontsize=12,
+        scaling=1,
+        rise=0,
+        text="x",
+        textwidth=10,
+        textdisp=(0, 1),
+        ncs=MagicMock(),
+        graphicstate=MagicMock(),
+    )
+
+
+def _make_interpreter(cur_item):
+    interp = object.__new__(CustomPDFPageInterpreter)
+    interp.device = MagicMock()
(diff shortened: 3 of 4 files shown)


unstructured/partition/pdf_image/pdfminer_utils.py
```diff
@@ -20,15 +20,19 @@ class CustomPDFPageInterpreter(PDFPageInterpreter):
...
+        objs = getattr(cur_item, "_objs", ())
+        render_mode = self.textstate.render
+        # Reset index when cur_item changes (new page or figure)
+        if getattr(self, "_patched_cur_item", None) is not cur_item:
+            self._last_patched_idx = 0
+            self._patched_cur_item = cur_item
+        for obj in objs[self._last_patched_idx:]:
+            if isinstance(obj, LTChar):
+                obj.rendermode = render_mode
+        self._last_patched_idx = len(objs)
```
qued: The way I understand this logic is, that `_last_patched_idx` can either be 0 or `len(objs)`, so the net effect will be to either iterate through all of objs, or to skip all of them. If so, why not just gate the whole process on whether cur_item has changed? Something like:

```suggestion
        if getattr(self, "_patched_cur_item", None) is not cur_item:
            self._patched_cur_item = cur_item
            for obj in objs:
                if isinstance(obj, LTChar):
                    obj.rendermode = render_mode
```
I think it's easier to understand what's going on this way.

qued: This also seems different than the previous code in that this code will "re-patch" objects that have already been patched, while the old code has a guard against that. This would only be relevant if `self.textstate.render` can change between calls of this method while simultaneously `cur_item` changes but later changes back, and I'm not familiar enough with this flow to know whether that's a thing that can happen.

@claude can you investigate this?

KRRT7: Thanks for the review! The key detail is that `_patch_current_chars_with_render_mode` is called after every [`do_TJ`/`do_Tj`](https://github.com/Unstructured-IO/unstructured/blob/ec868859a9948bf63ccbaed0706a996991cff51c/unstructured/partition/pdf_image/pdfminer_utils.py#L37-L43), and each `super()` call appends new `LTChar` objects to `cur_item._objs` before we get here. So for a page with multiple text operations, `_objs` grows between calls while `cur_item` stays the same.

With the suggested approach, only the first text operation per `cur_item` would patch its characters — all subsequent ones on the same page would be skipped since `cur_item` hasn't changed.

The `_last_patched_idx` tracking is what gives us O(N) amortized — each call slices from where the last one left off, so every `LTChar` is visited exactly once.

KRRT7: actually I think you're onto something, going to debug and come up with a fix.

qued:
Opened duplicate #4269 as an internal PR for full access to CI tools.

unstructured/partition/pdf_image/pdfminer_utils.py
qued: > So for a page with multiple text operations, `_objs` grows between calls while `cur_item` stays the same.

Gotcha, that's a piece I was missing.

PR #4268, opened 2026-02-26
Add a check for complex pdfs

aadland6:
This checks if a pdf file is likely a complex document like mini-holistic-3-v1-Eng_Civil-Structural-Drawing_p001.pdf that is mostly vector graphics by comparing the ratio of vector images to text elements. 

This limits the overhead to every file by setting a minimum file size before running the check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,18 +17,6 @@
 ### Enhancements
 - **Optimize PDF render mode patching performance**: Optimized `_patch_current_chars_with_render_mode` in `CustomPDFPageInterpreter` to avoid O(N²) re-scanning by tracking the last-patched index, so each `do_TJ`/`do_Tj` call only processes newly-added characters.
 
-## 0.21.10
-- **Add Form Class**: Adds a new form class in elements.py to deal with forms
-
-## 0.21.9
-- Add a fallback to use the filetype library to recover from incorrect results form libmagic
-
-## 0.21.8
-
-### Enhancements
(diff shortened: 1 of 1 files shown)


unstructured/partition/pdf.py
```diff
@@ -575,6 +591,158 @@ def check_pdf_hi_res_max_pages_exceeded(
...
+        rb"(?:^|(?<=\s))"
+        rb"(?:m|l|c|v|y|h|re|S|s|f|F|f\*|B|B\*|b|b\*|n|W|W\*|cm|q|Q|Do|"
+        rb"g|G|rg|RG|k|K|cs|CS|w|J|j|M|d|i|gs)"
+        rb"(?=\s|$)",
+        re.MULTILINE,
+    )
+    TEXT_OPS_PATTERN = re.compile(
+        rb"(?:^|(?<=\s))" rb"(?:Tj|TJ|'|\"|Tf|Td|TD|Tm|T\*|BT|ET)" rb"(?=\s|$)",
+        re.MULTILINE,
+    )
```
qued: I think it's better to compile these outside the function, so they aren't compiled every time.

unstructured/partition/pdf.py
```diff
@@ -53,7 +54,10 @@
     check_language_args,
     prepare_languages_for_tesseract,
 )
-from unstructured.partition.common.metadata import apply_metadata, get_last_modified_date
+from unstructured.partition.common.metadata import (
+    apply_metadata,
+    get_last_modified_date,
+)
```
qued: Not sure why this and other things in this file are getting reformatted, our [spec](https://github.com/Unstructured-IO/unstructured/blob/main/pyproject.toml#L214) is line length 100.

qued:
@claude , could you review this? Additionally, could you suggest how to unit test the function `is_pdf_too_complex`?

qued (approved):
LGTM!

PR #4273, opened 2026-03-03
add check if libmagic fails

aadland6:


unstructured/file_utils/filetype.py
```diff
@@ -265,7 +265,23 @@ def _file_type_from_guessed_mime_type(self) -> FileType | None:
...
         # -- if no more-specific rules apply, use the MIME-type -> FileType mapping when present --
         file_type = FileType.from_mime_type(mime_type)
-        return file_type if file_type != FileType.UNK else None
+        if file_type != FileType.UNK:
+            return file_type
+
+        # -- on some environments libmagic can return a generic/unhelpful MIME-type
+        # -- like octet-stream") for files that the `filetype` package identify.
+        # -- when that happens we retry using `filetype`  `FileType.UNK` results.
+        if LIBMAGIC_AVAILABLE:
```
badGarnet, line 274: we only need to call this when `file_type == FileType.UNK and LIBMAGIC_AVAILABLE`, right?

aadland6: Yes - only cases where libmagic is already available but failed to get the right file 

1. mime_type is not None (otherwise the function already returned None), and

2. FileType.from_mime_type(mime_type) returned FileType.UNK, and

3. LIBMAGIC_AVAILABLE is True.

qued (approved):

PR #4298, opened 2026-03-25
feat(chunking): repeat table headers on continuation chunks

cragwolfe:
### Before
- Oversized table chunks only preserved headers in the first chunk; continuation chunks could lose column context.
- Table header semantics (`<thead>` / `<th>`) were not retained as explicit row-level metadata after compactification.

### After
- Added `repeat_table_headers` (default `True`) to chunking APIs and strategy plumbing:
  - `chunk_elements(..., repeat_table_headers=...)`
  - `chunk_by_title(..., repeat_table_headers=...)`
  - `add_chunking_strategy(...)` forwarded args/docs
- `_TableChunker` now detects contiguous leading header rows and repeats them on non-initial continuation chunks.
- Repeated header rows are prepended to both continuation chunk text and `text_as_html`.
- First chunk behavior remains unchanged relative to legacy output.
- Added a guardrail: if a repeated header row would consume more than half the chunk window, splitter falls back to legacy non-repeating behavior.

## Invariants
- No body-row drop, duplication, or reordering across emitted continuation chunks.
- Opt-out behavior (`repeat_table_headers=False`) matches legacy table splitting behavior.
- Chunk windows still respect max-size constraints, including near-boundary continuation windows.
- Only contiguous leading header rows are repeated; later non-leading header-like rows are not promoted.

## Edge cases covered
- No headers, single leading header row, multiple leading header rows.
- Header detection from both `<thead>` and `<th>` rows.
- Exact-fit and near-boundary continuation sizing.
- Cascading repetition across 3+ continuation chunks.
- Pathologically large header rows trigger safe fallback to non-repeating behavior.
- Strategy-path forwarding validated through `partition_html(..., chunking_strategy="by_title")`.

## Test evidence
- `uv run --no-sync pytest -q test_unstructured/chunking/test_dispatch.py` (6 passed)
- `uv run --no-sync pytest -q test_unstructured/chunking/test_base.py -k "Describe_TableChunker"` (26 passed)
- `uv run --no-sync pytest -q test_unstructured/chunking/test_title.py::test_add_chunking_strategy_forwards_repeat_table_headers` (1 passed)
- `uv run --no-sync pytest -q test_unstructured/chunking/test_title.py -k "repeat_table_headers"` (5 passed)
- `uv run --with python-docx pytest -q test_unstructured/chunking/test_basic.py -k "repeat_table_headers"` (4 passed)
- `uv run --no-sync pytest -q test_unstructured/common/test_html_table.py` (26 passed)

authored by codex

### DIFF
--- a/unstructured/chunking/base.py
+++ b/unstructured/chunking/base.py
@@ -348,6 +348,41 @@ class ChunkingOptions:
 
 
 # ================================================================================================
+# TABLE ISOLATION (SHARED PRECHECKS)
+# ================================================================================================
+# Tables are always staged alone in a pre-chunk so downstream splitting can emit `Table` /
+# `TableChunk` elements instead of folding them into `CompositeElement` with surrounding text.
+# See GitHub issue #3921 and the chunking strategy docs for rationale.
+
+
+def _element_is_table_family(element: Element) -> bool:
+    """True when ``element`` is a `Table` or a concrete subtype such as `TableChunk`.
+
+    Subclasses share the same isolation contract: they must not share a pre-chunk with arbitrary
+    text elements, and two table-bearing sequences must not be merged by `PreChunkCombiner`.
+    """
+    return isinstance(element, Table)
+
+
+def _elements_contain_table_family(elements: Iterable[Element]) -> bool:
+    """True when ``elements`` already includes at least one table-family element."""
+    return any(_element_is_table_family(e) for e in elements)
+
+
+def _table_isolation_forbids_side_by_side_merge(
+    left: Iterable[Element],
+    right: Iterable[Element],
+) -> bool:
+    """True when a proposed merge of two element streams must be rejected for table isolation.
+
+    If either side already contains a table, the combiner must flush before accepting the other
+    side. This keeps `combine_text_under_n_chars` from concatenating a table pre-chunk with
+    neighboring narrative pre-chunks.
+    """
+    return _elements_contain_table_family(left) or _elements_contain_table_family(right)
+
+
+# ================================================================================================
 # PRE-CHUNKER
 # ================================================================================================
 
@@ -456,6 +491,12 @@ class PreChunkBuilder:
 
     def add_element(self, element: Element) -> None:
         """Add `element` to this section."""
+        # -- do not prefix a table-only pre-chunk with narrative overlap from the prior chunk --
+        if len(self._elements) == 0 and _element_is_table_family(element):
+            self._overlap_prefix = ""
+            self._text_segments = []
+            self._text_len = 0
+
         self._elements.append(element)
         if element.text:
             self._text_segments.append(element.text)
@@ -478,7 +519,11 @@ class PreChunkBuilder:
         pre_chunk = PreChunk(elements, self._overlap_prefix, self._opts)
         # -- clear builder before yield so we're not sensitive to the timing of how/when this
         # -- iterator is exhausted and can add elements for the next pre-chunk immediately.
-        self._reset_state(pre_chunk.overlap_tail)
+        overlap_for_next = pre_chunk.overlap_tail
+        # -- table tails must not prefix the following narrative pre-chunk (overlap_all) --
+        if len(elements) == 1 and _element_is_table_family(elements[0]):
+            overlap_for_next = ""
+        self._reset_state(overlap_for_next)
         yield pre_chunk
 
     def will_fit(self, element: Element) -> bool:
@@ -493,6 +538,14 @@ class PreChunkBuilder:
         - A text-element will not fit when together with the elements already present it would
           exceed the hard-max (aka. max_characters/max_tokens).
         """
+        # -- a `Table` can only start a pre-chunk; it is never appended to a non-empty pre-chunk --
+        if _element_is_table_family(element):
+            return len(self._elements) == 0
+
+        # -- no non-table element may share a pre-chunk with a `Table` --
+        if _elements_contain_table_family(self._elements):
+            return False
+
         # -- an empty pre-chunk will accept any element (including an oversized-element) --
         if len(self._elements) == 0:
             return True
@@ -574,6 +627,8 @@ class PreChunk:
 
     def can_combine(self, pre_chunk: PreChunk) -> bool:
         """True when `pre_chunk` can be combined with this one without exceeding size limits."""
+        if _table_isolation_forbids_side_by_side_merge(self._elements, pre_chunk._elements):
+            return False
         if len(self._text) >= self._opts.combine_text_under_n_chars:
             return False
         # -- avoid duplicating length computations by doing a trial-combine which is just as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
-## 0.22.9-dev0
-
+## 0.22.10
 ### Enhancements
 - **Repeat table headers across continuation chunks**: Add `repeat_table_headers` to basic/title chunking options and table chunking internals so leading header rows are detected once and carried forward when large tables spill across multiple chunks.
+
+## 0.22.9
+
+### Enhancements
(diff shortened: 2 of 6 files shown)


unstructured/chunking/base.py
```diff
@@ -1081,6 +1122,71 @@ def _iter_cell_splits(self, cell: HtmlCell) -> Iterator[TextAndHtml]:
...
+        # -- guard against pathological headers where a single repeated header row would consume
+        # -- more than half the chunking window.
+        return self._max_header_row_len <= (self._opts.hard_max + 1) // 2
+
+    @lazyproperty
+    def _max_header_row_len(self) -> int:
+        """Largest leading-header row text length."""
+        if not self._header_rows:
+            return 0
+        return max(row.text_len for row in self._header_rows)
```
badGarnet: here `text_len` is character based but elsewhere we use `self._opts.measure` to measure sizes (character or token based). we should use measure function to get the len here to be consistent

cragwolfe: Addressed in `eb9d91474` by replacing row-length checks with `self._opts.measure(" ".join(row.iter_cell_texts()))` so token-based and character-based chunking stay consistent. (authored by codex)

unstructured/chunking/base.py
```diff
@@ -902,8 +928,15 @@ def _iter_text_and_html_table_chunks(self) -> Iterator[TableChunk]:
...
 
+        header_row_count = self._leading_header_row_count if self._opts.repeat_table_headers else 0
+        splitter = _HtmlTableSplitter(
+            html_table,
+            self._opts,
+            header_row_count=header_row_count,
+        )
         yield from self._make_table_chunks(
-            _HtmlTableSplitter.iter_subtables(html_table, self._opts)
+            splitter._iter_subtables(),
```
qued: If we want to use `_HtmlTableSplitter` this way, we should probably make this method public.

unstructured/chunking/dispatch.py
```diff
@@ -173,23 +176,72 @@ def _chunk_sort_key(chunk: TableChunk) -> tuple[bool, int]:
...
     else:
         metadata.text_as_html = None
 
     return Table(text=text, metadata=metadata)
+
+
+def _num_carried_over_header_rows(chunk: TableChunk) -> int:
+    """Header rows prepended synthetically to this chunk."""
+    value = chunk.metadata.num_carried_over_header_rows
+    return value if value and value > 0 else 0
```
qued: `value > 0` is redundant here unless `chunk.metadata.num_carried_over_header_rows` can be negative (0 is falsy).

If we're just trying to avoid `None`s we could also do `return value or 0`.

unstructured/chunking/dispatch.py
```diff
@@ -173,23 +176,72 @@ def _chunk_sort_key(chunk: TableChunk) -> tuple[bool, int]:
...
+
+
+def _strip_carried_over_header_text(chunk: TableChunk) -> str:
```
qued, line 216: Claude demonstrated an issue that occurs when reconstructing the text. Certain kinds of headers fail to be properly recognized and stripped during reconstruction, for example:

```
from unstructured.chunking.basic import chunk_elements
from unstructured.chunking.dispatch import reconstruct_table_from_chunks
from unstructured.documents.elements import ElementMetadata, Table

# Header row has a cell with an <a> tag — text lives inside a child element.
# Body rows are large enough to force multiple chunks.
html = (
    "<table>"
    "<thead>"
    "<tr><th>ID</th><th><a href='#'>Category Link</a></th></tr>"
    "</thead>"
    "<tbody>"
    "<tr><td>1</td><td>Alpha data value here</td></tr>"
    "<tr><td>2</td><td>Bravo data value here</td></tr>"
    "<tr><td>3</td><td>Charlie data value here</td></tr>"
    "<tr><td>4</td><td>Delta data value here</td></tr>"
    "<tr><td>5</td><td>Echo data value here</td></tr>"
    "</tbody>"
    "</table>"
)

metadata = ElementMetadata(text_as_html=html)
table = Table(text="placeholder", metadata=metadata)
# Force small chunks so the table splits
chunks = chunk_elements([table], max_characters=80, include_orig_elements=False)

print("=== Chunks ===")
for i, c in enumerate(chunks):
    print(f"  [{i}] text={c.text!r}")
    print(f"       html={c.metadata.text_as_html!r}")
    print(f"       carried_headers={c.metadata.num_carried_over_header_rows}")
    print()

# Now reconstruct
tables = reconstruct_table_from_chunks(chunks)
print("=== Reconstructed ===")
for t in tables:
    print(f"  text={t.text!r}")
    print(f"  html={t.metadata.text_as_html!r}")
```

Output:
```
=== Chunks ===
  [0] text='ID 1 Alpha data value here 2 Bravo data value here 3 Charlie data value here'
       html='<table><tr><td>ID</td><td><a>Category Link</a></td></tr><tr><td>1</td><td>Alpha data value here</td></tr><tr><td>2</td><td>Bravo data value here</td></tr><tr><td>3</td><td>Charlie data value here</td></tr></table>'
       carried_headers=0

  [1] text='ID 4 Delta data value here 5 Echo data value here'
       html='<table><tr><td>ID</td><td><a>Category Link</a></td></tr><tr><td>4</td><td>Delta data value here</td></tr><tr><td>5</td><td>Echo data value here</td></tr></table>'
       carried_headers=1

=== Reconstructed ===
  text='ID 1 Alpha data value here 2 Bravo data value here 3 Charlie data value here ID 4 Delta data value here 5 Echo data value here'
  html='<table><tr><td>ID</td><td><a>Category Link</a></td></tr><tr><td>1</td><td>Alpha data value here</td></tr><tr><td>2</td><td>Bravo data value here</td></tr><tr><td>3</td><td>Charlie data value here</td></tr><tr><td>4</td><td>Delta data value here</td></tr><tr><td>5</td><td>Echo data value here</td></tr></table>'
```
(Note how "ID" isn't dropped from the beginning of the second chunk in the reconstructed text, but the reconstructed HTML is ok.)

That being said, I'm not sure embedded tags could end up in table cell HTML? Perhaps a precocious VLM might try injecting bold or italicized formatting into the cell text to mimic what it sees?

unstructured/chunking/base.py
```diff
@@ -1415,11 +1540,15 @@ def _iter_cell_texts(self) -> Iterator[str]:
...
+        """Number of chunk-size units remaining for accumulated row text."""
         # -- separators are one space (" ") at the end of each row's text, including last one to
         # -- account for space before prospective next row.
         separators_len = len(self._rows)
-        return self._maxlen - separators_len - sum(r.text_len for r in self._rows)
+        return self._maxlen - separators_len - sum(self._row_text_len(r) for r in self._rows)
+
+    def _row_text_len(self, row: HtmlRow) -> int:
+        """Length of `row` text in configured chunk-size units."""
+        return self._measure(" ".join(row.iter_cell_texts()))
```
qued, line 1613: We could avoid recalculating this by keeping a running total when we `add_row` and zeroing when we `flush`. Not sure how performance critical this is though (but it seems like we were trying to avoid recalculating before the change).

unstructured/chunking/base.py
cragwolfe: Addressed by switching this call site to the public splitter entrypoint in `9bc0f84a5` (`_HtmlTableSplitter.iter_subtables(...)`) and removing direct `_iter_subtables()` usage from `_TableChunker`.\n\n(authored by codex)

unstructured/chunking/dispatch.py
cragwolfe: Addressed in `3714c4da2` by simplifying this helper to `return value or 0`, so we avoid the redundant positive check while still normalizing `None` to `0`.\n\n(authored by codex)

unstructured/chunking/base.py
cragwolfe: Implemented in `bb6806618`: `_RowAccumulator` now keeps a running `_row_text_len` total, increments on `add_row()`, and resets on `flush()` to avoid recomputing row-length sums in `_remaining_space`.\n\n(authored by codex)

unstructured/chunking/dispatch.py
cragwolfe: Addressed in `12507587a` by preserving nested-markup header text during carried-header stripping, and revalidated with an E2E case using `<a>` in a header cell (reconstruction now yields a single `ID`/`Category Link` header in merged text).\n\n(authored by codex)

qued (approved):
LGTM!

PR #4306, opened 2026-03-28
fix: enable huge_tree for HTMLParser to handle large documents

### LINKED ISSUE #4289: Large HTML documents cannot be partitioned using the partition_html function.
partitioning large html documents leads to an empty result. This is due to the missing huge_tree option on HTML parser generation in:

https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/html/parser.py#L929

Fix:
Include the huge_tree option to solve this: `etree.HTMLParser(remove_comments=True, huge_tree=True)`

joaquinhuigomez:
Pass `huge_tree=True` to `lxml.html.HTMLParser` so that deeply nested or very large HTML documents can be parsed without hitting the default tree depth limit.

Closes #4289

### DIFF
--- a/unstructured/partition/html/parser.py
+++ b/unstructured/partition/html/parser.py
@@ -926,7 +926,7 @@ def derive_element_type_from_text(text: str) -> type[Text] | None:
 # ------------------------------------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


joaquinhuigomez:
Closing — will reopen if there's interest.

PR #4310, opened 2026-03-31
fix: pdfminer drops extractable text

qued:
> [!NOTE]
> **Medium Risk**
> Changes pdfminer integration to override CID font/CMap handling and introduces custom stream decoding/parsing, which can affect text extraction behavior and performance on diverse PDFs (mitigated by size/mapping caps).
> 
> **Overview**
> Fixes PDFs where **body text was silently dropped** because CIDFonts used an *embedded Encoding CMap stream* that `pdfminer.six` doesn’t resolve.
> 
> Adds a bounded embedded-CMap decoder/parser and wires it in via `CustomPDFCIDFont` + `CustomPDFResourceManager` so `init_pdfminer()` constructs CID fonts with a parsed CMap (including `WMode`), with DoS-oriented caps on decompression and total mappings.
> 
> Updates tests with a new fixture-driven regression for both `FAST` and `HI_RES` strategies plus targeted unit tests for CMap parsing/stream decoding, and bumps version to `0.22.12` with a changelog entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4326b15f6c400e81112f894576941d28fb150da7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/unstructured/partition/pdf_image/pdfminer_utils.py
+++ b/unstructured/partition/pdf_image/pdfminer_utils.py
@@ -1,16 +1,25 @@
 import os
 import re
 import tempfile
-from typing import BinaryIO, List, Optional, Tuple, Union
+import zlib
+from typing import BinaryIO, List, Mapping, Optional, Tuple, Union
 
-from pdfminer.cmapdb import CMap
+from pdfminer import settings as pdfminer_settings
+from pdfminer.cmapdb import CMap, CMapDB
 from pdfminer.converter import PDFPageAggregator
 from pdfminer.layout import LAParams, LTChar, LTContainer, LTImage, LTItem, LTTextLine
-from pdfminer.pdffont import PDFCIDFont
-from pdfminer.pdfinterp import PDFPageInterpreter, PDFResourceManager
+from pdfminer.pdffont import PDFCIDFont, PDFFontError
+from pdfminer.pdfinterp import LITERAL_FONT, PDFPageInterpreter, PDFResourceManager
 from pdfminer.pdfpage import PDFPage
-from pdfminer.pdftypes import PDFStream, resolve1
+from pdfminer.pdftypes import (
+    LITERALS_ASCII85_DECODE,
+    LITERALS_ASCIIHEX_DECODE,
+    LITERALS_FLATE_DECODE,
+    PDFStream,
+    resolve1,
+)
 from pdfminer.psexceptions import PSSyntaxError
+from pdfminer.psparser import literal_name
 from pydantic import BaseModel
 
 from unstructured.logger import logger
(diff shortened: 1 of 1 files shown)


unstructured/partition/pdf_image/pdfminer_utils.py
```diff
@@ -1,18 +1,142 @@
...
+    code2cid: dict[int, object] = {}
+    total_mappings = 0
+
+    for match in _RE_CIDRANGE_BLOCK.finditer(data):
+        entries = _RE_CIDRANGE_ENTRY.findall(match.group(1))
+        for start_hex, end_hex, cid_str in entries:
+            start_bytes = bytes.fromhex(start_hex.decode("ascii"))
+            end_bytes = bytes.fromhex(end_hex.decode("ascii"))
+            start_cid = int(cid_str)
+            code_len = len(start_bytes)
```
badGarnet, line 75: claude insists this can be a problem but it seem if pdf is malformed we might have other issues so failing it is not necessarily bad:

  code_len is set from start_bytes and never considers end_bytes. This is used later at line 67-76 to decide single-byte vs multi-byte dict structure:

```
  if code_len == 1:
      code2cid[code_val] = cid
  else:
      code_bytes = code_val.to_bytes(code_len, "big")
```

  In well-formed CMaps this is fine — the spec requires start and end to have the same byte width, e.g. <00> <FF> 1 or <0000> <00FF> 1.

  In a malformed CMap like <00> <0100> 1:
  - start_bytes = b'\x00' (1 byte), end_bytes = b'\x01\x00' (2 bytes)
  - code_len = 1
  - start_val = 0, end_val = 256
  - Range size = 257, passes the DoS cap
  - The loop generates codes 0–256 and stores each as a single-byte key (code_len == 1), but code values 256+ don't fit in one byte
  - code_val = 256 stored as code2cid[256] = 256 — this "works" in the Python dict but is wrong semantically: CMap single-byte lookup keys should be 0–255.
  pdfminer's CMap.decode() reads one byte at a time from the input, so it will only ever query keys 0–255. The entries for 256+ are dead mappings that waste
   memory but are never hit.

  The opposite case is worse: <0000> <00FF> 1:
  - start_bytes = b'\x00\x00' (2 bytes), end_bytes = b'\xff' (1 byte)
  - code_len = 2
  - start_val = 0, end_val = 255
  - Range size = 256
  - Each code is stored in the 2-level nested dict structure (code_len == 2), so code 0 becomes code2cid[0][0] = 0. But the input stream only has
  single-byte codes — pdfminer will look up code2cid[byte_val] and get a dict instead of a CID int, which would cause a decode failure for those characters.

  In practice, this only matters for malformed PDFs — real CMap producers always use matching lengths. But given the PR already guards against other
  malformed inputs (reversed ranges, DoS), adding if len(start_bytes) != len(end_bytes): continue would be consistent with the defensive style.

Issue #3578, opened 2024-08-29
bug/Wrong parsing of html, xml code blocks in markdown


cgjosephlee:
**Describe the bug**
HTML and XML code blocks in markdown are not parsed properly.

Results:
~~~
HTML Example
```html
Hello, World!
This is a simple HTML example.
```
XML Example
xml <note> <to>Tove</to> <from>Jani</from> <heading>Reminder</heading> <body>Don't forget me this weekend!</body> </note>
```xml
```
```xml
```
~~~
- HTML tags are not preserved.
- XML code is malformed. The blank lines may erase the context.
- `<?xml version='1.0' encoding='UTF-8'?>` line breaks the parser.
```
Traceback (most recent call last):
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/test.py", line 14, in <module>
    elems = partition_html(
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/documents/elements.py", line 605, in wrapper
    elements = func(*args, **kwargs)
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 706, in wrapper
    elements = func(*args, **kwargs)
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/file_utils/filetype.py", line 662, in wrapper
    elements = func(*args, **kwargs)
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/chunking/dispatch.py", line 74, in wrapper
    elements = func(*args, **kwargs)
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/partition/html/partition.py", line 103, in partition_html
    elements = list(
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/partition/lang.py", line 475, in apply_lang_metadata
    elements = list(elements)
  File "/Users/joseph.lee/Documents/repos_ncb/test_rag/.venv/lib/python3.10/site-packages/unstructured/partition/html/partition.py", line 222, in iter_elements
    yield from cls(opts)._iter_elements()
[...]

PR #4362, opened 2026-06-05
fix(partition): skip ProcessingInstruction nodes in HTML parser

### LINKED ISSUE #4358: bug/xml indentation causes parsing error
**Describe the bug**
This error occurs when xml opening tags are on the same level as the xml context itself

'lxml.etree._ProcessingInstruction' object has no attribute 'is_phrasing'

This breaks
```
    ```xml
    <?xml version="1.0" encoding="UTF-8"?>
```
This succeeds
```
```xml
   <?xml version="1.0" encoding="UTF-8"?>
```

**To Reproduce**
Use unstructured to process a markdown file that contains an xml block with the opening tags on the same level as the start of the content

**Expected behavior**
This should not be a breaking format, it should continue in the same way as is the xml opening tag was tabbed back

**Screenshots**

**Environment Info**
I am running unstructured version 0.18.24

michaelxer:
## Summary

Fixes #4358

The HTML parser raises `AttributeError: '_ProcessingInstruction' object has no attribute 'is_phrasing'` when lxml injects `ProcessingInstruction` or `_Comment` nodes into the element tree. This is a regression of #3578 (the fix from PR #4034 was lost during a refactor).

## Root Cause

`Flow.iter_elements()`, `Phrasing._iter_child_text_segments()`, and `Anchor._iter_phrases_and_elements()` build a deque from `self` children and access `.is_phrasing` on each — but PI/Comment nodes bypass lxml's `ElementNamespaceClassLookup` and are raw lxml types without that attribute.

## Fix

Filter children to only `Flow`/`Phrasing` instances at all three `deque(self)` call sites (lines 370, 641, 743). Non-Element nodes (PI, Comment) are silently skipped.

## Test

Added `DescribeProcessingInstructionHandling` with two tests verifying PI nodes in both flow and phrasing contexts don't crash the parser. All 135 tests pass.

---
## Summary by cubic
Prevent HTML parser crashes when `lxml` injects ProcessingInstruction or comment nodes by skipping non-`Flow`/`Phrasing` children during iteration. This fixes AttributeError on nodes without `is_phrasing` and keeps parsing stable.

- **Bug Fixes**
  - Filter non-`Flow`/`Phrasing` children when queuing in `Flow.iter_elements`, `Phrasing._iter_child_text_segments`, and `Anchor._iter_phrases_and_elements`.
  - Add tests covering PI nodes in flow and phrasing contexts; update changelog.

<sup>Written for commit 52877bbddfd03502ff1850467b6899f4cd187375. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 0.22.31
 
+### Fixes
+
+- **Fix HTML parser crash on ProcessingInstruction nodes**: The HTML parser raised `AttributeError` when encountering `lxml.etree._ProcessingInstruction` nodes (e.g. `<?xml ... ?>` inside markdown code blocks). PI and Comment nodes are now silently skipped during child iteration. Regression of #3578.
+
 ### Enhancements
 
 - **Rename `isolate_tables` chunking option to `isolate_table`**: the option added in 0.22.30 has been renamed for naming consistency. Callers passing `isolate_tables=` must update to `isolate_table=`.
--- a/test_unstructured/partition/html/test_parser.py
+++ b/test_unstructured/partition/html/test_parser.py
@@ -1517,3 +1517,45 @@ class DescribeDefaultElement:
             "O Deep Thought computer, he said, The task we have designed you to perform is this.",
             "We want you to tell us.... he paused,",
         ]
+
+
+class DescribeProcessingInstructionHandling:
+    """ProcessingInstruction nodes in HTML must not crash the parser."""
+
+    def it_skips_pi_nodes_in_flow_elements(self):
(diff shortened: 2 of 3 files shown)


michaelxer:
Cleaning up PR portfolio. Fix is still valid if maintainers are interested.

Issue #3871, opened 2025-01-17
Number getting converted into scientific notation in metadata.text_as_html


sahil0094:
## Problem Description
When using `partition_html()` and extracting table metadata via `chunk.metadata.text_as_html`, numeric values are being automatically converted to exponential notation.

### Example
- Input Number: 478923
- Converted Output: 4.7e+05

## Steps to Reproduce
1. Use `partition_html()` on an HTML file
2. Extract table chunks
3. Access `chunk.metadata.text_as_html`
4. Observe numeric value conversion

## Expected Behavior
- Numeric values should be preserved in their original format
- No automatic scientific notation conversion

## Environment Details
- Unstructured Library Version: 0.10.28
- Python Version: 3.11.0rc1
- Environment: databricks runtime 15.4 LTS ML

## Potential Impact
This automatic conversion can cause data integrity issues, especially in financial or scientific data processing.

## Suggested Investigation
- Review number parsing/serialization logic
- Check type conversion mechanisms in metadata handling

gyx09212214-prog:
I opened PR #4373 with regression coverage for this behavior on the `partition_html(..., chunking_strategy="by_title")` path.

I wasn't able to reproduce the scientific-notation conversion on current `main` with a minimal HTML table fixture, but the added test locks in the expected behavior that numeric-looking table cell text such as `478923`, `000001`, and `0.000123` remains unchanged in chunked `metadata.text_as_html`.

zeweihan:
Wanted to add a concrete angle on why this matters beyond general "data integrity": in legal and financial document pipelines (fee schedules, payment terms, loan amortization, M&A consideration tables) **every digit is a covenant term**, so the silent coercion `478923 → 4.7e+05` is not a cosmetic formatting bug — it's a contract-terms corruption bug.

Concrete failure mode we hit: a `text_as_html` table cell holding `4,923,847` (purchase price in USD) survives ingestion as a float, round-trips through `to_html()`, and on re-parse becomes `4,920,000`. A `3,847` delta is enough to be a material breach of a payment provision, and there is no signal anywhere in the element metadata that the value was lossy. That's the dangerous part — unlike an OCR error, a scientifically-notated number *looks clean* to every downstream validator, so it sails straight into the RAG store. Confidence scores (ref #4320) don't help here because the extraction "succeeded" with no structural ambiguity; the loss happened in numeric coercion, not layout.

[...]

PR #4373, opened 2026-06-12
fix: preserve numeric text in table HTML metadata

gyx09212214-prog:
## Summary

Refs #3871.

This updates table HTML generation so delimited table partitioners preserve numeric-looking cell text in `metadata.text_as_html` instead of letting pandas serialize numeric columns through `DataFrame.to_html()`.

Changes:

- read CSV and TSV data as strings with default NA coercion disabled, preserving source lexemes such as `478923`, `000001`, `000123`, `007`, `0.000123`, and very large integer-like values
- render CSV/TSV table HTML through a shared `htmlify_dataframe()` helper backed by `htmlify_matrix_of_cell_texts()`, avoiding pandas float/scientific-notation formatting
- preserve dataframe null values as empty table cells instead of serializing them as `nan`, `None`, or `<NA>` text
- add regression coverage for CSV and TSV `metadata.text_as_html`, including TSV file-like input and explicit leading-zero identifiers
- keep the existing `partition_html(..., chunking_strategy="by_title")` regression guard for numeric table cells, now including explicit leading-zero values

## Notes

I still could not reproduce the scientific-notation conversion through current `partition_html()` alone with a minimal HTML table fixture. I confirmed the current HTML table partition path does not do an `HTML -> DataFrame -> to_html()` round-trip: `TableBlock.iter_elements()` builds table data directly from DOM cell text and renders it through `htmlify_matrix_of_cell_texts()`.

The reproducible lossy path is the DataFrame-backed CSV/TSV table serialization path, which matches the pandas `to_html()` failure mode discussed in #3871.

## Testing

```bash
python -m pytest -q test_unstructured\common\test_html_table.py test_unstructured\partition\test_csv.py test_unstructured\partition\test_tsv.py
python -m ruff check unstructured\common\html_table.py unstructured\partition\csv.py unstructured\partition\tsv.py test_unstructured\common\test_html_table.py test_unstructured\partition\test_csv.py test_unstructured\partition\test_tsv.py test_unstructured\partition\html\test_partition.py
python -m ruff format --check unstructured\common\html_table.py unstructured\partition\csv.py unstructured\partition\tsv.py test_unstructured\common\test_html_table.py test_unstructured\partition\test_csv.py test_unstructured\partition\test_tsv.py test_unstructured\partition\html\test_partition.py
git diff --check
```

### DIFF
--- a/test_unstructured/common/test_html_table.py
+++ b/test_unstructured/common/test_html_table.py
@@ -53,6 +53,8 @@ class Describe_htmlify_dataframe:
             [
                 ["revenue", "478923"],
                 ["ticker", "000001"],
+                ["account", "000123"],
+                ["clause", "007"],
                 ["ratio", "0.000123"],
             ],
             columns=["metric", "value"],
@@ -63,6 +65,8 @@ class Describe_htmlify_dataframe:
             "<tr><td>metric</td><td>value</td></tr>"
             "<tr><td>revenue</td><td>478923</td></tr>"
             "<tr><td>ticker</td><td>000001</td></tr>"
+            "<tr><td>account</td><td>000123</td></tr>"
+            "<tr><td>clause</td><td>007</td></tr>"
             "<tr><td>ratio</td><td>0.000123</td></tr>"
             "</table>"
         )
--- a/test_unstructured/common/test_html_table.py
+++ b/test_unstructured/common/test_html_table.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import pandas as pd
 import pytest
 from lxml.html import fragment_fromstring
 
@@ -11,6 +12,7 @@ from unstructured.common.html_table import (
     HtmlCell,
     HtmlRow,
     HtmlTable,
+    htmlify_dataframe,
     htmlify_matrix_of_cell_texts,
 )
 
@@ -41,6 +43,29 @@ class Describe_htmlify_matrix_of_cell_texts:
 
     def test_htmlify_matrix_handles_empty_matrix(self):
         assert htmlify_matrix_of_cell_texts([]) == ""
+
+
+class Describe_htmlify_dataframe:
+    """Unit-test suite for `unstructured.common.html_table.htmlify_dataframe()`."""
+
+    def it_preserves_numeric_text_when_rendering_dataframe_cell_values(self):
+        dataframe = pd.DataFrame(
+            [
+                ["revenue", "478923"],
+                ["ticker", "000001"],
+                ["ratio", "0.000123"],
+            ],
+            columns=["metric", "value"],
+        )
+
+        assert htmlify_dataframe(dataframe, include_header=True) == (
+            "<table>"
+            "<tr><td>metric</td><td>value</td></tr>"
+            "<tr><td>revenue</td><td>478923</td></tr>"
+            "<tr><td>ticker</td><td>000001</td></tr>"
+            "<tr><td>ratio</td><td>0.000123</td></tr>"
+            "</table>"
+        )
 
 
 class DescribeHtmlTable:
--- a/test_unstructured/common/test_html_table.py
+++ b/test_unstructured/common/test_html_table.py
@@ -67,6 +67,23 @@ class Describe_htmlify_dataframe:
             "</table>"
         )
 
+    def it_renders_dataframe_null_values_as_empty_cells(self):
+        dataframe = pd.DataFrame(
+            [
+                ["real nulls", None, float("nan"), pd.NA],
+                ["literal text", "None", "nan", "<NA>"],
+            ],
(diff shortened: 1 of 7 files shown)


zeweihan:
Great direction — this is exactly the "carry the source lexeme as a string through the round-trip" fix that #3871 was missing, and reading CSV/TSV with NA coercion disabled so `478923` / `000001` / `0.000123` survive is the right call.

A couple of notes from ingesting a lot of structured-but-regulated documents, in case they're useful:

**Leading zeros are a silent reference-breakage case, not just cosmetic.** The `000001` example is more dangerous than it looks: account numbers, bond/CUSIP-like identifiers, statute section numbers, and clause IDs routinely carry leading zeros. Once a column is inferred as int64, `000001` becomes `1` with no metadata flag, and every downstream "find Section 0001" / "match account 00012345" retrieval query silently misses. It's worth adding an explicit leading-zero fixture to the regression suite (`"000123"`, `"007"`) so a future dtype-inference refactor can't silently regress it — the current string-everywhere approach fixes it, but it's the exact class of bug that creeps back when someone later re-enables `pd.read_csv(..., dtype=...)` inference for performance.

**The HTML-backed table path is the one I'd most want covered.** You noted you couldn't reproduce the lossy conversion through `partition_html()` alone with a minimal fixture. The dangerous real-world case for legal/financial PDFs isn't a hand-authored HTML table though — it's an HTML table that an OCR/layout model *synthesized* from a PDF page and handed to the partitioner, which then gets materialized through a DataFrame before `text_as_html` is rendered. If that path still goes `HTML → DataFrame (dtype inference) → to_html()`, the coercion can reappear even though the CSV/TSV path is now fixed. Worth confirming whether `htmlify_dataframe()` / `htmlify_matrix_of_cell_texts()` is the single funnel for *all* table-HTML emission (CSV, TSV, **and** the HTML partition path), or whether the HTML partitioner still has its own DataFrame round-trip. If it's the latter, the same `to_html()` failure mode is still reachable.

**Null-as-empty-cell is correct, but there's a semantic trap worth a comment in the code.** Rendering nulls as empty `<td></td>` is unambiguously better than literal `nan` / `None` / `<NA>` text leaking into the corpus. The one caveat: in regulatory/fee-schedule tables, an intentionally-blank cell usually means "not applicable" or "no penalty" rather than "missing/unknown". An empty string can't distinguish those, but that's acceptable — the important thing is the consumer sees *empty* (a parseable signal) instead of *garbage tokens that look like content*. If there's ever appetite, a separate `cell_metadata` with `{was_null: true}` would let a downstream consumer treat blanks as data-absent vs. value-absent, but empty-cell-first is the right default.

Thanks for picking this up — the `Refs #3871` + string-everywhere approach is the clean fix.

gyx09212214-prog:
Thanks for the review and the additional context here.

I’ve incorporated the feedback and tightened the PR around the original data-integrity issue:

- CSV/TSV inputs are now read with source lexemes preserved as strings, with default NA coercion disabled, so values like `478923`, `000001`, `000123`, `007`, `0.000123`, and very large integer-like identifiers survive into `metadata.text_as_html` unchanged.
- CSV/TSV table HTML generation now goes through a shared `htmlify_dataframe()` helper backed by `htmlify_matrix_of_cell_texts()`, avoiding pandas `DataFrame.to_html()` numeric formatting and scientific notation.
- Pandas null values are rendered as empty table cells instead of leaking `nan`, `None`, or `<NA>` tokens into the table text, with regression coverage for that behavior.
- I added explicit leading-zero fixtures across the shared renderer, CSV, TSV filename/file-like paths, and the existing `partition_html(..., chunking_strategy="by_title")` regression guard.
- I also confirmed the current `partition_html` table path does not do an `HTML -> DataFrame -> to_html()` round-trip; it builds table data directly from DOM cell text and renders through `htmlify_matrix_of_cell_texts()`.

Validation run locally:

```bash
python -m pytest -q test_unstructured\common\test_html_table.py test_unstructured\partition\test_csv.py test_unstructured\partition\test_tsv.py
python -m ruff check unstructured\common\html_table.py unstructured\partition\csv.py unstructured\partition\tsv.py test_unstructured\common\test_html_table.py test_unstructured\partition\test_csv.py test_unstructured\partition\test_tsv.py test_unstructured\partition\html\test_partition.py
python -m ruff format --check unstructured\common\html_table.py unstructured\partition\csv.py unstructured\partition\tsv.py test_unstructured\common\test_html_table.py test_unstructured\partition\test_csv.py test_unstructured\partition\test_tsv.py test_unstructured\partition\html\test_partition.py
git diff --check
```

The remaining Cubic blocker appears to be an outdated unresolved TSV coverage thread from an earlier review; the TSV regression coverage has since been added and the latest review reported no new issues. If this scope looks good to maintainers, I’d appreciate another review and merge when convenient.

PR #4385, opened 2026-07-02
Fix parallel PDF API script env handling

compair-steven:
## Summary

Fixes `scripts/user/process-pdf-parallel-through-api.sh` so it can be run with documented environment overrides and on the default Bash shipped with macOS.

Changes:
- Documents `UNST_API_KEY` as required.
- Adds `UNST_API_ENDPOINT` support instead of hard-coding the public endpoint.
- Gives `STRATEGY`, `BATCH_SIZE`, `PDF_SPLIT_PAGE_SIZE`, and `PDF_SPLITS_DIR` defaults before use.
- Avoids unbound-variable failures when `UNST_API_KEY` is missing.
- Replaces `mapfile` usage with a Bash 3.2-compatible loop.
- Supports both `md5sum` and BSD/macOS `md5 -q` in the parallel script and split helper.

## Validation

I tested the script against the hosted API with a valid `UNST_API_KEY` using a one-page split of `example-docs/pdf/layout-parser-paper-fast.pdf`:

```bash
PDF_SPLITS_DIR="$tmp" \
BATCH_SIZE=1 \
STRATEGY=fast \
scripts/user/process-pdf-parallel-through-api.sh "$pdf"
```

The script produced a valid split JSON file, combined it successfully, and the combined output contained 25 JSON elements:

```bash
jq length "$tmp/${base}-output-${md5}_split-10_strat-fast/${base}_combined.json"
# 25
```

I also checked the updated shell scripts with bash -n.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4385?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/user/process-pdf-parallel-through-api.sh
+++ b/scripts/user/process-pdf-parallel-through-api.sh
@@ -50,7 +50,20 @@ fi
 
 PDF_FILE="$1"
 PDF_NAME=$(basename "$PDF_FILE" .pdf)
-MD5_SUM=$(md5sum "$PDF_FILE" | awk '{ print $1 }')
+
+checksum_file() {
+  local file="$1"
+  if command -v md5sum >/dev/null 2>&1; then
+    md5sum "$file" | awk '{ print $1 }'
+  elif command -v md5 >/dev/null 2>&1; then
+    md5 -q "$file"
+  else
+    echo "Error: neither md5sum nor md5 is available." >&2
+    exit 1
+  fi
+}
+
+MD5_SUM=$(checksum_file "$PDF_FILE")
 PDF_DIR="$PDF_SPLITS_DIR/$PDF_NAME-${MD5_SUM}_split-${SPLIT_SIZE}"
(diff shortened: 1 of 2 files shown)


compair-steven:
Thanks for the quick review! I pushed a follow-up that validates BATCH_SIZE as a positive integer before the batching loop, tightens PDF_SPLIT_PAGE_SIZE to the same positive-integer shape, and checks that checksum generation produced a non-empty value before constructing split/output paths.

I also verified the failure paths for BATCH_SIZE=0, non-numeric BATCH_SIZE, PDF_SPLIT_PAGE_SIZE=0, and a missing PDF input. Hopefully this addresses the concerns :)

PR #4388, opened 2026-07-06
refactor: centralize URL fetching with host validation and default timeouts

william-u10d:
## Summary

Introduces `unstructured/safe_http.py` as a single, shared entry point for
outbound URL fetches, and routes the `url=` code paths in `partition`,
`partition_html`, and `partition_md` through it. Previously each of these
called `requests.get` directly with inconsistent (and in some cases absent)
timeout and validation behavior; this centralizes that logic in one place so
it stays consistent and testable.

## What changed

- **New `unstructured/safe_http.py`** — a `safe_get()` helper that all three
  partitioners now use instead of calling `requests.get` directly.
- The helper applies, consistently:
  - an `http`/`https` scheme allowlist
  - a hostname denylist with IDNA normalization
  - address validation performed at TCP connect time, so the address
    validated is the address actually connected to
  - manual redirect following with per-hop re-validation, dropping
    credential material (`Authorization`/`Cookie`/proxy auth, plus
    `auth=`/`cookies=`) on cross-origin hops via requests' own
    `should_strip_auth`
  - default `(connect, read)` timeouts, and refusal of proxied requests
  - an opt-out via `allow_private=` / the `UNSTRUCTURED_ALLOW_PRIVATE_URL`
    environment variable for controlled local usage

## Behavior changes (why this is a minor release)

Fetches that resolve to non-routable, loopback, or link-local addresses are
now rejected by default. Outbound fetches also now carry default timeouts
where some previously had none. Callers that legitimately need to reach a
private/internal host can opt out with `UNSTRUCTURED_ALLOW_PRIVATE_URL=1` (or
`allow_private=True`).

## Tests

Adds `test_unstructured/test_safe_http.py` covering IP/hostname
classification, connect-time validation, redirect re-validation, cross-origin
credential stripping, and the opt-out.

## Version

Bumps to `0.24.0` (see CHANGELOG) — minor, reflecting the behavior changes
above.

### DIFF
--- a/unstructured/safe_http.py
+++ b/unstructured/safe_http.py
@@ -1,0 +1,374 @@
+"""Centralized URL fetching with host validation and default timeouts.
+
+All HTTP fetches of user-supplied URLs should go through :func:`safe_get` so
+that scheme, hostname, and IP-range checks are applied consistently.
+"""
+
+from __future__ import annotations
+
+import ipaddress
+import logging
+import os
+import socket
+from typing import Any, Optional
+from urllib.parse import urljoin, urlparse
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.connection import HTTPConnection, HTTPSConnection
+from urllib3.connectionpool import HTTPConnectionPool, HTTPSConnectionPool
+from urllib3.exceptions import NameResolutionError
+from urllib3.poolmanager import PoolManager
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Blocked network ranges
+# ---------------------------------------------------------------------------
+
+# Ranges that ipaddress category flags don't classify but must still be blocked.
+_EXTRA_BLOCKED_NETWORKS = [
+    ipaddress.ip_network("100.64.0.0/10"),  # RFC 6598 CGNAT
+]
+
+# NAT64 (RFC 6052/8215) and IPv4-compatible IPv6 embed an IPv4 in the low 32 bits.
+_NAT64_NETWORKS = [
+    ipaddress.ip_network("64:ff9b::/96"),
+    ipaddress.ip_network("64:ff9b:1::/48"),
+]
+_IPV4_COMPAT_NETWORK = ipaddress.ip_network("::/96")
+
+_BLOCKED_HOSTNAMES = frozenset(
+    [
+        "localhost",
+        "metadata.google.internal",
+        "metadata.gke.internal",
+        "metadata.azure.com",
+        "kubernetes.default",
+        "kubernetes.default.svc",
+        "kubernetes.default.svc.cluster.local",
+    ]
+)
+
+_BLOCKED_IPS = frozenset(
+    [
+        "255.255.255.255",
+        "168.63.129.16",  # Azure wireserver
+    ]
+)
+
+_DEFAULT_TIMEOUT = (10, 300)  # (connect, read) seconds
+_MAX_REDIRECTS = 10
+_ENV_VAR = "UNSTRUCTURED_ALLOW_PRIVATE_URL"
+_TRUTHY_ENV_VALUES = frozenset({"1", "true", "yes"})
+
+
+def _env_allows_private() -> bool:
+    """Return True when the escape-hatch env var is set to a truthy value."""
+    return os.environ.get(_ENV_VAR, "").strip().lower() in _TRUTHY_ENV_VALUES
+
+
+class UnsafeURLError(ValueError):
+    """Raised when a URL targets a blocked or non-routable address."""
+
+
+# ---------------------------------------------------------------------------
+# IP / hostname helpers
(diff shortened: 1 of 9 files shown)


unstructured/safe_http.py
```diff
@@ -0,0 +1,374 @@
...
+            response = session.get(url, **kwargs)
+            if not response.is_redirect:
+                return response
+
+            location = response.headers.get("Location")
+            if not location:
+                return response  # malformed redirect — return as-is
+
+            new_url = urljoin(response.url, location)
+            _validate_url(new_url, allow_private=allow_private)
```
lawrence-u10d, line 409: Because `allow_redirects=False` and redirects are followed manually here, this loop bypasses `requests`' built-in cross-origin credential stripping. Re-validating `new_url` blocks redirects to internal IPs, but a redirect to a *different public origin* (e.g. `trusted.com` → an attacker-controlled host) still forwards any caller-supplied `Authorization`/`Cookie` header (and `auth=`/`cookies=`) to that target.

Suggest dropping credential-bearing headers on cross-origin hops, mirroring `requests.Session.should_strip_auth`:

```python
# module scope
_REDIRECT_SENSITIVE_HEADERS = frozenset({"authorization", "cookie", "proxy-authorization"})
_AUTH_PROBE = requests.Session()  # stateless; issues no requests

def _is_cross_origin(old_url: str, new_url: str) -> bool:
    try:
        return _AUTH_PROBE.should_strip_auth(old_url, new_url)
    except ValueError:
        return True  # fail safe

def _strip_sensitive_headers(kwargs: dict) -> None:
    headers = kwargs.get("headers")
    if headers:
        kwargs["headers"] = {
            k: v for k, v in headers.items() if k.lower() not in _REDIRECT_SENSITIVE_HEADERS
        }
    kwargs.pop("auth", None)
    kwargs.pop("cookies", None)
```

then in the loop:

```python
new_url = urljoin(response.url, location)
_validate_url(new_url, allow_private=allow_private)
if _is_cross_origin(url, new_url):
    _strip_sensitive_headers(kwargs)
url = new_url
```

That restores the credential-handling semantics a caller would otherwise get from `requests` with `allow_redirects=True`.

unstructured/safe_http.py
```diff
@@ -0,0 +1,374 @@
...
+def _validate_url(url: str, allow_private: bool = False) -> None:
+    """Pre-request validation. Connect-time check in _SafeHTTPConnection is authoritative.
+
+    Skipped entirely when ``allow_private`` is True or the
+    ``UNSTRUCTURED_ALLOW_PRIVATE_URL`` env var is truthy.
+    """
+    if allow_private or _env_allows_private():
+        return
+
+    parsed = urlparse(url)
```
lawrence-u10d: Two small fail-closed gaps in the pre-request validation:

1. A malformed/out-of-range port (e.g. `http://host:99999/`) makes `parsed.port` raise `ValueError`, but `.port` isn't accessed here — so it passes validation and only surfaces later as a lower-level error. Touching `.port` inside a guarded parse fails it closed up front:

```python
try:
    parsed = urlparse(url)
    _ = parsed.port  # out-of-range port raises ValueError here
except ValueError as exc:
    raise UnsafeURLError("URL could not be parsed") from exc
```

2. The `getaddrinfo` call below (line 171) only catches `socket.gaierror`. An IDNA-unencodable hostname can raise `UnicodeError` (and some inputs `ValueError`), which would propagate uncaught instead of failing closed. Widening the except keeps it safe:

```python
except (socket.gaierror, UnicodeError, ValueError) as exc:
    raise UnsafeURLError("Failed to resolve hostname") from exc
```

lawrence-u10d (commented):
A couple of notes on the URL-fetch hardening — one on redirect credential handling, one on fail-closed behavior in the pre-request check. Both are small additions; details inline.

lawrence-u10d (approved):
Both items from the earlier review are addressed:

- **Cross-origin redirects strip credentials** — `_is_cross_origin` (delegating to `requests`' `should_strip_auth`, failing safe on unparseable input) + `_strip_sensitive_headers` now drop `Authorization`/`Cookie`/`proxy-authorization` and the `auth=`/`cookies=` kwargs before a cross-origin hop, so a redirect to a different public origin no longer forwards caller credentials.
- **Pre-request validation fails closed** — `.port` is accessed inside a guarded `urlparse` (rejects malformed/out-of-range ports up front), and `getaddrinfo` now catches `UnicodeError`/`ValueError` in addition to `gaierror`.

Core SSRF defense remains solid: connect-time IP validation on the resolved address closes the DNS-rebinding window (TLS/SNI unchanged), plus the scheme allowlist, default `(10, 300)` timeout, proxy refusal, and the `allow_private=` / `UNSTRUCTURED_ALLOW_PRIVATE_URL` opt-out. New tests cover both additions and CI is green (52/52).

LGTM. 🚀

PR #4389, opened 2026-07-06
fix: add SSRF host validation to partition url= fetch paths

lawrence-u10d:
## Summary

`partition(url=)`, `partition_html(url=)`, and `partition_md(url=)` fetched caller-supplied URLs with a bare `requests.get` — no host validation, and no timeout on two of the three — then returned the response body as element text. That is a full-read SSRF: a caller-supplied URL can reach loopback, private, link-local, and cloud-metadata addresses and read back the response.

This routes all three sinks through a single shared `unstructured/safe_http.py::safe_get`.

## Approach

`safe_get` applies, in one place:

- an `http`/`https` scheme allowlist;
- rejection of non-routable/internal targets using the **resolved IP validated at TCP-connect time** — the address we validate is the address we connect to, which closes the DNS-rebinding (TOCTOU) window a resolve-then-connect check leaves open, while leaving TLS/SNI and certificate verification unchanged;
- manual redirect following (`allow_redirects` forced `False`) with per-hop re-validation and credential-header stripping on cross-origin hops, so a redirect can't bounce to an internal target or leak the caller's `Authorization`/`Cookie`;
- a default `(connect, read)` timeout; and
- an `allow_private=True` / `UNSTRUCTURED_ALLOW_PRIVATE_URL=1` opt-out for controlled internal fetches.

## Relationship to #4388

This is a leaner alternative to #4388 with the same core defense (connect-time IP validation + redirect re-validation + timeout + opt-out). Relative to that PR it intentionally drops:

- **the Unicode "visual confusables" fold table** — the connect-time IP check is authoritative, so hostname string-matching isn't needed; validating the *resolved* IP also covers obfuscated literals (e.g. `http://2130706433`) and case/IDN variants with no denylist;
- **the RFC 7230 header sanitizer** — header injection is a distinct vulnerability class, and the headers come from the trusted caller rather than the attacker-controlled URL, so it's better handled separately; and
- **the explicit hostname denylist / IDNA normalization** — subsumed by the resolved-IP check.

Result: a ~320-line helper + ~300-line test file (50 cases) vs ~597 + ~1130 (175 cases), with the same SSRF properties.

## Behavior change

Fetches resolving to non-routable, loopback, or link-local addresses are now rejected by default — hence the `0.24.0` minor bump. Set `UNSTRUCTURED_ALLOW_PRIVATE_URL=1` (or pass `allow_private=True`) to opt out. The secure path also sets `trust_env=False` and refuses a `proxies=` kwarg, since a proxy would bypass the connection-level IP check; proxied or internal egress must use the opt-out.

## Tests

`test_unstructured/test_safe_http.py` (50 cases): IP classification, scheme/host validation, split-DNS rejection, connect-time validation, redirect re-validation + credential stripping, timeout, and the opt-out. Existing `partition_html`/`partition_md` URL tests updated to mock `safe_get`. `ruff check` / `ruff format --check` and the version/CHANGELOG sync check pass.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4389?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.24.0
+
+### Fixes
+
+- **Validate outbound URLs against SSRF (`partition(url=...)`, `partition_html`, `partition_md`)**: the `url=` fetch paths previously called `requests.get` directly with no host validation — and, for two of the three, no timeout — letting a caller-supplied URL reach loopback/private/link-local/cloud-metadata addresses and return the response body (a full-read server-side request forgery). Fetches now route through a shared `unstructured.safe_http.safe_get`, which enforces an `http`/`https` scheme allowlist, blocks non-routable/internal targets using the *resolved* IP checked at TCP-connect time (closing the DNS-rebinding window), follows redirects manually with per-hop re-validation and credential-header stripping on cross-origin hops, and applies a default `(connect, read)` timeout. **Behavior change (hence the minor bump):** fetches resolving to non-routable, loopback, or link-local addresses are now rejected by default; set `UNSTRUCTURED_ALLOW_PRIVATE_URL=1` (or pass `allow_private=True`) to opt out for controlled internal usage.
+
 ## 0.23.3
 
 ### Fixes
--- a/test_unstructured/partition/html/test_partition.py
+++ b/test_unstructured/partition/html/test_partition.py
@@ -108,8 +108,8 @@ def test_partition_html_accepts_an_html_str():
     assert len(elements) > 0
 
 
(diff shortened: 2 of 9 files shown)


lawrence-u10d:
Closing — folded the substantive pieces into review feedback on #4388 so the fix lands in one place.

Issue #3365, opened 2024-07-09
bug(json): partition_json() does not preserve original element_id of elements


scanny:
**Summary**
The contract of `partition_json()` is to "rehydrate" the JSON elements serialized to a JSON array of element objects. However, it changes  the `element_id` and certain metadata fields from their original values.

**To Reproduce**
```python
file_path = example_doc_path("simple.json")
original_elements = elements_from_json(file_path)
partitioned_elements = partition_json(file_path)

assert elements_to_dicts(partitioned_elements) == elements_to_dicts(original_elements)
```
produces:
```diff
    [
        {
  -         'element_id': 'a06d2d9e65212d4aa955c3ab32950ffa',
  +         'element_id': 'dbc05298f7937a62027af643bd1c3c87',
            'metadata': {
                'category_depth': 0,
  -             'file_directory': 'unstructured/example-docs',
  +             'file_directory': '/Users/scanny/src/unstructured/example-docs',
  -             'filename': 'simple.docx',
  ?                                 ^ ^^
  +             'filename': 'simple.json',
  ?                                 ^^ ^
  -             'filetype': 'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
  +             'filetype': 'application/json',
                'languages': [
                    'eng',
                ],
  -             'last_modified': '2024-07-06T16:44:51',
  ?                                        ^ ^  ^^^^^
  +             'last_modified': '2024-07-08T23:06:02',
  ?                                        ^ ^^^^  ^^
            },
            'text': 'These are a few of my favorite things:',
            'type': 'Title',
        },
```

**Expected behavior**
Because `partition_json()` is the mechanism for step-wise processing of documents using the API, (like `elements_from_json()` is using `unstructured` open-source directly), identifiers and metadata should be unchanged from their serialized state. Note that step-wise processing includes chunking as a separate step, perhaps after filtering elements from the original payload or enhancing metadata.

**Additional context**
- Much of the problem appears to be caused by double-post-processing. `partition_json()` uses the same `@process_metadata()` and `@add_metadata_with_filetype()` decorators that other partitioners do, but since it is not actually a partitioner, those metadata and id post-processing steps are not needed and causes these and perhaps other unwelcome behaviors.

PR #4391, opened 2026-07-09
feat: partition arbitrary valid JSON and NDJSON files

simoncoombes:
## Summary

`partition_json()` and `partition_ndjson()` currently accept **only serialized Unstructured element output** - any other valid JSON is rejected with `"Schema does not match the Unstructured schema"` (and an array of scalars crashes with a raw `AttributeError`). This PR makes both partitioners **two-mode**:

- **Rehydration (unchanged):** a payload of serialized Unstructured elements is rehydrated back into its elements, exactly as before.
- **Arbitrary JSON (new):** any other valid JSON/NDJSON is converted to `Text` elements containing the pretty-printed JSON, instead of raising.

It also fixes a file-type detection bug this feature exposed: a **compact single-line JSON object** was misrouted to `FileType.NDJSON` and could never reach `partition_json` at all.

## Behavior change

| Input | Before | After |
|---|---|---|
| Object `{"customer": "Acme", ...}` | `ValueError: JSON cannot be partitioned. Schema does not match…` | 1 `Text` element, pretty-printed |
| Array of objects `[{"id":"one"},{"id":"two"}]` | same `ValueError` | one `Text` per object, array order |
| Array of scalars `[1,2,3]` | crash: `AttributeError: 'int' object has no attribute 'get'` | 1 `Text` with the whole array |
| Mixed array / top-level scalar | `ValueError` / crash | 1 `Text` |
| Mixed element-shaped + arbitrary array | partial rehydrate, arbitrary items silently dropped | whole array as arbitrary JSON - nothing dropped |
| Compact single-line object, `.json` file | misrouted to NDJSON → rejected | detected as `FileType.JSON` → 1 `Text` |
| Arbitrary NDJSON (one record per line) | `ValueError` | one `Text` per line, line order |
| Serialized element JSON / NDJSON (arrays / line-per-element) | rehydrates | rehydrates (unchanged, full regression suite green) |
| Element-shaped payload with corrupt contents (bad `metadata.coordinates`, non-gzip `orig_elements`) | raw `ValueError`/`binascii.Error`/`zlib.error` leaks | chained `ValueError("Payload resembles serialized Unstructured elements but could not be reconstructed: …")` |
| Serialized `TableChunk` elements (chunked output with split tables) | silently dropped by `elements_from_dicts()` | rehydrate as `TableChunk`; whole chunked payloads round-trip and can feed `reconstruct_table_from_chunks()` (#4291) |
| `{}` (JSON route) | error | one `Text` containing `{}` (per the output contract) |
| `[]` / empty string (JSON route) | `[]` / `[]` | `[]` |
| `{}` / `[]` as an NDJSON line | error | `Text("{}")` / `Text("[]")` (a line is a record) |
| Deeply nested payload (any depth) | `RecursionError` escapes | `ValueError` (`"Not a valid json"` / `"…nested too deeply…"`) |
| Malformed JSON `[{"hi":"there"}]]` | `ValueError("Not a valid json")` | unchanged |
| Malformed NDJSON line | `ValueError("Not a valid ndjson")` | unchanged |

## Design notes

- **Mode selection:** an explicit shape predicate (`is_element_shaped_dict` in `partition/common/json_partitioning.py`) - a list rehydrates only when every item is a dict with a recognized `str` `type`, the type's required field (`str` `text` / `bool` `checked` for CheckBox), and dict-or-absent `metadata`. Branches are exclusive; no exception-based control flow. Prefix/schema pre-gates are removed from `partition_json`, `partition_ndjson`, and `auto.py`; `is_json_processable`/`is_ndjson_processable` are **deprecated** (DeprecationWarning naming the replacement) but keep working for downstream importers. `unstructured.file_utils.ndjson.loads/load` are intentionally retained undeprecated as generic utilities.
- **Documented limitation** (pinned by tests): an array whose items *all* look like serialized elements rehydrates rather than being treated as arbitrary JSON. An element-shaped payload whose field contents fail rehydration raises a chained `ValueError` - loud, never a leaked low-level error.
- **`staging/base.py` deliberately not modified:** the shape predicate rejects payloads like `{"type": "Title"}` (no `str` text) before `elements_from_dicts` is ever called, so they partition as arbitrary JSON; hardening `item["text"]` → `.get()` in staging instead would silently rehydrate customer dicts as empty elements.
- **Filetype disambiguation:** whole-payload `json.loads` success → `FileType.JSON`; else ≥2 newline-delimited JSON values → `NDJSON`. The probe is **bounded to 1 MiB**, distinguishes an exact-bound-size payload from a truncated one, restores the file position (a `detect_filetype(file=f)` → `partition_json(file=f)` sequence works on the same handle), and treats `RecursionError` as a parse failure. For payloads exceeding the bound, one or more complete parsing lines classify as NDJSON (first-line semantics for oversized records); the residual degradation is an NDJSON file whose *first* record has no newline inside the bound, which classifies as JSON. One intentional flip: a *one-line serialized-element object* previously rehydrated via the NDJSON route; it now partitions as arbitrary JSON (rehydration applies only to arrays).
- **Output contract:** `pretty_json_text()` - `json.dumps(value, indent=2, sort_keys=True)`; stable, diffable output; `sort_keys` alphabetizes source field order (commented at the definition as the knob to revisit). NDJSON is strictly one `Text` per line; the only empty-container divergence is `[]` (an NDJSON line yields `Text("[]")`, a JSON-mode `[]` document yields no elements).
- **Deferred by design:** per-field metadata / JSONPath addressing and structure-aware tree walking. `elements_from_arbitrary_value()` (`partition/common/json_partitioning.py`) is the single swap-point for a future walker.

## TableChunk rehydration - intent

Review of this branch surfaced that serialized `TableChunk` dicts could not be deserialized at all: `TYPE_TO_TEXT_ELEMENT_MAP` has no `TableChunk` entry, so `elements_from_dicts()` silently dropped them on `main`, and with this branch's shape predicate one TableChunk flipped an entire serialized payload to arbitrary-JSON `Text`. The intent of the fix commit is narrow and explicit:

- **Complete #4291's design, using only existing mechanisms.** `reconstruct_table_from_chunks()` (added in #4291) filters `isinstance(e, TableChunk)` and its docstring states reconstruction "can be called on user-provided/deserialized chunks" - but no deserializer could produce a `TableChunk` until now. The reconstruction metadata (`table_id`, `chunk_index`, `is_continuation`, `num_carried_over_header_rows`) already round-trips via `ElementMetadata`; only the element-class dispatch was missing.
- **Special-case, not a map entry - deliberately.** `TYPE_TO_TEXT_ELEMENT_MAP` also feeds the COCO category vocabulary (`convert_to_coco` derives positional category ids from its keys), so adding TableChunk there would renumber existing category ids on every export, even for data with no TableChunks. Instead `elements_from_dicts()` special-cases `"TableChunk"` exactly like the existing `CheckBox` special case. Verified strictly additive: a 40-dict all-type sweep shows byte-identical output vs the pre-fix function for every non-TableChunk payload; `documents/elements.py` is untouched; COCO ids unchanged.
- **No new downstream exposure.** `type: "TableChunk"` records already flow to destinations today whenever live chunking splits a table (library output, ingest, hosted API responses); this changes only the read-back side.

## Testing

- Combined suites (partition json/ndjson/shared-predicate, filetype, staging serde, chunking dispatch/reconstruct): **691 passed, 11 skipped (pre-existing optional-dep skips), 1 xfailed**; `test_auto.py -k "json or ndjson"` **7 passed, 1 xfailed** (the #3365 strict-xfail - unaffected and kept).
- Full rehydration regression set green (round-trips, chunking, `last_modified`, metadata stamping), including end-to-end NDJSON rehydration through `partition()`.
- Five pre-existing tests intentionally flip expected behavior (`{}` no longer raises, empty NDJSON container lines emit `Text`, one-record payloads route JSON) - each renamed to describe the new behavior.
- Exact pretty-printed literals are confined to one canonical test per output shape; other tests assert structurally (element type + content containment) so a future formatting change doesn't invalidate dozens of assertions.
- TableChunk: end-to-end chunked-table round-trip (chunk -> serialize -> `partition_json` -> byte-identical re-serialization -> `reconstruct_table_from_chunks` returns the table), NDJSON TableChunk lines, predicate accept/reject, and staging serde tests. The exact-bound-size detection probe is pinned by a test proven to fail if the probe is reverted.
- Coverage includes `text=`/`file=`/`filename=` routes, corrupt-payload errors via `file=`, deep-nesting through both partitioners, boundary-size (1 MiB ± 1) disambiguation, detect-then-partition on one file handle, and chunking over arbitrary JSON **and** NDJSON output.
- New business-neutral fixtures: `example-docs/arbitrary-records.json`, `single-line-object.json`, `arbitrary-records.ndjson`.
- Hardened by two independent review passes (adversarial correctness + convention/nit pass), each finding verified by reproduction before being fixed.


simoncoombes:
**Note on the latest commit (`200c0e46`, TableChunk rehydration) - intent for reviewers:**

While validating this branch we found that serialized `TableChunk` elements (chunker output for oversized tables) have never been deserializable: `TYPE_TO_TEXT_ELEMENT_MAP` has no entry for them, so `elements_from_dicts()` silently dropped them - and with this PR's two-mode dispatch, a single TableChunk dict flipped a whole serialized payload to pretty-printed `Text`.

The fix intentionally uses only existing mechanisms:

1. `elements_from_dicts()` special-cases `"TableChunk"` exactly like the existing `CheckBox` special case. We deliberately did **not** add it to `TYPE_TO_TEXT_ELEMENT_MAP`: that map also feeds the COCO category vocabulary in `convert_to_coco`, where positional ids are derived from its keys - a map entry would renumber 5 existing category ids on every export, even for data containing no TableChunks. The special case is strictly additive (verified with an all-type sweep: byte-identical output for every non-TableChunk payload; `documents/elements.py` untouched).
2. This completes #4291: `reconstruct_table_from_chunks()` explicitly expects "user-provided/deserialized chunks" but had no deserialization path that could produce a `TableChunk`. Its metadata (`table_id`, `chunk_index`, `is_continuation`) already round-trips - only the class dispatch was missing. Rehydrated chunks now reconstruct identically to never-serialized ones.
3. No new downstream exposure: `type: "TableChunk"` records already reach destinations today from live chunking; this only fixes reading them back.

cc @qued as the author of #4291 in case you want to sanity-check the reconstruction interplay.

simoncoombes:
## End-to-end validation evidence

Validated on an internal single-node platform deployment running this branch's library on the partition API, with the companion plugin change on the ingestion path.

**Shape suite (60 fixtures, 6 categories):** pretty-printed/compact objects (up to 4.8 MB single-line), arrays (3 to 4,000 records), NDJSON (3 to 9,000 lines, crossing the 1 MiB detection bound), serialized-element files, and deliberately-malformed inputs. Result: 60/60 adjudicated PASS - 15,046 elements compared against this branch's expected output at type + full-text level; every valid file partitioned with order and content preserved; every malformed file failed per-file with a clear error and the batch completed.

**Metadata suite (11 fixtures):** all 37 recognized element types round-trip; parent_id references stay consistent across re-hashed element ids (#3365 behavior); links/emphasis, languages, page metadata, ad-hoc metadata keys, CheckBox states, and image_base64 payloads survive byte-exact. Serialized chunked output containing split tables rehydrates fully - `reconstruct_table_from_chunks()` rebuilds the original table byte-identically (text and html) from output that made the full round trip, exercising the TableChunk deserialization added in this PR end-to-end.

**Known items observed (all pre-existing or documented):** element ids are re-hashed on re-partition (#3365); metadata.coordinates is stripped unless requested via the API's coordinates parameter (verified preserved when requested); BOM-prefixed JSON still raises (documented deferral in this PR).

Full fixture/output pack and evaluation reports are linked from the internal companion PR.

simoncoombes:
## Rerun validation (2026-07-10, stack @ library `38d7ce43` + plugin `10128b96`)

Re-ran the full E2E suite on the SND after the empty-object contract fix (`{}` -> `Text("{}")`) and image refresh.

- **Shape suite: 61/61 PASS** (added an empty-object fixture this run). `{}` now returns one `Text` element containing `{}`, verified end-to-end through the platform - confirms the product-contract fix is live. 59 exact type+text matches; 2 NDJSON files match the one-Text-per-line contract exactly; the 8 deliberately-malformed files fail per-file with the job completing; empty/whitespace files emit nothing without error. Order/content preserved at scale (4,000-element array, 9,000-line NDJSON).
- **Metadata suite: 11/11 PASS** at full element-dict level - all 37 element types, parent_id referential integrity across re-hashed ids, TableChunks + carried headers, orig_elements decoded-equal, ad-hoc metadata keys, CheckBox states, byte-exact image_base64. Table reconstruction proven functionally.

SudSampath:
Product scope approved for this approach. The two-mode `partition_json()` behavior matches the PRD: serialized Unstructured element arrays remain the rehydration path, and valid arbitrary JSON falls back to readable `Text` elements.

I am also approving the explicit deviations: NDJSON in v1, deprecation warnings for the old processability helpers, and the documented one-record serialized-element routing limitation through `partition()` / `detect_filetype()`.

Before merge, please address two small items:
- Reject `NaN` / `Infinity` with `parse_constant`, since Python accepts those by default but they are not strict JSON.
- Fix the stale docstring in `json_partitioning.py` that describes `{}` behavior backwards.

With those fixed, I am comfortable with this as the library release basis for 0.25.0.

simoncoombes:
Thanks @SudSampath - scope approval much appreciated. Both items done in `3507d14f`:

- **NaN / Infinity / -Infinity** now reject as malformed. Added a shared strict loader used by both partitioners that passes `parse_constant` to raise, so it surfaces as the normal `Not a valid json` / `Not a valid ndjson` `ValueError` rather than a new error path. Covers nested (`{"x": NaN}`, `[1, Infinity]`) and NDJSON lines too, with parametrized tests on both.
- **Docstring** fixed - `{}` yields `Text("{}")` in both modes now, so the only mode difference left is empty arrays (`[]` line -> `Text("[]")` vs `[]` document -> no elements).

One design call worth flagging: I put the strict parse in the two partitioners (the validation gates) rather than in `detect_filetype`. A `NaN` file still gets rejected, just by the partitioner instead of at detection, so behaviour is correct and the change stayed minimal. Happy to extend it into detection too if you'd rather the two be symmetric.

All affected suites green. Good to go as the 0.25.0 basis from my side once CI's clean.

simoncoombes:
@cubic-dev-ai please re-review

badGarnet:
**Findings**

1. **High: `.ndjson` intent is lost when callers provide generic `application/json`**

Why: uploads often provide `Content-Type: application/json` even for `.ndjson`. The detector accepts the content type first, then `_disambiguate_json_file_type()` ignores the `.ndjson` extension and prefers any whole-payload JSON parse. A one-record `.ndjson` file such as `{"type":"Title","text":"x"}\n` is then routed to `partition_json()` and becomes arbitrary `Text` instead of rehydrating via `partition_ndjson()`.

Relevant code, `unstructured/file_utils/filetype.py:248`:

```python
if result_file_type == FileType.JSON:
    result_file_type = self._disambiguate_json_file_type
```

and `unstructured/file_utils/filetype.py:331`:

```python
try:
    json.loads(file_text)
    return FileType.JSON
```

Concrete fix: in the JSON/NDJSON disambiguation path, use an explicit `.ndjson` extension or `application/x-ndjson` as a tie-breaker before whole-document JSON wins. Add a regression test for one-record `.ndjson` with `content_type=application/json`.

2. **Medium: new scalar JSON/NDJSON support is not reachable through `partition()` in common detection paths**

Why: `partition_json()` and `partition_ndjson()` now accept scalars, but `detect_filetype()` still has old assumptions that JSON starts with `{` or `[` and that NDJSON lines are object-shaped. Valid scalar NDJSON in a `.ndjson` file can be detected as `TXT`, so `partition()` will not call the NDJSON partitioner.

Relevant code, `unstructured/partition/ndjson.py:90`:

```python
for line in file_text.splitlines():
    if not line.strip():
        continue
    values.append(loads_strict_json(line))
```

but detection still rejects scalar JSON starts in `unstructured/file_utils/filetype.py:921`:

```python
if text_head.lstrip()[0] not in "[{":
    return False
```

Concrete fix: update `_TextFileDifferentiator` to special-case `.ndjson` before the generic `text/plain` fallback. Add tests for scalar `.ndjson`, scalar `.json`, and scalar NDJSON with `metadata_file_path`.

3. **Low: detector uses permissive `json.loads()` while partitioners use strict JSON parsing**

Why: the PR rejects `NaN`, `Infinity`, and `-Infinity` in `partition_json()` / `partition_ndjson()`, but `_disambiguate_json_file_type()` still uses Python’s permissive `json.loads()`. That makes `detect_filetype()` inconsistent with the new validity contract.

Relevant code, `unstructured/partition/common/json_partitioning.py:45`:

```python
return json.loads(text, parse_constant=_reject_json_constant)
```

but `unstructured/file_utils/filetype.py:323` still does:

```python
for line in lines:
    json.loads(line)
```

Concrete fix: use the same strict parser semantics in disambiguation, ideally from a shared lower-level utility to avoid awkward dependency direction.

simoncoombes:
All review feedback is addressed and the branch is green:

- @SudSampath's two asks are in: `NaN`/`Infinity`/`-Infinity` now reject via `parse_constant` (top-level, nested, and NDJSON lines), and the `{}` docstring is corrected.
- @badGarnet's findings are fixed: `.ndjson`/`application/x-ndjson` now wins over the whole-parse rule; scalar JSON/NDJSON is reachable through `partition()`; and `loads_strict_json` moved to `unstructured.utils` so the detector shares strict parsing without an import cycle. The follow-on cubic nit (a dead `application/x-ndjson` operand) is gone too.

CI is green and cubic is at 0 issues. Guardrails re-verified: serialized-element arrays/TableChunks still rehydrate, compact-object routing unchanged, malformed input still raises. Could I get a formal approve so this can merge as the `0.25.0` release basis? The `unstructured-api` floor bump lands as a fast-follow once `0.25.0` is published.

unstructured/partition/json.py
```diff
@@ -66,22 +88,21 @@ def partition_json(
...
-                "JSON cannot be partitioned. Schema does not match the Unstructured schema.",
-            )
-    except json.JSONDecodeError:
+        value = loads_strict_json(file_text)
+    except (json.JSONDecodeError, RecursionError):
         raise ValueError("Not a valid json")
 
+    if isinstance(value, list) and value and all(is_element_shaped_dict(i) for i in value):
+        # -- Branch A: rehydrate serialized Unstructured elements --
+        elements = rehydrate_elements(value)
```
badGarnet, line 101: non-blocking: wonder if we should capture ValueError when `value` resembles elements but can't be parsed then fallback to `elements_from_arbitrary_value` method.

badGarnet (approved):

PR #4383, opened 2026-06-30
feat(pdf): flag invisible text in metadata

Success6666:
## Summary
- add `contains_invisible_text` metadata for PDF text objects that include render-mode-3 invisible characters
- keep the invisible-text flag scoped to each split pdfminer text snippet so visible snippets from the same container are not overflagged
- preserve optional metadata semantics during chunk consolidation: absent signals stay absent, explicit `False` stays `False`, and any `True` marks the chunk
- preserve the flag when list items are combined and when elements are chunked
- add regression coverage for pdfminer detection, fast PDF partitioning, snippet-level scoping, and chunk metadata consolidation

Refs #4377

## Notes
This is additive: extracted text output is unchanged, and the new metadata gives downstream RAG pipelines a signal they can filter, quarantine, or surface in audit logs.

## Tests
- `uv run --python 3.12 --group test --extra pdf python -m pytest test_unstructured\partition\pdf_image\test_pdfminer_processing.py::test_text_is_embedded test_unstructured\partition\pdf_image\test_pdfminer_processing.py::test_text_contains_invisible_text test_unstructured\partition\pdf_image\test_pdfminer_processing.py::test_split_pdfminer_text_by_paragraph_keeps_invisible_text_scoped_to_snippet test_unstructured\partition\pdf_image\test_pdf.py::test_partition_pdf_fast_marks_invisible_text test_unstructured\chunking\test_base.py::Describe_Chunker::it_forms_ElementMetadata_constructor_kwargs_by_applying_consolidation_strategies test_unstructured\chunking\test_base.py::Describe_Chunker::it_flags_invisible_text_when_any_element_contains_it test_unstructured\chunking\test_base.py::Describe_Chunker::it_omits_invisible_text_when_no_element_reports_it test_unstructured\chunking\test_base.py::Describe_Chunker::it_keeps_invisible_text_false_when_all_reporting_elements_clear_it test_unstructured\documents\test_elements.py::DescribeElementMetadata::it_can_find_the_consolidation_strategy_for_each_of_its_known_fields -q`
- `uv run --python 3.12 --group lint ruff check unstructured\chunking\base.py unstructured\partition\pdf.py unstructured\partition\pdf_image\pdfminer_processing.py test_unstructured\chunking\test_base.py test_unstructured\partition\pdf_image\test_pdfminer_processing.py`
[...]

PR #4382, opened 2026-06-30
feat: add max_page to chunk_by_title and remove multipage_sections

issahammoud:
## Summary

Adds a `max_page` parameter to `chunk_by_title()` that gives callers a hard upper bound on how many pages a single chunk may span. `max_page` is a strict superset of the existing `multipage_sections` flag: `max_page=None` (default) matches `multipage_sections=True`, and `max_page=1` matches `multipage_sections=False` — plus the new ability to express any page-count limit (`max_page=2`, `max_page=3`, …) that the old binary flag could not.

`multipage_sections` is kept for backward compatibility but deprecated. It emits a `DeprecationWarning` and will be removed in a future release.

### Motivation

`chunk_by_title` supported character/token-based size limits and an all-or-nothing page boundary toggle (`multipage_sections`). There was no way to say _"allow cross-page sections, but cap each chunk at N pages"_. This is a common need in RAG pipelines where page-level context windows matter — e.g. a retriever that attaches page images to each chunk should not pull more than a fixed number of page images per chunk.

### Behaviour

When `max_page=N` is specified, the chunker tracks the page where each chunk began and starts a new chunk as soon as an element's page number is `N` or more pages past that start page:

```python
chunks = chunk_by_title(elements, max_page=2)
# → each chunk spans at most 2 pages

chunks = chunk_by_title(elements, max_page=1)
# → a new chunk starts on every page change
```

- **Additive constraint** — works alongside all existing `by_title` boundaries (title, size, token).
- **Title-aware reset** — a `Title` element resets the page counter without double-firing, so each title-delimited section's page span is measured independently.
- **Hard limit, not a hint** — `PreChunk.can_combine()` also enforces `max_page`, so `PreChunkCombiner` cannot silently re-merge pre-chunks that were split by the page boundary.
- **No page-number, no problem** — elements without `metadata.page_number` are assumed to continue the current page.
- **Validation** — `max_page < 1` raises `ValueError`.

### `multipage_sections` deprecation

`multipage_sections` is kept but deprecated. Existing call sites continue to work without modification; they will receive a `DeprecationWarning` pointing to the `max_page` equivalent.

| Old | New | Notes |
|---|---|---|
| `multipage_sections=True` (default) | omit `max_page` (default) | No code change needed |
| `multipage_sections=False` | `max_page=1` | One-word change |

When both `multipage_sections` and `max_page` are provided, `max_page` takes priority silently — no error is raised.

### Example

```python
from unstructured.chunking.title import chunk_by_title
[...]

PR #4402, opened 2026-07-15
fix: use deepcopy instead of shallow copy when stripping orig_elements

chuenchen309:
## Problem

`_Chunker._orig_elements` copies each element before clearing its `metadata.orig_elements`, with an explicit comment stating this is because "these elements don't belong to us (the user may have downstream purposes for them)." But `copy.copy()` is a **shallow** copy — the copy's `.metadata` is the *same object* as the original's `.metadata` — so `orig_element.metadata.orig_elements = None` mutates the caller's own element in place, exactly contradicting the stated intent.

Sibling `_TableChunker._orig_elements` already does this correctly with `copy.deepcopy(self._table)`.

## Fix

Use `copy.deepcopy` instead of `copy.copy`, matching the already-correct sibling implementation.

## Testing

- Added a regression test confirming a caller's element still has its original `.metadata.orig_elements` value after `_Chunker._orig_elements` is computed.
- Confirmed red→green: reverting only `base.py` reproduces the caller mutation (`assert None == [inner]`); reapplying passes.
- Full `test_unstructured/chunking/` suite: 334 passed.
- `ruff check` — clean.
- xref: no open PR touches this shallow/deep-copy site (checked #4392, #4383, #4382, #4197 — none overlap; #4382/#4197 have unrelated `include_orig_elements` matches, a different parameter name).

## AI-Generated disclosure

I used an AI coding assistant (Claude) to help identify this bug and draft the fix. I independently reproduced the caller mutation with a throwaway script before writing the test, cross-checked against the already-correct sibling `_TableChunker` implementation, and ran the full local test suite before preparing this change.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4402?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - **Update README.md**: readme-only changes; added a link to Unstructured Pipelines to the README. No library behavior changes.
+- **`Chunker._orig_elements` no longer mutates the caller's original elements**: `copy.copy(e)` created a shallow copy whose `.metadata` was the *same object* as the original element's `.metadata`; clearing `orig_elements` on the copy (`orig_element.metadata.orig_elements = None`) mutated the caller's original element in place, contradicting the code's own comment that these elements "don't belong to us." Now uses `copy.deepcopy(e)`, matching the sibling `_TableChunker._orig_elements`'s existing correct pattern.
 
 ## 0.25.0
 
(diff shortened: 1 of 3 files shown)


zeweihan:
@/tmp/gh_comment_unstructured_4402.md

chuenchen309:
Heads up — your last comment came through as just a literal `@/tmp/gh_comment_unstructured_4402.md` path, so the intended content didn't attach. Happy to take a look if you re-post it.

PR #4406, opened 2026-07-16
Guard clean_ordered_bullets and extract_ordered_bullets against empty text

santhreal:
`clean_ordered_bullets` and `extract_ordered_bullets` call `text.split()` then index `text_sp[0]` without a guard, so an empty or whitespace-only string raises `IndexError: list index out of range`. Both are public cleaning bricks, and the sibling `clean_bullets` already returns the text unchanged when there is no match.

This returns the input unchanged (`clean_ordered_bullets`) / `(None, None, None)` (`extract_ordered_bullets`) for empty input, matching that graceful-degradation contract. Added empty and whitespace-only cases to the existing parametrized tests for both functions.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4406?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test_unstructured/cleaners/test_core.py
+++ b/test_unstructured/cleaners/test_core.py
@@ -59,6 +59,9 @@ def test_clean_bullets(text, expected):
         ("1..2.3 four", "1..2.3 four"),
         ("Fig. 2: The relationship", "Fig. 2: The relationship"),
         ("23 is everywhere", "23 is everywhere"),
+        ("", ""),
+        ("   ", "   "),
+        ("\n\t", "\n\t"),
     ],
 )
 def test_clean_ordered_bullets(text, expected):
--- a/test_unstructured/cleaners/test_extract.py
+++ b/test_unstructured/cleaners/test_extract.py
@@ -100,6 +100,9 @@ def test_extract_us_phone_number(text, expected):
         ("1..2.3 four", (None, None, None)),
         ("Fig. 2: The relationship", (None, None, None)),
         ("23 is everywhere", (None, None, None)),
+        ("", (None, None, None)),
+        ("   ", (None, None, None)),
+        ("\n\t", (None, None, None)),
     ],
 )
 def test_extract_ordered_bullets(text, expected):
--- a/unstructured/cleaners/core.py
+++ b/unstructured/cleaners/core.py
@@ -59,6 +59,8 @@ def clean_ordered_bullets(text) -> str:
     a.b This is a very important point -> This is a very important point
     """
     text_sp = text.split()
+    if not text_sp:
+        return text
     text_cl = " ".join(text_sp[1:])
     if any(["." not in text_sp[0], ".." in text_sp[0]]):
         return text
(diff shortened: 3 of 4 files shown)


zeweihan:
Nice catch — and the fix matches `clean_bullets`'s existing graceful-degradation contract, which is the right layer for it.

One regulated-domain angle worth raising beyond the IndexError itself: the sibling behavior `clean_bullets` returns the input **unchanged** on no-match is already a silent pass-through, and that's the more dangerous failure mode for contract corpora. Legal documents lean heavily on ordered/lettered enumerations (`(a)`, `(b)`, `(i)`, `(ii)`, `1.1.1`, `Section 4.2.3`) where the bullet **prefix carries operative meaning** (subordination level, cross-reference anchor). When `clean_ordered_bullets` matches a malformed prefix and returns text unchanged, the resulting chunk still indexes fine and retrieves fine — but the structural signal ("this is a sub-clause of 4.2") has been silently dropped. The IndexError is loud (good); the no-match-unchanged path is quiet and compounds downstream.

Two things worth considering before merge:

1. **The empty-input case is a symptom, not the root.** A `None`/empty string reaching `clean_ordered_bullets` means an upstream brick produced or passed through empty text — typically a splitter edge, a table cell with no content, or a header line that OCR'd to whitespace. Returning `None`/unchanged here is correct *for this brick*, but it propagates an empty chunk that looks healthy to the indexer. Consider logging (not throwing) when input is empty so the ingestion pipeline has a signal that *something upstream* produced nothing — otherwise empty chunks silently reach the store and contribute to the "silent zero-content" class of bugs.

2. **Coverage of the no-match path in tests.** The PR adds empty + whitespace cases (good). It would also be worth adding a no-match-but-non-empty case (e.g. a bullet-like string that doesn't match the regex) to lock in *which* contract is being honored — "return unchanged" vs "return None" — since `clean_ordered_bullets` (unchanged) and `extract_ordered_bullets` (None tuple) diverge here, and downstream callers that switch between them need to know the empty/no-match shape is *intentional* and not a regression.

For document-AI consumers: the invariant worth tracking is `cleaning_result{text_in/text_out/changed/no_match}` per brick — turns "did cleaning silently do nothing" from infer-by-diffing-the-corpus into a queryable property. Same shape as the text-coverage / formula-render-status companions in other doc-parsing pipelines.

(Working on AI Workdeck, where legal-document cleaning pipelines run before retrieval — empty-input propagation and silent no-match pass-through are both recurring sources of "the chunk looks fine but the structure is gone" issues.)

santhreal:
Thanks. The no-match contract is already pinned on identical inputs: `23 is everywhere` returns the string unchanged in test_core and `(None, None, None)` in test_extract, so both shapes are locked. On logging empty input, I'd keep it out of the pure string cleaner to avoid a log on every empty cell; happy to add a debug log if you'd prefer it here.

PR #4408, opened 2026-07-17
perf: avoid spaCy statistical pipeline for word tokenization

KRRT7:
## Summary

Use spaCy's tokenizer-only `make_doc()` path in `word_tokenize()` instead of running the complete statistical pipeline.

Word tokenization only consumes `token.text`, so running the tagger and dependency parser adds substantial work without affecting the result.

This PR also:

- Preserves the existing oversized-input truncation behavior.
- Adds regression tests comparing tokenizer-only output with the full pipeline.
- Adds a representative multi-format benchmark with exact output-parity checks.

## Performance

Benchmarked on an 8-vCPU AMD EPYC host with 15 GiB RAM and Python 3.12.

The benchmark covers large TXT, HTML, DOCX, and PPTX documents through their format-specific partitioners. The spaCy model is warmed before measurement, while token, sentence, and POS result caches are cleared before every measured iteration. This measures the algorithmic improvement rather than repeated-document cache hits.

Each result is the median of three iterations.

| Document | `upstream/main` | This PR | Improvement |
|---|---:|---:|---:|
| 1,225-page War and Peace TXT | 57.393s | 49.215s | 14.2% |
| 230-page 10-K HTML | 1.129s | 1.044s | 7.5% |
| 872-page handbook DOCX | 23.553s | 20.628s | 12.4% |
| 369-page science presentation PPTX | 3.726s | 2.350s | 36.9% |
| **Aggregate median** | **85.801s** | **73.238s** | **14.6% / 1.17x** |

Output parity was exact for every document, covering:

- 11,751 TXT elements
- 314 HTML elements
- 4,759 DOCX elements
- 5,116 PPTX elements

The benchmark verifies every element's concrete type and text using a stable SHA-256 fingerprint.

## Benchmark reproduction

```bash
# On upstream/main
uv run --no-sync python scripts/performance/benchmark_text_partition.py \
  /tmp/main.json

# On this branch
uv run --no-sync python scripts/performance/benchmark_text_partition.py \
  /tmp/candidate.json --compare /tmp/main.json
```

The benchmark defaults to three iterations over the representative four-document matrix. Custom workloads can be supplied by repeating `--input`:

```bash
uv run --no-sync python scripts/performance/benchmark_text_partition.py \
  /tmp/custom.json \
  --input example-docs/example-10k-230p.html \
  --input example-docs/handbook-872p.docx
```

## Testing

- 121 focused NLP, text-classification, and benchmark tests pass.
[...]

PR #4409, opened 2026-07-17
perf: skip dependency parser during POS tagging

KRRT7:
## Stack dependency

This draft is stacked on #4408 and should be reviewed and merged after that PR. Once #4408 lands, this branch will be rebased onto `main` so the final diff contains only the parser-free POS change.

## Summary

Skip spaCy's dependency parser in `pos_tag()` while retaining the `tok2vec` and `tagger` components that produce `token.tag_`.

The dependency parser runs after the tagger, and `pos_tag()` immediately discards its output. Disabling it removes unnecessary statistical work without changing returned POS tags.

## Performance

Benchmarked against the tokenizer-only parent branch on an 8-vCPU AMD EPYC host with 15 GiB RAM and Python 3.12. The spaCy model was warmed, NLP result caches were cleared before every iteration, and each full-matrix result is the median of three runs.

| Document | Tokenizer-only parent | This PR | Incremental change |
|---|---:|---:|---:|
| 1,225-page War and Peace TXT | 48.986s | 47.237s | 3.6% faster |
| 230-page 10-K HTML | 1.038s | 1.048s | effectively neutral |
| 872-page handbook DOCX | 20.582s | 20.128s | 2.2% faster |
| 369-page science presentation PPTX | 2.327s | 2.586s | noisy short workload |
| **Aggregate median** | **72.932s** | **70.998s** | **2.7% faster / 1.03x** |

A seven-iteration follow-up on the short workloads measured HTML 1.5% faster and PPTX 1.6% slower, confirming both are effectively neutral. The repeatable gains are in the text-heavy TXT and DOCX paths.

Compared with `upstream/main`, the combined tokenizer-only and parser-free-POS stack improves the representative aggregate from 85.801s to 70.998s: approximately **17.3% faster / 1.21x**.

Output parity was exact for all four documents.

## Testing

- 128 focused NLP, text-classification, and benchmark tests pass.
- POS tags are compared with and without the dependency parser across six representative inputs.
- A regression test verifies that `pos_tag()` disables the parser.
- Existing tokenizer parity, truncation, caching, and benchmark tests continue to pass.
- Ruff lint and formatting checks pass.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4409?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 - **Speed up word tokenization**: `word_tokenize()` now uses spaCy's tokenizer-only path instead of running the statistical tagger and dependency parser when only token text is needed. This preserves token and element output while reducing representative text-heavy partition time by 14.6% across TXT, HTML, DOCX, and PPTX benchmarks.
+- **Speed up POS tagging**: `pos_tag()` now skips spaCy's dependency parser, whose output is not used when returning token tags. POS tags and partition output remain unchanged, while representative text-heavy partition time improves by a further 2.7% over the tokenizer-only path.
 
 ## 0.25.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
-## 0.25.1
+## 0.25.2-dev0
 
 ### Enhancements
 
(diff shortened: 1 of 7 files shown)


cragwolfe:
LGTM. will wait for #4408 to land before formal approval

PR #4418, opened 2026-07-21
perf(html): carry parsed fragments through merging

KRRT7:
## Summary

Carry parsed ontology fragments alongside HTML elements during ontology conversion so adjacent inline elements can be checked for mergeability without reparsing their serialized HTML.

## Performance

The inline merge path now parses each element fragment once and reuses the parsed representation across merge checks, avoiding repeated BeautifulSoup parsing for long HTML element streams.

## Compatibility

The standalone `can_unstructured_elements_be_merged()` helper retains its existing behavior and parsing interface.

## Testing

- 339 HTML tests passed
- Ruff checks passed

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Enhancements
+
+- **Avoid repeated HTML fragment parsing during inline merging**: the HTML ontology conversion now carries parsed fragments forward while combining adjacent elements.
+
 ## 0.25.1
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## 0.25.2-dev0
 
 ### Enhancements
 
--- a/unstructured/__version__.py
+++ b/unstructured/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.25.1"  # pragma: no cover
+__version__ = "0.25.2-dev0"  # pragma: no cover
--- a/unstructured/partition/html/transformations.py
+++ b/unstructured/partition/html/transformations.py
@@ -217,11 +217,15 @@ def combine_inline_elements(
 
         current_element, current_depth, current_ontology_elements = current
         next_element, next_depth, next_ontology_elements = nxt
-        if _can_ontology_elements_be_merged(
-            current_ontology_elements,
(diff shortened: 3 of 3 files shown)


cragwolfe:
### Severity: High — the optimization eagerly reparses and retains every non-layout subtree, including content that can never be merged

**Location:** `unstructured/partition/html/transformations.py:172-178`, `unstructured/partition/html/transformations.py:220-232`, `unstructured/partition/html/transformations.py:293-298`

**Description:** `_ontology_to_unstructured_elements()` now serializes every non-layout ontology element and immediately reparses that HTML into a second ontology subtree. This happens before determining whether the element has an adjacent merge candidate or is structurally eligible for inline merging.

**Why this is risky:** The parsed fragments are full recursive `OntologyElement`/Pydantic object graphs and are retained alongside the output elements through the enclosing conversion. Elements with children, tables, forms, media, and other non-inline types are guaranteed to fail `_can_ontology_elements_be_merged()`, so their additional parse and retained object graph provide no benefit. In particular, a layout containing only one large child previously performed no mergeability parse because the combine loop never compared a pair; it now reparses the entire child unconditionally.

**Concrete failure scenario:** A request containing a single multi-thousand-row table causes the already-parsed table to be serialized and rebuilt as a duplicate ontology tree while the original tree and serialized HTML remain live. With many such requests processed concurrently, the extra CPU and peak memory can produce severe latency, worker OOM termination, and retry amplification. This makes an important production document shape materially worse despite the change being presented as a performance optimization.

**Missing high-value tests:** Add a differential performance and peak-memory test for a single large non-mergeable subtree and for mixed inline/block streams. Assert that fragments with no neighbor, with children, or with a categorically non-mergeable type do not incur mergeability parsing/model construction. Include a concurrent-worker stress case rather than testing only long mergeable inline runs.

**Verdict: Not Merge.**

(authored by Codex)

PR #4422, opened 2026-08-01
fix(chunking): measure inter-chunk overlap in tokens under token chunking

eeshsaxena:
`overlap` is documented as a size in the chunking window's units, and under `max_tokens` chunking every other size (`max_tokens`, `new_after_n_tokens`, `hard_max`, `soft_max`) is measured in tokens via `ChunkingOptions.measure`. The text splitter follows that: when it splits an oversized element it calls `_get_token_overlap_tail(fragment, overlap)` and hands the next fragment approximately `overlap` **tokens**.

`PreChunk.overlap_tail`, which produces the overlap between two whole chunks, did not:

```python
overlap = self._opts.inter_chunk_overlap
return self._text[-overlap:].strip() if overlap else ""
```

That is a character slice regardless of mode, so the same `overlap` value means two different things depending on which boundary it lands on.

With `max_tokens=100, tokenizer="cl100k_base", overlap=50, overlap_all=True` over a long paragraph:

```
split boundary  -> 50 tokens   (as requested)
chunk boundary  -> 50 chars = 13 tokens
```

So inter-chunk overlap is silently about a quarter of what was asked for, and it varies with the text, since characters-per-token is not constant. Overlap exists to keep context across a boundary for retrieval, so quietly shrinking it degrades results without any error.

The fix pulls the existing binary-search tail lookup out of `_TextSplitter` into a module-level `_token_overlap_tail(text, target_tokens, measure)` and calls it from both places. `_TextSplitter._get_token_overlap_tail` now delegates to it, so the splitter's behaviour is byte-for-byte the same. Character-based chunking still takes the character slice and is untouched.

After the change the chunk boundary carries approximately 50 tokens, matching the split boundary.

Added two cases to `DescribePreChunk` in `test_unstructured/chunking/test_base.py`: one asserting the tail measures the requested token count (and is a real suffix of the text), one covering text shorter than the overlap. Verified with `pytest test_unstructured/chunking/test_base.py`: the token case fails against the current code, and all 226 pass with the fix. The existing character-mode `overlap_tail` cases pass unchanged either way.

I kept the approximate semantics rather than making the tail exact, since `_get_token_overlap_tail` already rounds to a word boundary and the docstring says the overlap "will not exceed this number ... but may be less as required to respect splitting-character boundaries".

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test_unstructured/chunking/test_base.py
+++ b/test_unstructured/chunking/test_base.py
@@ -876,6 +876,50 @@ class DescribePreChunk:
         )
         assert pre_chunk.overlap_tail == expected_value
 
+    def it_measures_its_overlap_tail_in_tokens_under_token_chunking(self):
+        """Under token chunking `overlap` is a token count, so the tail is too.
+
+        A character slice would carry far fewer tokens than requested.
+        """
+        pytest.importorskip("tiktoken")
+        import tiktoken
+
+        enc = tiktoken.get_encoding("cl100k_base")
+        text = (
+            "In rhoncus ipsum sed lectus porta volutpat morbi tincidunt augue"
+            " interdum velit euismod in pellentesque massa placerat."
+        )
+        pre_chunk = PreChunk(
+            [Text(text)],
+            overlap_prefix="",
+            opts=ChunkingOptions(
+                max_tokens=100, tokenizer="cl100k_base", overlap=10, overlap_all=True
+            ),
+        )
+
+        tail = pre_chunk.overlap_tail
+
+        # -- the tail is a real suffix of the chunk text --
+        assert text.endswith(tail)
+        tail_tokens = len(enc.encode(tail))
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #4423, opened 2026-08-01
feat: add lazy chunking entry points

badGarnet:
## Summary

Adds `iter_chunk_elements()` and `iter_chunks_by_title()`, generator counterparts to the existing `chunk_elements()` and `chunk_by_title()`. Same options, same chunks, same order — the only difference is that chunks are yielded as they are formed rather than accumulated into a list.

## Why

Chunking has always been lazy internally. `PreChunker.iter_pre_chunks()` and `PreChunk.iter_chunks()` are generators, and the public functions only wrapped them in a list comprehension. There was no public way to reach that pipeline, so a caller that wanted to stream chunks had to either buffer the whole document or import private symbols (`_ByTitleChunkingOptions`, `_BasicChunkingOptions`) and reassemble the pipeline by hand.

Combined with a lazy `elements` source, the new entry points keep peak memory proportional to the largest pre-chunk instead of the whole document. That matters when elements carry large `metadata.image_base64` payloads, where a document's element list can be far larger than the source file.

## Implementation

This exposes the existing pipeline rather than adding a second one:

- `_iter_chunk_elements()` / `_iter_chunks_by_title()` hold the generator logic.
- `_chunk_elements()` / `_chunk_by_title()` become `list()` over those generators, so there is a single implementation to keep correct and the list and generator forms cannot drift.
- Both public entry points keep their existing `(elements, opts)` call into the private implementation, so the option-plumbing unit tests are unchanged.

Options are validated **eagerly**, when the function is called, not when the returned iterator is first advanced. `iter_chunk_elements` and `iter_chunks_by_title` are therefore plain functions returning an iterator, not generator functions — a generator function would have deferred the whole body, so an invalid option combination would surface at an unrelated point in the caller, or not at all for a document that yields no chunks.

## Tests

`test_unstructured/chunking/test_lazy.py`:

- **Equivalence** with the list form for both strategies, with `include_orig_elements` both on and off, comparing serialized chunks field-by-field (`orig_elements` included; `element_id` is a fresh UUID per chunk and is excluded).
- **Laziness** — yielding the first chunk does not drain the source iterator.
- **Eager option validation** — an invalid option raises at the call, not at first advance.
- **Signature parity** — guards against the generator and list signatures drifting as options are added.

The existing chunking suite (343 tests) passes unchanged.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aadland6:
---

## Issues and proposed fixes (most to least severe)

### 1. `iter_chunks_by_title` can materialize the rest of the source before the first `yield`

**Severity:** Medium — breaks the advertised memory bound for the by-title strategy.

**Where:** `unstructured/chunking/title.py` → `_iter_chunks_by_title` →
`unstructured/chunking/base.py` `PreChunkCombiner.iter_combined_pre_chunks`
(lines 1741–1752).

**Problem:** The by-title path runs pre-chunks through `PreChunkCombiner`, which greedily combines
small pre-chunks and therefore always reads **one pre-chunk ahead** before it can flush the
accumulator (it must see a successor that does not fit, or hit EOF). The basic path has no combiner
and stops after roughly one window plus one element of lookahead.

The docstrings and CHANGELOG claim peak memory is "bounded by the largest pre-chunk rather than by
the whole document." For by-title that bound is really **≈ two pre-chunks** (accumulator + one
lookahead), and in the degenerate case where the whole document is only one or two chunking windows,
that one-pre-chunk lookahead *is* the remainder of the source — so the first `next()` builds every
remaining element (and every `image_base64` payload) before returning a single chunk. This is
exactly the short-text / image-heavy shape the feature's motivation cares about.

**Reproduction (confirmed on branch):**

```text
source: 300 × NarrativeText("a", image_base64=200KB), max_characters=500, include_orig_elements=False
  iter_chunk_elements:  first next() materializes 168 elements
  iter_chunks_by_title: first next() materializes 300 elements (entire source)

source: 2000 elements (many windows), same options
  iter_chunk_elements:  first next() materializes 168 elements
  iter_chunks_by_title: first next() materializes 335 elements  (≈ two windows — the inherent lookahead)
```

The two-window (`335`) case is the inherent, acceptable cost of greedy combining; the whole-document
(`300`) case is the one that nullifies the advertised win.

**Proposed fix (primary — safe, validated):** Correct the guarantee wording in the
`iter_chunks_by_title` docstring (`title.py:136`) and the CHANGELOG entry so it is accurate for
by-title: the combiner reads one pre-chunk ahead, so first-chunk peak memory is bounded by roughly
**two** pre-chunks, and when the whole document fits in one or two chunking windows the lookahead can
materialize the remainder of the source. Keep the stronger "largest pre-chunk" phrasing only for
`iter_chunk_elements`, where it is accurate.

**Proposed fix (optional deeper change):** Make `PreChunkCombiner.iter_combined_pre_chunks` flush the
accumulator eagerly the moment it provably cannot combine with any successor, instead of waiting to
pull the next pre-chunk. Note this is **not a drop-in win**: `can_combine()` rejects a successor on
the length test only when `len(accumulated_text) >= combine_text_under_n_chars`, but `PreChunker`
fills pre-chunks to *just under* the hard max (e.g. 499 chars against a 500 ceiling), so the simple
`>= ceiling` guard does not fire in the common default case, and a looser guard is not
behavior-preserving (a zero-length successor pre-chunk can still legitimately combine). If pursued,
it must be gated behind a condition that is provably equivalent to the current combining decision and
covered by the existing equivalence tests. Given that risk, prefer the documentation fix unless the
memory bound must hold literally for by-title.

---

### 2. Tokenizer/setup errors are deferred to the first `next()`, violating the eager-validation contract

**Severity:** Medium — list and iterator forms diverge on a contract the PR explicitly advertises.

**Where:** `unstructured/chunking/base.py` `ChunkingOptions._validate` (lines 326–385) and
`TokenCounter._encoder` (lines 63–73); surfaced through the `iter_*` wrappers in
`basic.py` / `title.py`.

**Problem:** The PR states options are validated "when the function is called, not when the returned
iterator is first advanced." That holds for the `ValueError` combination checks in `_validate`
(e.g. `new_after_n_chars=-3`), but **not** for the tokenizer. `TokenCounter` imports tiktoken and
resolves the encoding lazily on first `count()`. The list APIs hit that during the call because they
immediately exhaust the generator; the new `iter_*` APIs return a generator successfully and only
raise on first advance — the exact "error shows up somewhere unrelated" failure mode the PR says it
avoids.

**Reproduction (confirmed on branch, tiktoken installed):**

```python
from unstructured.chunking.basic import chunk_elements, iter_chunk_elements
els = [Title("A"), NarrativeText("hello world " * 50)]

chunk_elements(els, max_tokens=20, tokenizer="not-a-real-tokenizer-xyz")
# -> ValueError("Unknown encoding not-a-real-tokenizer-xyz") raised DURING the call

it = iter_chunk_elements(els, max_tokens=20, tokenizer="not-a-real-tokenizer-xyz")
# -> call SUCCEEDS, returns a generator
next(it)
# -> ValueError raised here instead
```

Without tiktoken installed the same split shows up as `ModuleNotFoundError` on `next(it)` vs. during
`chunk_elements(...)`.

**Proposed fix:** Resolve the tokenizer eagerly inside `ChunkingOptions._validate()`, at the same
moment as the other eager checks, so list and iter fail at the same call site. Since `token_counter`
is a cached property, resolving it in `_validate` caches the encoder for reuse — no double
resolution:

```python
# in ChunkingOptions._validate(), after the existing checks
if self._kwargs.get("tokenizer") is not None:
    # -- force tokenizer/encoder resolution now so an invalid tokenizer (or a missing
    # -- tiktoken install) raises at the call site, matching the list APIs and the
    # -- eager-validation contract, instead of on the iterator's first advance. --
    _ = self.token_counter._encoder  # noqa: SLF001
```

Validated: this makes `iter_chunk_elements(..., tokenizer="not-a-real-tokenizer-xyz")` raise
`ValueError` at the call, while a valid tokenizer (`"gpt-4"`) and the character-only path (no
`tokenizer`) are unaffected. (If reaching into `_encoder` is undesirable, add a small public
`TokenCounter.validate()` / `ensure_ready()` that forces resolution, and call that instead.)

---

### 3. Memory guidance overstates the per-chunk benefit under the default `include_orig_elements=True`

**Severity:** Medium — misleading operational guidance for the primary use case; not a correctness bug.

**Where:** Docstrings in `unstructured/chunking/basic.py` (`iter_chunk_elements`, lines 127–138) and
`unstructured/chunking/title.py` (`iter_chunks_by_title`, lines 136–147); the CHANGELOG enhancement
entry.

**Problem:** The APIs are promoted as keeping peak memory proportional to the largest pre-chunk, with
`metadata.image_base64` as the motivating case. But with the default `include_orig_elements=True`,
every emitted chunk embeds its source elements — including their full `image_base64` payloads — in
`metadata.orig_elements`. Streaming still helps (you avoid buffering *all* output chunks, and you can
pair with a lazy element source), but a caller who follows the docstring literally for large
`image_base64` workloads sees little per-chunk relief unless they also pass
`include_orig_elements=False`, which the docstring never mentions next to the motivating example.

**Reproduction (confirmed on branch):**

```python
from unstructured.chunking.basic import iter_chunk_elements
from unstructured.documents.elements import NarrativeText, ElementMetadata

big_b64 = "A" * 500_000
elements = [NarrativeText(f"t{i}", metadata=ElementMetadata(image_base64=big_b64)) for i in range(10)]

chunk = next(iter_chunk_elements(elements, max_characters=100, new_after_n_chars=0))
assert chunk.metadata.orig_elements[0].metadata.image_base64 == big_b64  # payload still carried
```

**Proposed fix:** Qualify the memory claim in both docstrings and the CHANGELOG — note that
`include_orig_elements=False` is typically required when large element metadata (e.g.
`image_base64`) is the bottleneck, and that the streaming wins are (a) not buffering all output
chunks and (b) pairing with a lazy element iterator. This wording change can be made together with
the by-title correction in Issue 1, since both touch the same docstrings/CHANGELOG memory claim.

---

aadland6 (approved):

PR #4425, opened 2026-08-02
fix(cleaners): handle empty text in ordered-bullet cleaners

asjad3:
### Summary

`clean_ordered_bullets()` and `extract_ordered_bullets()` read the first token of
`text.split()` without checking that a token exists, so empty or whitespace-only text
raises `IndexError` instead of passing through:

```python
>>> from unstructured.cleaners.core import clean_ordered_bullets
>>> clean_ordered_bullets("")
IndexError: list index out of range
```

Every other cleaner in `unstructured/cleaners/core.py` returns `''` for empty input —
`clean_bullets`, `clean_extra_whitespace`, `clean_dashes`, `clean_trailing_punctuation`,
`clean_non_ascii_chars`, `clean_ligatures`, `replace_unicode_quotes`,
`remove_punctuation` — so this looks like an oversight rather than an intended contract.

It matters because elements with empty text are common, and cleaners are applied per
element. A single empty element aborts the whole run:

```python
from unstructured.cleaners.core import clean_ordered_bullets

for el in [Text("1.1 Introduction"), Text(""), Text("2.1 Methods")]:
    el.apply(clean_ordered_bullets)   # raises IndexError on the empty one
```

### Changes

Both functions now return early when `text.split()` is empty — the text unchanged for
`clean_ordered_bullets`, and `(None, None, None)` for `extract_ordered_bullets`, which is
what it already returns for any text that carries no bullet.

### Testing

The snippet above runs clean on this branch. Or via the existing parametrized tests,
which now cover `""`, `"   "` and `"\n"` for both functions:

```
pytest test_unstructured/cleaners/test_core.py test_unstructured/cleaners/test_extract.py
```

- 133 passed on this branch (127 before, plus the 6 new cases)
- The 6 new cases fail on `main` with `IndexError`, pass here
- `ruff check` / `ruff format --check` clean on the changed files

CHANGELOG and `__version__.py` bumped to `0.25.2-dev1`.

---

Written with AI assistance (Claude Code); I reviewed the change and ran the checks above.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4425?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
-## 0.25.2-dev0
+## 0.25.2-dev1
 
 ### Enhancements
 
 - **Speed up HTML element hierarchy reconstruction**: `elements_to_html()` now indexes elements by ID before attaching children, avoiding repeated linear parent scans.
+
+### Fixes
+
+- **Fix `IndexError` on empty text in the ordered-bullet cleaners**: `clean_ordered_bullets()` and `extract_ordered_bullets()` indexed the first token of `text.split()` without checking that a token existed, so empty or whitespace-only text raised `IndexError`. Applying either to a document containing an empty element (e.g. `element.apply(clean_ordered_bullets)`) aborted the run. They now return the text unchanged and `(None, None, None)` respectively, matching the other cleaners in the module.
 
 ## 0.25.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,20 @@
-## 0.25.2-dev1
+## 0.25.3-dev0
+
+### Fixes
+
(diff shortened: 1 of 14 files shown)


eeshsaxena (commented):
Clean fix, and it looks complete for this pattern rather than a spot patch. The unguarded `text.split()[0]` / `[1:]` indexing only lives in `clean_ordered_bullets` (core.py) and `extract_ordered_bullets` (extract.py), and both are guarded here, so there is no sibling in the cleaners module still exposed to the same `IndexError`. Returning the text unchanged / `(None, None, None)` on empty input also matches how the other cleaners in the module behave on no-op input, so `element.apply(clean_ordered_bullets)` over a document with an empty element no longer aborts the whole run. The added empty/whitespace/newline test cases cover it well.

PR #4437, opened 2026-08-14
fix(security): bound quadratic array-stream decoding in is_pdf_too_complex (SEC-146)

aadland6:
## Summary

Fixes [SEC-146](https://linear.app/unstructured/issue/SEC-146/quadratic-memorycpu-in-is-pdf-too-complex-array-stream-decoding-pypdf). `is_pdf_too_complex()` in `unstructured/partition/pdf.py` re-implemented pypdf's array-based `/Contents` decoding using the same quadratic `raw_data += obj.get_data()` accumulation that pypdf patched under **CVE-2026-33123 / GHSA-qpxp-75px-xjcp** ("Inefficient decoding of array-based streams"). A crafted PDF whose page `/Contents` is an array of many small stream objects could force excessive CPU/memory. This function runs on every partitioned PDF, so it sits on the untrusted-input path (`pdf.py:309`).

Per the ticket, this keeps the intentional lightweight raw-bytes approach (added in #4268 to cheaply detect vector-heavy CAD/engineering PDFs) — it does **not** switch to pypdf's expensive `ContentStream`. It just makes the accumulation efficient and bounded.

## Changes

- **Accumulate into a `bytearray` with `.extend()`** instead of rebinding a `bytes` object (amortized O(1) vs O(n²)).
- **Per-page caps, fail closed** (a single pathological page → treat as too complex, skip PDFMiner):
  - `max_raw_stream_bytes` (50 MB) — decoded bytes per page, checked *before* each stream is copied so an oversized stream is never accumulated or regex-scanned.
  - `max_content_stream_array_entries` (10,000, pypdf's number) — bounds an array of many empty/tiny streams that a byte cap alone misses.
- **Document-level caps** so total work is a property of the function, not the page count (pages can share one indirect `/Contents` array → tiny file, unbounded scan). Charged incrementally per stream so a mid-page decode error can't discard the accounting. Set far above any plausible real document and logged at `warning`:
  - `max_total_stream_bytes` (1 GB) — decoded bytes per document.
  - `max_total_array_entries` (1,000,000) — decoded entries per document (zero-byte entries never advance the byte total but still cost a decode).
- **Dereference an indirect `/Contents`** before the array check — `DictionaryObject.get` (unlike `__getitem__`) does not resolve references, so an indirect array of streams was silently skipping the array branch entirely (both the heuristic and the caps). 87 of 1,105 corpus pages reach their content array this way.
- **Count operators with `finditer`** instead of `findall`, so counting no longer allocates a match list proportional to stream size.
- **Bump `pypdf` to `>=6.9.1`** (from `>=6.6.2`) so the library's own code path is patched too; lock resolves 6.10.0.

## Audit (AC4)

Grepped `unstructured/partition/pdf.py` and `pdf_image/` for other `+=`-on-stream-bytes loops — none found; this was the only instance. A separate O(n²) string-accumulation pattern in `unstructured/partition/html/transformations.py` (element-merge loop) is out of SEC-146's scope (not on the PDF path) and is tracked in its own ticket.

## Testing

Rewrote the regression tests around real `PdfWriter`/`PdfReader` fixtures (FlateDecode-compressed streams so the file stays small while decoded output is huge — the actual attack shape). Coverage: direct + indirect graphics-heavy arrays, a many-small-streams array (~2 MB file → ~900 MB decoded) that runs for minutes / OOMs on the pre-fix code and returns in ~0.05 s here, the entry cap, the pre-copy byte cap, the cross-page byte and entry budgets, and budget survival across a mid-page decode error. The key tests were confirmed to fail on the pre-fix code.

- Full `test_unstructured/partition/pdf_image/test_pdf.py`: **178 passed, 1 skipped**
- `ruff check` + `ruff format --check`: clean

## Acceptance criteria

- [x] `is_pdf_too_complex` accumulates via `bytearray`, not `bytes +=`
- [x] Total-length bounds cap worst-case work on array-based content streams (per-page + per-document, bytes + entries)
- [x] Regression test with a many-small-streams array PDF completes in bounded time/memory (fails on the old code)
- [x] Repo audited for other `+=`-on-stream-bytes copies; findings fixed or ticketed
- [x] `pypdf` dependency confirmed ≥ 6.9.1

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4437?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


badGarnet:
a few comments:

1. High — Small compressed PDFs bypass every new safety limit
   `unstructured/partition/pdf.py:704`

   Files under 1 MB return `False` before their content structure is inspected. A 55 KB PDF containing 9,000 references to one 100 KB compressed stream represents about 900 MB of content but bypasses the check.

   Reproduce: create that PDF with `PdfWriter`; the default call returns `False`, while passing `min_file_size_bytes=1` returns `True`.

   Suggested fix: remove the file-size shortcut, or at least inspect `/Contents` arrays and compressed streams before allowing small files through.

2. Medium — One decode error skips the rest of the page and fails open
   `unstructured/partition/pdf.py:809`

   Any exception other than `LimitReachedError` causes the entire page to be skipped. Later streams may therefore exceed the byte limits without inspection.

   Reproduce: create `/Contents` as `[unsupported-filter stream, valid 10-byte stream]`, set `max_total_stream_bytes=1`, and call with `min_file_size_bytes=1`. It incorrectly returns `False`.

   Suggested fix: return `True` when a content stream cannot be inspected, or catch errors per item and fail closed for that page.

3. Medium — The byte cap is checked only after allocating decoded data
   `unstructured/partition/pdf.py:762` and `unstructured/partition/pdf.py:783`

   `get_data()` fully decompresses the stream before its size is compared with `max_raw_stream_bytes`. Thus, the configured cap does not bound peak memory.

   Reproduce: use a 59 KB PDF containing one stream that expands to 60 MB, set `max_raw_stream_bytes=1_000_000`, and measure with `tracemalloc`. The call returns `True`, but peaks around 129 MiB.

   Suggested fix: decode incrementally with the remaining page/document budget, or add/use a `pypdf` decoding API that accepts a per-call output limit.

badGarnet (approved):

aadland6:
Thanks @badGarnet — all three reproduced and addressed in c8b495b0.

**1. Small files bypass the new limits (High).** Fixed. `min_file_size_bytes` now defaults to `0`, so every file is inspected; skipping small files is opt-in. Inspection is bounded, and normal small files stay cheap (~0.44 ms/call) because the per-page `min_raw_stream_bytes` skip avoids the regex on small content. Regression test: a <1 MB file whose array decodes to ~900 MB is now flagged by default.

Tradeoff worth a look: this reverts the small-file fast path from #4268 (~0.44 ms/PDF on the partition path). The knob remains, so a perf-sensitive caller can set `min_file_size_bytes` back and accept the bypass. Happy to restore the old default if you'd rather handle small-file bombs elsewhere.

**2. One decode error skips the rest of the page (Medium).** Fixed. Decoding is now per stream: a bomb still fails closed (`LimitReachedError` → `True`), but an otherwise-unreadable stream skips only itself, so the remaining streams on the page are still inspected and charged. Regression test: `[unsupported-filter stream, over-cap stream]` now returns `True` (was `False`).

**3. Byte cap checked only after decode (Medium).** Confirmed, but bounded and not separately fixable here. `get_data()` must decode a stream to learn its size, and a single stream's decode is already capped by pypdf's own per-filter limit (`ZLIB_MAX_OUTPUT_LENGTH`, 75 MB), which raises `LimitReachedError` — now caught and failed closed. pypdf exposes no per-call output limit, and lowering its module-global limit around `get_data()` isn't thread-safe (it would affect concurrent pypdf work in the same process). So peak per-stream memory is bounded by pypdf (≈75 MB) rather than by `max_raw_stream_bytes`, and the docstring no longer overclaims otherwise. Glad to add a tighter bound if pypdf ever grows a per-call limit.

Full `test_pdf.py`: 183 passed, 1 skipped.

PR #4441, opened 2026-08-18
fix(cleaners): prevent UnboundLocalError in _get_indexed_match when pattern is absent

### LINKED ISSUE #4427: extract_text_before/extract_text_after raise UnboundLocalError when the pattern is absent

mayuriphad:
### Summary

Closes #4427

`extract_text_before()` and `extract_text_after()` raised an `UnboundLocalError` instead of a `ValueError` when the target pattern was not present in the input text. The helper `_get_indexed_match()` referenced loop counter `i` in its formatted error message, which remained unassigned when `re.finditer` produced 0 matches.

### Solution
- Initialized `i = -1` before the match iteration loop.
- Added a clear error message `Result with index {index} was not found. No matches were found.` when no matches are detected.
- Added unit regression tests in `test_unstructured/cleaners/test_extract.py`.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test_unstructured/cleaners/test_extract.py
+++ b/test_unstructured/cleaners/test_extract.py
@@ -15,8 +15,21 @@ def test_get_indexed_match_raises_with_bad_index():
 
 
 def test_get_indexed_match_raises_with_index_too_high():
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="The largest index was 2"):
         extract._get_indexed_match("BLAH BLAH BLAH", "BLAH", 4)
+
+
+def test_get_indexed_match_raises_with_no_matches():
+    with pytest.raises(ValueError, match="No matches were found"):
+        extract._get_indexed_match("BLAH BLAH BLAH", "XYZ", 0)
+
+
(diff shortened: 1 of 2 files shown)


mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

PR #4449, opened 2026-08-20
perf(nlp): run the spaCy pipeline once per distinct text

r0h1tb:
## Problem

`_process()` is the only expensive call in `unstructured/nlp/tokenize.py` and the only one without a cache. `word_tokenize()`, `pos_tag()` and `sent_tokenize()` each memoize their own extracted view, which hides the repetition from their callers but not from spaCy.

`is_possible_narrative_text()` reaches all three while classifying a single element:

```
_tokenize_for_cache <- sent_tokenize <- sentence_count <- exceeds_cap_ratio <- is_possible_narrative_text
word_tokenize <- exceeds_cap_ratio <- is_possible_narrative_text
pos_tag <- is_possible_narrative_text
```

Three identical `Doc`s per element. On a 400-paragraph document that is 88% of `partition_html()` (`is_possible_narrative_text` 1.60s of 1.81s in cProfile).

## Fix

Memoize the `Doc` so the three extractors share one pipeline run.

Texts over 8 KiB bypass the cache. A `Doc` is much heavier than the token lists the existing caches hold (~8 KB serialised for a 68-char paragraph) and `_process()` accepts input up to spaCy's 1M-char limit, so one oversized element could otherwise pin an outsized object for the life of the process.

All three consumers only read from the `Doc` and each returns a fresh list/tuple, so sharing is safe. `_process()` has no callers outside this module.

## Numbers

| | before | after |
|---|---|---|
| `partition_html`, 400 `<p>` | 1127 ms | 395 ms |
| `partition_text`, 400 paragraphs | 1129 ms | 403 ms |

Same shared classification path, so this is not HTML-specific. Scaling was already linear and stays linear — this is a constant-factor change. The test suite itself drops 38.2s to 33.0s.

## Tests

Four tests in `test_unstructured/nlp/test_tokenize.py`, counting entries into spaCy below every cache in the module rather than timing anything. With the source change reverted on this branch:

```
FAILED test_the_spacy_pipeline_runs_once_per_distinct_text - assert 3 == 1
FAILED test_repeated_text_does_not_re_run_the_pipeline     - assert 2 == 1
FAILED test_oversized_text_bypasses_the_doc_cache
FAILED test_text_at_the_threshold_is_still_cached
```

Suite before 146 failed / 2531 passed, after 146 failed / 2533 passed — the same 146 IDs, all pre-existing here (pandoc and the ML extras aren't installed locally). 14 modules that can't be collected without those extras were excluded from both runs.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4449?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Enhancements
 
-- **Run the spaCy pipeline once per distinct text**: `_process()` in `unstructured.nlp.tokenize` was uncached while `word_tokenize()`, `pos_tag()` and `sent_tokenize()` each cached only their own extracted view, so every helper paid for its own identical pipeline run. Classifying one element via `is_possible_narrative_text()` reached all three and ran spaCy three times on the same string. The `Doc` is now memoized and shared, cutting `partition_html()` on a text-heavy document by ~2.8x (400 paragraphs: 1127ms -> 395ms). Texts above 8 KiB bypass the cache so a single very large element cannot pin an outsized `Doc` in memory.
+- **Run the spaCy pipeline once per distinct text**: `_process()` in `unstructured.nlp.tokenize` was uncached while `word_tokenize()`, `pos_tag()` and `sent_tokenize()` each cached only their own extracted view, so every helper paid for its own identical pipeline run. Classifying one element via `is_possible_narrative_text()` reached all three and ran spaCy three times on the same string. The `Doc` is now memoized and shared, cutting `partition_html()` on a text-heavy document by ~2.8x (400 paragraphs: 1127ms -> 395ms). Texts longer than 8,192 characters bypass the cache so a single very large element cannot pin an outsized `Doc` in memory.
 
 ## 0.26.3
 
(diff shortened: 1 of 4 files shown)


r0h1tb:
Both addressed or answered in 468ec3b.

**P2, `tokenize.py:171` — characters vs bytes.** Correct, and the comment was the thing that was wrong. `len(text)` counts characters, so calling the bound "8 KiB" understates it for non-ASCII — 8,192 CJK characters is roughly 24 KB of UTF-8. Characters are the right unit to bound on, though: `Doc` size tracks token count, not UTF-8 width, so switching the comparison to encoded length would bound the wrong quantity. Reworded the comment and the changelog to say 8,192 characters; the constant was already named `MAX_CACHEABLE_CHARS`.

**P2, `tokenize.py:153` — concurrent misses.** Real but left alone. `lru_cache` isn't single-flight, so two threads asking for the same uncached string can both run the pipeline. The cost is one redundant run that would have happened anyway before this change — not a correctness problem, and the cache converges immediately after. The three existing caches in this module (`word_tokenize`, `pos_tag`, `_tokenize_for_cache`) have the same property. Adding per-key locking here would introduce a lock-ordering surface for a transient duplicate, and only in this one of the four.

cragwolfe:
Content review of `bdc90f76e74823f8ebb1ec22831e151ace0e7b02` looks good; no actionable content issues found. The sentence, word, and POS consumers only read the shared `Doc`, and the sequential narrative-classification path reduces three pipeline executions to one while the entry is resident. The 8,192-character limit and concurrent-miss behavior are reasonable for this implementation.

This is a content-only conclusion: exact-head CI is currently `action_required` with no CI jobs created. Branch conflicts and release metadata still need separate reconciliation.

(authored by codex)

cragwolfe (approved):

PR #4450, opened 2026-08-20
fix(cleaners): raise clear ValueError when extract pattern is absent

### LINKED ISSUE #4427: extract_text_before/extract_text_after raise UnboundLocalError when the pattern is absent

ayam04:
Fixes #4427

`_get_indexed_match` referenced the enumerate loop variable `i` inside the "not found" error message, but `i` is never bound when `re.finditer` yields zero matches. So `extract_text_before` / `extract_text_after` raised an `UnboundLocalError` (crash) instead of the intended `ValueError` when the pattern does not occur in the text.

Track the largest match index in a dedicated variable initialized to `-1`, so the no-match path always raises a clear message: "Result with index 0 was not found. The largest index was -1."

Tests added:
- `test_get_indexed_match_no_match_raises_clear_error`
- `test_extract_text_before_no_match_raises_clear_error`

Validation: `python3 -m pytest test_unstructured/cleaners/test_extract.py -q` (41 passed), `uv run ruff check` clean.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test_unstructured/cleaners/test_extract.py
+++ b/test_unstructured/cleaners/test_extract.py
@@ -19,6 +19,16 @@ def test_get_indexed_match_raises_with_index_too_high():
         extract._get_indexed_match("BLAH BLAH BLAH", "BLAH", 4)
 
 
+def test_get_indexed_match_no_match_raises_clear_error():
+    with pytest.raises(ValueError, match="not found"):
+        extract._get_indexed_match("BLAH BLAH BLAH", "XYZ")
+
+
+def test_extract_text_before_no_match_raises_clear_error():
+    with pytest.raises(ValueError, match="not found"):
+        extract.extract_text_before("BLAH BLAH BLAH", "XYZ")
+
+
 def test_extract_text_before():
     text = "Teacher: BLAH BLAH BLAH; Student: BLAH BLAH BLAH!"
     assert extract.extract_text_before(text, "BLAH", 1) == "Teacher: BLAH"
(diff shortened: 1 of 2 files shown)


ayam04:
Hi @cragwolfe @badGarnet — quick ping on this one: Fixes #4427 (`extract_text_before`/after raised `UnboundLocalError` when the pattern is absent; now raises a clear `ValueError` with the largest-index context, plus regression tests). Cubic review passed with 'No issues found'. Would appreciate a look when you have time.

ayam04:
Hi @cragwolfe @badGarnet, gentle bump: this is complete and mergeable. `extract_text_before`/`extract_text_after` now raise a clear `ValueError` naming the pattern when it is absent, instead of the `UnboundLocalError` from #4427, with the regression test. Anything needed to move it forward?

PR #4451, opened 2026-08-21
fix(html): extract <details>/<summary> content instead of discarding it

r0h1tb:
## Problem

`<details>` and `<summary>` were both mapped to `RemovedBlock`, so `partition_html()` discarded the whole subtree. On the FAQ page from the issue, every question and answer disappears and the caller gets a shorter element list rather than an error:

```python
>>> partition_html(text="<h1>Support FAQ</h1>"
...   "<details><summary>Which Bluetooth profiles are supported?</summary>"
...   "<p>Your device supports A2DP and HFP profiles.</p></details>")
[<Title 'Support FAQ'>]          # both the question and the answer are gone
```

## Fix

`<details>` is an ordinary block container, so it becomes `Flow`, like `<div>`.

`<summary>` becomes `Heading`. The spec calls it "a summary, caption, or legend for the rest of the contents", the issue asks for the questions as titles, and it is what makes the fix useful downstream — `chunk_by_title()` opens a new section at each `Title`, so a question stays attached to its own answer instead of blending into the next entry.

`<summary>` was also listed twice, under both phrasing and the removed-form group; there is now one mapping.

## Scope

Only these two tags. `<dl>`, `<figure>` and the form/nav tags stay in `RemovedBlock`:

- `<figure>`/`<figcaption>` removal is asserted by `DescribeRemovedBlock.it_is_skipped_during_parsing`, so it is a deliberate decision, not an oversight.
- `<dl>`/`<dt>`/`<dd>` also drop their content, and `ListBlock`'s docstring already says "maybe a `<dl>` element at some point". That looks worth doing, but reclassifying it changes the ancestor chain that `_category_depth` walks (there is a parametrised case at `test_parser.py:298` relying on `<dl><dd>` nesting), so it belongs in its own PR rather than riding along here.

## Tests

Four tests in `test_unstructured/partition/html/test_partition.py`. With the parser change reverted on this branch:

```
FAILED test_partition_html_extracts_details_and_summary_content        - assert [] == [(...5 elements)]
FAILED test_partition_html_makes_a_summary_start_its_own_chunk_section
FAILED test_partition_html_extracts_a_details_block_that_has_no_summary
FAILED test_partition_html_extracts_nested_details_blocks
```

Also checked by hand and behaving: unclosed `<details>`, `<summary>` not the first child, two `<summary>` siblings, empty `<summary>`, `<summary>` outside any `<details>`, attributes on `<details>`, inline markup inside `<summary>`, `<details>` inside a table cell and inside a list item, 20-deep nesting, and a JSON round-trip.

Suite before 146 failed / 2531 passed, after 146 failed / 2535 passed — the same 146 IDs, all pre-existing here (pandoc and the ML extras aren't installed locally). 14 modules that cannot be collected without those extras were excluded from both runs.

Note: this bumps to `0.26.4`, as does #4449 — whichever lands second needs the version and changelog re-pointed.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
 
 - **Add privacy-bounded partition runtime telemetry**: Each top-level public partition call now makes one best-effort local attempt to send an event with fixed-enum runtime characteristics and aggregate final-element counts. Nested dispatch is suppressed. Delivery uses one non-queued daemon worker with tight connect/read timeouts; it disables redirects, retries, response-body downloads, proxy settings, and netrc credentials. A stuck transport cannot block document processing or process exit, at most one daemon can remain stranded, and later events drop while that slot is occupied. The event contains no document content, filenames, paths, URLs, caller MIME strings, exception details, credentials, or persistent identifiers. Set either `DO_NOT_TRACK` or `SCARF_NO_ANALYTICS` to any non-empty value after trimming to disable both startup and runtime telemetry before any runtime telemetry collection or delivery work.
 
+## 0.27.2
+
+### Fixes
+
+- **Extract `<details>`/`<summary>` content instead of discarding it**: both tags were mapped to `RemovedBlock`, so `partition_html()` silently dropped every disclosure widget and everything inside it. FAQ and documentation pages built from accordions lost all of their questions and answers, and the caller saw a shorter element list rather than an error. `<details>` is now `Flow` (an ordinary block container) and `<summary>` is `Heading`, which emits a `Title` -- so `chunk_by_title()` opens a new section per entry and a question stays attached to its own answer. Resolves #3919.
+
 ## 0.26.3
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


cragwolfe:
**High — a leading image or block wrapper makes a `<summary>` lose its `Title` boundary.**

`unstructured/partition/html/parser.py:1023` maps `summary` to `Heading`, but `Heading` supplies `Title` only to the initial text/phrasing run in `Flow.iter_elements()` (`parser.py:365-376`). Since `img` is a non-phrasing `ImageBlock`, `<summary>Second question?</summary>` first emits the image, then processes `Second question?` through the generic tail path without `Heading._ElementCls`; it becomes `Text`, not `Title`. A `<div>`/`<p>`-wrapped question follows the same child-block path. As a result, `chunk_by_title(..., combine_text_under_n_chars=0)` never sees the second FAQ boundary and can merge that question/answer into the preceding section.

Please add summary-specific traversal that preserves the title role for the summary label across leading images and ordinary block wrappers, while keeping the details body and post-summary tail outside the title and avoiding a duplicate title for a nested heading. The regression should assert exact element types/order and two separate chunks; covering link/emphasis and page metadata on the label would also protect the annotations carried by this path.

(authored by codex)

Issue #2449, opened 2024-01-24
UnicodeDecodeError in logger.info during the execution of partition_doc


MinnyKuan:
hi,
I'm using version 0.11.8.
I use the following code to execute partition_doc :
```

from unstructured.partition.doc import partition_doc

filename = ""D:\\Testcase\\test.doc""
elements = partition_doc(filename=filename)

```

However, I encountered a UnicodeDecodeError occurring while executing convert_office_doc in ` logger.info(output.decode().strip())`
The error message is as follows:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xaa in position 38: invalid start byte`

MthwRobinson:
@MinnyKuan - Any chance you can share the file that causes the error?

MinnyKuan:
Sure, because I cannot upload a doc file here, I used a compressed file.
This is a compressed file of a doc document:
[test.zip](https://github.com/Unstructured-IO/unstructured/files/14045576/test.zip)

MthwRobinson:
Thanks @MinnyKuan ! We'll look into this as soon as we can

MthwRobinson:
Hi @MinnyKuan - I was able to process the example doc using `partition_doc` on a Mac. Are you able to try this on Mac or Linux? From your example code it looks like you're on Windows.

MinnyKuan:
I tried to test in a Linux environment. 
The doc file shows the following error message, but the docx file does not.
`FileNotFoundError                         Traceback (most recent call last)`
![image](https://github.com/Unstructured-IO/unstructured/assets/74747729/bd7c79fa-2f7e-43d3-b67c-f0f959aaa236)


I also tried using the UnstructuredFileLoader method, but the result was the same.
I am currently still searching for a solution.

srjchauhan:
@MinnyKuan is this still a issue for you? I was able to process the example doc using `partition_doc` on my linux

MinnyKuan:
@scanny 
Yes, but currently I am converting the doc files to docx format locally before reading them. This allows me to continue using them.

scanny:
@MinnyKuan A couple things:
1. If you have already converted a document from `.doc` to `.docx`, you will need to call `partition_docx(...)` rather than `partition_doc(...)`.
2. Please post a minimum-reproducible-example that produces this problem in your environment: https://stackoverflow.com/help/minimal-reproducible-example
[...]

PR #4465, opened 2026-08-31
fix: tolerate non-UTF-8 soffice output in doc/ppt conversion

### LINKED ISSUE #3652: bug/Cannot partition doc files with multi-byte names
**Describe the bug**
When calling `unstructured.partition.doc.partition_doc` with a doc file with multi-byte name (I checked: `文章.doc`/`风格.doc`), it fails with an error.

**To Reproduce**

1. Create empty `*.doc` file using Word, and name it `文章.doc` or `风格.doc`.  I attached that files below:
[empty_docs.zip](https://github.com/user-attachments/files/17084675/empty_docs.zip)

2. Run the code below: 
```python

doc_path = "文章.doc"
# doc_path = "风格.doc"
elements = partition_doc(filename=doc_path)
```

3. It will throw: 
```
  File "<home>\utf8error.py", line 4, in <module>
    elements = partition_doc(filename=doc_path)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<python>\site-packages\unstructured\documents\elements.py", line 605, in wrapper
               ^^^^^^^^^^^^^^^^^^^^^
  File "<python>\site-packages\unstructured\file_utils\filetype.py", line 731, in wrapper
               ^^^^^^^^^^^^^^^^^^^^^
  File "<python>\site-packages\unstructured\file_utils\filetype.py", line 687, in wrapper
               ^^^^^^^^^^^^^^^^^^^^^
  File "<python>\site-packages\unstructured\chunking\dispatch.py", line 74, in wrapper
               ^^^^^^^^^^^^^^^^^^^^^
  File "<python>\site-packages\unstructured\partition\doc.py", line 89, in partition_doc
    convert_office_doc(
  File "<python>\site-packages\unstructured\partition\common.py", line 429, in convert_office_doc
    message = output.stdout.decode().strip()
              ^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8a in position 32: invalid start byte
```

(I replaced private directory with \<home>. Also replaced directory Python installed with \<python>)

**Expected behavior**
The code should not throw an exception. 

**Screenshots**
**Environment Info**
- Windows 11 Home
- Python 3.11.9
- unstructured: 0.15.13

**Additional context**
English(alphabet only) filename didn't cause an exception to be thrown.

huuufu:
## Summary

`convert_office_doc()` — shared by `partition_doc()` and `partition_ppt()` — decoded `soffice` stdout and stderr with a strict UTF-8 `.decode()` at three sites, used only for logging and to check whether stdout was empty (which drives the retry loop and the final failure check). LibreOffice echoes the input path in the console encoding, which on Windows is the locale codepage rather than UTF-8, so a name or path with multi-byte characters makes the decode raise and aborts a conversion that had already succeeded:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8a in position 32: invalid start byte
```

The traceback in #3652 is on the first stdout decode. The stderr site is reachable and broken the same way when a conversion fails, and sits outside the surrounding `try` entirely.

Fixing the child process's output encoding is not an option — @Snowman-s confirmed on the issue that `PYTHONIOENCODING=utf-8:surrogateescape` and `sys.stdout.reconfigure()` both leave the exception unchanged.

## Fix

All three sites go through one `_decode_soffice_output()` helper decoding with `errors="backslashreplace"`.

@scanny weighed three options on the issue and voted for option 1, `try/except UnicodeDecodeError` falling back to `str(bytes)`. A codec error handler gets the same tolerance in one expression with no rarely-exercised second branch, and escapes only the offending bytes rather than turning a perfectly good line into a `repr`. Option 2 — detect Windows and decode with the locale encoding — would render the filename correctly when the guess is right, but the console output codepage need not match `locale.getpreferredencoding()` — anything that ran `chcp` upstream moves one and not the other — and a wrong guess is silent mojibake rather than a visible escape. It layers cleanly on top of this later if you want it.

`backslashreplace` rather than `replace` because U+FFFD is not encodable in the very codepages this bug is about: a `logging.FileHandler` whose stream uses cp932 or cp936 cannot encode it and drops the whole record, so `replace` would stop the abort and then throw away the diagnostic. `backslashreplace` keeps the line pure ASCII and shows the actual bytes — closer to what option 1 was reaching for.

One behavior change worth flagging: stdout consisting *entirely* of undecodable bytes now yields a non-empty string, so with `returncode == 0` it no longer trips the empty-stdout heuristic the code uses to detect a silent failure. Real `soffice` output carries ASCII scaffolding around the path, so this needs a degenerate payload, but it is a real narrowing of that heuristic — one option 1 shares, since `str()` of non-empty bytes is never empty either. The `b""` that the heuristic actually watches for still decodes to `""`.

## Testing

No LibreOffice needed — stub `subprocess.run` with the bytes `soffice` emits on a Japanese console:

```python
from unittest import mock
from unstructured.partition.common import common

out = mock.Mock(returncode=0, stdout="convert 文章.doc -> 文章.docx".encode("cp932"), stderr=b"")
with mock.patch.object(common.subprocess, "run", return_value=out):
    common.convert_office_doc("文章.doc", "fake-directory", target_format="docx")
```

Raises `UnicodeDecodeError` on `main`; returns normally here.

`test_convert_office_doc_survives_non_utf8_soffice_output` is parametrized over stdout and stderr so both decode paths are covered, and asserts the escaped bytes reach the log and that the record is ASCII-encodable — it fails on `main`, and also fails if the helper is switched to `errors="replace"`.

`test_common.py`: 63 passed / 2 failed here, 61 passed / 4 failed with the same file on `main` — the extra two are the new test's stdout and stderr cases. The other two fail identically on both (`FileNotFoundError: soffice command was not found`; no LibreOffice on this machine). `ruff check .` passes repo-wide and `ruff format --check` passes on both changed files (repo-wide it exits non-zero on the `example-docs/umlauts-non-utf8.md` fixture, identically on `main`).

Not verified against a real `soffice`: the root cause is from @Snowman-s's and @scanny's analysis on the issue, and the CP932 payload is my own reconstruction of it.

Resolves #3652

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4465?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


huuufu:
Prior art: #3830 patched the first of these three sites in 2024; the open question there was a 1-page `.doc` that reproduces this in a unittest. As far as I can determine, no checked-in fixture can: the undecodable bytes are not in the document, they are the input path as `soffice` echoes it back in the console's encoding. The failing files in #3652 were empty docs and renaming to an ASCII name clears the error; the failing file shared in #2449 partitioned fine on a Mac; and under a UTF-8 locale on Ubuntu 24.04 (LibreOffice 24.2.7, what CI installs) I verified `soffice` echoes a multi-byte path as valid UTF-8, so a committed fixture converts cleanly with or without this fix. The test here instead stubs `subprocess.run` with the CP932 bytes a Japanese Windows console emits, exercising both the stdout and stderr sites; it fails on `main` unpatched. Happy to also add a multi-byte-named `.doc` through `partition_doc()` as a live-`soffice` smoke test if that coverage is wanted — it just can't fail on this particular bug.

cragwolfe:
The substantive implementation looks good. The helper covers the initial and retry stdout decodes plus failure-path stderr, malformed bytes remain visible as escapes, and the existing return-code/empty-output behavior is preserved. The parameterized regression also covers both a successful conversion with malformed stdout and a nonzero-exit conversion with malformed stderr.

Before merge, please refresh the release metadata against current `main`: this branch sets `unstructured/__version__.py` to `0.27.6-dev0` and adds a competing `0.27.6-dev0` changelog section, while `main` is already released at `0.27.6` with an existing `0.27.6` section. Please move this entry/version to the current development version while reconciling those metadata files.

(authored by codex)

huuufu:
Rebased onto current `main` (69d50a74). `__version__` is now `0.27.7-dev0` and the changelog entry moved into a new `## 0.27.7-dev0` section above the released `0.27.6`. The code and test are unchanged.

cragwolfe (approved):

PR #4467, opened 2026-09-01
fix: stop encoding detection from returning mojibake

### LINKED ISSUE #4466: bug/encoding-detection-returns-mojibake
**Describe the bug**

`detect_file_encoding()` accepts whatever `charset_normalizer.detect()` returns as long as confidence clears `ENCODE_REC_THRESHOLD`, without checking it against `COMMON_ENCODINGS` (`unstructured/file_utils/encoding.py:74-78`). When detection lands on a codec outside that set that still decodes the bytes, the mojibake becomes the element text and nothing raises. `validate_encoding()` (`encoding.py:53`) exists for exactly this check but has no call sites in the package or the test suite. Affects everything reading text through `read_txt_file()`: `partition_text`, `partition_md`, `partition_html`.

**To Reproduce**

```python
import os, tempfile

for text, encoding in [
    ("会議室予約", "shift_jis"),
    ("café très à Paris", "iso_8859_1"),
    ("Bonjour, ça va très bien. Le café est déjà prêt.", "iso_8859_1"),
]:
    fd, path = tempfile.mkstemp(suffix=".txt"); os.close(fd)
    with open(path, "wb") as f:
        f.write(text.encode(encoding))
    print(repr(partition_text(filename=path)[0].text), "  expected:", repr(text))
    os.unlink(path)
```

Output on `main` (d68ab0fd):

```
'괦귺꿬듚뒴'   expected: '会議室予約'
'café trčs ŕ Paris'   expected: 'café très à Paris'
'Bonjour, ça va trčs bien. Le café est déjŕ pręt.'   expected: 'Bonjour, ça va très bien. Le café est déjà prêt.'
```

`detect()` returns `johab` at 0.8 for the first and `windows-1250` at 1.0 for the third; both decode cleanly, so the `UnprocessableEntityError` guard at `encoding.py:107-111` never fires.

**Second defect, same function**

The inconclusive-detection fallback walks `COMMON_ENCODINGS` in order and takes the first codec that decodes (`encoding.py:80-90`). `iso_8859_1` sits at index 1 and maps all 256 byte values, so it never raises and the remaining 18 of 20 entries — including every CJK codec the list was added for — are unreachable. Reordering does not fix it: with `iso_8859_1` moved last, the latin-1 French bytes above are claimed by `iso_8859_6` as `'cafى trوs ـ Paris'`. The only test covering this path, `test_charset_detection_failure`, mocks both `detect` and `COMMON_ENCODINGS`, so the real list is never exercised.

**Expected behavior**

Text partitions to its original content, or fails loudly — not silently to mojibake.

**Environment Info**

```
OS version:  Windows-11-10.0.26200-SP0
Python version:  3.12.13
unstructured version:  0.27.5
unstructured-inference version:  1.6.13
pytesseract is not installed
Torch version:  2.13.0
Detectron2 is not installed
PaddleOCR is not installed
Libmagic version: file-5.45
magic file from /usr/share/misc/magic
```

`charset_normalizer` 3.5.1. Which encodings get misreported varies by version; accepting an unchecked out-of-set result does not.

**Additional context**

I tried to fix this and backed out, which is why this is an issue and not a PR.

Wiring up `validate_encoding()` and re-running detection constrained to the supported set (`from_bytes(..., cp_isolation=COMMON_ENCODINGS)`) repairs the cases above, but regresses others that work today. `Türkçe bir cümle: şu ğüzel çiçekler İstanbul.` written as cp1254 is detected as `Windows-1254` at confidence 1.0 — correct, but outside `COMMON_ENCODINGS`, so the constrained pass overrides it with `cp1252` and returns `þu ðüzel çiçekler Ýstanbul.`. Letting the detected encoding compete inside the candidate set only swaps which languages break. Widening `COMMON_ENCODINGS` far enough to hold the right answers also admits the wrong ones: `windows-1250` is the correct detection for Czech and the incorrect one for the French line above.

The French case looks unfixable by any rule over these bytes alone — `cp1252` and `windows-1250` agree on most positions and both yield plausible text — so the question is less "which encoding" than which way to be wrong, and whether to stay silent about it. That reads like a policy call for the maintainers rather than something an outside patch should decide. Happy to implement whichever direction you prefer.

Worth flagging alongside this: #4434 proposes routing file-like objects through this same `detect_file_encoding()` fallback, which would widen the blast radius.

BetterAndBetterII:
## Summary

`detect_file_encoding()` accepted a high-confidence `charset_normalizer.detect()` result even when that codec was outside `COMMON_ENCODINGS`. Those guesses (`johab`, `windows-1250`) still decoded, so the `UnprocessableEntityError` guard never fired and the mojibake became element text. That hits everything that reads text through `read_txt_file()` (`partition_text`, `partition_md`, `partition_html`).

A second path in the same function walked `COMMON_ENCODINGS` and took the first codec that decoded. `iso_8859_1` maps every byte, so later CJK codecs on that list were unreachable.

## Fix

- Reject a detected codec that fails `validate_encoding()` (already present, previously unused).
- When detection is inconclusive or out of set, rank candidates with `from_bytes(..., cp_isolation=COMMON_ENCODINGS)` instead of first-success decode.

## Tests

- `detect_file_encoding` recovers the issue's `shift_jis` and `iso_8859_1` samples via filename and file.
- Fallback path with mocked inconclusive `detect()` reaches `shift_jis` instead of stopping at `iso_8859_1`.
- `partition_text` recovers the same three samples.

Fixes #4466

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.27.6
+
+### Fixes
+
+- **Stop encoding detection from returning mojibake for shift_jis and iso_8859_1 text**: `detect_file_encoding()` no longer accepts a high-confidence `charset_normalizer` result outside `COMMON_ENCODINGS` (johab, windows-1250), which decoded cleanly as garbage. When detection is inconclusive or out of set, candidates are ranked inside `COMMON_ENCODINGS` instead of taking the first codec that decodes, so `iso_8859_1` no longer shadows later CJK codecs.
+
 ## 0.27.5
 
 ### Fixes
--- a/test_unstructured/file_utils/test_encoding.py
+++ b/test_unstructured/file_utils/test_encoding.py
@@ -11,6 +11,15 @@ import pytest
 from unstructured.errors import UnprocessableEntityError
 from unstructured.file_utils.encoding import detect_file_encoding
 
+# Samples from Unstructured-IO/unstructured#4466. charset_normalizer.detect() reports
+# johab / windows-1250 for these (outside COMMON_ENCODINGS); both still decode, so the
+# UnprocessableEntityError guard never fires and the mojibake becomes the element text.
+_OUT_OF_SET_DETECT_CASES = [
+    ("会議室予約", "shift_jis"),
+    ("café très à Paris", "iso_8859_1"),
+    ("Bonjour, ça va très bien. Le café est déjà prêt.", "iso_8859_1"),
+]
+
 
 def test_charset_detection_failure():
     """Test encoding detection failure with memory safety checks."""
@@ -69,3 +78,35 @@ def test_decode_failure():
 
         assert exception_memory < 10_000  # Small in-memory footprint
         assert serialized_size < 10_000  # Small serialization footprint
+
+
+@pytest.mark.parametrize(("text", "encoding"), _OUT_OF_SET_DETECT_CASES)
+@pytest.mark.parametrize("via", ["filename", "file"])
+def test_detect_file_encoding_recovers_text_when_detect_returns_out_of_set_codec(
+    text: str, encoding: str, via: str, tmp_path
+):
+    payload = text.encode(encoding)
+    if via == "filename":
+        path = tmp_path / "sample.txt"
+        path.write_bytes(payload)
+        detected, file_text = detect_file_encoding(filename=str(path))
+    else:
+        detected, file_text = detect_file_encoding(file=payload)
+
+    assert file_text == text
+    assert detected not in {"johab", "windows-1250"}
+
+
+def test_detect_file_encoding_fallback_reaches_cjk_instead_of_stopping_at_latin1(tmp_path):
+    """COMMON_ENCODINGS fallback must not take iso_8859_1 just because it never raises."""
+    text = "会議室予約"
+    encoding = "shift_jis"
+    path = tmp_path / "sample.txt"
+    path.write_bytes(text.encode(encoding))
+    detect_result = {"encoding": None, "confidence": None}
+    with patch("unstructured.file_utils.encoding.detect", return_value=detect_result):
+        detected, file_text = detect_file_encoding(filename=str(path))
+
+    assert file_text == text
+    assert file_text != text.encode(encoding).decode("iso_8859_1")
+    assert detected not in {"iso-8859-1", "latin-1"}
--- a/test_unstructured/partition/test_text.py
+++ b/test_unstructured/partition/test_text.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import uuid
+from pathlib import Path
 from typing import Optional, Type
 
 import pytest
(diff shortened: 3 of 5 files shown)


huuufu:
Thanks for picking this up so quickly. I had prototyped the same approach while drafting #4466, so I already had a corpus and harness on hand and ran this branch against it: 28 short/medium plain-text samples in their real-world legacy encodings, `f879f683` vs `main` (`d68ab0fd`), charset_normalizer 3.5.1, Python 3.12.

The fixes are real: the issue's shift_jis and latin-1 samples come back correct, and so do Spanish latin-1, Hebrew cp1255 and Arabic iso-8859-6 samples that are mojibake on `main` today. Raising on actual binary junk instead of returning iso-8859-1 noise is also an improvement. But these rows, all correct on `main`, come back worse:

| sample (correct on `main` today) | on this branch |
|---|---|
| ru cp1251 | raises `UnprocessableEntityError` |
| el cp1253 | raises |
| th cp874 | raises |
| ar cp1256 | raises |
| pl cp1250 / cs cp1250 | raises |
| en cp1252 (curly quotes, em dash) | raises |
| ko cp949 (UHC extension syllables) | raises, or gb18030 mojibake, depending on the sample |
| ja cp932 (IBM extension kanji 髙, 﨑) | mojibake: `郄粼様と粼山様…` |
| tr cp1254 (longer sample) | mojibake: `Iðdýr ovasý…` |

The mechanism is the same in every row: `detect()` returns an encoding that decodes the file correctly (confidence 1.0 everywhere except the cp932 row's 0.875), but that codec is not in `COMMON_ENCODINGS`, so the gate rejects the right answer. From there it either raises — nothing in-set survives the `cp_isolation` pass, so `.best()` returns `None` — or a wrong in-set codec wins: `shift_jis_2004` claims the cp932 bytes and silently swaps the extension kanji (髙→郄, 﨑→粼).

```python
import os, tempfile

samples = [
    ("ru cp1251", "Привет, как дела? Это тестовое сообщение для проверки.", "cp1251"),
    ("th cp874 ", "สวัสดีครับ ยินดีต้อนรับสู่ห้องประชุมของเรา", "cp874"),
    ("ja cp932 ", "髙﨑様と﨑山様が来社されました。", "cp932"),
    ("ko cp949 ", "점심 메뉴는 똠얌꿍으로 예약했습니다.", "cp949"),
]
for name, text, enc in samples:
    fd, path = tempfile.mkstemp(suffix=".txt"); os.close(fd)
    with open(path, "wb") as f:
        f.write(text.encode(enc))
    try:
        got = partition_text(filename=path)[0].text
        print(name, "->", "ok" if got == text else f"MOJIBAKE {got!r}")
    except Exception as e:
        print(name, "->", f"RAISED {type(e).__name__}")
    finally:
        os.unlink(path)
```

On this branch:

```
ru cp1251 -> RAISED UnprocessableEntityError
th cp874  -> RAISED UnprocessableEntityError
ja cp932  -> MOJIBAKE '郄粼様と粼山様が来社されました。'
ko cp949  -> RAISED UnprocessableEntityError
```

On `main` all four print `ok`.

One measurement that may help: changing only the `.best() is None` branch to fall back to the detected encoding when detection was conclusive — raising only when detection failed too, which still covers binary junk — fixed all seven raise rows on my corpus and did not un-fix anything this branch fixes. `None` from the constrained pass turns out to be a decent signal that the file's encoding family just isn't in the supported set, and at that point the unconstrained detection is the best hypothesis available.

The mojibake rows are harder and I don't have a rule that solves them; on my corpus the same codec name at the same confidence can be the right answer for one input and the wrong one for another, so no membership check on the name separates them (`windows-1250` at 1.0 is correct for the pl/cs rows and incorrect for latin-1 French; `cp932` at 1.0 is correct for a shift_jis sample and incorrect for a big5 one — and koi8-r comes back as `shift_jis_2004` at 1.0, which is *in* the set, so the gate passes it on both trees). I also tried deciding these on chaos/coherence margins, and the numbers order the wrong way: wrong cp775-for-Spanish beats right latin-1 by a larger coherence margin (+0.107) than right cp1254-for-Turkish beats wrong latin-1 (+0.077). That residue looks like the policy half of #4466 rather than something this patch has to settle — the raise rows are the part that seems clearly fixable. Happy to share the full corpus and harness if useful.

PR #4469, opened 2026-09-03
Fix DOCX text_as_html duplicating merged-cell text instead of colspan/rowspan

qued:
## Summary

DOCX table extraction produced two different representations of a merged cell: `.text` included its content once, but `.metadata.text_as_html` repeated the content into every `<td>` the merge visually covered, with no `colspan`/`rowspan` attribute marking that a merge had happened at all.

Root cause: `_convert_table_to_html` built its HTML from `python-docx`'s `row.cells`, which resolves both horizontal (`gridSpan`) and vertical (`vMerge="continue"`) merges by yielding the *same* underlying cell content at every grid position the merge spans. That already-expanded grid was serialized straight into HTML with one `<td>` per matrix position and no span-collapsing step.

Fix: track the underlying `tc` XML element's identity (not cell text) when building the matrix — the same `tc` object appears at every grid position a merge covers — then collapse runs of identical identity into `(colspan, rowspan)` before emitting `<td>`s. DOCX only permits rectangular merges, so there's no irregular-region case to handle.

A new shared helper (`collapse_matrix_of_keyed_cells_to_spans`) does the collapsing; the existing `htmlify_matrix_of_cell_texts` (still used unchanged by the pptx and HTML-parser partitioners) now shares its cell-escaping logic with the new span-aware path via an extracted `_format_td` helper.

## Test plan

- [x] `partition_docx` on `example-docs/docx-tables.docx` (the merged-cell fixture) now produces `text_as_html` with correct `colspan`/`rowspan` and no duplicated cell text — added a behavioral regression test through the public `partition_docx` API.
- [x] Existing `test_docx.py` fixtures pinning the old duplicated-text/no-span output updated to the new expected output.
- [x] Checked downstream consumers that assumed DOCX tables never carry spans:
  - `unstructured/metrics/table/table_extraction.py`'s span-aware grid reconstruction already handles `colspan`/`rowspan` correctly (written for other span-producing sources) — traced by hand against the merged-cell fixture.
  - Chunking's table splitter handles real spanned DOCX tables without crashing at multiple `max_characters` values. Found one pre-existing (not introduced here) limitation: splitting between rows that share a `rowspan` can drop that cell's data in the later chunk — this was always latent for any spanned HTML source, just never exercised for DOCX before since DOCX never emitted spans until now. Not fixed in this PR; flagging as a possible follow-up.
- [x] Full relevant test suites green (partition/docx, common/html_table, chunking, metrics/table), lint clean.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4469?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.27.9
+
+### Fixes
+
+- **Scope `rowspan` chunk-boundary protection to its actual `<thead>`/`<tbody>`/`<tfoot>` row-group, not the whole table.** A `rowspan="0"` cell in a short `<thead>` (or any positive `rowspan`) was previously resolved as reaching to the end of the entire table rather than the end of its own row-group, so a real, bounded `<tbody>` following it could get swallowed into one unbounded chunk alongside the header. Rows are now grouped by their actual containing section (or the whole table, when there is none), and a span can no longer bind rows across a real section boundary.
+
 ## 0.27.8
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.27.13
+
+### Fixes
+
+- **Preserve a cell's actual HTML when correcting a clipped or `rowspan="0"` `rowspan`, and bound the `<thead>` row's own original occurrence even when header repetition is configured.** Correcting an overreaching `rowspan` previously reconstructed the cell from its plain text, discarding any nested table, hyperlink, image, or other markup and attribute the source cell carried. Only the `rowspan` attribute is now rewritten on a copy of the real cell; everything else survives unchanged. Separately, a `<thead>` row's ORIGINAL, wrapper-less occurrence was being exempted from this correction whenever header repetition was merely configured — not only for an actual repeated/carried copy (a different artifact, already safely wrapped in its own real `<thead>`) — leaving that first occurrence still vulnerable to exactly the cross-row-group corruption this mechanism exists to prevent. The exemption is now scoped correctly: every row is bounded to its own row-group, full stop.
+
 ## 0.27.12
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,56 +1,8 @@
-## 0.27.14
-
-### Fixes
-
-- **Bound a singleton oversized row's `rowspan` before splitting it cell-by-cell.** A rowspan-bound row too large to fit any chunk even alone was handed to the cell splitter with its original, uncorrected `rowspan` still attached — the self-correcting rewrite only applied to rows that pass through the normal row accumulator. A public caller reassembling separately-emitted chunks (`reconstruct_table_from_chunks()`) could see that stale span reach into rows from a later chunk, the exact corruption this mechanism exists to prevent. The row's own bound is now applied before it's split.
-
-## 0.27.13
-
-### Fixes
-
-- **Preserve a cell's actual HTML when correcting a clipped or `rowspan="0"` `rowspan`, and bound the `<thead>` row's own original occurrence even when header repetition is configured.** Correcting an overreaching `rowspan` previously reconstructed the cell from its plain text, discarding any nested table, hyperlink, image, or other markup and attribute the source cell carried. Only the `rowspan` attribute is now rewritten on a copy of the real cell; everything else survives unchanged. Separately, a `<thead>` row's ORIGINAL, wrapper-less occurrence was being exempted from this correction whenever header repetition was merely configured — not only for an actual repeated/carried copy (a different artifact, already safely wrapped in its own real `<thead>`) — leaving that first occurrence still vulnerable to exactly the cross-row-group corruption this mechanism exists to prevent. The exemption is now scoped correctly: every row is bounded to its own row-group, full stop.
-
-## 0.27.12
-
-### Fixes
-
-- **Rewrite chunked-table `rowspan` handling to self-correct emitted span values, instead of tracking which chunk boundaries are unsafe to cross.** A clipped or `rowspan="0"` cell's emitted HTML previously always carried its original declared value, relying on chunk-boundary bookkeeping to keep it away from rows it didn't truly cover. That value is now rewritten, at the point each chunk is assembled, to the number of rows actually present in that same chunk — an emitted `rowspan` can no longer overreach regardless of what else the chunker decides to pack alongside it. Also fixes two remaining gaps in the boundary bookkeeping this replaces: an overdeclared `<thead>` `rowspan` was exempted from clipping even when header repetition isn't active for it (so nothing else protected it), and the accumulator compared a candidate row-group against the FIRST row it had accumulated rather than the most recent one, which could miss a real transition when an earlier and later row happened to share the same row-group identity.
-
-## 0.27.11
-
-### Fixes
-
-- **Also treat a clipped positive `rowspan` as a hard chunk boundary, not only `rowspan="0"`.** A positive `rowspan` that declares more rows than its own `<thead>`/`<tbody>`/`<tfoot>` row-group actually has is clipped to that row-group when grouping rows for chunking, but its emitted HTML still carries the original, uncorrected declared value. If such a group got packed into the same chunk as a following row-group's rows, that value would legitimately reach into rows it was never meant to bind, once section wrappers are stripped. A row-group change is now also a hard chunk boundary whenever the preceding group's span was clipped this way — a `<thead>` row's positive span is unaffected, since it's already handled separately as a repeated/carried-forward header.
-
-## 0.27.10
-
-### Fixes
-
-- **Never let a chunk's rows span more than one source `<thead>`/`<tbody>`/`<tfoot>` row-group when a `rowspan="0"` cell is involved.** Row-group boundaries were correctly used to compute rowspan-bound groups, but the chunk accumulator could still pack an already-grouped `rowspan="0"` header together with a following row-group's rows when both fit the character budget — since section wrappers are stripped from emitted chunk HTML, the reparsed, flattened result let that span reach into rows it was never meant to bind. A row-group change is now a hard chunk boundary whenever a `rowspan="0"` row is involved; ordinary rows (and rows with a real, positive `rowspan` meant to carry across a boundary, e.g. a repeated header) are unaffected.
-
-## 0.27.9
-
-### Fixes
-
-- **Scope `rowspan` chunk-boundary protection to its actual `<thead>`/`<tbody>`/`<tfoot>` row-group, not the whole table.** A `rowspan="0"` cell in a short `<thead>` (or any positive `rowspan`) was previously resolved as reaching to the end of the entire table rather than the end of its own row-group, so a real, bounded `<tbody>` following it could get swallowed into one unbounded chunk alongside the header. Rows are now grouped by their actual containing section (or the whole table, when there is none), and a span can no longer bind rows across a real section boundary.
-
-## 0.27.8
-
-### Fixes
-
-- **Don't drop a table's trailing rows when a `rowspan` reaches past the last row, and honor `rowspan="0"`.** The rowspan-aware chunking boundary added in 0.27.7 could silently lose an entire table's tail when a declared `rowspan` named more rows than the table had, and treated `rowspan="0"` (HTML's "span every remaining row") as no span at all, letting a chunk boundary fall through it. Both now resolve to "the rest of the table" and are never split or dropped.
-
-## 0.27.7
-
-### Fixes
-
-- **Keep rows bound by an active `rowspan` in the same table chunk.** Table chunking split purely on row *text* length, unaware that a `rowspan` crossing the boundary would overclaim rows in one chunk and shift cells into the wrong column in the next. Such rows are now kept together as one unit, same tolerance already given a single oversized row or cell.
-
 ## 0.27.6
 
 ### Fixes
 
-- **Stop duplicating merged-cell text in DOCX `text_as_html`.** A merged cell (`gridSpan`/`vMerge`) was repeated into every `<td>` its merge visually covered, with no `colspan`/`rowspan` attribute marking the merge. Merged cells are now emitted once, with `colspan`/`rowspan` reflecting the true geometry.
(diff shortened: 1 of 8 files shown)


aadland6:
---

## Summary

| # | Severity | Issue | Primary location |
| --- | --- | --- | --- |
| 1 | High | Rowspan crossing a `<thead>`/`<tbody>`/`<tfoot>` boundary is clipped at the section edge, shifting columns | `unstructured/chunking/base.py` (`_iter_rowspan_bound_row_groups`, `_group_last_idx`) |
| 2 | High | Repeated headers re-emit the pre-clip `rowspan` on every continuation chunk, shifting columns (and reconstruction bakes it in) | `unstructured/chunking/base.py` (`_as_header_row_html`); `unstructured/chunking/dispatch.py` (`_first_carried_header_rows`) |
| 3 | Medium-High | A rowspan-bound row group that can't fit is emitted whole, exceeding the hard `max_characters`/`max_tokens` limit | `unstructured/chunking/base.py` (`_iter_subtables`, final `else` branch) |
| 4 | Medium | Nested DOCX tables still duplicate merged-cell text in `text_as_html` | `unstructured/partition/docx.py` (`_convert_table_to_html` → `iter_cell_block_items`) |
| 5 | Medium | Splitting an oversized *non-empty* merged cell drops its `colspan` | `unstructured/chunking/base.py` (`_iter_cell_splits`) |
| 6 | Medium | `HtmlCell.html` drops `colspan`/`rowspan` on *empty* cells, so cell-split chunks lose merge geometry | `unstructured/common/html_table.py` (`HtmlCell.html`); consumed by `_CellAccumulator.flush` |

---

## 1. (High) Rowspan crossing a section boundary is clipped, shifting every following column

**Where:** `unstructured/chunking/base.py`, `_HtmlTableSplitter._iter_rowspan_bound_row_groups` and `_group_last_idx`; the shift is then made permanent by `reconstruct_table_from_chunks()` in `unstructured/chunking/dispatch.py`.

**What's wrong:** A positive `rowspan` is valid HTML when it extends from `<thead>` into `<tbody>` (or `<tbody>` into `<tfoot>`). The continuation rows in the next section deliberately omit the covered column. `_group_last_idx()` groups rows strictly by their row-group parent, and `_iter_rowspan_bound_row_groups()` treats any span that reaches past its own row-group as *clipped*, rewriting it down to the section's own row count and making the section boundary a hard chunk split. The covering cell is then gone from the continuation rows, so every remaining cell in those rows slides one column to the left. This fires whenever such a table is split — including at the default `max_characters=500` — not just at tiny windows.

**Reproduced:** For

```html
<table>
  <thead><tr><th rowspan="2">Name</th><th>Field</th></tr></thead>
  <tbody>
    <tr><td>Age</td></tr>
    <tr><td>person00</td><td>val00 extra padding text</td></tr>
    <tr><td>person01</td><td>val01 extra padding text</td></tr>
  </tbody>
</table>
```

`chunk_by_title(..., max_characters=80, repeat_table_headers=False)` produces chunk 0 `<table><tr><td>Name</td><td>Field</td></tr></table>` (the `rowspan` is stripped) and chunk 1 beginning `<table><tr><td>Age</td></tr>...`. Reconstruction yields `<table><tr><td>Name</td><td>Field</td></tr><tr><td>Age</td></tr>...`. `Age` is now in column 0; it belongs in column 1 under `Field` (the `Name` cell originally covered column 0 of that row).

**Proposed fix:** Stop clamping a *positive* `rowspan` to its row-group's last row. Bind rows by the span's true reach (clamped only to the table's last row, `n-1`) so a section-crossing span keeps every row it covers in the same atomic group, and never clips the covering cell out of a continuation row. Retain the row-group clamp only for `rowspan="0"` (`HtmlRow.max_rowspan is None`), which the HTML spec scopes to the containing row group. Concretely, in `_iter_rowspan_bound_row_groups`, for the positive-span branch set `reach[idx] = min(idx + row.max_rowspan - 1, n - 1)` and `clipped[idx] = declared_reach > n - 1`, and set each row's `bound` from the resulting merged group extent rather than from `own_group_last`.

**Note on the ambiguous case:** An *overdeclared* header span whose body rows are already full-width (a cell in every column) is genuinely different from a span that legitimately covers a short row, but the two cannot be distinguished from row-group membership alone. Binding by true reach is the safe choice: it never shifts columns. The trade-off is that an overdeclared span may keep a slightly larger group together than strictly necessary, which is acceptable since correctness outranks split granularity. The existing in-tree assertions only cover the overdeclared-with-full-body case, so they will need to be updated to reflect the corrected behavior.

---

## 2. (High) Repeated headers re-emit the pre-clip `rowspan` on continuation chunks

**Where:** `unstructured/chunking/base.py`, `_HtmlTableSplitter._as_header_row_html` (via `_header_rows_html` / `_prepend_repeated_headers`); reconstruction in `unstructured/chunking/dispatch.py`, `_first_carried_header_rows` / `_merge_table_chunks`.

**What's wrong:** Header repetition is on by default (`repeat_table_headers=True`). Continuation `<thead>` markup is rebuilt from `HtmlRow.source_html`, which is captured *before* compactification and *before* rowspan clipping, so it still carries the original, possibly over-reaching `rowspan`. The first-chunk header is correctly clipped, but every continuation chunk re-injects the unclipped span. When the span is larger than the number of header rows actually prepended, it reaches down into the first body row of the continuation and eats that row's first column. Reconstruction then makes it permanent: `_first_carried_header_rows()` prefers the carried (unclipped) header rows as the canonical `<thead>` and discards the clipped first-chunk header, so the reconstructed table also carries the over-reaching span. First-chunk and continuation-chunk HTML additionally disagree about the same table's header geometry.

**Reproduced:** For

```html
<table>
  <thead><tr><th rowspan="2">Region</th><th>Quarter</th></tr></thead>
  <tbody>
    <tr><td>Northwest Territory</td><td>Q1 FY2026</td></tr>
    <tr><td>Southwest Territory</td><td>Q2 FY2026</td></tr>
    <tr><td>Midwest Territory</td><td>Q3 FY2026</td></tr>
  </tbody>
</table>
```

`chunk_by_title(..., max_characters=55)` (default header repetition) emits chunk 0 `<table><tr><td>Region</td><td>Quarter</td></tr></table>` (clipped), but each continuation chunk is `<table><thead><tr><th rowspan="2">Region</th><th>Quarter</th></tr></thead><tr><td>Northwest Territory</td><td>Q1 FY2026</td></tr></table>`. The one-row `<thead>` declares `rowspan="2"`, so once the `<thead>` is dropped, `Region` covers into the body row and `Northwest Territory` / `Q1 FY2026` shift right. Reconstruction preserves the `rowspan="2"`.

**Proposed fix:** Build the repeated header from the clipped header row, not from `source_html`. In `_as_header_row_html` (or in `_header_rows_html`), clip each header cell's `rowspan` down to the number of header rows actually prepended — the count in `_header_rows` / `carried_over_header_row_count` — using the existing `HtmlRow.row_clipped_to_rows()` / `html_clipped_to_rows()` before converting `<td>`→`<th>`. With a single prepended header row the `rowspan` is clipped to 1 (removed), so the continuation `<thead>` matches the clipped first-chunk header and reconstruction stays correct.

---

## 3. (Medium-High) A rowspan-bound group that can't fit is emitted whole, exceeding the hard limit

**Where:** `unstructured/chunking/base.py`, `_HtmlTableSplitter._iter_subtables`, final `else` branch (the "emit whole" fallback for a group that doesn't fit an empty window).

**What's wrong:** When a rowspan-bound group does not fit even in an empty chunking window, the splitter adds it and flushes it unchanged, producing a chunk larger than `max_characters`/`max_tokens`. This contradicts `_TableChunker`'s "`maxlen` or smaller" contract and can yield arbitrarily large chunks for tables with long vertical merges, which downstream embedding/model calls (that enforce the size limit) can reject. The in-code comment claims this matches "the same tolerance the codebase already grants a single oversized row/cell," but that is inaccurate: a single oversized row is cell-split and a single oversized cell is text-split — this group path is the only one that emits over-limit content.

**Reproduced:** A three-row table with a `rowspan="3"` covering cell and ~55 characters of total text, chunked with `chunk_by_title(..., max_characters=50)`, returns a single `TableChunk` of text length 55 (HTML length 142) instead of chunks at or below the hard maximum.

**Proposed fix:** Replace the "emit whole" fallback with a size-respecting split, reusing the oversized-row path (`_iter_row_splits` / `_iter_cell_splits`). Split the group row-by-row (then cell-by-cell / text-split as needed), and for each emitted fragment rewrite the covering cell's `rowspan` to the number of rows actually present in that fragment (via `row_clipped_to_rows`), re-injecting the covering cell as `rowspan="1"` into any continuation row that would otherwise be missing its covered column so columns stay aligned. Splitting a vertical merge below its own size unavoidably repeats the covering cell's text across fragments rather than preserving it as a single merge; that is the correct trade-off for honoring the hard limit, and it matches how the other oversized paths behave.

---

## 4. (Medium) Nested DOCX tables still duplicate merged-cell text in `text_as_html`

**Where:** `unstructured/partition/docx.py`, `_convert_table_to_html` → `iter_cell_block_items` (the `DocxTable` branch, lines ~521-523).

**What's wrong:** The merge-collapse fix is applied only to the outermost table. A nested table is flattened into its parent `<td>` by walking `row.cells` directly (`yield from (text for text, _ in iter_row_cells(row))`), and `python-docx` repeats a merged cell's origin `tc` at every grid position it covers (for both `gridSpan` and `vMerge="continue"`). So a merged cell's text is concatenated once per covered position. `.text` (which walks `tr.tc_lst` and skips `vMerge="continue"`) does not duplicate, so `.text` and `metadata.text_as_html` disagree.

**Reproduced:** `python-docx` `row.cells` for a 2×2 nested table whose top row is horizontally merged to `MERGED` yields `['MERGED', 'MERGED', 'c', 'dd']`; the parent `<td>` therefore receives `parent MERGED MERGED c dd` (`MERGED` twice), while `.text` contains a single `MERGED`. A fully-merged 2×2 nested table repeats its text four times.

**Proposed fix:** Apply the same merge-collapse to the nested table before flattening its text. Replace the nested-table branch with a keyed-collapse pass and emit each origin cell's text once:

```python
elif isinstance(table := block_item, DocxTable):
    matrix = [list(iter_row_merge_keyed_texts(row)) for row in table.rows]
    for spanned_row in collapse_matrix_of_keyed_cells_to_spans(matrix):
        yield from (text for text, _colspan, _rowspan in spanned_row)
```

This dedups merged cells exactly as the outer table already does and recurses correctly for tables nested more than one level deep (via `iter_row_merge_keyed_texts` → `iter_row_cells` → `cell_text` → `iter_cell_block_items`), bringing `text_as_html` back into agreement with `.text`.

---

## 5. (Medium) Splitting an oversized non-empty merged cell drops its `colspan`

**Where:** `unstructured/chunking/base.py`, `_HtmlTableSplitter._iter_cell_splits` (lines ~1411 and ~1416).

**What's wrong:** When a single cell's text is too large for the window, `_iter_cell_splits()` text-splits it and hardcodes each fragment as a bare `<table><tr><td>{text}</td></tr></table>`, discarding the cell's `colspan`/`rowspan`. `reconstruct_table_from_chunks()` then rebuilds the fragments as plain single-column rows, losing the merge geometry this PR exists to preserve.

**Reproduced:** A one-row table containing `<td colspan="2">` with thirty `word` tokens, chunked with `max_characters=50`, produces ten fragments, each `<table><tr><td>word word word</td></tr></table>` with no `colspan`. Reconstruction yields ten single-column rows and zero `colspan` attributes instead of a two-column-spanning cell.

**Proposed fix:** Carry the cell's span onto each emitted fragment. Emit the `<td>` using the shared `_format_td(text, cell.colspan, rowspan=1)` helper (from `unstructured/common/html_table.py`) instead of a hardcoded `<td>` string. Preserve `colspan` on every fragment (each fragment is its own one-row table, so the columns line up on reconstruction); clip `rowspan` to `1` on fragments, since a vertical span across the newly created fragment-rows would over-reach.

---

## 6. (Medium) `HtmlCell.html` drops `colspan`/`rowspan` on empty cells

**Where:** `unstructured/common/html_table.py`, `HtmlCell.html`; consumed by `unstructured/chunking/base.py`, `_CellAccumulator.flush` (the cell-split path used when a rowspan-bound group is a single row that doesn't fit the window).

**What's wrong:** `HtmlCell.html` serializes any cell with no text as a bare `"<td/>"`, discarding span attributes:

```python
return etree.tostring(self._td, encoding=str) if self.text else "<td/>"
```

Row-level emission (`row.html` / `html_clipped_to_rows`) keeps the attributes, but `_CellAccumulator.flush()` uses `HtmlCell.html`, so when an oversized row is split on cell boundaries an empty merged cell loses its span and the remaining cells in that sub-row shift left. This now matters because DOCX empty horizontal merges emit `<td colspan="2"/>`.

**Reproduced:** `HtmlCell` on `<td colspan="2" rowspan="3"></td>` reports `colspan=2, rowspan=3` but `.html == "<td/>"`. For `<table><tr><td colspan="2"></td><td>aaaaaaaaaaaaaaaaaaaa</td><td>bbbbbbbbbbbbbbbbbbbb</td></tr></table>` split at `max_characters=30`, the first sub-chunk is `<table><tr><td/><td>aaaaaaaaaaaaaaaaaaaa</td></tr></table>` — the `colspan="2"` is gone, so `aaaa…` lands in column 1 instead of column 2.

**Proposed fix:** Preserve span attributes even when the cell is empty. Emit the empty cell via the shared `_format_td(self.text, self.colspan, self.rowspan)` helper (which already renders `<td colspan="2"/>` for an empty spanned cell), or, minimally, only shortcut to `"<td/>"` when the cell has no span attributes:

```python
return _format_td(self.text, self.colspan, self.rowspan)
```

This keeps compact `<td/>` output for genuinely attribute-free empty cells while retaining `colspan`/`rowspan` on empty merged cells.

*(Findings 5 and 6 share a theme — the cell-split path loses spans — but they are distinct bugs in different functions with different fixes: #5 is the hardcoded `<td>` string in `_iter_cell_splits` for oversized non-empty cells, #6 is the empty-cell shortcut in `HtmlCell.html`. Both should be fixed.)*

---

aadland6 (approved):

PR #4471, opened 2026-09-05
fix: partition multi-section DOCX files in linear time

### LINKED ISSUE #3592: bug(docx): O(N^2) time on large section+paragraph count
**Summary**
A DOCX with a large number of sections combined with a large number of paragraphs triggers an O(N^2) process to determine the paragraphs in each section. This greatly slows partitioning of these documents.

**To Reproduce**
The original document in this case had 470 sections and 31,723 paragraphs. Partitioning took in excess of 40 minutes.

**Expected behavior**
Partitioning time should be O(N) with the number of paragraphs.

**Diagnosis**
The upstream `python-docx` `SectBlockIterator` algorithm is effectively O(N^2) when called once for each section.
https://github.com/python-openxml/python-docx/blob/master/src/docx/oxml/section.py#L437

It gets all the block items before the section end and then subtracts all the block items before the prior section end. This is effectively N^2 with the number of block-items (paragraphs/tables) in the document.

Work out an algorithm to get only the block items between the prior section end and the current section end. Something roughly like:
```python
section_p = Section._sectPr
prior_section_p = section_p.xpath(./[prior-sibling]::w:sectPr)
block_item = prior_section_p
while block_item := block_item.next():
    yield block_item
    if block_item = section_p:
        break
```
There will be some edge-cases to work out in tests because of `w:p/w:sectPr` and `w:sectPr` diversity between body and document-end `w:sectPr`s.

linhongyu510:
## Summary

This pull request fixes #3592 by replacing per-section `Section.iter_inner_content()` calls with a single document-body traversal. Section boundaries are detected at paragraph-level `sectPr` elements so page breaks, headers, body content, and footers retain their existing order.

On the current main branch with python-docx 1.2.0, traversing synthetic documents with 20 paragraphs per section took:

- 100 sections: 0.43s
- 200 sections: 3.61s
- 400 sections: 40.37s

With this change, the isolated traversal path took 0.008s, 0.010s, and 0.024s respectively. The full partition pipeline still includes text classification and metadata work; these numbers intentionally isolate the repeated section-prefix scan fixed here.

## Tests

- `uv run --no-sync pytest -q test_unstructured/partition/test_docx.py` (79 passed)
- `uv run --no-sync ruff check .`
- `uv run --no-sync ruff format --check .`
- `git diff --check`
- Regression test verified against the previous implementation: it fails because `Section.iter_inner_content()` is called.

`make test-extra-docx` additionally ran 96 tests successfully, while 6 legacy `.doc` conversion tests failed because the local LibreOffice process did not produce the requested `.docx` files. The complete DOCX test module passes independently.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4471?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/release-version-alert.yml
+++ b/.github/workflows/release-version-alert.yml
@@ -10,11 +10,9 @@ on:
 jobs:
   check-version:  
     runs-on: ubuntu-latest
-    env:
-      SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
     steps:
       - name: Checkout codes
-        uses: actions/checkout@v7
+        uses: actions/checkout@v4 
       - name: Get PR information
         id: pr-info
         run: |
@@ -51,7 +49,7 @@ jobs:
       - name: Restore Message from Cache
         if: env.SKIP_STEPS != 'true'
         id: restore-cache
-        uses: actions/cache/restore@v6
+        uses: actions/cache/restore@v4
         with: 
           path: message_cache.txt
           key: message-cache-${{ env.MESSAGE_HASH  }}
@@ -71,16 +69,15 @@ jobs:
           cat message_cache.txt
       - name: Store Message in Cache
         if: env.SKIP_STEPS != 'true'
-        uses: actions/cache/save@v6
+        uses: actions/cache/save@v4
         with:
           path: message_cache.txt
           key: message-cache-${{ env.MESSAGE_HASH }}
       - name: Slack Notification
-        if: env.SKIP_STEPS != 'true' && env.SLACK_BOT_TOKEN != ''
-        uses: slackapi/slack-github-action@v4
+        if: env.SKIP_STEPS != 'true'
+        uses: slackapi/slack-github-action@v1.24.0
         with:
-          method: chat.postMessage
-          token: ${{ env.SLACK_BOT_TOKEN }}
-          payload: |
-            channel: C05S1QMKL5D
-            text: ${{ toJSON(env.SLACK_MESSAGE) }}
+          channel-id: 'C05S1QMKL5D'
+          slack-message: ${{ env.SLACK_MESSAGE }}
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
--- a/.github/workflows/release-version-alert.yml
+++ b/.github/workflows/release-version-alert.yml
@@ -10,9 +10,11 @@ on:
 jobs:
   check-version:  
     runs-on: ubuntu-latest
+    env:
+      SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
     steps:
       - name: Checkout codes
-        uses: actions/checkout@v4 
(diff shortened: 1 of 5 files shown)


cragwolfe:
One correctness issue:

- **High — a document whose last section ends in a paragraph-level `sectPr` now fails instead of completing.** In `unstructured/partition/docx.py:425`, the boundary transition unconditionally calls `next(sections)`. A body-level `w:sectPr` is optional, and python-docx exposes the paragraph-level section properties that exist without synthesizing a terminal body section. For that accepted shape, the old `for section in document.sections` loop completed normally; the new generator advances past the final section and surfaces `RuntimeError: generator raised StopIteration`, causing `partition_docx()` to fail the whole document. Please use `next(sections, None)` and return when there is no next section, immediately after emitting the completed section's footer. Returning there is important so the footer is not emitted again after the body loop. A focused save/reload regression should remove the body-level `sectPr` while retaining a paragraph-level one and assert the full header/body/footer sequence exactly once. The new 25-section test does not cover this because `Document.add_section()` retains the terminal body sentinel.

(authored by codex)

linhongyu510:
Addressed the paragraph-level terminal `sectPr` case in `819e7588` and rebased the branch onto current `main` (resolving the changelog/version conflict). The new save/reload regression moves the final body `sectPr` onto the last paragraph and asserts the exact `Header`, `Body`, `Footer` sequence once. It fails on the previous head with `RuntimeError: generator raised StopIteration`; the fix uses `next(sections, None)` and returns immediately after emitting the final footer when no next section exists. Verification on the rebased head: `test_unstructured/partition/test_docx.py` 85 passed; Ruff check and format check passed; `git diff --check` passed.

cragwolfe:
One production-readiness issue remains:

- **Medium — release metadata is still attached to an already committed version, so required lint CI cannot pass.** `CHANGELOG.md:5` adds this fix under `0.27.6`, while `unstructured/__version__.py:1` remains `0.27.6`. `make check` fails on this exact head with `Error: there is already a commit associated with version 0.27.6.` Move the changelog entry under the next version heading and update `unstructured/__version__.py` to the same version, following the repository’s current versioning convention.

(authored by codex)

linhongyu510:
Moved this fix to the next release version in f31f9621: `CHANGELOG.md` and `unstructured/__version__.py` now both use `0.27.7`. Verification on the new head: `test_unstructured/partition/test_docx.py` 85 passed; Ruff check and format check passed; version metadata agrees and is newer than `origin/main` (`0.27.6`); `git diff --check` passed. The full local `make check` now gets past the previous committed-version error and stops only at the script’s macOS GNU sed requirement (`sed version must be >= 4.3`), which is environment-specific.

cragwolfe (approved):

PR #4478, opened 2026-09-07
fix: preserve MathML content in HTML partitioning

fireblade2534:
## Summary

This PR adds support for preserving inline MathML content in the v1 HTML parser used by `partition_html()`.

Previously, `<math>` was not assigned a custom element class. As a result, it fell back to `DefaultElement`, whose contents are intentionally skipped. This could silently remove mathematical expressions from otherwise valid extracted text.

For example, HTML such as:

```html
<p>
    The error rate is
    <math>
        <semantics>
            <annotation encoding="application/x-tex">
                10^{-4}
            </annotation>
        </semantics>
    </math>
    per token.
</p>
```

could be extracted as:

```text
The error rate is per token.
```

This PR introduces a `Math` phrasing element that preserves a textual representation of the expression when one is available.

## Changes

* Adds a `Math` class derived from `Phrasing`.
* Registers `<math>` elements with the v1 HTML parser.
* Searches for recognized TeX/LaTeX annotations within the MathML subtree.
* Supports the following annotation encodings:

  * `application/x-tex`
  * `tex`
  * `latex`
  * `application/x-latex`
* Falls back to the `<math>` element's `alttext` attribute when no supported TeX annotation is available.
* Preserves text following the closing `</math>` element by emitting the element tail using the existing phrasing behavior.

The parser prefers an explicit TeX representation because flattening presentation MathML with `itertext()` can lose mathematical structure. For example, a superscript expression such as `10^-4` could otherwise be flattened into ambiguous plain text such as `10-4`.

## Why

The HTML parser's documented principle is that the browser rendering is the document.

However, before this change, standards-based MathML content could be rendered correctly by a browser while being completely removed during HTML partitioning because `<math>` was treated as an unrecognized element.

This is especially problematic because the failure is silent. Surrounding paragraph text remains intact, so consumers of `partition_html()` may receive syntactically valid but incomplete text.

Treating `<math>` as a specialized phrasing element allows mathematical content to participate naturally in surrounding paragraph text without requiring the parser to understand every individual Presentation MathML element such as `<mrow>`, `<msup>`, `<mfrac>`, `<mi>`, or `<mn>`.

## Scope

This PR intentionally does not attempt to implement a complete MathML-to-text converter.

It currently preserves mathematical content when a reliable textual representation is directly available through:

1. a recognized TeX/LaTeX `<annotation>`, or
2. the `<math>` element's `alttext` attribute.

Other representations, such as Content MathML, arbitrary `annotation-xml` formats, OpenMath, SVG, or unsupported custom annotation encodings, are outside the scope of this change.

## Example

Input:

```html
<p>
    Complexity decreases from
    <math alttext="O(n^2)"></math>
    to
    <math>
        <semantics>
            <annotation encoding="application/x-tex">
                O(n^2/r)
            </annotation>
        </semantics>
    </math>.
</p>
```

Before:

```text
Complexity decreases from to .
```

After:

```text
Complexity decreases from O(n^2) to O(n^2/r).
```

## Testing

Tests have been added for MathML behavior including:

* extracting `application/x-tex` annotations,
* supported TeX/LaTeX encoding variants,
* falling back to `alttext`,
* preferring a supported TeX annotation when both annotation and `alttext` are present,
* preserving surrounding and tail text,
* multiple inline `<math>` elements in one paragraph,
* ignoring unsupported annotation encodings,
* realistic nested Presentation MathML,
* interaction with existing phrasing/emphasis behavior,
* and regression coverage for ordinary phrasing elements.

Reviewer verification can also be performed with:

```python

html = r"""
<p>
    The loss difference is on the order of
    <math>
        <semantics>
            <annotation encoding="application/x-tex">
                10^{-4}
            </annotation>
        </semantics>
    </math>.
</p>
"""

elements = partition_html(text=html)

for element in elements:
    print(element.text)
```

Expected output:

```text
The loss difference is on the order of 10^{-4}.
```

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4478?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.27.6
+
+### Fixes
+
+- **Stop duplicating merged-cell text in DOCX `text_as_html`.** A merged cell (`gridSpan`/`vMerge`) was repeated into every `<td>` its merge visually covered, with no `colspan`/`rowspan` attribute marking the merge; merged cells are now emitted once, with `colspan`/`rowspan` reflecting the true geometry. Since DOCX tables can now carry real spans, table chunking was also made rowspan-aware, so a chunk boundary can no longer split a table in a way that misattributes a spanned cell's rows to the wrong columns.
+
 ## 0.27.5
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - **Stop duplicating merged-cell text in DOCX `text_as_html`.** A merged cell (`gridSpan`/`vMerge`) was repeated into every `<td>` its merge visually covered, with no `colspan`/`rowspan` attribute marking the merge; merged cells are now emitted once, with `colspan`/`rowspan` reflecting the true geometry. Since DOCX tables can now carry real spans, table chunking was also made rowspan-aware, so a chunk boundary can no longer split a table in a way that misattributes a spanned cell's rows to the wrong columns.
 
+### Enhancements
+
+- **Add basic MathML support.** `<math>` elements now extract a textual representation from alttext or supported TeX/LaTeX `<annotation>` values, including case-insensitive annotation encodings. Inline math remains part of surrounding text, while display="block" math is emitted as a separate element boundary. KaTeX's paired MathML/HTML output is also handled without duplicating the same expression.
+
 ## 0.27.5
 
 ### Fixes
--- a/test_unstructured/chunking/test_base.py
+++ b/test_unstructured/chunking/test_base.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import io
 import logging
 from typing import Any, Sequence
 
@@ -28,6 +29,7 @@ from unstructured.chunking.base import (
     is_title,
 )
 from unstructured.chunking.dispatch import reconstruct_table_from_chunks
+from unstructured.chunking.title import chunk_by_title
 from unstructured.common.html_table import HtmlCell, HtmlRow, HtmlTable
 from unstructured.documents.elements import (
     CheckBox,
@@ -1576,7 +1578,14 @@ class Describe_TableChunker:
             repeat_table_headers=True,
         )
 
+        # -- Region's rowspan="2" legitimately reaches into Northwest's row, its own group, so
+        # -- both land in the leading chunk together --
         assert len(chunks) == 3
+        original_html = chunks[0].metadata.text_as_html
+        assert original_html is not None
+        original_table = fragment_fromstring(original_html)
+        assert original_table.xpath("./tr[1]/td[1]/@rowspan") == ["2"]
+
         continuation_html = chunks[1].metadata.text_as_html
         assert continuation_html is not None
         continuation_table = fragment_fromstring(continuation_html)
@@ -1584,7 +1593,9 @@ class Describe_TableChunker:
         assert continuation_table.xpath("./thead/tr[1]/@data-role") == ["header-row"]
         assert continuation_table.xpath("./thead/tr[1]/th[1]/@scope") == ["col"]
         assert continuation_table.xpath("./thead/tr[1]/th[1]/@abbr") == ["region-code"]
-        assert continuation_table.xpath("./thead/tr[1]/th[1]/@rowspan") == ["2"]
+        # -- only one header row is ever prepended, so a repeated copy's rowspan="2" -- which
+        # -- would otherwise reach into the continuation's own body row -- is clipped away --
+        assert continuation_table.xpath("./thead/tr[1]/th[1]/@rowspan") == []
         assert continuation_table.xpath("./thead/tr[1]/th[2]/@class") == ["sales-cell"]
         assert continuation_table.xpath("./thead/tr[1]/th[2]/@data-k") == ["1"]
         assert continuation_table.xpath("./thead/tr[1]/th[2]/@colspan") == ["2"]
@@ -2452,7 +2463,10 @@ class Describe_TableChunker:
         assert reconstructed.xpath("./thead/tr[1]/th[1]/@abbr") == ["region-code"]
         assert reconstructed.xpath("./thead/tr[1]/th[2]/@colspan") == ["2"]
         assert reconstructed.xpath("./thead/tr[2]/th[1]/@headers") == ["sales-group"]
-        assert reconstructed.xpath("./thead/tr[2]/th[2]/@rowspan") == ["2"]
+        # -- "Revenue"'s rowspan="2" reaches one row past the header block into the first body
+        # -- row (Northwest's); only 2 header rows are ever carried into a repeated copy, so that
+        # -- reach is clipped away rather than claiming an arbitrary continuation's body row --
+        assert reconstructed.xpath("./thead/tr[2]/th[2]/@rowspan") == []
         assert reconstructed.xpath("./tr[1]/th") == []
         assert self._row_texts(table.metadata.text_as_html) == expected_rows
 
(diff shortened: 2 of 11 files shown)


cragwolfe:
I found three correctness issues that should be addressed before merging:

1. **Default KaTeX output is duplicated.** KaTeX's default `htmlAndMathml` output places a `.katex-mathml` branch beside a `.katex-html` visual branch. `Math.iter_text_segments()` now emits the TeX annotation, and normal `Phrasing` traversal then emits the sibling HTML as well (`parser.py:735-741`, registration at `1040-1041`). A simple rendered `2` therefore becomes `22`; structured formulas concatenate TeX with a flattened visual copy. Please recognize the paired producer structure and emit exactly one representation, while retaining HTML-only, MathML-only, unusable-MathML fallback, and two genuinely separate identical equations. Add KaTeX-shaped regression fixtures with exact occurrence counts.

2. **The annotation lookup can select a subexpression or discard a later usable representation.** `self.find(".//annotation[...]")` searches every descendant, so an annotation belonging to a nested `<semantics>` operand can replace the entire outer `<math>` even when the outer expression has its own annotation or `alttext` (`parser.py:724-733`). The loop also breaks as soon as it finds an element, before checking whether its text is non-empty, so an empty first candidate hides later supported annotations. Finally, the PR advertises lowercase `encoding="tex"`, but the exact XPath matching accepts only `"TeX"`. Please limit candidates to annotations that represent the whole expression, normalize supported encoding labels, and continue until a non-empty candidate is found before falling back to `alttext`. Cover nested semantics, lowercase `tex`, and empty-first/later-nonempty cases.

3. **Display math is merged into surrounding prose.** Every `Math` instance is unconditionally `Phrasing`, so `<div>Before<math display="block" alttext="x=1"/>After</div>` is accumulated as `Beforex=1After`, and adjacent display equations become one invented string (`parser.py:702-741`). `display="block"` is a defined MathML boundary and matches this parser's own block-item invariant. Please emit display math through an element-boundary path (preserving order, tails, metadata, links/emphasis) while keeping absent/inline display values inline; test prose/display/prose and adjacent display equations, including case-insensitive `block`.

The branch also needs the repository's release metadata update: it has no `CHANGELOG.md` entry and still reports `0.27.5`, while current `main` is `0.27.6`. Please update the changelog and synchronized package version after rebasing.

(authored by codex)

fireblade2534:
@cragwolfe Your issues should be fixed now

cragwolfe:
Follow-up content review at `1cd816fe85da019b878efb7237551ac4dffcd448`: the original fixes are substantially improved, but four correctness gaps remain.

1. **High — KaTeX pruning can delete or duplicate equations in table cells** (`unstructured/partition/html/parser.py:785-793`, `parser.py:556-565`). `_normalize_katex()` removes the visual branch when `_latex` is usable, but `TableBlock` bypasses `Math` and reads cells with `td.itertext()`. For an alttext-only pair, pruning leaves no text node and can make the table disappear; with presentation MathML plus a TeX annotation, raw `itertext()` can emit both forms. Please make table-cell traversal consume the selected math representation exactly once while retaining HTML fallback, and cover `Table.text` plus `text_as_html` for alttext-only, annotated, MathML-only, and unusable/HTML-only pairs.

2. **High — a direct `<semantics>` child is not necessarily a whole-expression representation** (`parser.py:713-732`). In `<math alttext="x+y"><semantics><mi>x</mi><annotation encoding="TeX">x</annotation></semantics><mo>+</mo><mi>y</mi></math>`, the annotation describes only `x`, but `_latex` emits it for the entire root and suppresses the valid `x+y` fallback (and can authorize KaTeX visual removal). Require the selected semantics node to cover the root expression, including sibling-before/after and multiple-semantics cases. Also correct `test_partition_html_prefers_math_alttext_over_tex_annotation` (`test_partition.py:1618-1640`): its expected `alttext` contradicts both the implementation and PR description's annotation-first contract, so the checked-in test cannot pass as written.

3. **High — unusable display MathML still loses its block boundary inside KaTeX/phrasing wrappers** (`parser.py:646-664`, `752-758`, `785-788`). If `<math display="block">` has neither annotation nor `alttext`, its block emitter yields nothing; the retained `.katex-html` sibling is then ordinary phrasing, producing `Beforex=1After` instead of three elements. Preserve the Math node's explicit display intent when selecting the visual fallback, including adjacent display fallbacks and non-whitespace tails, while keeping inline fallback inline.

4. **Medium — span-wrapped display math loses page metadata** (`parser.py:365-376`, `752-757`). `_page_number` only recurses through a `Flow` parent, so the normal span-wrapped/KaTeX shape stops lookup before an enclosing `data-page-number`; the formula gets `None` while surrounding prose gets the page. This can also cause default page-grouped HTML export to skip the formula. Resolve the nearest valid page number through phrasing wrappers and add wrapped KaTeX/anchor/emphasis cases, including invalid nearer values falling back outward.

(authored by codex)

fireblade2534:
@cragwolfe  Thanks for the follow-up review. I agree that #2 is a correctness issue in the MathML representation-selection logic added by this PR, and I’ll address that by only treating a `<semantics>` annotation as authoritative when it represents the whole `<math>` expression.

For #1, #3, and #4, though, I think we are expanding beyond the intended scope of this PR.

The goal of this change is to add basic MathML/KaTeX support to a parser that previously had no MathML-specific handling at all. Previously, `<math>` was not registered and fell through to the generic unknown-element behavior, so MathML content was generally skipped or only happened to be recovered indirectly through surrounding KaTeX HTML.

For #1, `TableBlock` has an independent extraction path based on raw `td.itertext()` and does not use the parser's `Flow`/`Phrasing` traversal. Making table cells understand MathML representation selection, annotations, alttext, and fallback semantics would require changing table parsing itself. I agree that the new KaTeX normalization should not make existing table behavior worse, so I can avoid pruning the visual KaTeX branch inside tables. That preserves the existing table behavior without turning this PR into a table-parser refactor.

For #3, preserving `display="block"` when MathML is unusable would require associating the failed `<math>` node with a separate `.katex-html` sibling and transferring the Math node's layout semantics onto that phrasing element. There is currently no parser abstraction for transferring block/inline semantics between alternate DOM representations. That seems like a broader fallback/layout feature rather than a correctness fix to the basic MathML extraction introduced here.

For #4, page-number inheritance currently only walks through `Flow` parents. That behavior predates this PR and is not specific to MathML. Changing it to walk transparently through spans, anchors, emphasis, invalid nearer page values, etc. changes generic metadata inheritance behavior across the parser and should be handled separately.

I’d prefer to keep this PR bounded to introducing basic MathML support and fixing regressions directly caused by that implementation. Otherwise this is effectively growing from basic MathML extraction into a much broader MathML/KaTeX parser integration, including tables, fallback layout propagation, and generic metadata ancestry.

fireblade2534:
@cragwolfe any updates

cragwolfe:
Thanks for the thoughtful scope breakdown and the updates at `89be3beaa578b8e83a3ce6e749752554590cff63`. I rechecked the current head. The whole-expression annotation selection is fixed, and the table guard is sufficient to avoid regressing existing table extraction. I also agree that carrying display semantics onto an unusable MathML visual fallback can be handled separately.

There is one narrow blocker remaining. In the normal `div[data-page-number] > span.katex > span.katex-mathml > math[display=block]` structure, this change prunes the visual branch and emits a separate `Math` element, but page-number lookup stops at the immediate phrasing wrapper. The emitted formula therefore receives `page_number=None`; default page-grouped `elements_to_html()` skips elements without a page number, so the formula can disappear from that output. Although the generic ancestry limitation predates this PR, the content loss becomes newly observable when this PR replaces the former page-bearing visual representation.

Please add a narrowly scoped page-number ancestor lookup for `Math` through phrasing wrappers, plus a regression using the normal span-wrapped KaTeX structure and page-grouped HTML export. This does not require changing generic metadata ancestry. With that focused fix, the content changes otherwise look ready from my review.

(authored by codex)

fireblade2534:
@cragwolfe Done

fireblade2534:
@cragwolfe

PR #4484, opened 2026-09-11
fix: recognize single-block list items as ListItem

### LINKED ISSUE #3499: feat(html): list-item adopts single child block element as its text
**Describe the bug**
I'm using loose lists in markdown *(each item is separated by a blank line.)* and the html parser fails to identify the list. Depending on the context it either categorises the elements as `Title` or `Narrative Text`.

Comparing the loose vs tight lists it seems it's got something to do with the paragraph tag but I'm not sure how exactly that's affecting the parser.

I can mostly work around this by parsring the markdown text first and remove any newlines that prepend a list item. Or by removing the paragraph tags from the html. But I'm not sure if either of those will run up against edge cases at some point so it would be nice if the html parser could handle this.

**To Reproduce**
```python
import markdown
from unstructured.partition.md import partition_md

loose_list = """
1. list item one.

2. list item two.

3. list item three.
"""

tight_list = """
1. list item one.
2. list item two.
3. list item three.
"""

print("markdown_loose:")
elements = partition_md(text=loose_list)
for el in elements:
    print(f"{el.category}: {el}")

print("\nmarkdown_tight:")
elements = partition_md(text=tight_list)
for el in elements:
    print(f"{el.category}: {el}")

print("\nhtml_loose:")
html = markdown.markdown(loose_list, extensions=["tables"])
print(html)

print("\nhtml_tight:")
html = markdown.markdown(tight_list, extensions=["tables"])
print(html)
```

**Expected behavior**
Both the tight and the loose lists should result in the `ListItem` element.

**Screenshots**
image

**Environment Info**
```bash
unstructured/scripts/collect_env.py:5: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  import pkg_resources
OS version:  Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python version:  3.11.9
unstructured version:  0.15.1
unstructured-inference version:  0.7.36
pytesseract version:  0.3.10
Torch version:  2.3.1.post300
Detectron2 version:  0.6
PaddleOCR is not installed
Libmagic version: file-5.41
magic file from /etc/magic:/usr/share/misc/magic
LibreOffice version:  LibreOffice 24.2.5.2 420(Build:2)
```

**Additional context**

huuufu:
### Summary

Resolves #3499.

A `<li>` whose only content is a single ordinary text block (`<li><p>text</p></li>`) now produces a `ListItem`. This is the HTML that Markdown *loose lists* (items separated by blank lines) compile to, with both `markdown` and `markdown-it-py`, so `partition_md()` on a loose list returned `Text`/`NarrativeText`/`Title` elements instead of `ListItem`s.

This implements the approach proposed in the issue thread: when a `<li>` contains exactly one block element, adopt that block's content as the list item's text.

### Change

`ListItemBlock.iter_elements()` in `unstructured/partition/html/parser.py` adopts the child block only when all of the following hold, otherwise it falls through to the existing `Flow.iter_elements()` traversal:

- the `<li>` has no text of its own and exactly one child element,
- the child is a plain `Flow`/`BlockItem` (`<p>`, `<div>`, `<blockquote>`) — not a heading, `<pre>`, image, table, or nested list,
- the child has no tail text and contains only phrasing content (so multi-paragraph items, including paragraphs nested inside inline markup, keep their current behavior).

The adopted element is built with the `<li>`'s own accumulator, so `category_depth` reflects list nesting, and inline annotations (links, emphasis) and `data-page-number` on the child are preserved.

### Before / after

```python
from unstructured.partition.md import partition_md

loose = "1. list item one.\n\n2. list item two.\n\n3. list item three.\n"
for el in partition_md(text=loose):
    print(f"{el.category}: {el}")
```

Before:
```
UncategorizedText: list item one.
UncategorizedText: list item two.
UncategorizedText: list item three.
```

After:
```
ListItem: list item one.
ListItem: list item two.
ListItem: list item three.
```

Tight lists are unchanged.

### Testing

- `test_unstructured/partition/html/test_parser.py::DescribeListItemBlock` — adoption for `<p>`/`<div>`/`<blockquote>`, annotation and nested-depth preservation, page-number inheritance, and 14 cases that must keep normal traversal (multiple paragraphs, prefix/suffix text, nested list, `<pre>`, heading, table, image, `<figure>`, whitespace-only block, plain text).
- `test_unstructured/partition/test_md.py::test_partition_md_recognizes_loose_and_tight_lists` — end-to-end for ordered and unordered lists, tight and loose.
- Full `test_unstructured/partition/html` + `test_md.py` suites pass locally; the broader `test_unstructured` run has the same result set as `main` in my environment.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4484?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Fixes
 
+- **Recognize HTML and Markdown loose-list items.** A list item containing a single ordinary text block (such as `<li><p>text</p></li>`) now produces a `ListItem`, preserving inline annotations and list depth. Multi-paragraph items and specialized blocks retain their existing behavior. Resolves #3499.
+
 - **`partition_doc()` and `partition_ppt()` no longer fail on a document whose name contains multi-byte characters.** `convert_office_doc()` decoded `soffice` stdout and stderr with a strict UTF-8 decode purely to log them and to check whether stdout was empty. LibreOffice echoes the input path using the console encoding, which on Windows is the locale codepage, so a document whose name or path contains multi-byte characters raised `UnicodeDecodeError` and aborted a conversion that would otherwise have succeeded. All three decode sites now go through one helper using `errors="backslashreplace"`, which keeps the message pure ASCII -- readable, still loggable by a handler using the locale codepage, and showing the offending bytes. Resolves #3652.
 
 - **A stray processing instruction no longer crashes HTML partitioning.** `partition_html` (and formats that route through it, such as `.md`) raised `AttributeError: 'lxml.etree._ProcessingInstruction' object has no attribute 'is_phrasing'` when the HTML contained a processing-instruction node like a `<?xml ...?>` declaration. The parser now drops processing instructions at parse time, the same way it already drops comments.
--- a/test_unstructured/partition/html/test_parser.py
+++ b/test_unstructured/partition/html/test_parser.py
@@ -374,6 +374,80 @@ class Describe_PreElementAccumulator:
 # -- FLOW (BLOCK-ITEM) ELEMENTS ------------------------------------------------------------------
 
 
+class DescribeListItemBlock:
+    @pytest.mark.parametrize("tag", ["p", "div", "blockquote"])
+    @pytest.mark.parametrize("text", ["list item one.", "A", "* literal bullet"])
+    def it_adopts_a_single_text_block(self, tag: str, text: str):
+        root = etree.fromstring(f"<ul><li>\n<{tag}>{text}</{tag}>\n</li></ul>", html_parser)
+
+        (element,) = root.find("body").iter_elements()
+
+        assert element == ListItem(text)
+        assert element.metadata.category_depth == 1
+
+    def it_preserves_annotations_and_nested_list_depth(self):
+        root = etree.fromstring(
+            '<ul><li>outer<ul><li data-page-number="3"><p>'
+            '<a href="https://example.com">link</a> <b>bold</b>'
+            "</p></li></ul></li></ul>",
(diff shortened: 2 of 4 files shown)


cragwolfe:
**Low, non-blocking — preserve cross-page nested items during page-grouped HTML export.**

This change can expose an existing exporter edge case when a nested list item is assigned to a different page from its parent. `set_element_hierarchy()` gives the child the parent item's ID, but default page-grouped `elements_to_html()` processes that child without its parent in the page subset. Because the child still has a `parent_id`, the exporter excludes it from the page roots and its text is omitted from the generated HTML. The partitioned element and Unstructured JSON remain intact.

This is limited to cross-page hierarchy plus page-grouped HTML export and should not block this PR. A follow-up can preserve missing-parent items as page-local roots or continuation children, with a partition-to-export regression covering same-page, cross-page, and `no_group_by_page=True` cases.

(authored by codex)

cragwolfe (approved):
The core list-item classification change looks good. The cross-page page-grouped HTML-export edge case noted separately is non-blocking and can be handled as a follow-up.

(authored by codex)

PR #4485, opened 2026-09-13
fix: avoid getpgid crash for GLOBAL_WORKING_PROCESS_DIR on Windows

sbguangha:
## Problem

Enabling `GLOBAL_WORKING_DIR` or reading the default `GLOBAL_WORKING_PROCESS_DIR` called `os.getpgid(0)`. That API is POSIX-only, so Windows raised `AttributeError`.

## Solution

Resolve the process key with `os.getpgid` when it exists, otherwise `os.getpid()`. POSIX process-group sharing is unchanged.

## User impact

Windows users can enable the global working directory without a crash.

## Evidence

`uv run --with pytest python -m pytest test_unstructured/partition/utils/test_config.py -q` — 6 passed, including `test_get_tempdir_without_getpgid`.

CHANGELOG updated under `0.27.8-dev0`.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4485?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cragwolfe (approved):
Reviewed exact head `1e9301bd592527bc4d39257168705e9985e49d95`.

No actionable findings. The fallback is correctly scoped: POSIX retains process-group-keyed sharing through `os.getpgid(0)`, while platforms without that API use `os.getpid()` for per-process isolation. Override precedence, caching, directory creation, and direct consumers remain unchanged, and the regression test exercises the missing-API path.

Execution evidence at this head is limited to two Socket Security checks, both successful; no native-Windows or general test workflow is reported. The branch is currently conflicted with `main`, which remains operational follow-up before merge.

(authored by codex)

PR #4486, opened 2026-09-14
test(partition): add unit test suite for document element ID hash determinism

gcoinstash-cmd:
### Summary
Adds unit tests verifying element ID determinism based on text content and document coordinates:
- Asserts that identical text content produces consistent, deterministic element identifiers.
- Confirms distinct text payloads generate unique hash digests.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured/pull/4486?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cragwolfe (changes requested):
High: the new tests do not exercise production element-ID generation and assert an incompatible surrogate contract.

`test_element_id_invariants.py:4-16` defines a local helper that hashes `element_type` plus stripped text into a 64-character digest, then tests that helper against itself. Production `Element.id_to_hash()` instead hashes filename, raw text, page number, and per-page sequence number, and truncates the digest to 32 characters. It does not use element type or coordinates. A regression that drops a production input, breaks duplicate-text disambiguation, changes output length, or breaks ID remapping would therefore leave both new tests green. Existing tests already cover determinism and golden hashes through the real implementation, so these tests add no regression protection while documenting the wrong contract.

Please remove the local helper and exercise the production APIs (`Element.id_to_hash()` and/or the partition metadata hash-assignment path) with real elements. The replacement should establish the intended invariant for filename, page number, raw text, sequence number, duplicate text, and 32-character output. The PR description's coordinate claim also needs to be reconciled explicitly: under the current implementation coordinate-only changes do not affect the ID, so either test that behavior or change the production contract separately.

(authored by codex)

